PR #2117, opened 2026-05-26
[PR #230] Add lint to enforce humantime_serde usage for Duration in config structs

orivaris:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#230](https://github.com/cyberfabric/cyberware-rust/pull/230) | **Author:** nanoandrew4 | **Opened:** 2026-01-13T11:05:16Z | **Status:** closed (not merged)

---

---


il10241024:
**nanoandrew4** commented on 2026-01-13T11:25:52Z:

---

The lints are failing because the `modkit-transport-grpc` lib has a config struct that is missing the now mandatory annotations for duration serialization/deserialization. We can either add them despite them not being necessary for the fields in that config struct, since it doesn't seem like it is populated from a file, or we can weaken the lint to only look at files called `config.rs` or something like that, so that not all config structs need to have their Durations marked with the serde macro. I'm leaning towards the first, but am interested in collecting some feedback before making any changes.

il10241024:
**nanoandrew4** commented on 2026-01-13T17:25:13Z:

---

After speaking with &#2369;Bechma , we have decided to close this PR in favour of adding the recommendation to the guidelines. The issue is that no lint will reliably catch all durations that need to use the macro, and it would also force struct fields which don't need it to add it, which isn't helpful.

il10241024:
**Bechma** reviewed `dylint_lints/de12_config_layer/de12_enforce_humantime_serde/src/lib.rs` line 25 on 2026-01-13T14:13:02Z:

---

what if used after a 
```rust
use modkit_utils::humantime_serde
```

il10241024:
**Bechma** reviewed `dylint_lints/de12_config_layer/de12_enforce_humantime_serde/src/lib.rs` line 91 on 2026-01-13T14:17:20Z:

---

it's possible to have more than one attribute, so we need to locate the segment with ident name with_sym

il10241024:
**Bechma** reviewed `dylint_lints/de12_config_layer/de12_enforce_humantime_serde/src/lib.rs` line 127 on 2026-01-13T14:22:46Z:

---

regarding your comment, checking the name of the struct ends with Config maybe it's not the best approach, but to call the struct FileConfig or something like that. We could have Config names everywhere without the need of having this rule

Issue #3433, opened 2026-06-04
[PR #1974] feat(usage-collector): reset specs to simplified design and drop SDK scaffold


il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#1974](https://github.com/cyberfabric/cyberware-rust/pull/1974) | **Author:** capybutler | **Opened:** 2026-05-27T09:49:55Z | **Status:** merged on 2026-06-04T08:04:01Z
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

## Summary

Resets the Usage Collector specification stack to a simplified design and removes the prior SDK scaffold so the rewrite can land cleanly.

### What changes

- **Specs rewritten**: PRD, DESIGN, DECOMPOSITION updated to the simplified
  design (Usage Collector as a centralized metering store with a pluggable
  storage SPI; business logic stays in downstream consumers).
- **ADRs replaced**: the two prior ADRs (`scoped-emit-source`,
  `two-phase-emit-authz`) are dropped in favor of seven scoped decisions:
  - `0001` PDP-centric authorization
  - `0002` Pluggable storage
  - `0003` Caller-supplied attribution
  - `0004` Mandatory idempotency
  - `0005` Monotonic deactivation
  - `0006` Contract stability
  - `0007` Gateway-local metric catalog
- **Features replaced**: the single `sdk-and-ingest-core` feature is split
  into five focused features: `foundation`, `metric-lifecycle`,
  `usage-emission`, `usage-query`, `event-deactivation`.
- **New design docs**: `domain-model.md`, `plugin-spi.md`, `sdk-trait.md`,
  and an OpenAPI v1 sketch (`usage-collector-v1.yaml`).
- **Code**: removes the `usage-collector-sdk` crate (workspace member, root
  `Cargo.toml` dependency entry, and sources). The crate predates the
  simplified design and will be re-introduced from the rewritten
  `sdk-trait.md` in a follow-up PR.
- **Cypilot config**: removes the temporary ignore patterns that covered
  the deleted code paths and prior feature directory.

### What is *not* in this PR

- No production behavior changes; the removed SDK crate had no consumers
  in-tree.
- No implementation of the new specs — those land in follow-up PRs that
  reintroduce the SDK, ingest path, and storage plugin against the
  simplified design.

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
[...]

PR #3497, opened 2026-06-04
[PR #1919] feat(usage-collector): [3/6] add REST client crate for remote ingest delivery

il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#1919](https://github.com/cyberfabric/cyberware-rust/pull/1919) | **Author:** capybutler | **Opened:** 2026-05-16T07:06:53Z | **Status:** closed (not merged)

---

- Adds `cf-usage-collector-rest-client` crate: a ModKit module for out-of-process
  binaries that must deliver usage records to a remote collector gateway over HTTP
  instead of in-process.
- `UsageCollectorRestClientModule::init()` loads config, acquires `AuthNResolverClient`
  and `AuthZResolverClient` from `ClientHub`, builds `UsageCollectorRestClient`
  with a configured `HttpClient`, wires it into `UsageEmitter`, and registers
  `UsageEmitterV1` — identical surface API to the in-process path.
- `UsageCollectorRestClient` implements `UsageCollectorClientV1`:
  - `create_usage_record`: acquires a fresh bearer token per attempt, POSTs to
    `POST /usage-collector/v1/records`; maps `204→Ok`, `401/403→AuthorizationFailed`,
    `429/5xx→PluginTimeout` (Retry), other `4xx→Internal` (Reject),
    timeouts→`PluginTimeout`, transport errors→`Unavailable`.
  - `get_module_config`: GETs `GET /usage-collector/v1/modules/{module_name}/config`
    with percent-encoded module name; maps `200→ModuleConfig`, `404→ModuleNotFound`.
  - AuthN errors: `Unauthorized/NoPluginAvailable` → permanent reject;
    all others → transient retry.
- `DatabaseCapability::migrations()` returns `outbox_migrations()` — the module owns
  the source's local outbox queue.
- TLS check: startup WARN when `base_url` uses `http://` with a non-localhost host
  (cpt-cf-dod-rest-ingest-tls-config).
- Adds feature design document `0002-cpt-cf-usage-collector-feature-rest-ingest` (v1.3)
  covering actor flows, business logic, DoD, and acceptance criteria.

## PR Train
| # | PR | Description | Lines |
|---|---|------------|------:|
| 1 | [1907](https://github.com/constructorfabric/cyberfabric-core/issues/1907) | Feature 1 (Core SDK, Emitter & In-Process Ingest): SDK + docs | ~2500 |
| 2 | [1908](https://github.com/constructorfabric/cyberfabric-core/issues/1908) | Feature 1 (Core SDK, Emitter & In-Process Ingest): `usage-collector` gateway + `usage-emitter` + `noop-usage-collector-storage-plugin` | ~8000 |
| 3 | [1919 (**this**)](https://github.com/constructorfabric/cyberfabric-core/issues/1919) | Feature 2 (REST Client & Remote Ingest Delivery) | ~2800 |
| 4 | [1926](https://github.com/constructorfabric/cyberfabric-core/issues/1926) | Feature 3 (Query API) | ~6000 |
| 5 | [1927](https://github.com/constructorfabric/cyberfabric-core/issues/1927) | Feature 4 (TimescaleDB plugin) | ~7700 |
| 6 | [1835](https://github.com/constructorfabric/cyberfabric-core/issues/1835) | e2e tests | ~1500 |
|   |  | ... to be continued ... | |

## Checklist
- [x] Self-review completed
- [x] No linting errors (`cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::perf`)
- [x] `cpt validate` — all checks passed
- [x] Tests pass (`cargo nextest run --workspace`)

## Release Notes

* **New Features**
  * Added REST client module for remote usage collection with OAuth2 bearer token authentication and configurable request timeouts.
  * Implemented TLS/HTTPS security validation for production deployments.

* **Documentation**
  * Added feature specification and configuration guide for the REST-based usage collector module.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3127,6 +3127,39 @@ dependencies = [
 ]
 
 [[package]]
+name = "cyberware-usage-collector-rest-client"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "bytes",
+ "chrono",
+ "cyberware-authn-resolver-sdk",
+ "cyberware-authz-resolver-sdk",
+ "cyberware-modkit",
+ "cyberware-modkit-db",
+ "cyberware-modkit-http",
+ "cyberware-modkit-macros",
+ "cyberware-modkit-security",
+ "cyberware-usage-collector-sdk",
+ "cyberware-usage-emitter",
+ "http",
+ "http-body-util",
+ "httpmock",
+ "sea-orm-migration",
+ "secrecy",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tower",
+ "tracing",
+ "tracing-test",
+ "url",
+ "uuid",
+ "zeroize",
+]
+
+[[package]]
 name = "cyberware-usage-collector-sdk"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ members = [
     "modules/mini-chat/mini-chat",
     "modules/system/usage-collector/plugins/noop-usage-collector-plugin",
     "modules/system/usage-collector/usage-collector",
+    "modules/system/usage-collector/usage-collector-rest-client",
     "modules/system/usage-collector/usage-collector-sdk",
     "modules/system/usage-collector/usage-emitter",
 ]
@@ -249,6 +250,7 @@ cyberware-system-sdk-directory = { version = "0.1.35", path = "libs/system-sdks/
 account-management-sdk = { package = "cyberware-account-management-sdk", version = "0.1.1", path = "modules/system/account-management/account-management-sdk" }
 types-registry-sdk = { package = "cyberware-types-registry-sdk", version = "0.2.2", path = "modules/system/types-registry/types-registry-sdk" }
 tenant-resolver-sdk = { package = "cyberware-tenant-resolver-sdk", version = "0.3.6", path = "modules/system/tenant-resolver/tenant-resolver-sdk" }
+authn-resolver-sdk = { package = "cyberware-authn-resolver-sdk", version = "0.3.16", path = "modules/system/authn-resolver/authn-resolver-sdk" }
 authz-resolver-sdk = { package = "cyberware-authz-resolver-sdk", version = "0.3.6", path = "modules/system/authz-resolver/authz-resolver-sdk" }
 resource-group-sdk = { package = "cyberware-resource-group-sdk", version = "0.1.3", path = "modules/system/resource-group/resource-group-sdk" }
 oagw-sdk = { package = "cyberware-oagw-sdk", version = "0.5.1", path = "modules/system/oagw/oagw-sdk" }
@@ -261,6 +263,7 @@ mini-chat-sdk = { package = "cyberware-mini-chat-sdk", version = "0.10.6", path 
 
 # usage-collector
 usage-collector = { package = "cyberware-usage-collector", version = "0.1.0", path = "modules/system/usage-collector/usage-collector" }
+usage-collector-rest-client = { package = "cyberware-usage-collector-rest-client", version = "0.1.0", path = "modules/system/usage-collector/usage-collector-rest-client" }
 usage-collector-sdk = { package = "cyberware-usage-collector-sdk", version = "0.1.0", path = "modules/system/usage-collector/usage-collector-sdk" }
 usage-emitter = { package = "cyberware-usage-emitter", version = "0.1.0", path = "modules/system/usage-collector/usage-emitter" }
 
--- a/modules/system/usage-collector/docs/features/0002-cpt-cf-usage-collector-feature-rest-ingest.md
+++ b/modules/system/usage-collector/docs/features/0002-cpt-cf-usage-collector-feature-rest-ingest.md
@@ -1,0 +1,312 @@
+---
+cpt:
+  version: "1.11"
+  changelog:
+    - version: "1.11"
+      date: "2026-05-16"
+      changes:
+        - "Re-align with the updated `cpt-cf-usage-collector-feature-sdk-and-ingest-core` emitter and error semantics after rebase: (a) `ClientHub` registration key is `dyn UsageEmitterRuntimeV1` (the layer-1 runtime) — `UsageEmitterFactoryV1` is not a trait that exists, the factory is the cloneable layer-2 `UsageEmitterFactory` obtained via `runtime.factory(module_name)`. Out-of-process sources now drive the three-layer chain `runtime.factory(MODULE_NAME).with_*().authorize(ctx, resource_id, resource_type)?` then `usage_record_builder()?.build()? → enqueue(record)` / `enqueue_in(db, record)`. Updated §1.1, §1.2, §2 Module Init Flow output, §3 `inst-init-6` (`UsageEmitter` → `UsageEmitterRuntime::build(...)`) and `inst-init-7` (registration key), §5 DoD body for `UsageCollectorRestClientModule`, and §6 AC. (b) Module Config Retrieval Flow (REST) — the factory's `.authorize()` step propagates non-`NotFound` variants from `get_module_config()` unchanged per Feature 1 `inst-authz-5a`; the emitter no longer collapses them to `UsageEmitterError::Internal`. §2 Error Scenarios rewritten to list `ServiceUnavailable` (transport / 401 / 5xx), `DeadlineExceeded` (timeout), `ResourceExhausted` (429), and `PermissionDenied` (403) propagation paths; `inst-cfg-rem-6a` now enumerates the explicit HTTP-status → canonical-variant mappings (`401`→`ServiceUnavailable`, `403`→`ModuleConfigError::permission_denied()`, `429`→`ModuleConfigError::resource_exhausted()`, `5xx`→`ServiceUnavailable`, timeout→`ModuleConfigError::deadline_exceeded()`, transport / residual `HttpError` variants→`ServiceUnavailable`, any other unexpected status→`Internal`) and corrects the stale `inst-authz-5b` cross-reference to `inst-authz-5a`; §5 DoD body and §6 AC for `get_module_config()` extended with the same explicit mapping list. (c) `UsageRecord` JSON shape — the legacy `subject_id` / `subject_type` field pair has been replaced by a single nested `subject: Option<Subject>` object (Feature 1 `inst-enq-8`); `inst-rem-5` updated to describe the nested serialization (absent when `None`, `\"subject\": { \"id\": \"...\", \"type\": \"...\" }` when `Some`, with the inner `type` itself absent when `Subject.r#type` is `None`)."
+    - version: "1.10"
+      date: "2026-05-10"
+      changes:
+        - "§5 DoD `BearerTokenAuthLayer` bullet: widen the AuthN-resolver-originated outcome collapse so the narrative matches `bearer_token_auth_layer.rs`. All AuthN-resolver-originated outcomes — transient resolver failures, permanent credential rejection (`Unauthorized` / `NoPluginAvailable`), and the post-exchange missing-token guard when the returned `SecurityContext` carries no bearer token — collapse to `HttpError::Transport(Box<AuthNResolverError>)`. In addition, an invalid token byte sequence yields `HttpError::InvalidHeaderValue`, which the REST client maps to `ServiceUnavailable` via the residual non-Transport, non-Timeout arm (already covered by the `create_usage_record()` mapping list — this is now stated explicitly in the layer description). No code change."
+    - version: "1.9"
+      date: "2026-05-10"
+      changes:
+        - "Document the bearer-token Tower middleware architecture (RESEARCH-diverge issue G) and the flat `infra/` crate layout (RESEARCH-diverge issue J). Token acquisition + `Authorization` header injection are now a per-request `tower::Layer` (`BearerTokenAuthLayer`), wired into the shared `HttpClient` via `HttpClientBuilder::with_auth_layer(...)`; the REST client's `create_usage_record()` / `get_module_config()` issue plain `.post(...).send()` / `.get(...).send()` calls. The layer collapses transient AuthN failures and permanent credential rejection alike to `HttpError::Transport`, which `rest_client.rs` maps to `ServiceUnavailable` (preserves the previous `inst-rem-3a` / `inst-rem-4a` mapping). Updates: §2 Remote Usage Emission Flow steps `inst-rem-2` / `inst-rem-3` / `inst-rem-4` and Module Config Retrieval Flow step `inst-cfg-rem-2`; §3 REST Client Module Initialization step `inst-init-5` (HTTP client built with `BearerTokenAuthLayer` injected via `with_auth_layer`); §5 DoD body for `UsageCollectorRestClient` (token acquisition is performed by the `BearerTokenAuthLayer`; the REST client method itself only issues the HTTP request). Crate layout: §5 DoD body now reflects the flat `infra/` module (`rest_client.rs`, `bearer_token_auth_layer.rs`, plus tests + `test_support.rs`); the previous `api/rest/dto.rs` DTO layer has been removed — the client serializes `usage_collector_sdk::models::UsageRecord` directly. Public surface: `pub use config::UsageCollectorRestClientConfig; pub use infra::UsageCollectorRestClient;`."
+    - version: "1.8"
+      date: "2026-05-10"
+      changes:
+        - "Align emitter trait name with the implemented crate surface (RESEARCH-diverge issue C): `UsageEmitterV1` → `UsageEmitterFactoryV1` across §1 Overview, §2 Module Init Flow output, §3 inst-init-7, §5 DoD body for `UsageCollectorRestClientModule`, and §6 AC for `ClientHub` availability. The factory trait is the `ClientHub` registration key shared with Feature 1."
+    - version: "1.7"
+      date: "2026-05-10"
+      changes:
+        - "Complete v1.6 canonical-taxonomy alignment by fixing two missed sites that still referenced the nonexistent `ModuleNotFound` variant: §5 DoD body for the `usage-collector-rest-client` crate (`maps 404 to ModuleNotFound` → `maps 404 to NotFound (built via ModuleConfigError::not_found(module_name))`) and §6 AC for `get_module_config()` (`returns ModuleNotFound on 404` → `returns NotFound (built via ModuleConfigError::not_found(module_name)) on 404`). v1.6 updated §2 and `inst-cfg-rem-5a`/`-6a` but missed the DoD body and AC."
+    - version: "1.6"
+      date: "2026-05-10"
+      changes:
+        - "Align Module Config Retrieval Flow (REST) with the canonical taxonomy (`UsageEmitterError = UsageCollectorError = CanonicalError`; resource-scoped builder `ModuleConfigError`): §2 Error Scenarios — `UsageEmitterError::ModuleNotConfigured` → `UsageEmitterError::NotFound` (built via `ModuleConfigError::not_found()`); `inst-cfg-rem-5a` — `UsageCollectorError::ModuleNotFound(module_name)` (nonexistent variant) → `UsageCollectorError::NotFound` (built via `ModuleConfigError::not_found(module_name)`); `inst-cfg-rem-6a` — clarify that infrastructure failures are surfaced as `UsageCollectorError::Internal` / `ServiceUnavailable` / `DeadlineExceeded` per the canonical taxonomy."
+    - version: "1.5"
+      date: "2026-05-08"
+      changes:
+        - "Replace legacy `UsageCollectorError` variant names with their `CanonicalError` equivalents across §2 Error Scenarios, §2 Steps (`inst-rem-3a / 4a / 6a / 7a / 9a / 10a`), §5 DoD mapping list, §5 Permanent credential rejection, and §6 AC: `Unavailable` → `ServiceUnavailable`; `PluginTimeout` (HTTP timeout) → `DeadlineExceeded`; `PluginTimeout` (HTTP 429) → `ResourceExhausted`; `PluginTimeout` (HTTP 5xx) → `ServiceUnavailable`; `AuthorizationFailed` (HTTP 401) → `ServiceUnavailable`. `UsageCollectorError = CanonicalError` no longer exposes `PluginTimeout` / `AuthorizationFailed` builders. `HandlerResult::Retry` vs `Reject` routing is preserved by `delivery_handler.rs` (`DeadlineExceeded` / `ResourceExhausted` / `ServiceUnavailable` → Retry; everything else → Reject)."
+    - version: "1.4"
+      date: "2026-05-08"
+      changes:
+        - "Ratify AuthN permanent-rejection retry semantics in `inst-rem-4 / 4a`: both transient AuthN failures and permanent credential rejection (`Unauthorized` / `NoPluginAvailable`) map to `Unavailable` and Retry; permanent rejection is NOT auto-dead-lettered — operator recovery via token-acquisition `WARN` logs and retry-rate metrics; §2 Error Scenarios and §5 DoD updated"
+        - "Split 401 (Retry / fresh token) from 403 (Reject / `PermissionDenied`) across DoD mapping list, §2 Error Scenarios, and `inst-rem-11a` — 403 now dead-letters as `PermissionDenied`; `inst-rem-9` (401 retry) unchanged in identity"
+        - "Remove `request_timeout` from §5 Configuration table and from `inst-init-5` — `UsageCollectorRestClientConfig` does not declare it; HTTP client built with `HttpClientConfig::default()`"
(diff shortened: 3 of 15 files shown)


il10241024:
**coderabbitai[bot]** commented on 2026-05-16T07:06:59Z:

---




> [!IMPORTANT]
> ## Review skipped
> 
> Auto reviews are disabled on base/target branches other than the default branch.
> 
> 
> 
> 
> <details>
> 
> 
> **Review profile**: CHILL
> 
> **Plan**: Free
> 
> **Run ID**: `05ef00a6-6a79-4977-a719-ada29da210fc`
> 
> </details>
> 
> 
> - [ ]  🔍 Trigger review





---

> [!NOTE]
> <details>
> <summary>🎁 Summarized by CodeRabbit Free</summary>
> 
> Your organization is on the Free plan. CodeRabbit will generate a high-level summary and a walkthrough for each pull request. For a comprehensive line-by-line review, please upgrade your subscription to CodeRabbit Pro by visiting <https://app.coderabbit.ai/login>.
> 
> </details>


<sub>Comment `@coderabbitai help` to get the list of available commands and usage tips.</sub>

il10241024:
**qodo-code-review[bot]** commented on 2026-05-16T19:33:27Z:

---

### Qodo reviews are paused for this user.

Troubleshooting steps vary by plan [Learn more →](https://docs.qodo.ai/review-eligibility)


**On a Teams plan?**
Reviews resume once this user has a paid seat *and* their Git account is linked in Qodo.
[Link Git account →](https://app.qodo.ai)

**Using GitHub Enterprise Server, GitLab Self-Managed, or Bitbucket Data Center?**
These require an Enterprise plan - Contact us
[Contact us →](https://app.qodo.ai)

il10241024:
**capybutler** commented on 2026-05-27T09:57:10Z:

---

The approach to usage collector is revised in #3433

il10241024:
**Artifizer** submitted review **COMMENTED** on 2026-05-20T08:17:28Z:

---



---

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 193 on 2026-05-16T07:08:12Z:

---

&#2369;MikeFalcon77 commented:

> 
> **HIGH**
> 
> Comment on the preceding line claims `AuthNResolverError::Internal` is "Permanent: Retrying will not fix this", but the code maps it to `UsageCollectorError::unavailable(...)` — and the very next match arm at line 200 explicitly documents `Unavailable` as the *transient retry* path ("Retrying after backoff is appropriate"). The two comments contradict each other for the same target variant, and the PR description states `Internal → transient retry`.
> 
> A future maintainer reading the "Permanent" comment will likely change the mapping to `internal(...)`, silently flipping retry behavior on AuthN subsystem failures.
> 
> Delete the misleading "Permanent: ... Retrying will not fix this" comment, or — if the intent really is permanent — change the mapping to `UsageCollectorError::internal(...)` and add a regression test through `DeliveryHandler` that asserts `MessageResult::Reject`.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 128 on 2026-05-16T07:08:46Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> The non-204 branch discards the response body before constructing the error, so the server's RFC 9457 Problem details (or any diagnostic payload) never reach logs or the returned `UsageCollectorError`.
> 
> When the collector returns 400/401/403/422 in production, operators see only `"unexpected HTTP status from usage collector: <status>"` with no reason, making it impossible to distinguish bad-payload bugs from auth-config bugs without server-side correlation.
> 
> Read the body (with a small size cap) before mapping the status, and include it as context: e.g. `let body = response.text().await.unwrap_or_default(); ... UsageCollectorError::internal(format!("unexpected HTTP {status} from usage collector: {body}"))`. Apply the same change in the 401/403 and 429/5xx arms inside `http_status_to_usage_collector_error`.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 173 on 2026-05-16T07:09:15Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> `get_module_config` discards the response body for every non-200/non-404 status, so a 401, 403, 422, or 5xx from the config endpoint surfaces only as `"unexpected HTTP status"` without the server-side reason.
> 
> Same production-debugging gap as `create_usage_record`: operators cannot tell a bad token from a malformed query without going to collector logs.
> 
> Capture the response body with a size limit and include it in the error string returned by the status-mapping helper, or pass body content through to the variant constructors.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 74 on 2026-05-16T07:09:43Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> `bearer_token` exits the `SecretString` wrapper via `.expose_secret().to_owned()` and returns plain `String`, then line 89 builds the auth header with `format!("Bearer {token}")`. From this point the token lives in unwrapped `String` values that participate in `Debug`/`Display` like any other string.
> 
> The `secrecy` crate exists specifically to prevent accidental leaks via tracing, error formatting, or future `request.headers()` dumps. Once the token is unwrapped here, that protection is gone for the rest of the request path.
> 
> Return `Result<SecretString, UsageCollectorError>` from `bearer_token`, then build the header value at the call site using `format!("Bearer {}", token.expose_secret())` only inline, ideally feeding it into a `HeaderValue` flagged sensitive (`HeaderValue::set_sensitive(true)`) so it stays out of debug output.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/tests/rest_client_tests.rs` line 1 on 2026-05-16T07:10:21Z:

---

&#2369;MikeFalcon77 commented:

> **HIGH**
> 
> This integration test file duplicates the unit tests in `src/infra/rest_client_tests.rs` almost line-for-line — same `MockAuthN` outcomes, same `make_client`, same scenarios (`*_returns_ok_on_204` ↔ `*_success_on_204`, bearer header check, 401/403/500/429, percent-encoding, etc.). `tests/common/mod.rs` is a verbatim copy of the test scaffolding embedded in the unit tests file.
> 
> This inflates the test count without adding behavioral coverage, doubles maintenance cost (any contract change must be edited in two places), and makes it harder to spot real coverage gaps — e.g. neither file exercises 5xx response bodies, retry-after headers, or non-JSON 200 responses for `create_usage_record`.
> 
> Pick one location: either keep the unit tests in `src/infra/rest_client_tests.rs` (closer to the implementation) and delete the duplicates here, or move all HTTP-driven scenarios into `tests/` and leave only the pure error-mapping unit tests inside `src/infra/`. Then extract `MockAuthN` and helpers into a single `pub(crate)` test-support module rather than duplicating them.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/Cargo.toml` line 76 on 2026-05-16T07:10:51Z:

---

&#2369;MikeFalcon77 commented:

> **LOW**
> 
> `usage-emitter` is already declared in `[dependencies]` at line 25 with the same `{ workspace = true }` specifier; repeating it in `[dev-dependencies]` adds nothing — Cargo already exposes the production dep to integration tests.
> 
> Delete this entry.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 193 on 2026-05-16T08:27:12Z:

---

Resolved by refactor — the offending function no longer exists.

The `authn_error_to_usage_collector_error` helper (with the contradictory `Internal` arm) was deleted as part of moving token acquisition into a dedicated Tower middleware: `infra/bearer_token_auth_layer.rs`. The REST client now only sees `HttpError`, never `AuthNResolverError`.

**New mapping (`bearer_token_auth_layer.rs:103-106`):** every `AuthNResolverError` variant — `Internal`, `Unauthorized`, `NoPluginAvailable`, `ServiceUnavailable`, `TokenAcquisitionFailed`, etc. — is uniformly wrapped into `HttpError::Transport(Box::new(e))`. There is no per-variant branching, so the comment-vs-code mismatch is structurally impossible.

**Downstream (`rest_client.rs:90-112`):** `HttpError::Transport` (and other non-timeout transport variants) are mapped to `UsageCollectorError::service_unavailable()`, i.e. the transient-retry path. This is now consistent with the comment, which states all of these are transient from the outbox's point of view because the request never reached the server.

**Why this is intentional, not an accidental flip:** the comment block at `bearer_token_auth_layer.rs:94-102` explicitly documents the policy: both transient AuthN failures *and* permanent credential rejection collapse to `Transport` → `ServiceUnavailable` and are retried. Permanent rejection is deliberately **not** auto-dead-lettered; operators recover via the `WARN` log emitted on each token-acquisition failure plus retry-rate metrics. The original PR description's "Internal → transient retry" line is consistent with this; it was the deleted inline comment that was wrong.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 128 on 2026-05-16T08:30:25Z:

---

Resolved in the refactor — see `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs`.

The response body is now read once (with a 4 KiB cap) via a shared helper before status mapping, and included in every non-success arm:

```rust
async fn truncated_response_body(response: HttpResponse) -> String {
    const MAX: usize = 4_096;
    let mut body = response
        .bytes()
        .await
        .map(|b| String::from_utf8_lossy(&b).into_owned())
        .unwrap_or_default();
    if body.len() > MAX {
        body.truncate(body.floor_char_boundary(MAX));
    }
    body
}
```

In both `create_usage_record` and `get_module_config`, the non-success match now reads the body up-front:

```rust
status => {
    let body = truncated_response_body(response).await;
    match status {
        StatusCode::UNAUTHORIZED => /* ... "...HTTP 401: {body}" */,
        StatusCode::FORBIDDEN    => /* ... "...HTTP 403: {body}" */,
        StatusCode::TOO_MANY_REQUESTS => /* ... quota violation with "{body}" */,
        s if s.is_server_error() => /* ... "HTTP {s}: {body}" */,
        status => /* ... "unexpected HTTP status ...: {status}: {body}" */,
    }
}
```

So 400/401/403/422/429/5xx and the catch-all arm all surface the collector's RFC 9457 Problem details (or any diagnostic payload) in the returned `UsageCollectorError` / `UsageRecordError` / `ModuleConfigError`, capped at 4 KiB and truncated on a UTF-8 char boundary to avoid unbounded memory use or invalid UTF-8.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 173 on 2026-05-16T08:35:21Z:

---

Resolved.

`get_module_config` no longer discards the body for non-200/non-404 responses. The status arm at `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs:222` now reads the response via the shared `truncated_response_body` helper (4 KiB UTF-8 cap, `rest_client.rs:259`) and threads the captured body into every error variant:

- `401` &rarr; `UsageCollectorError::service_unavailable().with_detail("... HTTP 401: {body}")`
- `403` &rarr; `ModuleConfigError::permission_denied().with_reason("... HTTP 403: {body}")`
- `429` &rarr; `ModuleConfigError::resource_exhausted(...).with_quota_violation("requests", "... {body}")`
- `5xx` &rarr; `UsageCollectorError::service_unavailable().with_detail("... HTTP {s}: {body}")`
- residual 4xx / unknown &rarr; `UsageCollectorError::internal("unexpected HTTP status ...: {status}: {body}")`

`create_usage_record` was refactored symmetrically (`rest_client.rs:119`), so a 401, 403, 422, or 5xx from either endpoint now surfaces the server-side reason directly in the error, without operators having to dig into collector logs. The body read is size-limited and truncated on a UTF-8 char boundary, so a hostile or oversized payload cannot blow up memory or panic the client.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 74 on 2026-05-16T08:37:26Z:

---

Resolved. The bearer-token handling moved out of `rest_client.rs` into a dedicated Tower layer at `modules/system/usage-collector/usage-collector-rest-client/src/infra/bearer_token_auth_layer.rs`, and the old `bearer_token` helper that returned a plain `String` is gone.

In the new layer (lines 110–123):

```rust
let token = auth_result.security_context.bearer_token().ok_or_else(|| { /* ... */ })?;
// token is `&SecretString` (from `modkit-security`), never unwrapped to `String`.

let raw = zeroize::Zeroizing::new(format!("Bearer {}", token.expose_secret()));
let mut hv = HeaderValue::from_str(&raw).map_err(HttpError::InvalidHeaderValue)?;
hv.set_sensitive(true);

req.headers_mut().insert(AUTHORIZATION, hv);
```

This matches every point of the review:

- `SecurityContext::bearer_token()` returns `Option<&SecretString>`, so the secret stays wrapped on the path from AuthN to the format site. `.expose_secret()` is called exactly once, inline inside the `format!`.
- The intermediate `"Bearer …"` string is wrapped in `zeroize::Zeroizing`, so it is zeroed on drop instead of lingering as a plain `String` (stronger than the reviewer's suggestion of "inline only").
- The resulting `HeaderValue` is flagged with `set_sensitive(true)`, which keeps `hyper`'s HTTP/2 HPACK from indexing it and, more importantly here, makes `Debug`/header-dump output redact it.
- Because the token never leaves the layer's async closure, no other code in the request path (including `rest_client.rs`) ever sees an unwrapped token.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/tests/rest_client_tests.rs` line 1 on 2026-05-16T08:39:19Z:

---

Resolved. The duplication has been removed in the latest revision:

- Deleted the duplicate `tests/rest_client_tests.rs` and `tests/common/mod.rs` — the `tests/` directory no longer exists in the crate.
- Kept the HTTP scenarios in a single location next to the implementation: `src/infra/rest_client_tests.rs` (option 1 from the review).
- Extracted the shared scaffolding (`MockAuthN`, `test_cfg`, `test_record`) into one test-support module at `src/infra/test_support.rs`, declared `#[cfg(test)] pub mod test_support;` in `src/infra/mod.rs`.
- Both `rest_client_tests.rs` and `bearer_token_auth_layer_tests.rs` now import from that single module (`use super::super::test_support::...`), so any contract change is edited in one place.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector-rest-client/Cargo.toml` line 76 on 2026-05-16T08:40:27Z:

---

Resolved.

The duplicate `usage-emitter = { workspace = true }` entry in `[dev-dependencies]` has been removed. `usage-emitter` is now declared only once, in `[dependencies]` at `modules/system/usage-collector/usage-collector-rest-client/Cargo.toml:25`, and remains available to integration tests via Cargo's normal exposure of production deps.

Current `[dev-dependencies]` (lines 66–78) only contains `modkit-security`, `tokio`, `httpmock`, `http-body-util`, and `bytes` — no `usage-emitter` line.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/module.rs` line 19 on 2026-05-19T21:43:13Z:

---

**HIGH**

Making this a standalone `#[modkit::module]` whose only job is to host a REST implementation of `UsageCollectorClientV1` contradicts the framework rule documented in `docs/modkit_unified_system/09_oop_grpc_sdk_pattern.md` ("single `*-sdk` crate containing API trait, types, gRPC client, and wiring helpers"). The sibling `usage-emitter` README repeats this explicitly — it ships as a library crate, not a modkit module, precisely because there is no SDK-only consumer shape for it.

No other crate in the workspace introduces a modkit module solely to publish a transport client (no other `*-rest-client` modkit module exists). A future binary that wants the REST collector path now has to depend on this module *and* declare it in its `modules:` config — even though it has no transport, no DB tables of its own, no API. The runtime registration in `ClientHub` is the only thing tying it to ModuleCtx.

Move `UsageCollectorRestClient` and the `UsageEmitterRuntime` wiring into a plain library crate (mirroring `usage-emitter`) and let each host binary's own module build the runtime and call `client_hub().register::<dyn UsageEmitterRuntimeV1>(...)` from its own `init()`. This also resolves the migration-ownership problem flagged separately on `migrations()` and the duplicated bearer-auth layer flagged on `bearer_token_auth_layer.rs` — both fall out naturally once the crate stops being a module.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/module.rs` line 99 on 2026-05-19T21:43:13Z:

---

**HIGH**

`DatabaseCapability::migrations()` returns `modkit_db::outbox::outbox_migrations()` from *this* module. The outbox table is consumed by `UsageEmitterRuntime`, which the rest of this `init()` registers in `ClientHub` for use by other modules in the same binary.

If two host modules in one process both want to use the REST emit path (e.g. an agent module + a forwarder module), this module is a singleton that owns the migrations on behalf of all of them — but the source modules cannot easily express that dependency. And if the binary already has another module that runs migrations against the same outbox queue name (the in-process `usage-collector` module configured with the same `outbox_queue`), you get conflicting migration ownership.

Migrations belong to whichever module owns the data. If this crate becomes a library (per the module-shape finding), the host module declares `capabilities = [db]` and pulls `outbox_migrations()` from `modkit_db` directly — single owner, no cross-module migration coupling.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/bearer_token_auth_layer.rs` line 91 on 2026-05-19T21:43:13Z:

---

**HIGH**

`libs/modkit-auth/src/oauth2/token.rs` (`Token`) already implements all of that — background refresh, jitter, `min_refresh_period`, auto-refresh-on-401 via `BearerAuthAutoRefreshLayer`. Bypassing it means an outbox burst-drain after an IdP outage will issue one full client-credentials exchange per queued record (×retry count), hammering both the AuthN resolver and the upstream IdP in lockstep with the queue depth.

Either extend `libs/modkit-auth::Token` to source tokens from `AuthNResolverClient::exchange_client_credentials` (so this layer can use the existing `BearerAuthLayer` with caching), or wrap `AuthNResolverClient` in a similar `expires_in`-aware cache here. Per-request token acquisition is acceptable for a tutorial; it is not acceptable as the bearer source for an unattended outbox emitter.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 54 on 2026-05-19T21:43:13Z:

---

**MEDIUM**

`HttpClientConfig::default()` enables `RetryConfig::default()` and `RateLimitConfig::default()` with a 30s request timeout. Stacked on top of the outbox-level retry (the delivery handler already retries `ServiceUnavailable` / `ResourceExhausted` / `DeadlineExceeded`), a single 5xx burst from the collector now triggers `modkit-http retry × outbox retry × per-attempt token refresh`. Each retry inside `modkit-http` also re-runs `BearerTokenAuthLayer`, so the IdP gets hit n times per logical record too.

The README states "HTTP transport tuning is not exposed; the client is built with `HttpClientConfig::default()`", which freezes this multiplicative behavior into the configuration contract.

The outbox is the durable retry mechanism here — let it own retry. Build the client with retries disabled (`HttpClientConfig { retry: None, ..default() }`), or expose `RetryConfig` in the module config so operators can tune it explicitly. Either way, document the per-record amplification factor so operators sizing the collector know what burst load to expect.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/test_support.rs` line 74 on 2026-05-19T21:43:13Z:

---

**LOW**

`MockAuthN::authenticate` panics rather than returning an `AuthNResolverError`. The `MockAuthN` is meant to drive layer/client tests that only exercise `exchange_client_credentials`, but the trait surface is fixed — any future test path that calls `authenticate` (e.g. a regression test for an inbound bearer-token flow through the same mock) gets a panic with no `#[should_panic]` annotation, which obscures the real assertion.

Return `Err(AuthNResolverError::NoPluginAvailable)` (or a similar typed error) so the test still fails meaningfully if `authenticate` is reached unintentionally.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/bearer_token_auth_layer.rs` line 102 on 2026-05-19T21:43:13Z:

---

The doc-comment at lines 96-102 states "operators recover via the `WARN` log here plus retry-rate metrics", but the code at line 109 emits `tracing::debug!`, not `warn!`. The justification block at lines 103-108 contradicts the doc-comment above it — both DEBUG-because-burst-amplification *and* WARN-as-operator-signal cannot both hold.

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/lib.rs` line 30 on 2026-05-20T08:16:41Z:

---

**MEDIUM**

`lib.rs` exposes no public surface: all three modules are declared private (`mod config; mod infra; mod module;`) with no `pub use` re-exports anywhere. The DoD (§5) explicitly specifies the crate's public surface as `pub use config::UsageCollectorRestClientConfig;` and `pub use infra::UsageCollectorRestClient;` — neither is exported.

Any workspace member that needs to reference these types (e.g. a host binary that wants to construct a client in its own `init()` or a cross-crate integration test) cannot name them. If the ModKit macro covers everything and the types genuinely need no external visibility, the DoD text is wrong and will mislead readers.

Add the two `pub use` statements the DoD specifies, or update the DoD to state the crate has no public surface and all wiring goes through the module macro.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/infra/rest_client.rs` line 132 on 2026-05-20T08:16:57Z:

---

**MEDIUM**

The `other` arm maps `HttpError::InvalidHeaderValue` (among other variants) to `UsageCollectorError::internal(...)`, which dead-letters the outbox record permanently. The DoD (§5) places `InvalidHeaderValue` on the `ServiceUnavailable` (retry) path: "maps `HttpError::Transport(_)` … and any residual non-Transport, non-Timeout `HttpError` variants (`InvalidHeaderValue`, `BodyTooLarge`, `Tls`, …) to `ServiceUnavailable`."

`HttpError::InvalidHeaderValue` originates in `BearerTokenAuthLayer` when the IdP returns a token containing non-ASCII bytes — a transient IdP-side encoding bug, not a fault in the outbox record itself. Dead-lettering a valid record because the IdP momentarily returned a malformed token is the wrong failure mode.

Match `HttpError::InvalidHeaderValue` explicitly before the `other` arm and map it to `ServiceUnavailable`, or update the DoD to document the `Internal` mapping with rationale so the divergence is intentional and visible.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/docs/features/0002-cpt-cf-usage-collector-feature-rest-ingest.md` line 288 on 2026-05-20T08:17:11Z:

---

**MEDIUM**

This step states "Any path component of `collector_url` is replaced by the fixed REST routes … only the scheme + host[:port] are honored." The implementation in `build_url()` (`rest_client.rs:83–95`) does the opposite — it **appends** API segments to any existing path prefix via `path.pop_if_empty().extend(segments)`. The struct-level doc-comment on `UsageCollectorRestClient`, the README, and two dedicated tests (`create_usage_record_preserves_collector_url_path_prefix`, `get_module_config_preserves_collector_url_path_prefix`) all confirm the append/preserve behavior.

An operator reading this step and configuring `collector_url: "https://gw/usage/"` will expect the prefix to be discarded, so requests hit `https://gw/usage-collector/v1/records`. The client actually hits `https://gw/usage/usage-collector/v1/records`. If the gateway routes on the full path this silently misdirects all traffic with no obvious error.

Update `inst-init-5` to: "any existing path prefix in `collector_url` is preserved; the API segments (`/usage-collector/v1/...`) are appended to it."

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector-rest-client/src/config.rs` line 14 on 2026-05-20T08:17:23Z:

---

**MEDIUM — Architecture**

`OAuthConfig` (`client_id`, `client_secret`, `scopes`) duplicates the config surface that `libs/modkit-auth` already owns. `modkit-auth::Token` provides cached OAuth2 client-credentials exchange with background refresh via `ArcSwap` and a `Token::invalidate()` for 401-triggered re-fetch — the exact capability that `BearerTokenAuthLayer` reimplements synchronously per outbox record without caching (flagged separately on `bearer_token_auth_layer.rs:91`).

By owning its own OAuth config this crate: (1) diverges from the platform OAuth2 config schema if `modkit-auth` evolves, (2) forces every binary using the REST collector path to carry its own `client_id`/`client_secret` block rather than sharing a `modkit-auth`-managed token, (3) adds direct `secrecy`/`zeroize` dependencies that are already transitive through `modkit-auth`.

Replace `OAuthConfig` with `modkit-auth`'s equivalent config type. Build the `HttpClient` in `module.rs` via `HttpClientBuilderExt::with_bearer_auth(token)` — the token is then cached, background-refreshed, and consistent with every other OAuth2 client in the codebase.

---

Issue #2247, opened 2026-05-26
[PR #1] Port from sandbox


orivaris:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#1](https://github.com/cyberfabric/cyberware-rust/pull/1) | **Author:** modcrafter77 | **Opened:** 2025-09-02T09:18:35Z | **Status:** merged on 2025-09-06T05:29:27Z
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

Move all code from sandbox to mainline

---

orivaris:
**modcrafter77** commented on 2025-09-05T00:09:43Z:

---

TODO:
 1. ~~Eliminate `panic!/unwrap()` from libraries (registry, ingress, db locks, operation builder). Use proper error types.~~
2. Stabilize OpenAPI schema deduplication — canonical names + conflict checks.
3. Ensure graceful HTTP shutdown with lifecycle integration.
4. Keep `api_ingress` thin (gateway + OpenAPI registry); move sample routes to `examples/`.
5. Enforce domain → contract → HTTP error mapping into `ProblemResponse`.
6. CI: add workflows (fmt, clippy, tests, deny, audit, sccache).
7. Add `rust-toolchain.toml`, `justfile/Makefile`, clean up docs (remove “latest updates” logs).

** Nice-to-haves: **

* Metrics/logs for router rebuilds.
* Trace propagation (`request_id`, domain fields).
* Feature-gate heavy deps/assets.

orivaris:
**striped-zebra-dev** reviewed `modules/api_ingress/src/lib.rs` line 56 on 2025-09-03T17:09:12Z:

---

Minor for future:
I'd rather  use builder here, plus I think we'd better have some `restkit` library for that.

orivaris:
**striped-zebra-dev** reviewed `modules/users_info/src/infra/storage/mapper.rs` line 5 on 2025-09-03T17:25:47Z:

---

Don't we want to use `Into` for such cases?

orivaris:
**modcrafter77** reviewed `modules/users_info/src/infra/storage/mapper.rs` line 5 on 2025-09-03T18:59:41Z:

---

Yes. We can. It's a half-generated sample.

orivaris:
**modcrafter77** reviewed `modules/api_ingress/src/lib.rs` line 56 on 2025-09-04T22:03:31Z:

---

remove it at all.

orivaris:
[...]

Issue #2243, opened 2026-05-26
[PR #5] Refactor/move examples


orivaris:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#5](https://github.com/cyberfabric/cyberware-rust/pull/5) | **Author:** modcrafter77 | **Opened:** 2025-09-08T20:40:49Z | **Status:** merged on 2025-09-09T22:43:34Z
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

- renamed request_id to traceId for HAPIG compliance
- #module macro attribute caps renamed to capabilities for better readability
- users_info module is moved to modkit examples

---

Issue #2240, opened 2026-05-26
[PR #8] docs and ci: unify module docs and add generation prompt


orivaris:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#8](https://github.com/cyberfabric/cyberware-rust/pull/8) | **Author:** modcrafter77 | **Opened:** 2025-09-11T21:56:05Z | **Status:** merged on 2025-09-12T20:53:19Z
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

- Add docs/MODULE_CREATION_PROMT.md for module creation via LLM-based editors
- Few redundant documents are removed (all information presented in MODKIT_UNIFIED_SYSTEM.md)
- Refresh README.md and CONTRIBUTING.md (Rust stable, Makefile targets, links but a lot of TBDs)
- Align Makefile usage referenced across docs (quickstart/dev targets)
- gitworkflow CI is optimized. Do not run Rust CI for *.MD and /docs changes

---

PR #3503, opened 2026-06-04
[PR #1908] feat(usage-collector): [2/6] add emitter, gateway, REST client, and noop storage plugin crates

il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#1908](https://github.com/cyberfabric/cyberware-rust/pull/1908) | **Author:** capybutler | **Opened:** 2026-05-15T11:26:46Z | **Status:** closed (not merged)

---

Introduces the core ingest-path crates completing Feature 1 (SDK & Ingest Core):

- `usage-emitter`: two-phase PDP authorization (`authorize_for` / `authorize`), `UsageRecordBuilder` fluent API, transactional outbox enqueue (`enqueue` / `enqueue_in`), and async delivery worker forwarding to `UsageCollectorClientV1`. The security invariant ensures `UsageCollectorClientV1` is never published to `ClientHub`; source modules can only reach the collector through a PDP-authorized, tenant/resource-bound `AuthorizedUsageEmitter`.

- `usage-collector`: gateway ModKit module; resolves the active storage plugin by vendor, registers `UsageCollectorLocalClient` as `dyn UsageCollectorClientV1`, wires in the embedded emitter, and serves the REST ingestion API for out-of-process callers.

- `noop-usage-collector-storage-plugin`: discard-all storage plugin for local dev, CI, and smoke tests.

Updates DESIGN, DECOMPOSITION, and Feature 1 spec (→ v1.4) to reflect the implemented emitter API (`authorized.build_usage_record(...).enqueue()`), correct entity locations, and mark all p1 DoD items and CDSL blocks complete (status: IMPLEMENTED).

## PR Train
| # | PR | Description | Lines |
|---|---|------------|------:|
| 1 | [1907](https://github.com/constructorfabric/cyberfabric-core/issues/1907) | Feature 1 (Core SDK, Emitter & In-Process Ingest): SDK + docs | ~2500 |
| 2 | [1908 (**this**)](https://github.com/constructorfabric/cyberfabric-core/issues/1908) | Feature 1 (Core SDK, Emitter & In-Process Ingest): `usage-collector` gateway + `usage-emitter` + `noop-usage-collector-storage-plugin` | ~8000 |
| 3 | [1919](https://github.com/constructorfabric/cyberfabric-core/issues/1919) | Feature 2 (REST Client & Remote Ingest Delivery) | ~2800 |
| 4 | [1926](https://github.com/constructorfabric/cyberfabric-core/issues/1926) | Feature 3 (Query API) | ~6000 |
| 5 | [1927](https://github.com/constructorfabric/cyberfabric-core/issues/1927) | Feature 4 (TimescaleDB plugin) | ~7700 |
| 6 | [1835](https://github.com/constructorfabric/cyberfabric-core/issues/1835) | e2e tests | ~1500 |
|   |  | ... to be continued ... | |

## Checklist
- [x] Self-review completed
- [x] No linting errors (`cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::perf`)
- [x] `cpt validate` — all checks passed
- [x] Tests pass (`cargo nextest run --workspace`)

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added usage collection infrastructure with REST API endpoints for recording and querying usage metrics
  * Introduced usage emission pipeline with authorization and validation for metric tracking
  * Added development/testing storage backend for usage record handling

* **Chores**
  * Registered new usage collection workspace modules and dependencies

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,6 +2663,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "cyberware-noop-usage-collector-plugin"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "chrono",
+ "cyberware-modkit",
+ "cyberware-modkit-macros",
+ "cyberware-modkit-security",
+ "cyberware-types-registry-sdk",
+ "cyberware-usage-collector-sdk",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
+ "uuid",
+]
+
+[[package]]
 name = "cyberware-oagw"
 version = "0.2.25"
 dependencies = [
@@ -3075,6 +3094,39 @@ dependencies = [
 ]
 
 [[package]]
+name = "cyberware-usage-collector"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "axum",
+ "base64 0.22.1",
+ "chrono",
+ "cyberware-authz-resolver-sdk",
+ "cyberware-modkit",
+ "cyberware-modkit-canonical-errors",
+ "cyberware-modkit-db",
+ "cyberware-modkit-macros",
+ "cyberware-modkit-odata",
+ "cyberware-modkit-security",
+ "cyberware-modkit-utils",
+ "cyberware-types-registry-sdk",
+ "cyberware-usage-collector-sdk",
+ "cyberware-usage-emitter",
+ "http",
+ "parking_lot",
+ "sea-orm-migration",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tokio",
+ "tower",
+ "tracing",
+ "utoipa",
+ "uuid",
+]
+
+[[package]]
 name = "cyberware-usage-collector-sdk"
 version = "0.1.0"
 dependencies = [
@@ -3091,6 +3143,26 @@ dependencies = [
  "serde_json",
  "tokio",
  "utoipa",
+ "uuid",
+]
+
+[[package]]
+name = "cyberware-usage-emitter"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "chrono",
+ "cyberware-authz-resolver-sdk",
+ "cyberware-modkit-canonical-errors",
+ "cyberware-modkit-db",
+ "cyberware-modkit-security",
+ "cyberware-usage-collector-sdk",
+ "sea-orm-migration",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
  "uuid",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,10 @@ members = [
     "modules/system/oagw/oagw-sdk",
     "modules/mini-chat/mini-chat-sdk",
     "modules/mini-chat/mini-chat",
+    "modules/system/usage-collector/plugins/noop-usage-collector-plugin",
+    "modules/system/usage-collector/usage-collector",
     "modules/system/usage-collector/usage-collector-sdk",
+    "modules/system/usage-collector/usage-emitter",
 ]
 exclude = ["tools/fuzz"]
 resolver = "3"
@@ -257,7 +260,9 @@ credstore-sdk = { package = "cyberware-credstore-sdk", version = "0.1.24", path 
 mini-chat-sdk = { package = "cyberware-mini-chat-sdk", version = "0.10.6", path = "modules/mini-chat/mini-chat-sdk" }
 
 # usage-collector
+usage-collector = { package = "cyberware-usage-collector", version = "0.1.0", path = "modules/system/usage-collector/usage-collector" }
 usage-collector-sdk = { package = "cyberware-usage-collector-sdk", version = "0.1.0", path = "modules/system/usage-collector/usage-collector-sdk" }
+usage-emitter = { package = "cyberware-usage-emitter", version = "0.1.0", path = "modules/system/usage-collector/usage-emitter" }
 
 # system modules
 grpc_hub = { package = "cyberware-grpc-hub", version = "0.2.6", path = "modules/system/grpc-hub" }
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/Cargo.toml
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/Cargo.toml
@@ -1,0 +1,51 @@
+[package]
+name = "cyberware-noop-usage-collector-plugin"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+description = "No-op usage-collector plugin for development and testing"
+readme = "README.md"
+keywords = ["cyberfabric", "cyberware", "cyberware-system"]
+categories = ["development-tools"]
+
+[lib]
+name = "noop_usage_collector_plugin"
+
+[lints]
+workspace = true
+
+[dependencies]
+# Local dependencies
+types-registry-sdk = { workspace = true }
+usage-collector-sdk = { workspace = true }
+
+# ModKit dependencies
+modkit = { workspace = true }
+modkit-macros = { workspace = true }
+
+# Async runtime
+async-trait = { workspace = true }
+
+# Error handling
+anyhow = { workspace = true }
+
+# Serialization
+serde = { workspace = true }
+serde_json = { workspace = true }
+
+# Logging
+tracing = { workspace = true }
+
+[dev-dependencies]
+# ModKit dependencies
+modkit-security = { workspace = true }
+
+# Data structures
+chrono = { workspace = true }
+uuid = { workspace = true }
+
+# Async runtime
+tokio = { workspace = true }
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/README.md
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/README.md
@@ -1,0 +1,35 @@
+# No-op Usage Collector Storage Plugin
+
+> **Development and testing plugin** — discards all usage records. It is not a production storage backend.
+
+No-op implementation of the usage-collector storage plugin contract: the gateway can resolve a plugin instance and call `create_usage_record`, but no data is retained.
+
+## Purpose
+
+Use this plugin when you want the usage-collector module and emitters to run without configuring a real storage implementation. Typical cases:
+
+- Local development and smoke tests
+- CI pipelines that exercise the ingestion path only
+- Demos where durable usage history is unnecessary
+
+**Do not use when you need metering data to be stored or queried.**
+
+## Configuration
+
+Add the plugin section under your module configuration:
+
+```yaml
+modules:
+  noop-usage-collector-plugin:
+    config:
+      vendor: "cyberfabric"
+      priority: 100
+```
+
+`vendor` must match the usage-collector gateway configuration so the gateway selects this instance.
+
+## Testing
+
+```bash
+cargo test -p cyberware-usage-collector-plugin
+```
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/config.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/config.rs
@@ -1,0 +1,28 @@
+//! Configuration for the no-op usage-collector storage plugin.
+
+use serde::Deserialize;
+
+/// Plugin configuration.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct NoopUsageCollectorConfig {
+    /// Vendor name (must match the usage-collector gateway `vendor`).
+    pub vendor: String,
+
+    /// Plugin priority (lower = higher priority when multiple instances exist).
+    pub priority: i16,
+}
+
+impl Default for NoopUsageCollectorConfig {
+    fn default() -> Self {
+        Self {
+            vendor: "cyberfabric".to_owned(),
+            priority: 100,
+        }
+    }
+}
+
+#[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
+#[path = "config_tests.rs"]
+mod config_tests;
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/config_tests.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/config_tests.rs
@@ -1,0 +1,31 @@
+use super::*;
+
+#[test]
+fn configured_values_override_serde_defaults() {
+    // Pair each field with a value distinct from its Default so the assertion fails if
+    // serde silently falls back to the Default impl instead of using the configured value.
+    let default_cfg = NoopUsageCollectorConfig::default();
+    assert_ne!(default_cfg.vendor, "acme");
+    assert_ne!(default_cfg.priority, 10);
+
+    let json = r#"{"vendor": "acme", "priority": 10}"#;
+    let cfg: NoopUsageCollectorConfig = serde_json::from_str(json).unwrap();
+    assert_eq!(cfg.vendor, "acme");
+    assert_eq!(cfg.priority, 10);
+}
+
+#[test]
+fn serde_default_applies_default_vendor_and_priority() {
+    let cfg: NoopUsageCollectorConfig = serde_json::from_str("{}").unwrap();
+    assert_eq!(
+        cfg.vendor, "cyberfabric",
+        "serde(default) must use Default impl"
+    );
+    assert_eq!(cfg.priority, 100, "serde(default) must use Default impl");
+}
+
+#[test]
+fn rejects_unknown_fields() {
+    let json = r#"{"vendor": "x", "priority": 1, "unexpected": true}"#;
+    assert!(serde_json::from_str::<NoopUsageCollectorConfig>(json).is_err());
+}
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/client.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/client.rs
@@ -1,0 +1,13 @@
+use async_trait::async_trait;
+
+use usage_collector_sdk::models::UsageRecord;
+use usage_collector_sdk::{UsageCollectorError, UsageCollectorPluginClientV1};
+
+use super::service::Service;
+
+#[async_trait]
+impl UsageCollectorPluginClientV1 for Service {
+    async fn create_usage_record(&self, _record: UsageRecord) -> Result<(), UsageCollectorError> {
+        Ok(())
+    }
+}
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/mod.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/mod.rs
@@ -1,0 +1,4 @@
+mod client;
+pub mod service;
+
+pub use service::Service;
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/service.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/service.rs
@@ -1,0 +1,7 @@
+//! No-op usage-collector storage domain service.
+
+use modkit_macros::domain_model;
+
+/// Stateless service backing the storage plugin client; records are not retained.
+#[domain_model]
+pub struct Service;
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/lib.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/lib.rs
@@ -1,0 +1,20 @@
+//! No-op Usage Collector Storage Plugin
+//!
+//! Registers a [`usage_collector_sdk::UsageCollectorPluginClientV1`] implementation that
+//! accepts `create_usage_record` calls and drops all data. Use when the usage-collector gateway should run
+//! end-to-end without a real storage backend.
+//!
+//! ## Configuration
+//!
+//! ```yaml
+//! modules:
+//!   noop_usage_collector_storage_plugin:
+//!     config:
+//!       vendor: "cyberfabric"
+//!       priority: 100
+//! ```
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+
+mod config;
+mod domain;
+mod module;
--- a/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/module.rs
+++ b/modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/module.rs
@@ -1,0 +1,73 @@
+//! No-op usage-collector storage plugin module.
+//!
+//! Registers a GTS plugin instance in the types registry and exposes
+//! [`usage_collector_sdk::UsageCollectorPluginClientV1`] under a scope keyed by that instance.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use modkit::Module;
+use modkit::client_hub::ClientScope;
+use modkit::context::ModuleCtx;
+use modkit::gts::BaseModkitPluginV1;
+use tracing::info;
+use types_registry_sdk::{RegisterResult, TypesRegistryClient};
+use usage_collector_sdk::{UsageCollectorPluginClientV1, UsageCollectorPluginSpecV1};
+
+use crate::config::NoopUsageCollectorConfig;
+use crate::domain::Service;
+
+/// No-op usage-collector storage plugin for development and testing.
+#[modkit::module(
+    name = "noop-usage-collector-plugin",
+    deps = ["types-registry", "usage-collector"]
+)]
+#[derive(Default)]
+struct NoopUsageCollectorPlugin;
+
+// @cpt-dod:cpt-cf-usage-collector-dod-sdk-and-ingest-core-noop-plugin:p1
+#[async_trait]
+impl Module for NoopUsageCollectorPlugin {
+    async fn init(&self, ctx: &ModuleCtx) -> anyhow::Result<()> {
+        let cfg: NoopUsageCollectorConfig = ctx.config_or_default()?;
+        info!(
+            %cfg.vendor,
+            cfg.priority,
+            "Loaded {} configuration",
+            Self::MODULE_NAME,
+        );
+
+        let instance_id = UsageCollectorPluginSpecV1::gts_make_instance_id(
+            "cf.core._.noop_usage_collector_storage_plugin.v1",
+        );
+
+        let registry = ctx.client_hub().get::<dyn TypesRegistryClient>()?;
+        let instance = BaseModkitPluginV1::<UsageCollectorPluginSpecV1> {
+            id: instance_id.clone(),
+            vendor: cfg.vendor.clone(),
+            priority: cfg.priority,
+            properties: UsageCollectorPluginSpecV1,
+        };
+        let instance_json = serde_json::to_value(&instance)?;
+
+        let results = registry.register(vec![instance_json]).await?;
+        RegisterResult::ensure_all_ok(&results)?;
+
+        let service = Service;
+
+        let api: Arc<dyn UsageCollectorPluginClientV1> = Arc::new(service);
+        ctx.client_hub()
+            .register_scoped::<dyn UsageCollectorPluginClientV1>(
+                ClientScope::gts_id(&instance_id),
+                api,
+            );
+
+        info!(
+            %instance_id,
+            "Registered {}",
+            Self::MODULE_NAME,
+        );
+
+        Ok(())
+    }
+}
--- a/modules/system/usage-collector/usage-collector/Cargo.toml
+++ b/modules/system/usage-collector/usage-collector/Cargo.toml
@@ -1,0 +1,78 @@
+[package]
+name = "cyberware-usage-collector"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+description = "Usage Collector gateway"
+readme = "README.md"
+keywords = ["cyberfabric", "cyberware", "cyberware-system"]
+categories = ["development-tools"]
+
+[lib]
+name = "usage_collector"
+
+[lints]
+workspace = true
+
+[dependencies]
+# Local dependencies
+authz-resolver-sdk = { workspace = true }
+types-registry-sdk = { workspace = true }
+usage-collector-sdk = { workspace = true }
+usage-emitter = { workspace = true }
+
+# ModKit dependencies
+modkit = { workspace = true }
+modkit-canonical-errors = { workspace = true }
+modkit-db = { workspace = true, features = ["preview-outbox"] }
+modkit-macros = { workspace = true }
+modkit-security = { workspace = true }
+modkit-utils = { workspace = true, features = ["humantime-serde"] }
+
+# REST
+axum = { workspace = true }
+http = { workspace = true }
+utoipa = { workspace = true }
+
+# Data structures
+chrono = { workspace = true }
+uuid = { workspace = true }
+
+# Async runtime
+async-trait = { workspace = true }
+tokio = { workspace = true }
+
+# Synchronous locking
+parking_lot = { workspace = true }
+
+# Error handling
+anyhow = { workspace = true }
+thiserror = { workspace = true }
+
+# Serialization
+base64 = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+
+# Logging
+tracing = { workspace = true }
+
+# DB
+sea-orm-migration = { workspace = true }
+
+[dev-dependencies]
+# Local dependencies
+types-registry-sdk = { workspace = true, features = ["test-util"] }
+
+# ModKit dependencies
+modkit-db = { workspace = true, features = ["sqlite"] }
+modkit-odata = { workspace = true }
+
+# HTTP testing
+tower = { workspace = true }
+
+# Async runtime
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
--- a/modules/system/usage-collector/usage-collector/README.md
+++ b/modules/system/usage-collector/usage-collector/README.md
@@ -1,0 +1,46 @@
+# Usage Collector
+
+> **Gateway module** — hosts the Usage Collector gateway in-process: registers the storage-plugin GTS schema with `types-registry`, resolves the active storage plugin by `vendor`, exposes `dyn UsageCollectorClientV1` in `ClientHub` for the outbox delivery path, and serves a REST API for out-of-process emitters.
+
+ModKit module `usage-collector`: central ingest for usage observations (`UsageRecord`) from the outbox pipeline (`usage-emitter`) and delegation to the selected `UsageCollectorPluginClientV1` implementation.
+
+## Overview
+
+At startup the module:
+
+- Registers `UsageCollectorPluginSpecV1` in the types registry so storage backends can be declared as GTS instances.
+- Builds the domain `Service` that lists plugin instances, picks one for the configured `vendor`, and forwards `create_usage_record` calls with a bounded timeout.
+- Registers **`UsageCollectorLocalClient`** as `dyn UsageCollectorClientV1` so code in the **same binary** can call `create_usage_record` and `get_module_config` (which filters the metrics whitelist to only the metrics allowed for the calling module) without a network hop.
+- Initializes the embedded **`UsageEmitterRuntime`** (from `usage-emitter`) and registers it as `dyn UsageEmitterRuntimeV1` in `ClientHub`; the runtime owns the outbox worker and vends module-scoped `UsageEmitterFactory` instances via `runtime.factory(module_name)`; each factory's `.with_*().authorize(subject)` chain produces a per-call `UsageEmitter` that handles PDP authorization and the outbox delivery path before forwarding records to the local client.
+
+Source modules should emit through **`usage-emitter`** (PDP + outbox). This crate implements the gateway side of `UsageCollectorClientV1::create_usage_record` and the storage-plugin selection logic.
+
+## Dependencies
+
+- **At least one storage plugin** — a module that registers `dyn UsageCollectorPluginClientV1` in `ClientHub` for the GTS instance id chosen for your `vendor`. For dev/tests, see **`cyberware-noop-usage-collector-plugin`**.
+
+## Configuration
+
+`vendor` must match the `vendor` field on the storage plugin GTS content so `choose_plugin_instance` selects the right instance. `plugin_timeout` bounds each `create_usage_record` call; exceeded waits become `UsageCollectorError::DeadlineExceeded`. `emitter` holds nested `UsageEmitterConfig` for outbox and authorization tuning. `metrics` is a whitelist of allowed metric names, each with a `kind` (`gauge` or `counter`) and an optional `modules` list (if absent, all modules may emit that metric). `max_metadata_bytes` caps the serialized size of `UsageRecord.metadata` (default `8192`, max `1_048_576`; `0` disables metadata entirely); the value is published via `get_module_config` and enforced by the emitter.
+
+```yaml
+modules:
+  usage-collector:
+    config:
+      vendor: "cyberfabric"
+      plugin_timeout: "5s"
+      emitter:
+        # UsageEmitterConfig fields (outbox, auth tuning)
+      metrics:
+        "storage.bytes_used":
+          kind: "gauge"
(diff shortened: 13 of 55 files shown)


il10241024:
**coderabbitai[bot]** commented on 2026-05-15T11:27:03Z:

---




> [!IMPORTANT]
> ## Review skipped
> 
> Auto reviews are disabled on base/target branches other than the default branch.
> 
> 
> 
> 
> <details>
> 
> 
> **Review profile**: CHILL
> 
> **Plan**: Free
> 
> **Run ID**: `3006333e-3b9e-42b2-b365-2398d136c5d8`
> 
> </details>
> 
> 
> - [ ]  🔍 Trigger review





---

> [!NOTE]
> <details>
> <summary>🎁 Summarized by CodeRabbit Free</summary>
> 
> Your organization is on the Free plan. CodeRabbit will generate a high-level summary and a walkthrough for each pull request. For a comprehensive line-by-line review, please upgrade your subscription to CodeRabbit Pro by visiting <https://app.coderabbit.ai/login>.
> 
> </details>


<sub>Comment `@coderabbitai help` to get the list of available commands and usage tips.</sub>

il10241024:
**qodo-code-review[bot]** commented on 2026-05-15T11:34:50Z:

---

### Qodo reviews are paused for this user.

Troubleshooting steps vary by plan [Learn more →](https://docs.qodo.ai/review-eligibility)


**On a Teams plan?**
Reviews resume once this user has a paid seat *and* their Git account is linked in Qodo.
[Link Git account →](https://app.qodo.ai)

**Using GitHub Enterprise Server, GitLab Self-Managed, or Bitbucket Data Center?**
These require an Enterprise plan - Contact us
[Contact us →](https://app.qodo.ai)

il10241024:
**qodo-code-review[bot]** commented on 2026-05-15T16:38:11Z:

---

### Qodo reviews are paused for this user.

Troubleshooting steps vary by plan [Learn more →](https://docs.qodo.ai/review-eligibility)


**On a Teams plan?**
Reviews resume once this user has a paid seat *and* their Git account is linked in Qodo.
[Link Git account →](https://app.qodo.ai)

**Using GitHub Enterprise Server, GitLab Self-Managed, or Bitbucket Data Center?**
These require an Enterprise plan - Contact us
[Contact us →](https://app.qodo.ai)

il10241024:
**capybutler** commented on 2026-05-27T09:57:19Z:

---

The approach to usage collector is revised in #3433

il10241024:
**Artifizer** submitted review **COMMENTED** on 2026-05-19T21:34:01Z:

---



---

il10241024:
**Artifizer** submitted review **COMMENTED** on 2026-05-20T08:38:17Z:

---



---

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/usage_builder.rs` line 119 on 2026-05-15T11:38:29Z:

---

&#2369;MikeFalcon77 commented:

> HIGH
> 
> Doc on with_idempotency_key (line 43) states "If omitted, a new UUID string is used on enqueue", but unwrap_or_default() produces an empty String, not a UUID.
> 
> For a Counter without an idempotency key the emitted record has idempotency_key = "", which validate_counter_idempotency_key then rejects with InvalidRecord. Callers relying on the documented contract will see unexpected 422s.
> 
> Either generate a UUID here when kind == Counter && idempotency_key.is_none() (matching the doc), or update the docstring on line 43 and on dto.rs:30 to say the key is required for counters.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/scoped_emitter.rs` line 138 on 2026-05-15T11:39:52Z:

---

&#2369;MikeFalcon77 commented:

> **HIGH**
> 
> `Err(e) => return Err(UsageEmitterError::internal(e.to_string()))` collapses every non-`ModuleNotFound` `UsageCollectorError` (`PluginTimeout`, `CircuitOpen`, `Unavailable`) into a stringified `Internal` error.
> 
> For source modules using `UsageEmitterV1` over the REST client, a transient gateway outage becomes an opaque 500 instead of being mappable to 503/504. The categorization the collector deliberately exposes is lost in transit.
> 
> Match the relevant variants and surface them as distinct emitter errors (e.g. an `Unavailable`/`Timeout` variant on `UsageEmitterError`), or at minimum preserve the source via `#[from]` on a new variant.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/routes.rs` line 46 on 2026-05-15T11:40:44Z:

---

&#2369;MikeFalcon77 commented:

> **HIGH**
> 
> `.json_response(StatusCode::NO_CONTENT, ...)` advertises an `application/json` body for HTTP 204, but the handler returns `StatusCode::NO_CONTENT` with no body. RFC 7230 §3.3.3 forbids a body on 204, and the OpenAPI doc will tell generated clients to expect JSON.
> 
> `OperationBuilder` exposes `no_content_response(...)` precisely for this case (see `libs/modkit/src/api/operation_builder.rs:1084`).
> 
> Replace the call with `.no_content_response(http::StatusCode::NO_CONTENT, "Record accepted and stored")`.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/emitter.rs` line 102 on 2026-05-15T11:41:45Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> `EnforcerError::CompileFailed(_)` and `EvaluationFailed(_)` discard the inner error, leaving the operator with only a fixed string.
> 
> When PDP misbehaves in production, logs show "authorization constraint compilation failed" with no clue what failed. That defeats the point of having structured PDP errors.
> 
> Format the inner cause into the message, e.g. `UsageEmitterError::internal(format!("authorization constraint compilation failed: {e}"))` for both arms; or add a dedicated variant that preserves `#[source]`.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 53 on 2026-05-15T11:42:15Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> If `serde_json::to_vec(meta)` fails, `map_or(0, |v| v.len())` returns 0 and the metadata silently passes the size check. The same metadata is then re-serialized in `AuthorizedUsageEmitter::validate_metadata_size`, which surfaces a generic `Internal` error rather than the more accurate "metadata too large / unserializable" 422.
> 
> Replace with an explicit match: on serialization error return a 422 `Problem` here ("Invalid metadata"), so the failure path is consistent and never bypasses validation.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 77 on 2026-05-15T11:42:41Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> The comment claims "No subject — skip PDP authorization", but the code still calls `authorize_for(...)` which always invokes the PDP via `PolicyEnforcer::access_scope_with` (see `scoped_emitter.rs:105`). Only the `subject_id`/`subject_type` resource properties are omitted from the request.
> 
> The misleading comment will lead future readers to believe this branch is unauthorized, which it is not.
> 
> Rewrite as: `// No subject context — call PDP without subject_id/subject_type properties.`

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/infra/delivery_handler.rs` line 62 on 2026-05-15T11:43:29Z:

---

&#2369;MikeFalcon77 commented:

> **MEDIUM**
> 
> `if let Err(...) { return ...; } let record = record_result.unwrap();` is unidiomatic and forces a `#[allow(clippy::unwrap_used)]` and a SAFETY comment to explain why the unwrap is sound.
> 
> Replace the block with a direct `match` that binds in the `Ok` arm:
> 
> ```rust
> let record = match serde_json::from_slice::<UsageRecord>(&msg.payload) {
>     Ok(r) => r,
>     Err(err) => {
>         warn!(msg.seq, msg.partition_id, %err, "usage record deserialization failed");
>         return MessageResult::Reject(err.to_string());
>     }
> };
> ```
> 
> Drop the `#[allow]` and the SAFETY note.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/usage_builder_tests.rs` line 257 on 2026-05-15T11:45:19Z:

---

&#2369;MikeFalcon77 commented:

> MEDIUM
> 
> The test name claims "uses_uuid_fallback", but the assertion (line 351-354) is assert!(record.idempotency_key.is_empty(), ...) — it verifies the opposite (empty string, not UUID).
> 
> This indicates either the production behaviour regressed without the test being renamed, or the test was written against an aspirational contract that was never implemented. Either way the name lies about what is verified.
> 
> If the empty-string behaviour is intentional, rename to gauge_with_blank_idempotency_key_stays_blank. If a UUID fallback is the intended contract (per usage_builder.rs:43 doc), fix usage_builder.rs:116-120 and update the assertion accordingly.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/error_tests.rs` line 8 on 2026-05-15T11:46:04Z:

---

&#2369;MikeFalcon77 commented:

> LOW
> 
> The display_* tests in this file (lines 8-57) only snapshot the #[error(...)] format strings produced by thiserror. They verify framework-generated Display output, not domain behavior — a contract no consumer relies on. Renaming the message strings would require lock-step test updates with zero defensive value (snapshot abuse, anti-pattern https://github.com/constructorfabric/gears-rust/issues/2240 in the test guidelines).
> 
> Delete the display_* tests, or replace them with a single test that asserts the message includes the dynamic field (metric, value, len, etc.) when that field is the diagnostic contract.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/config_tests.rs` line 7 on 2026-05-15T11:46:44Z:

---

&#2369;MikeFalcon77 commented:

> LOW
> 
> default_authorization_max_age_is_30_seconds, default_outbox_queue_is_usage_records, default_outbox_partition_count_is_4, and default_outbox_backoff_max_is_10_minutes (lines 5-21, 78-82) only read back constants from the Default impl. They are constructor-echo tests (anti-pattern https://github.com/constructorfabric/gears-rust/issues/2247 in the test guidelines): tightening or loosening any default to a different reasonable value breaks them with no behavior actually broken.
> 
> Delete them. The remaining validate_* and serde tests in this file already cover the meaningful contracts (validation bounds, serde compatibility).

il10241024:
**capybutler** reviewed `modules/system/usage-collector/plugins/noop-usage-collector-storage-plugin/src/domain/client_tests.rs` line 34 on 2026-05-15T12:00:14Z:

---

&#2369;MikeFalcon77 commented:

> LOW
> 
> create_usage_record_multiple_records_all_return_ok adds no coverage over create_usage_record_always_returns_ok: the noop service is stateless, so calling the same Ok-returning function in a loop verifies nothing new (anti-pattern https://github.com/constructorfabric/gears-rust/issues/2243: redundant test).
> 
> Delete this test.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/usage_builder.rs` line 119 on 2026-05-15T12:19:26Z:

---

The feedback is fully resolved. The old `usage_builder.rs:119` no longer exists; the builder is now `usage-emitter/src/domain/usage_record_builder.rs` and the contract has been split kind-aware:

  ## Documented contract (the part that was wrong)

  - `with_idempotency_key` doc at `usage_record_builder.rs:110-115`:
    > Required (non-empty) for counter metrics — enforced by `UsageEmitter::enqueue_in`. Optional for gauge metrics; when omitted or blank, the emitter generates a UUID at enqueue time.
  - Wire DTO at `dto.rs:28-30` mirrors this: "Required for counter metrics (non-empty); optional for gauge — when omitted for a gauge, the gateway generates a UUID … Counter records without a key are rejected
  with `422 Unprocessable Entity`."

  So the doc no longer overpromises a UUID for every record.

  ## Behavior (the part that has to match)

  `UsageRecordBuilder::build` still uses `idempotency_key.unwrap_or_default()` (`usage_record_builder.rs:186`) and emits `""` as a sentinel. The kind-aware logic lives one layer down in
  `UsageEmitter::enqueue_in` (`emitter.rs:60-119`), in this specific order:

  1. `validate_counter_idempotency_key` (`emitter.rs:235-246`) — if `kind == Counter` and the key is blank/whitespace, reject with `InvalidArgument` ("counter records require a non-empty idempotency_key"). 422
   here is now the *documented* contract, not a surprise.
  2. `generate_gauge_idempotency_key` (`emitter.rs:276-282`) — if `kind == Gauge` and the key is blank/whitespace, set it to `Uuid::new_v4().to_string()`.

  The order matters: the comment at `emitter.rs:77-80` calls it out — generate runs *after* validate, so counter records can't be silently rescued by a generated UUID.

  ## Net effect

  Both cases the reviewer flagged are correct now:

  - A gauge without a key gets a UUID (matches the doc).
  - A counter without a key gets the canonical 422 (also matches the doc, on both the builder side and the REST DTO side).

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/scoped_emitter.rs` line 138 on 2026-05-15T12:27:52Z:

---

The feedback is resolved. Here's how:

## The original problem (commit `daf14524`, `scoped_emitter.rs:138`)

```rust
Err(e) => return Err(UsageEmitterError::internal(e.to_string())),
```

Every non-`ModuleNotFound` `UsageCollectorError` variant got stringified into a generic Internal error, so `DeadlineExceeded` / `ServiceUnavailable` / `ResourceExhausted` etc. were no longer distinguishable to
callers.

The refactor

1. `UsageEmitterError` is now a type alias for `CanonicalError` (`usage-emitter/src/error.rs:16`), the same canonical taxonomy that `UsageCollectorError` already uses (`usage-collector-sdk/src/error.rs:3`).
2. In `domain/factory.rs:187-202`, the old match-on-`ModuleNotFound`/stringify-everything-else block is gone. The collector call is now:

```rust
let Ok(module_cfg_result) = timeout(
    self.config.authorize_call_timeout,
    self.collector.get_module_config(&self.module),
).await else {
    return Err(UsageRecordError::deadline_exceeded(...).create());
};
let module_cfg = module_cfg_result?;
```

2. The ? propagates the collector error verbatim — same enum, same variant, no string collapse. `DeadlineExceeded`, `ServiceUnavailable`, `ResourceExhausted`, `NotFound` (module-not-found), `PermissionDenied`, etc.
all reach the caller intact and remain mappable to REST 503/504/etc.
3. Coverage is locked in by factory_tests.rs:
  - `authorize_propagates_collector_deadline_exceeded` (line 421) asserts `Err(UsageEmitterError::DeadlineExceeded { .. })`.
  - `authorize_propagates_collector_service_unavailable` (line 443) asserts `Err(UsageEmitterError::ServiceUnavailable { .. })`.
  - `authorize_propagates_collector_resource_exhausted` (line 465) for the rate-limit variant.

## Summary

So the feedback — "preserve the source / surface them as distinct variants" — is satisfied structurally: unifying both crates on `CanonicalError` makes the propagation lossless by construction rather than
needing a per-variant `#[from]` mapping.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/routes.rs` line 46 on 2026-05-15T12:32:28Z:

---

Resolved in the latest revision of this branch.

`modules/system/usage-collector/usage-collector/src/api/rest/routes.rs:49` now reads:

```rust
.no_content_response(http::StatusCode::NO_CONTENT, "Record accepted and stored")
```

instead of the original `.json_response(http::StatusCode::NO_CONTENT, "Record accepted and stored")`. The handler at `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs:59` still returns `Ok(StatusCode::NO_CONTENT)` with no body, so the wire behavior and the OpenAPI doc are now consistent: clients are told to expect an empty 204, matching RFC 7230 §3.3.3.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/emitter.rs` line 102 on 2026-05-15T12:40:51Z:

---

**Status:** Resolved.

The mapping moved from `emitter.rs` into a dedicated helper at
`modules/system/usage-collector/usage-emitter/src/error.rs:24`
(`enforcer_error_to_emitter_error`). Both arms now bind the inner
`source` and log it via `tracing::error!` before returning a
fail-closed canonical error:

```rust
EnforcerError::CompileFailed(source) => {
    error!(
        pdp_error_variant = "CompileFailed",
        error = %source,
        "PDP constraint compilation failed; access denied (fail-closed)",
    );
    UsageRecordError::permission_denied()
        .with_reason("AUTHORIZATION_DENIED")
        .create()
}
// EvaluationFailed handled the same way.
```

The reviewer's concern — *"logs show a fixed string with no clue what
failed"* — is addressed: every infra failure now produces a structured
log line carrying the variant tag and the full `Display` of the
underlying `ConstraintCompileError` / `AuthZResolverError`.

**Deliberate divergence:** the cause is not folded into the returned
error (no `format!("...: {e}")`, no `#[source]`). All three
`EnforcerError` outcomes collapse to `PermissionDenied` with
`AUTHORIZATION_DENIED` to keep the fail-closed posture — a PDP infra
fault must be indistinguishable from a policy denial at the API
boundary. Diagnosability lives in logs; the returned error stays safe.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 53 on 2026-05-15T12:45:37Z:

---

Resolved by the refactor — the handler-side pre-check that this feedback targeted no longer exists.

**What changed**

The handler (`modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs`) no longer calls a local `validate_metadata_size` helper. It just attaches metadata to the builder and lets the emitter validate at enqueue time, with errors mapped through one canonical path.

```rust
if let Some(meta) = req.metadata {
    builder = builder.with_metadata(meta);
}
let record = builder.build().map_err(|e| canonical_error_to_problem(&e))?;
emitter.enqueue(record).await.map_err(|e| canonical_error_to_problem(&e))?;
```

**Why the original concern is gone**

1. **No more `map_or(0, |v| v.len())`.** The lossy `to_vec(...).map_or(0, ...)` line was deleted; there is no path that treats a serialization failure as 0 bytes anymore.
2. **Single source of truth for size validation.** `UsageEmitter::validate_metadata_size` in `usage-emitter/src/domain/emitter.rs:248` is now the only place this check runs. Serialization failure there is explicit — it produces `UsageRecordError::invalid_argument()` with constraint `"metadata is not serializable: {e}"`, not a generic `Internal`:
   ```rust
   let len = serde_json::to_vec(metadata)
       .map_err(|e| UsageRecordError::invalid_argument()
           .with_constraint(format!("metadata is not serializable: {e}"))
           .create())?
       .len();
   ```
3. **Mapped to a proper client error, not 500.** `canonical_error_to_problem` routes `InvalidArgument` through `modkit_canonical_errors::Problem::from_error`, where `InvalidArgument` → HTTP 400 with the canonical title/detail/context preserved as RFC 9457 extension members. So an unserializable metadata payload surfaces as a structured client error, not a generic `Internal`.
4. **No double-serialization risk.** Because the handler no longer serializes metadata to size-check it, the value is serialized exactly once (inside the emitter), so there is no second call that could disagree with the first.

Net result: the failure mode the reviewer flagged (silent bypass → re-serialize → generic `Internal`) is structurally impossible in the current code.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 77 on 2026-05-15T12:48:50Z:

---

Resolved in the refactor — the misleading comment is gone and the code now makes it structurally obvious that PDP is always invoked.

`modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs:63-82` (`authorize_request`) is now a single, unconditional path:

```rust
let factory = runtime.factory(&req.module).with_tenant(req.tenant_id);
let factory = match req.subject.clone() {
    Some(s) => factory.with_subject(s),
    None    => factory.without_subject(),
};
factory.authorize(ctx, req.resource_id, &req.resource_type).await
```

Key points:

- `factory.authorize(...)` is called on every branch — there is no "skip PDP" path anymore. The subject choice only toggles whether `subject_id` / `subject_type` are included as resource properties in the PDP request, exactly as the original feedback requested.
- The new comment above the `match` (lines 68–72) explicitly states that `None` is forwarded as the caller's explicit "no subject" intent — not a default-from-context fallback — and warns that substituting the gateway's own `SecurityContext` subject would silently misattribute the principal at the PDP boundary.
- The three-state contract (`Explicit(Some)` / `Explicit(None)` / `DefaultFromCtx`) is documented on `SubjectChoice` in `usage-emitter/src/domain/factory.rs` and covered by tests in `factory_tests.rs` (`authorize_without_subject_omits_subject_properties_from_pdp_request`, `authorize_pdp_request_includes_subject_id_and_subject_type_resource_properties`, `authorize_default_subject_choice_falls_back_to_ctx_subject_in_pdp`).

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/infra/delivery_handler.rs` line 62 on 2026-05-15T12:50:41Z:

---

Resolved in the current branch (`modules/system/usage-collector/usage-emitter/src/infra/delivery_handler.rs:52-66`).

The `if let Err(...) { return ...; } let record = record_result.unwrap();` block has been replaced with a direct `match` that binds in the `Ok` arm, exactly as suggested:

```rust
let record = match serde_json::from_slice::<UsageRecord>(&msg.payload) {
    Ok(r) => r,
    Err(err) => {
        warn!(
            msg.seq,
            msg.partition_id,
            %err,
            "usage record deserialization failed"
        );
        return MessageResult::Reject(format!("{:#}", anyhow::Error::new(err)));
    }
};
```

- No `unwrap()` call remains.
- `#[allow(clippy::unwrap_used)]` is gone.
- The `// SAFETY:` comment justifying the unwrap is gone.

Verified via `grep -nE "unwrap|allow\(clippy|SAFETY"` on the file — no matches.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/usage_builder_tests.rs` line 257 on 2026-05-15T12:52:41Z:

---

Resolved in the follow-up refactor. The aspirational UUID-fallback contract is now actually implemented, and the test name matches the assertion.

**What changed**

- The file `usage_builder_tests.rs` was renamed to `domain/usage_record_builder_tests.rs`. The misleading test was removed and replaced by `gauge_with_blank_idempotency_key_gets_uuid_at_enqueue` (`domain/usage_record_builder_tests.rs:416`).
- The new test goes through the full enqueue path (real outbox + sqlite + capture handler), then asserts:
  - `!record.idempotency_key.is_empty()` (`:505`)
  - `Uuid::parse_str(&record.idempotency_key).is_ok()` (`:509`)
- The production contract from the doc comment is now implemented in the emitter, not the builder, which is the right boundary (build-time records can still be blank; enqueue fills in the UUID):
  - `domain/emitter.rs:81` — `Self::generate_gauge_idempotency_key(&mut record);` called from `enqueue_in`
  - `domain/emitter.rs:276-280` — generates `Uuid::new_v4().to_string()` when `kind == Gauge && idempotency_key.trim().is_empty()`
  - `domain/emitter.rs:235-240` — counters with blank keys are rejected with `InvalidArgument`
- The doc comment on `with_idempotency_key` (`domain/usage_record_builder.rs:110-115`) accurately states the contract: required for counters, generated at enqueue for blank gauges.
- The remaining `is_empty()` assertion at `domain/usage_record_builder_tests.rs:132` is inside `build_succeeds_with_all_required_fields` — its name is honest, and it correctly verifies that `build()` itself does *not* fabricate a key (that is the emitter's job).

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/error_tests.rs` line 8 on 2026-05-15T12:59:35Z:

---

**Resolved.**

The entire `modules/system/usage-collector/usage-emitter/src/error_tests.rs` file (all seven `display_*` snapshot tests) has been deleted on this branch. `error.rs` no longer carries a `#[cfg(test)]` block, and `grep` for any of the removed test names across the crate returns nothing — so no equivalent assertions were re-introduced elsewhere.

The only remaining `err.to_string()` assertions in the crate are in `config_tests.rs`, which use `assert!(err.to_string().contains("<field>"))` to verify the dynamic field name surfaces in the message — i.e. exactly the "assert the message includes the dynamic field" pattern you suggested as the acceptable alternative, not full-string snapshots of thiserror output.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/usage-emitter/src/config_tests.rs` line 7 on 2026-05-15T13:08:15Z:

---

Resolved.

All four constructor-echo tests have been deleted from `modules/system/usage-collector/usage-emitter/src/config_tests.rs`:

- `default_authorization_max_age_is_30_seconds`
- `default_outbox_queue_is_usage_records`
- `default_outbox_partition_count_is_4`
- `default_outbox_backoff_max_is_10_minutes`

The file now contains only `validate_*` tests that exercise the meaningful contracts (validation bounds for `outbox_queue`, `outbox_partition_count`, `authorization_max_age`, `outbox_backoff_max`, and `authorize_call_timeout`), plus `validate_accepts_defaults` which asserts the behavioral invariant "defaults must pass validation" rather than echoing specific constants. Changing any default to another reasonable value no longer breaks the test suite.

il10241024:
**capybutler** reviewed `modules/system/usage-collector/plugins/noop-usage-collector-storage-plugin/src/domain/client_tests.rs` line 34 on 2026-05-15T13:10:16Z:

---

Resolved.

`create_usage_record_multiple_records_all_return_ok` has been deleted. `modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/domain/client_tests.rs` now contains only `create_usage_record_always_returns_ok` — the loop variant that exercised the same stateless `Ok`-returning path is gone.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 176 on 2026-05-19T21:48:02Z:

---

**HIGH**

`canonical_error_to_problem` reinvents the canonical→`Problem` bridge and bypasses the documented migration prelude.

The standard pattern in the codebase (see `modules/file-parser/src/api/rest/error.rs:60`, `modules/system/types-registry/.../api/rest/error.rs:65`, `modules/system/nodes-registry/.../api/rest/error.rs:56`, `modules/simple-user-settings/.../api/rest/error.rs:52`) is:

```rust
use modkit::api::canonical_prelude::*;  // Problem == modkit_canonical_errors::Problem, ApiResult == Result<T, Problem>

impl From<DomainError> for Problem {
    fn from(err: DomainError) -> Self {
        Problem::from(CanonicalError::from(err)).with_temporary_request_context("/")
    }
}
```

Handlers then return `ApiResult<T>` and use `?` on `CanonicalError` / `DomainError` directly — no per-call `map_err(|e| canonical_error_to_problem(&e))?` boilerplate, and the result already carries `trace_id` / `instance` via `with_temporary_request_context`, which is what the canonical error middleware will key off when it lands (`libs/modkit/src/api/canonical_trace.rs:5-13`).

This file instead imports the legacy `modkit::api::problem::Problem`, hand-rolls the canonical→legacy field copy, and skips `with_temporary_request_context`, so:
- every handler gets a 6-line `map_err(|e| canonical_error_to_problem(&e))?` wrapper repeated 4× in `handle_create_usage_record`
- responses ship without `trace_id` / `instance` — operators lose request-correlation for every error path
- when the canonical error middleware (`cpt-cf-errors-component-error-middleware`) lands, this module won't pick it up automatically; the bespoke helper has to be deleted by hand

Replace `use modkit::api::problem::Problem` with `use modkit::api::canonical_prelude::*`, return `ApiResult<StatusCode>` from the handler, add `impl From<DomainError> for Problem` (the canonical one) using `Problem::from(CanonicalError::from(err)).with_temporary_request_context("/")`, and drop `canonical_error_to_problem` plus the four `map_err` wrappers.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/dto.rs` line 20 on 2026-05-19T21:48:02Z:

---

**MEDIUM**

Doc on `subject` says "PDP subject validation is skipped in that case", but the PDP is still invoked on the `None` branch — only the `SUBJECT_ID` / `SUBJECT_TYPE` resource properties are omitted from the request (see `handlers.rs:90-97` → `factory.without_subject().authorize(...)` → `factory.rs:188-193`).

This is the same misleading wording resolved in prior-review comment #3247933305 (where it lived on `handlers.rs:63`) — it migrated into the DTO docstring instead of being removed. Future readers (and generated OpenAPI consumers, since this DTO renders into the schema) will conclude that `subject: null` means "no authorization at all" rather than "authorize without subject properties".

Replace with: `\`None\` means the request is authorized without \`SUBJECT_ID\` / \`SUBJECT_TYPE\` resource properties on the PDP request; tenant / resource / module checks still run.`

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-emitter/src/domain/emitter.rs` line 62 on 2026-05-19T21:48:02Z:

---

**MEDIUM**

`db.conn()` failure is mapped to `Internal`, but pool/connection acquisition failure is a transient infrastructure condition, not a code defect.

The consequence is the same retry-classification mismatch the rest of this PR is careful about: in `delivery_handler.rs:88-101`, `Internal` is dead-lettered (permanent) while `ServiceUnavailable` is retried. A pool-exhausted enqueue from an in-process source module surfaces to the caller as a 500 (RUST-ERR-001) and, when the caller is the gateway forwarding via REST, the outbox worker will dead-letter the resulting `Internal` instead of retrying — exactly the failure mode the canonical-variant alignment work in this PR is designed to prevent.

Return `UsageEmitterError::service_unavailable().with_detail(format!("db.conn() failed: {:#}", anyhow::Error::new(e))).create()` so transient pool errors stay classified as transient throughout the pipeline.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-emitter/src/lib.rs` line 22 on 2026-05-19T21:48:02Z:

---

**MEDIUM**

This docstring (and the matching one in `usage-collector/src/lib.rs:21`, plus the gateway `README.md` cross-reference) enshrines a design where a separate ModKit module (`usage-collector-rest-client`) exists solely to register `dyn UsageCollectorClientV1` over HTTP.

A standalone ModKit module per transport multiplies registration surface (one `init()` per transport, one set of config keys, one set of deps) for a type that the framework intentionally hides behind `dyn UsageCollectorClientV1`. Move the REST client implementation into `usage-collector-sdk` and let the existing host module (or a thin in-process wiring helper) pick local-vs-remote, mirroring the `wire_client` / `build_client` pattern in `09_oop_grpc_sdk_pattern.md`.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector/tests/common/mod.rs` line 394 on 2026-05-19T21:48:02Z:

---

**LOW**

`AppHarness::build` mounts handlers via raw `Router::new().route(...).layer(Extension(...))` instead of calling `crate::api::rest::routes::register_routes`. That bypasses every effect of `OperationBuilder.register(...)`:
- response-shape registration (`.no_content_response` vs `.json_response` for 204 — the exact bug Mike flagged in prior comment #3247921402 would have passed silently through these tests),
- `.allow_content_types(["application/json"])` enforcement,
- `.require_license_features::<License>([])`,
- `.authenticated()` middleware and `Extension<SecurityContext>` injection,
- the OpenAPI registry receiving the routes.

These tests are labelled "integration" (`tests/create_record_tests.rs`) but only exercise handler bodies. Replace the manual `Router::new().route(...)` chain with `let openapi = …; let router = register_routes(Router::new(), &openapi, runtime, service);` and inject `Extension(SecurityContext)` via the existing layer so the production wiring is what's under test.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector/tests/common/mod.rs` line 396 on 2026-05-19T21:48:02Z:

---

**LOW**

`.layer(Extension(authz))` adds an `Extension<Arc<dyn AuthZResolverClient>>` that no handler in this module extracts (PDP access goes through `runtime` → `UsageEmitterRuntime` → `PolicyEnforcer`, which holds its own `Arc<dyn AuthZResolverClient>` set in `UsageEmitterRuntime::build`).

The extension is dead weight, kept alive solely by the `AppHarness::build` signature taking `authz: Arc<dyn AuthZResolverClient>`. Drop the `authz` parameter and the `.layer(Extension(authz))` call, or extract it in `handle_create_usage_record` if it was supposed to be used.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-collector/src/config_tests.rs` line 148 on 2026-05-19T21:48:02Z:

---

**LOW**

`serde_default_applies_default_vendor` asserts `cfg.vendor == "cyberfabric"` (a literal copy of the value in `Default::default()` at `config.rs:119`). `serde_default_applies_default_plugin_timeout` does the same for `Duration::from_secs(5)`.

This is the constructor-echo anti-pattern Mike flagged in prior comment #3247953687 against `usage-emitter/src/config_tests.rs` (where four equivalent `default_*_is_X` tests were deleted in the follow-up refactor). The same family is alive here — changing either default to another reasonable value would break these without breaking any real contract.

If the intent is to pin the `#[serde(default)]` wiring (i.e. "missing field inherits from `Default`"), assert `cfg.vendor == UsageCollectorConfig::default().vendor` and `cfg.plugin_timeout == UsageCollectorConfig::default().plugin_timeout` instead — exactly the pattern used at `:196-201` for `max_metadata_bytes`. Otherwise delete both tests.

il10241024:
**MikeFalcon77** reviewed `modules/system/usage-collector/usage-emitter/src/error.rs` line 65 on 2026-05-19T21:48:02Z:

---

**LOW**

`enforcer_error_to_emitter_error` accepts the `EnforcerError` by value (`#[allow(clippy::needless_pass_by_value)]`) and the three arms collapse all PDP outcomes — including infra failures (`CompileFailed`, `EvaluationFailed`) — to `permission_denied("AUTHORIZATION_DENIED")`. The original `DenyReason.error_code` is preserved via `map_denied`, but infra failures lose all variant identity at the canonical-error layer (only the structured `tracing::error!` carries it).

The fail-closed posture is justified in the comment, but `permission_denied("AUTHORIZATION_DENIED")` for `EnforcerError::CompileFailed` means a misconfigured policy and a legitimate `Denied { deny_reason: None }` are indistinguishable at the API boundary *and* in the logs `reason` field. If an operator has to triage a spike of `AUTHORIZATION_DENIED` errors, they can't tell from the metric / Problem `reason` whether it's a policy outage or a real deny — they have to grep `pdp_error_variant` in `error!` traces.

Use a stable distinct reason code (e.g. `"PDP_INFRA_FAILURE"` for the two non-`Denied` arms) so the canonical Problem's `reason` extension differentiates infra-failure from policy-deny. Status stays 403; only the `reason` string changes, keeping the fail-closed posture intact.

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/module.rs` line 88 on 2026-05-20T08:36:59Z:

---

**CRITICAL**

`UsageCollectorLocalClient` is constructed here but never registered in `ClientHub` as `dyn UsageCollectorClientV1`.

Any module in the same binary that tries to resolve `dyn UsageCollectorClientV1` from the hub will get a runtime "not found" — despite the README advertising this as a supported in-process usage pattern.

Add `ctx.client_hub().register::<dyn UsageCollectorClientV1>(Arc::clone(&collector))` after `Arc::new(collector)` and before passing it into `UsageEmitterRuntime::build`.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/module.rs` line 33 on 2026-05-20T08:37:02Z:

---

**MEDIUM**

`OnceLock<Arc<Service>>` on the module struct is used to pass state from `init()` to `register_rest()` — that's non-idiomatic for ModKit and prevents clean re-initialization.

The canonical pattern is to register the service in `ClientHub` under an internal key during `init()` and retrieve it in `register_rest()` via `ctx.client_hub()`. The `OnceLock` also panics silently on a second `init()` call rather than producing a meaningful lifecycle error.

Register the service via the hub or a typed internal client, and drop the `OnceLock` field.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-emitter/src/domain/factory.rs` line 169 on 2026-05-20T08:37:06Z:

---

**HIGH**

`BarrierMode::Ignore` is applied to every PDP request unconditionally — there's no opt-in for `BarrierMode::Respect`.

If a source module incorrectly gains access to `dyn UsageEmitterRuntimeV1`, it gets implicit cross-barrier emit capability it was never supposed to have. The whole security model here rests on PDP policy alone, with no defense-in-depth at the framework layer.

Expose `BarrierMode` as a config field on `UsageEmitterFactory` (defaulting to `Ignore` for platform modules), so the widened barrier is an explicit choice rather than an invisible invariant.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-emitter/src/api.rs` line 28 on 2026-05-20T08:37:09Z:

---

**HIGH**

`UsageEmitterRuntimeV1::factory()` returns a concrete `UsageEmitterFactory`, not a trait object.

Any crate that wants to mock `UsageEmitterRuntimeV1` in tests must hold a real `UsageEmitterRuntime` backed by in-memory SQLite — dragging `modkit-db` and `sea-orm-migration` into test dependencies of every consumer. In practice this makes the trait non-mockable in isolation.

Define a `UsageEmitterFactoryV1` trait and return `Box<dyn UsageEmitterFactoryV1>` (or an associated type). Test stubs can then provide a lightweight in-memory fake without touching the DB stack.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-emitter/src/domain/emitter.rs` line 312 on 2026-05-20T08:37:12Z:

---

**HIGH**

`validate_metadata_size` serializes `metadata` to a `Vec<u8>` purely to measure `len()` — then the same value gets serialized again when building the outbox payload a few lines later.

On a high-frequency gauge path this is two allocations per call where one would do.

Serialize once, capture the `Vec<u8>`, check `.len()` against the limit, then pass the pre-serialized bytes into the outbox rather than serializing again downstream.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/module.rs` line 88 on 2026-05-20T08:37:21Z:

---

**MEDIUM**

`UsageEmitterRuntimeV1` is a trait defined in the `cyberware-usage-emitter` impl crate, yet it's registered directly in `ClientHub` here.

Per ModKit conventions, hub-registered types must live in an SDK crate so other modules can resolve them without depending on the full impl. Right now any consumer of this hub key has to add `cyberware-usage-emitter` to its deps — that's an impl-crate leak into the dependency graph.

Move `UsageEmitterRuntimeV1` (and the upcoming `UsageEmitterFactoryV1`) into a `cyberware-usage-emitter-sdk` crate and register the hub key from there.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/routes.rs` line 96 on 2026-05-20T08:37:26Z:

---

**MEDIUM**

`Extension<Arc<dyn UsageEmitterRuntimeV1>>` and `Extension<Arc<Service>>` are layered on the outer router, not scoped to individual routes.

When this router is later composed with other module routers at the server level, both extensions inject into every route in the combined set — including ones that don't need them — and can shadow identically-typed extensions from other modules.

Scope the `.layer(Extension(...))` calls to the specific routes that use them (via a nested sub-router or per-route layering) rather than the top-level router.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/api/rest/handlers.rs` line 92 on 2026-05-20T08:37:31Z:

---

**MEDIUM**

`req.subject.clone()` allocates on every ingest request — `req` is already owned here (moved out of `Json`), so the `Option<Subject>` can be moved directly.

Replace `req.subject.clone()` with `req.subject`. Adjust any subsequent access to `req` fields that the move would break (or reorder the field reads so `subject` is consumed last).

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/usage-collector/src/domain/circuit_breaker.rs` line 189 on 2026-05-20T08:37:36Z:

---

**MEDIUM**

`PluginNotFound` is classified as a health failure and trips the circuit breaker — but `PluginNotFound` means the configured vendor string matched no registered GTS instances. That's a configuration error, not a transient infrastructure fault.

A misconfigured vendor will hold the circuit permanently open, masking any real availability signal from a correctly-registered plugin. It also makes it impossible to distinguish "plugin is down" from "plugin was never registered."

Map `PluginNotFound` to `false` alongside `InvalidPluginInstance` so only genuine transient failures (timeouts, connection errors) trigger the breaker.

---

il10241024:
**Artifizer** reviewed `modules/system/usage-collector/plugins/noop-usage-collector-plugin/src/module.rs` line 23 on 2026-05-20T08:38:11Z:

---

**CRITICAL**

`deps = ["types-registry", "usage-collector"]` creates a circular initialization dependency.

`usage-collector` can't finish `init()` until this plugin has registered its `dyn UsageCollectorPluginClientV1`, but ModKit requires `usage-collector` to be initialized before `noop-usage-collector-plugin` starts — deadlock.

Remove `"usage-collector"` from `deps`. The plugin registers a client the gateway consumes; it must not declare the gateway as its own dependency.

---

PR #3662, opened 2026-06-05
[PR #1295] Feature/resource group features

il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyberware-rust#1295](https://github.com/cyberfabric/cyberware-rust/pull/1295) | **Author:** ffedoroff | **Opened:** 2026-03-24T10:04:39Z | **Status:** closed (not merged)

---

## Summary by CodeRabbit

* **New Features**
  * Resource Group module with full CRUD operations for types, groups, and memberships.
  * Hierarchy management with closure-table support for group relationships.
  * REST API endpoints under `/resource-group/v1` and `/types-registry/v1` with OData `$filter` support.
  * Dual authentication modes: JWT with full authorization and MTLS with endpoint allowlisting.

* **Documentation**
  * Comprehensive architecture, design, and PRD documentation.
  * Integration test plan and E2E test specifications.
  * Feature decomposition and implementation roadmap.

* **Dependencies**
  * Added `resource-group-sdk` and `resource-group` modules to workspace.
  * Updated multiple dependency versions for security and compatibility.

* **Chores**
  * Added linting rule DE1101 for test file separation in resource-group module.
  * Database migrations for resource-group tables.
  * Configuration updates for new module initialization.

---


il10241024:
**coderabbitai[bot]** commented on 2026-03-24T10:04:50Z:

---




> [!IMPORTANT]
> ## Review skipped
> 
> Draft detected.
> 
> 
> <details>
> 
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> **Run ID**: `1c0eaaf8-9597-4673-971a-a1a18108a515`
> 
> </details>
> 
> 
> - [ ]  🔍 Trigger review





<details>
<summary>📝 Walkthrough</summary>

## Walkthrough

This pull request introduces a comprehensive Resource Group system module for managing hierarchical resource grouping, type definitions, and membership relationships. It includes complete SDK contracts, domain services with authorization integration, REST API endpoints, database persistence with SeaORM migrations, and integration test coverage, while updating workspace dependencies and authorization documentation to reflect group-scoping semantics.

## Changes

| Cohort / File(s) | Summary |
|---|---|
| **Documentation & PRD Updates** <br> `docs/arch/authorization/*`, `modules/system/resource-group/docs/*`, `modules/system/authz-resolver/README.md` | Updated authorization design documents to reflect resource-group projection constraints (membership not projected to domain services; only hierarchy projected); added comprehensive decomposition, design, and feature specifications for resource-group module covering SDK foundation, type management, entity hierarchy, membership, integration auth modes, and E2E testing; marked feature/acceptance criteria as complete in PRD. |
| **Workspace & Dependency Management** <br> `Cargo.toml`, `apps/hyperspot-server/Cargo.toml`, `config/*.yaml`, `deny.toml` | Added `resource-group-sdk` and `resource-group` modules to workspace; bumped versions for `serde-saphyr`, `chrono-tz`, `nvml-wrapper`, `criterion`, `calamine`, `enable-ansi-support`; configured `jsonschema` features; added RG module database config to e2e/quickstart YAML; updated RustSec advisory suppressions. |
| **Core Libraries Enhancement** <br> `libs/modkit-*`, `modules/system/authz-resolver/authz-resolver-sdk/src/*` | Added group predicate variants (`InGroup`, `InGroupSubtree`) to authorization SDK constraints; extended scope-filter and security access APIs with resource-group table constants and filter constructors; added transaction-config support to DB provider; added deadlock/serialization-failure detection utilities; enhanced OData filtering with `in` list operator and pagination metadata (`has_next_page`, `has_previous_page`). |
| **Error Handling Standardization** <br> `*/src/domain/error.rs`, `*/src/.../error.rs` | Replaced `.to_string()` with `format!("{...}")` calls across multiple error-conversion implementations for consistency in error message formatting. |
| **Resource-Group SDK** <br> `modules/system/resource-group/resource-group-sdk/src/*` | Defined public SDK contract: `ResourceGroupClient` and `ResourceGroupReadHierarchy` traits; `ResourceGroupError` enum with specific variants; type-safe models (`ResourceGroupType`, `ResourceGroup`, `ResourceGroupMembership`, `GtsTypePath`); OData filter field enums for types, groups, hierarchy, and memberships. |
| **Resource-Group Implementation: Domain Layer** <br> `modules/system/resource-group/resource-group/src/domain/*` | Implemented domain services (`TypeService`, `GroupService`, `MembershipService`, `RgService`, `RgReadService`) with CRUD operations, closure-table hierarchy management, SERIALIZABLE transaction retry logic, authorization integration via `PolicyEnforcer`, and seeding utilities; defined repository traits and error taxonomy. |
| **Resource-Group Implementation: REST API & Routing** <br> `modules/system/resource-group/resource-group/src/api/rest/*` | Implemented REST handlers for types, groups, and memberships (list/create/get/update/patch/delete); added dual-auth-mode routing (JWT vs MTLS) with endpoint allowlisting and client CN validation; wired DTOs with conversions from/to domain models; registered OpenAPI-documented routes. |
| **Resource-Group Implementation: Persistence Layer** <br> `modules/system/resource-group/resource-group/src/infra/storage/*` | Implemented `GroupRepository`, `TypeRepository`, `MembershipRepository` using SeaORM; added six-table schema via migration (types, allowed-parents/memberships junctions, resource-groups, closure table, memberships); implemented OData-to-SQL mapping and hierarchy depth/type filtering. |
| **Resource-Group Module Integration** <br> `modules/system/resource-group/resource-group/src/module.rs`, `modules/system/resource-group/resource-group/src/lib.rs`, `modules/system/resource-group/resource-group/Cargo.toml`, `apps/hyperspot-server/src/registered_modules.rs` | Registered resource-group module with `hyperspot-server` via `inventory` crate; implemented `Module::init` to wire services, resolve AuthZ client, register `ResourceGroupClient` and `ResourceGroupReadHierarchy` into `ClientHub`; exposed database migrations and REST API capabilities. |
| **Testing Infrastructure** <br> `modules/system/resource-group/resource-group/tests/*`, `modules/system/resource-group/resource-group/src/api/rest/dto_tests.rs`, `modules/system/resource-group/resource-group/src/infra/storage/odata_mapper_tests.rs` | Added comprehensive REST API integration tests (2300+ lines) covering CRUD, partial updates, metadata handling, error responses, input validation; added authorization integration tests; provided common test helpers for DB setup, context creation, assertions, and service construction. |
| **Linting & Code Quality** <br> `dylint_lints/*`, `modules/system/authn-resolver/`, `modules/system/nodes-registry/`, `modules/system/oagw/`, `modules/mini-chat/`, `modules/credstore/` | Added DE1101 lint rule enforcing separate test files for resource-group module; enabled DE1302 linting; updated lint_utils formatting; updated various module error conversions for consistency. |
| **Configuration & Artifact Management** <br> `.cypilot/config/artifacts.toml`, `.claude/commands/update-docs.md` | Added ignore rule for `modules/system/oagw/*` (traceability markers not restored); added documentation procedure for synchronizing docs with code post-implementation. |

## Estimated code review effort

🎯 4 (Complex) | ⏱️ ~90 minutes

## Possibly related issues

- Resource Group feature implementation and integration across multiple design/authorization layers and module subsystems.

## Possibly related PRs

- cyberfabric/cyberfabric-core#282: Both modify authorization design documentation around resource-group projection semantics and PDP/PEP scoping.
- cyberfabric/cyberfabric-core#382: Main PR depends on modkit-db refactor (DBProvider, transaction_with_config, secure DB APIs) introduced in this related PR.
- cyberfabric/cyberfabric-core#898: Both modify `libs/modkit-security/src/access_scope.rs` to extend scope filtering with resource-group predicates.

## Suggested labels

`feature`, `chore`

## Poem

> 🐰 A rabbit hops through the resource-group dell,\
> Hierarchies bloom, closures compile so well!\
> Types stack and memberships gracefully bind,\
> AuthZ integration with scopes perfectly aligned—\
> The module is born, with tests standing tall,\
> A feature complete, a triumph for all! 🌿

</details>






<details>
<summary>✨ Finishing Touches</summary>

<details>
<summary>🧪 Generate unit tests (beta)</summary>

- [ ]    Create PR with unit tests

</details>

</details>

il10241024:
**aviator5** commented on 2026-04-01T13:39:34Z:

---

&#2369;coderabbitai review

il10241024:
**coderabbitai[bot]** commented on 2026-04-01T13:39:44Z:

---


<details>
<summary>✅ Actions performed</summary>

Review triggered.

> Note: CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused.

</details>

il10241024:
**coderabbitai[bot]** submitted review **COMMENTED** on 2026-04-01T13:58:57Z:

---

**Actionable comments posted: 3**

> [!NOTE]
> Due to the large number of review comments, Critical severity comments were prioritized as inline comments.

> [!CAUTION]
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> 
> 
> <details>
> <summary>⚠️ Outside diff range comments (1)</summary><blockquote>
> 
> <details>
> <summary>docs/arch/authorization/DESIGN.md (1)</summary><blockquote>
> 
> `1335-1346`: _⚠️ Potential issue_ | _🟡 Minor_
> 
> **Point `resource_group_membership.group_id` at `resource_group`, not the closure table.**
> 
> This section and the SQL example below treat `group_id` as a resource-group ID, but the column description currently says it is an FK to `resource_group_closure`. That changes the data model and will mislead schema/query implementations.
> 
> <details>
> <summary>🤖 Prompt for AI Agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> In `@docs/arch/authorization/DESIGN.md` around lines 1335 - 1346, The column
> description for resource_group_membership.group_id is incorrect: change its FK
> target from resource_group_closure to resource_group so it clearly references
> the resource group ID (as used by the surrounding text and SQL examples); update
> the table description for resource_group_membership and any mention of the FK to
> reference resource_group (not resource_group_closure), and ensure the notes
> about predicate mapping (`in_group` / `in_group_subtree`) remain consistent with
> resource_group being the referenced table.
> ```
> 
> </details>
> 
> </blockquote></details>
> 
> </blockquote></details>

<details>
<summary>🟠 Major comments (21)</summary><blockquote>

<details>
<summary>.cypilot/config/artifacts.toml-36-39 (1)</summary><blockquote>

`36-39`: _⚠️ Potential issue_ | _🟠 Major_

**Over-broad ignore disables OAGW docs validation**

`patterns = ["modules/system/oagw/*"]` suppresses **everything** under OAGW (including `docs/`), which conflicts with the stated intent that docs stay validated. This creates a CI blind spot for PRD/DESIGN/FEATURE artifacts.



<details>
<summary>Suggested fix</summary>

```diff
 [[ignore]]
 reason = "Ignore module 'oagw' (temporary: traceability markers are intentionally not restored in this cycle)."
-patterns = ["modules/system/oagw/*"]
+patterns = [
+  "modules/system/oagw/oagw/src/*",
+  "modules/system/oagw/oagw/tests/*",
+  "modules/system/oagw/oagw-sdk/src/*",
+]
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In @.cypilot/config/artifacts.toml around lines 36 - 39, The ignore pattern
"modules/system/oagw/*" in the [[ignore]] block is too broad and currently
suppresses OAGW docs; update the [[ignore]] block so it only targets the
intended implementation files and not docs — e.g., replace the single pattern
with more specific patterns that target the implementation paths (for example
"modules/system/oagw/src/**" or "modules/system/oagw/**/*traceability*" ) and
ensure the docs path "modules/system/oagw/docs/**" is explicitly excluded or not
matched; edit the patterns array (the string "modules/system/oagw/*" in the
block) to narrow scope and preserve validation of docs.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs-30-31 (1)</summary><blockquote>

`30-31`: _⚠️ Potential issue_ | _🟠 Major_

**`Type` field kind is incorrect (`I64` instead of string).**

`HierarchyFilterField::Type` represents a type path and should be `FieldKind::String`; current mapping will reject valid `$filter` string values.


<details>
<summary>💡 Proposed fix</summary>

```diff
     fn kind(&self) -> FieldKind {
         match self {
-            Self::HierarchyDepth | Self::Type => FieldKind::I64,
+            Self::HierarchyDepth => FieldKind::I64,
+            Self::Type => FieldKind::String,
         }
     }
 }
@@
-        assert_eq!(HierarchyFilterField::Type.kind(), FieldKind::I64);
+        assert_eq!(HierarchyFilterField::Type.kind(), FieldKind::String);
```
</details>


Also applies to: 48-49

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs`
around lines 30 - 31, The match mapping incorrectly assigns
HierarchyFilterField::Type to FieldKind::I64; update the match arms that include
HierarchyFilterField::Type (both occurrences where Type is grouped with
HierarchyDepth/other fields around the existing mappings) to use
FieldKind::String instead of FieldKind::I64 so that Type is treated as a string
path for $filter parsing.
```

</details>

</blockquote></details>
<details>
<summary>libs/modkit-db/src/deadlock.rs-34-39 (1)</summary><blockquote>

`34-39`: _⚠️ Potential issue_ | _🟠 Major_

**`is_deadlock()` returns `true` for PostgreSQL serialization failures, violating its contract.**

The function's contract (lines 26–30) explicitly states it should return `false` for PostgreSQL errors. However, the implementation checks only `msg.contains("40001")`, which includes both MySQL deadlocks *and* PostgreSQL serialization failures—both use SQLSTATE `40001`. PostgreSQL deadlocks use `40P01`, which this function correctly ignores.

The proposed fix in the review is incorrect: checking for `"40001" && "deadlock"` is illogical since `40001` in PostgreSQL means serialization failure, not deadlock. The correct fix should exclude PostgreSQL serialization messages:

```diff
pub fn is_deadlock(err: &DbErr) -> bool {
    match err {
        DbErr::Exec(runtime_err) | DbErr::Query(runtime_err) => {
            let msg = runtime_err.to_string();
-           msg.contains(SERIALIZATION_FAILURE_SQLSTATE)
+           msg.contains(SERIALIZATION_FAILURE_SQLSTATE) && !msg.contains(PG_SERIALIZATION_MSG)
        }
        _ => false,
    }
}
```

This aligns with how `is_serialization_failure()` disambiguates: it checks for the code `40001` *and* the PostgreSQL-specific message pattern `"could not serialize access"` to distinguish the two cases.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@libs/modkit-db/src/deadlock.rs` around lines 34 - 39, is_deadlock currently
returns true for any error whose runtime message contains
SERIALIZATION_FAILURE_SQLSTATE ("40001"), which incorrectly includes PostgreSQL
serialization failures; update is_deadlock (the function handling
DbErr::Exec/DbErr::Query) to treat "40001" as a deadlock only when the runtime
message does NOT match the PostgreSQL serialization pattern (e.g., does not
contain "could not serialize access"); in other words, require
msg.contains(SERIALIZATION_FAILURE_SQLSTATE) && !msg.contains("could not
serialize access") so Postgres serialization failures are excluded while MySQL
40001 deadlocks still return true.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/api/rest/error.rs-89-91 (1)</summary><blockquote>

`89-91`: _⚠️ Potential issue_ | _🟠 Major_

**Do not expose raw authorization failure detail in client-facing Problem responses.**

Returning `message` directly can leak policy/debug context. Emit a generic forbidden detail and keep specifics in logs.



<details>
<summary>Suggested fix</summary>

```diff
-            DomainError::AccessDenied { message } => {
-                Problem::new(http::StatusCode::FORBIDDEN, "Access denied", message)
+            DomainError::AccessDenied { message } => {
+                tracing::warn!(error = %message, "Access denied");
+                Problem::new(
+                    http::StatusCode::FORBIDDEN,
+                    "Access denied",
+                    "You do not have permission to perform this operation",
+                )
             }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/api/rest/error.rs` around
lines 89 - 91, DomainError::AccessDenied currently returns the raw `message`
into the client-facing Problem via Problem::new(http::StatusCode::FORBIDDEN,
"Access denied", message) which may leak auth/policy details; change this to
return a generic detail (e.g. "You do not have permission to access this
resource") in the Problem::new call and move the original `message` into a
server-side log entry (use your logger in this module) so the specific failure
is recorded but not exposed to clients.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group-sdk/src/error.rs-10-55 (1)</summary><blockquote>

`10-55`: _⚠️ Potential issue_ | _🟠 Major_

**Add an explicit SDK authz error variant instead of collapsing to internal.**

Without a dedicated permission-denied variant, authorization failures get flattened into `Internal`, which breaks caller-side handling (e.g., deny vs retry paths).



<details>
<summary>Suggested fix</summary>

```diff
 pub enum ResourceGroupError {
@@
+    /// Caller is not authorized to perform this operation.
+    #[error("Access denied")]
+    AccessDenied,
@@
 impl ResourceGroupError {
@@
+    /// Create an `AccessDenied` error.
+    #[must_use]
+    pub fn access_denied() -> Self {
+        Self::AccessDenied
+    }
+
     /// Create an `Internal` error.
     #[must_use]
     pub fn internal() -> Self {
         Self::Internal
     }
 }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group-sdk/src/error.rs` around lines
10 - 55, Add a specific authorization failure variant to the ResourceGroupError
enum (e.g., PermissionDenied { message: String } with #[error("Permission
denied: {message}")]) instead of collapsing authz failures into Internal; then
update all places that currently map authorization/permission errors to
ResourceGroupError::Internal (or generic internal conversion) to return
ResourceGroupError::PermissionDenied with a clear message (or implement
From<AuthzError> for ResourceGroupError to map authz errors to the new
PermissionDenied variant).
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group-sdk/src/odata/memberships.rs-31-37 (1)</summary><blockquote>

`31-37`: _⚠️ Potential issue_ | _🟠 Major_

**Fix incorrect OData filter type for `resource_type` field.**

`resource_type` is defined as `String` across all membership models (seeding.rs, dto.rs, handlers, models.rs), but the filter metadata at line 34 incorrectly maps it to `FieldKind::I64`. This breaks OData filter validation and parsing for string-typed queries.

Correct both the implementation and test:

<details>
<summary>Fix</summary>

```diff
     fn kind(&self) -> FieldKind {
         match self {
             Self::GroupId => FieldKind::Uuid,
-            Self::ResourceType => FieldKind::I64,
+            Self::ResourceType => FieldKind::String,
             Self::ResourceId => FieldKind::String,
         }
     }
 }
@@
         assert_eq!(MembershipFilterField::ResourceType.name(), "resource_type");
-        assert_eq!(MembershipFilterField::ResourceType.kind(), FieldKind::I64);
+        assert_eq!(
+            MembershipFilterField::ResourceType.kind(),
+            FieldKind::String
+        );
```

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group-sdk/src/odata/memberships.rs`
around lines 31 - 37, The FieldKind for the ResourceType variant in the
memberships OData metadata is wrong: in memberships.rs update the match arm in
fn kind(&self) so Self::ResourceType returns FieldKind::String (instead of
FieldKind::I64) to match models/dto/seeding definitions; then update any
unit/integration tests that assert the filter type for "resource_type" (OData
filter metadata tests) to expect a string field type instead of i64 so
parsing/validation tests align with the corrected metadata.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/migrations/m20260306_000001_initial.rs-190-205 (1)</summary><blockquote>

`190-205`: _⚠️ Potential issue_ | _🟠 Major_

**Down migration may fail on SQLite due to `DROP DOMAIN` statement.**

The `down` method executes raw SQL including `DROP DOMAIN IF EXISTS gts_type_path` which is Postgres-specific. SQLite will error on this unknown syntax.


<details>
<summary>🔧 Proposed fix: backend-specific down migration</summary>

```diff
     async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let backend = manager.get_database_backend();
         let conn = manager.get_connection();
-        conn.execute_unprepared(
-            r"
+
+        let sql = match backend {
+            sea_orm::DatabaseBackend::Postgres => {
+                r"
 DROP TABLE IF EXISTS resource_group_membership;
 DROP TABLE IF EXISTS resource_group_closure;
 DROP TABLE IF EXISTS resource_group;
 DROP TABLE IF EXISTS gts_type_allowed_membership;
 DROP TABLE IF EXISTS gts_type_allowed_parent;
 DROP TABLE IF EXISTS gts_type;
 DROP DOMAIN IF EXISTS gts_type_path;
-            ",
-        )
-        .await?;
+                "
+            }
+            sea_orm::DatabaseBackend::Sqlite => {
+                r"
+DROP TABLE IF EXISTS resource_group_membership;
+DROP TABLE IF EXISTS resource_group_closure;
+DROP TABLE IF EXISTS resource_group;
+DROP TABLE IF EXISTS gts_type_allowed_membership;
+DROP TABLE IF EXISTS gts_type_allowed_parent;
+DROP TABLE IF EXISTS gts_type;
+                "
+            }
+            sea_orm::DatabaseBackend::MySql => {
+                return Err(DbErr::Migration(
+                    "Only PostgreSQL and SQLite are supported".to_owned(),
+                ));
+            }
+        };
+
+        conn.execute_unprepared(sql).await?;
         Ok(())
     }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In
`@modules/system/resource-group/resource-group/src/infra/storage/migrations/m20260306_000001_initial.rs`
around lines 190 - 205, The down() raw SQL includes a Postgres-only "DROP DOMAIN
IF EXISTS gts_type_path" which will break on SQLite; modify the down()
implementation in the migration so it checks the database backend (use the
connection from manager.get_connection()) and only executes the DROP DOMAIN
statement for Postgres (e.g., inspect conn.get_database_backend() or equivalent
and compare to DatabaseBackend::Postgres), otherwise omit that DROP DOMAIN
clause for SQLite—keep the rest of the DROP TABLE statements unchanged and
ensure the conditional logic surrounds only the domain drop.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group-sdk/src/odata/groups.rs-38-44 (1)</summary><blockquote>

`38-44`: _⚠️ Potential issue_ | _🟠 Major_

**Fix `type` field to use `FieldKind::String` for the public OData contract.**

The `type` field is documented as a GTS type path (a string identifier), but mapped to `FieldKind::I64`. This mismatch affects OpenAPI schema generation and exposes an incorrect contract to clients. While `GroupRepository::resolve_type_filter()` does convert string type paths to SMALLINT IDs internally, this conversion should not be reflected in the public API contract; clients must send string GTS paths, not numeric literals.

<details>
<summary>Suggested fix</summary>

```diff
-            // Type is SMALLINT in DB; filter values are resolved from string to
-            // integer at the persistence boundary before OData processing.
-            Self::Type => FieldKind::I64,
+            // Public OData filters accept GTS type path strings; the persistence
+            // layer resolves them to SMALLINT IDs internally.
+            Self::Type => FieldKind::String,
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group-sdk/src/odata/groups.rs` around
lines 38 - 44, The kind() implementation currently maps the enum variant
Self::Type to FieldKind::I64 but the public OData contract expects the GTS type
path as a string; change the mapping in the kind(&self) -> FieldKind function so
that Self::Type returns FieldKind::String instead of FieldKind::I64 (leave
GroupRepository::resolve_type_filter() unchanged since it handles internal
conversion).
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/module.rs-78-83 (1)</summary><blockquote>

`78-83`: _⚠️ Potential issue_ | _🟠 Major_

**Load the query profile from module config instead of hardcoding the default.**

`QueryProfile::default()` ignores the `max_depth` / `max_width` knobs this PR introduces. As written, operators cannot enforce the documented hierarchy guardrails because the service always boots with the library default.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/module.rs` around lines 78 -
83, The code currently constructs the GroupService using
QueryProfile::default(), which ignores the module configuration knobs (e.g.,
max_depth and max_width); change the initialization to load the QueryProfile
from the module's config (read the module config struct/values) and build a
QueryProfile with those values before calling GroupService::new (replace
QueryProfile::default() with a QueryProfile constructed from the module config's
max_depth/max_width or via a helper like config.to_query_profile()), ensuring
GroupService receives the configured profile rather than the library default.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/module.rs-31-35 (1)</summary><blockquote>

`31-35`: _⚠️ Potential issue_ | _🟠 Major_

**Don’t make `init()` depend on AuthZ being pre-registered.**

This module declares `deps = []`, but `init()` hard-fails if `dyn AuthZResolverClient` is not already in `ClientHub`. That makes phase-1 startup order-sensitive and prevents RG from registering its own clients first, which is exactly the cycle the phased RG↔AuthZ bootstrap is supposed to avoid.

Based on learnings module `init()` failures are expected to fail fast rather than retry, so this ordering issue becomes a hard startup blocker.



Also applies to: 59-64

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/module.rs` around lines 31 -
35, The init() function in module.rs currently hard-fails if dyn
AuthZResolverClient is not present in ClientHub, which makes startup
order-sensitive; change init() to treat the AuthZ resolver as optional: query
ClientHub for dyn AuthZResolverClient and if missing do NOT panic or return an
error—either proceed without it (deferring AuthZ-dependent wiring) or register
the resource-group's own clients first and subscribe/resolve the AuthZ client
when it becomes available; specifically modify the init() codepaths that call
into ClientHub for dyn AuthZResolverClient (and the similar checks around lines
~59-64) to use a conditional get/option, log a debug/info message when absent,
and avoid failing fast so the module can register its clients during phase-1
bootstrap.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/type_repo.rs-141-155 (1)</summary><blockquote>

`141-155`: _⚠️ Potential issue_ | _🟠 Major_

**Boolean/non-object `metadata_schema` is lost on readback.**

`build_stored_schema()` stores non-object schemas under `__user_schema`, but `load_full_type()` strips every `__*` key before returning the user-facing value. A valid JSON Schema like `true` or `false` therefore comes back as `None`, which corrupts round-trips and seed/update diffing.

<details>
<summary>💡 Suggested fix</summary>

```diff
-        let metadata_schema = type_model.metadata_schema.as_ref().and_then(|ms| {
-            if let serde_json::Value::Object(map) = ms {
-                let filtered: serde_json::Map<String, serde_json::Value> = map
-                    .iter()
-                    .filter(|(k, _)| !k.starts_with("__"))
-                    .map(|(k, v)| (k.clone(), v.clone()))
-                    .collect();
-                if filtered.is_empty() {
-                    None
-                } else {
-                    Some(serde_json::Value::Object(filtered))
-                }
-            } else {
-                Some(ms.clone())
-            }
-        });
+        let metadata_schema = type_model.metadata_schema.as_ref().and_then(|ms| {
+            if let serde_json::Value::Object(map) = ms {
+                if let Some(user_schema) = map.get("__user_schema") {
+                    return Some(user_schema.clone());
+                }
+
+                let filtered: serde_json::Map<String, serde_json::Value> = map
+                    .iter()
+                    .filter(|(k, _)| !k.starts_with("__"))
+                    .map(|(k, v)| (k.clone(), v.clone()))
+                    .collect();
+                if filtered.is_empty() {
+                    None
+                } else {
+                    Some(serde_json::Value::Object(filtered))
+                }
+            } else {
+                Some(ms.clone())
+            }
+        });
```
</details>


Also applies to: 221-237

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/infra/storage/type_repo.rs`
around lines 141 - 155, build_stored_schema stores boolean/non-object
metadata_schema values under "__user_schema" but load_full_type strips all "__*"
keys and thus returns None for valid boolean schemas; update load_full_type to
detect and return a preserved non-object schema stored under "__user_schema"
(i.e., if the deserialized object has a "__user_schema" key, return its value
instead of dropping it), and apply the same fix for the duplicate logic region
referenced around the other occurrence (the block matching lines 221-237) so
boolean schemas round-trip correctly.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/type_repo.rs-75-83 (1)</summary><blockquote>

`75-83`: _⚠️ Potential issue_ | _🟠 Major_

**Return `allowed_*` lists in a deterministic order.**

These helpers depend on database iteration order twice, so the same type can round-trip with different vector ordering across backends. That makes API responses unstable and can break the idempotent “definition matches” check during type seeding.

<details>
<summary>💡 Suggested fix</summary>

```diff
-        Ok(parent_types.into_iter().map(|m| m.schema_id).collect())
+        let mut parent_codes: Vec<_> = parent_types.into_iter().map(|m| m.schema_id).collect();
+        parent_codes.sort();
+        Ok(parent_codes)

-        Ok(membership_types.into_iter().map(|m| m.schema_id).collect())
+        let mut membership_codes: Vec<_> =
+            membership_types.into_iter().map(|m| m.schema_id).collect();
+        membership_codes.sort();
+        Ok(membership_codes)
```
</details>


Also applies to: 109-117

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/infra/storage/type_repo.rs`
around lines 75 - 83, The returned Vec of schema IDs is currently derived from
database iteration order (via GtsTypeEntity::find() -> parent_types -> map(|m|
m.schema_id).collect()) which is non-deterministic across backends; make the
output deterministic by collecting into a Vec, deduplicating if needed, and
sorting the IDs before returning (apply the same change where a similar pattern
appears around lines 109-117). Target the code that builds parent_types from
GtsTypeEntity::find() and the other analogous helper so the final allowed_*
lists are consistently ordered.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/docs/features/0002-type-management.md-203-208 (1)</summary><blockquote>

`203-208`: _⚠️ Potential issue_ | _🟠 Major_

**Don't describe type seeds as independent or safe for blanket `JoinSet` parallelism.**

Create/update validation requires referenced `allowed_parents` and `allowed_memberships` types to exist first, so a naive parallel pass makes valid seed sets fail nondeterministically whenever one seeded type depends on another. This needs dependency ordering or a two-phase seed flow instead.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/docs/features/0002-type-management.md` around
lines 203 - 208, The doc incorrectly claims seed_def entries are safe for
blanket parallel execution via JoinSet; because create/update validation
requires referenced types in allowed_parents and allowed_memberships,
update/create can fail nondeterministically if dependencies aren’t present.
Change the seeding flow described in inst-seed-2 to either perform
dependency-aware ordering (topologically sort seed_def by
allowed_parents/allowed_memberships and then process in dependency order) or
adopt a two-phase seed: phase 1 create all types with minimal skeletal
definitions (no parent/membership refs), phase 2 update each type to apply
allowed_parents/allowed_memberships; keep JoinSet only within independent groups
after ordering or in the second-phase updates where dependencies are already
satisfied.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/membership_repo.rs-45-52 (1)</summary><blockquote>

`45-52`: _⚠️ Potential issue_ | _🟠 Major_

**Trait is missing `AccessScope` parameter required by project-wide security pattern.**

The `MembershipRepositoryTrait::list_memberships` method has no `scope` parameter, violating the project convention that repository methods enforce tenant isolation via `scope: &AccessScope` (per learnings). Consequently, `MembershipRepository::list_memberships` applies `AccessScope::allow_all()` unconditionally, and the service layer's computed scope (via `enforcer.access_scope()`) is discarded rather than passed to the repository. This leaves row-level AuthZ enforcement to the caller.

Add `scope: &AccessScope` to the trait signature and pass it from the service layer, or document why this method intentionally returns unscoped results and relies on caller-side filtering.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In
`@modules/system/resource-group/resource-group/src/infra/storage/membership_repo.rs`
around lines 45 - 52, The repository trait method
MembershipRepositoryTrait::list_memberships is missing the required AccessScope
parameter so the impl (MembershipRepository::list_memberships) currently uses
AccessScope::allow_all() and ignores the service-layer enforcer.access_scope();
update the trait signature to accept scope: &AccessScope, update the
implementation signature and callers to pass the computed
enforcer.access_scope(), and ensure the DB query built via
MembershipEntity::find().secure().scope_with(&scope) uses that passed-in scope
instead of a hardcoded allow_all; if intentional, add documentation on why
unscoped results are returned instead of enforcing tenant isolation in the repo.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/group_service.rs-629-663 (1)</summary><blockquote>

`629-663`: _⚠️ Potential issue_ | _🟠 Major_

**Validate root metadata against the type schema too.**

The child branch calls `validate_metadata_against_schema(...)` before insert, but the root-group branch inserts immediately. That lets invalid metadata bypass the type schema whenever `parent_id` is `None`.

<details>
<summary>Suggested fix</summary>

```diff
         } else {
             // Root group: validate can_be_root
             if !rg_type.can_be_root {
                 return Err(DomainError::invalid_parent_type(format!(
                     "Type '{}' cannot be a root group (can_be_root=false)",
                     req.type_path
                 )));
             }
+
+            validation::validate_metadata_against_schema(
+                req.metadata.as_ref(),
+                rg_type.metadata_schema.as_ref(),
+            )?;
 
             // Insert group
             let group_id = Uuid::now_v7();
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/group_service.rs`
around lines 629 - 663, The root-group creation branch currently skips schema
validation and inserts directly; before calling group_repo.insert for the root
(where parent_id is None and you checked rg_type.can_be_root), call
validate_metadata_against_schema(&rg_type.schema, req.metadata.as_ref()) (same
validator used in the child branch) and propagate/convert any validation errors
into a DomainError before proceeding to generate group_id and calling
group_repo.insert; ensure you reference req.metadata, type_id and rg_type in the
validation call so invalid root metadata cannot bypass the type schema.
```

</details>

</blockquote></details>
<details>
<summary>libs/modkit-security/src/access_scope.rs-335-350 (1)</summary><blockquote>

`335-350`: _⚠️ Potential issue_ | _🟠 Major_

**Keep subquery operands out of the generic property-value helpers.**

Routing `InGroup` / `InGroupSubtree` through `ScopeFilter::values()` makes helpers like `AccessScope::all_uuid_values_for(pep_properties::RESOURCE_ID)` and `contains_uuid("id", ...)` report group/ancestor IDs as if they were resource IDs. That changes the meaning of the existing helper API for any caller inspecting `RESOURCE_ID` scopes. Please expose these operands through dedicated accessors instead of feeding them into `values()/uuid_values()`.

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/type_service.rs-59-149 (1)</summary><blockquote>

`59-149`: _⚠️ Potential issue_ | _🟠 Major_

**Make `create_type` and `update_type` atomic.**

Both paths mutate multiple tables over a plain connection. If a later statement fails, the type registry is left partially applied: a created type without its junction rows, or an updated type with old relations deleted but new ones only partly inserted. Wrap each operation in a DB transaction so the definition change commits all-or-nothing.



Also applies to: 181-279

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/type_service.rs`
around lines 59 - 149, The create_type (and update_type) flow mutates multiple
tables (calls to type_repo::insert, insert_allowed_parents,
insert_allowed_memberships, delete/insert relations on update, etc.) on a plain
connection and must be executed in a DB transaction so failures roll back; wrap
the whole validation + persistence sequence in a single transaction (begin a
transaction from the existing connection/DB, perform
find_by_code/resolve_ids/insert/insert_allowed_parents/insert_allowed_memberships/load_full_type
inside it, and commit at the end, ensuring any error causes rollback) and apply
the same transactional wrapping to the update_type path that deletes and
reinserts relations so the change is atomic.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/group_service.rs-708-731 (1)</summary><blockquote>

`708-731`: _⚠️ Potential issue_ | _🟠 Major_

**Use the requested parent when type and parent change together.**

`existing.parent_id.or(req.parent_id)` prefers the current parent, so one update that both changes type and moves the group/root-checks it against the old parent. Valid transitions get rejected whenever the new type disallows the old parent but allows the requested one.

<details>
<summary>Suggested fix</summary>

```diff
-            if let Some(parent_id) = existing.parent_id.or(req.parent_id) {
+            let effective_parent_id = if parent_changed {
+                req.parent_id
+            } else {
+                existing.parent_id
+            };
+
+            if let Some(parent_id) = effective_parent_id {
                 let parent = group_repo
                     .find_model_by_id(tx, parent_id)
                     .await?
                     .ok_or_else(|| DomainError::group_not_found(parent_id))?;
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/group_service.rs`
around lines 708 - 731, The validation incorrectly prefers the existing parent
when type and parent are updated together: replace
existing.parent_id.or(req.parent_id) with logic that prefers the requested
parent (e.g., req.parent_id.or(existing.parent_id)) so checks in the type-change
block (the parent lookup via group_repo.find_model_by_id and
resolve_type_path_from_id, and the rg_type.can_be_root check) use the requested
new parent when provided; update the conditional binding and any subsequent uses
so the new-type validation uses the requested parent_id rather than the old one
and keep DomainError::invalid_parent_type messages as-is.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group-sdk/src/models.rs-315-324 (1)</summary><blockquote>

`315-324`: _⚠️ Potential issue_ | _🟠 Major_

**Add `skip_serializing_if = "Option::is_none"` to all patch request fields.**

`PatchGroupRequest` implements tri-state semantics where absent fields should be unchanged, not cleared. Without `skip_serializing_if = "Option::is_none"`, the outer `None` serializes as JSON `null`, causing clients that omit fields to inadvertently clear them instead. This breaks the documented behavior: "Fields not present in the request body are left unchanged."

<details>
<summary>Suggested fix</summary>

```diff
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[allow(clippy::option_option)]
 pub struct PatchGroupRequest {
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
-    #[serde(default, deserialize_with = "option_option::deserialize")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "option_option::deserialize"
+    )]
     pub parent_id: Option<Option<Uuid>>,
-    #[serde(default, deserialize_with = "option_option::deserialize")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "option_option::deserialize"
+    )]
     pub metadata: Option<Option<serde_json::Value>>,
 }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group-sdk/src/models.rs` around lines
315 - 324, The PatchGroupRequest currently serializes outer Option fields as
null which breaks tri-state semantics; update the struct so each field (name,
parent_id, metadata) includes serde attribute skip_serializing_if =
"Option::is_none" (in addition to existing rename/default/deserialize attrs) so
absent fields are omitted from JSON; modify the PatchGroupRequest definition to
add serde(skip_serializing_if = "Option::is_none") to those fields to preserve
"leave unchanged" behavior when a field is omitted.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/repo.rs-251-287 (1)</summary><blockquote>

`251-287`: _⚠️ Potential issue_ | _🟠 Major_

**Make membership repository methods scope-aware.**

`list_memberships` and membership composite-key read/write methods have no `AccessScope` in the contract. Add `scope` so tenant filtering is enforced at the repo layer, not only by service-level assumptions.

<details>
<summary>Suggested contract update (excerpt)</summary>

```diff
     async fn list_memberships<C: DBRunner>(
         &self,
         db: &C,
+        scope: &AccessScope,
         query: &ODataQuery,
     ) -> Result<Page<ResourceGroupMembership>, DomainError>;

     async fn insert<C: DBRunner>(
         &self,
         db: &C,
+        scope: &AccessScope,
         group_id: Uuid,
         gts_type_id: i16,
         resource_id: &str,
     ) -> Result<membership_entity::Model, DomainError>;
```
</details>

  
Based on learnings: in this repository, repository methods are expected to enforce tenant isolation via `scope: &AccessScope` with `.secure().scope_with(scope)`.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/repo.rs` around lines
251 - 287, The repository contract for ResourceGroup membership methods is
missing tenant scope enforcement: add a parameter scope: &AccessScope to the
signatures of list_memberships, insert, delete, find_by_composite_key, and
get_existing_membership_tenant_ids so the repo can enforce tenant isolation;
update their implementations to pass scope into DB queries and use the standard
pattern (.secure().scope_with(scope)) when building queries so tenant filtering
happens at the repo layer rather than relying on service-level checks.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/repo.rs-51-146 (1)</summary><blockquote>

`51-146`: _⚠️ Potential issue_ | _🟠 Major_

**Add `scope: &AccessScope` to group write/closure methods.**

These mutating APIs (`update`, `delete_by_id`, closure-row mutations, membership cleanup helpers) are currently unscoped, so tenant isolation depends on upstream discipline instead of the repository boundary. Please include `scope` in these signatures and enforce it in implementations.

<details>
<summary>Suggested contract update (excerpt)</summary>

```diff
-    async fn update<C: DBRunner>(
+    async fn update<C: DBRunner>(
         &self,
         db: &C,
+        scope: &AccessScope,
         id: Uuid,
         parent_id: Option<Uuid>,
         gts_type_id: i16,
         name: &str,
         metadata: Option<&serde_json::Value>,
     ) -> Result<rg_entity::Model, DomainError>;

-    async fn delete_by_id<C: DBRunner>(&self, db: &C, id: Uuid) -> Result<(), DomainError>;
+    async fn delete_by_id<C: DBRunner>(
+        &self,
+        db: &C,
+        scope: &AccessScope,
+        id: Uuid,
+    ) -> Result<(), DomainError>;
```
</details>

  
Based on learnings: in this repository, repository methods are expected to enforce tenant isolation via `scope: &AccessScope` with `.secure().scope_with(scope)`.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/repo.rs` around lines
51 - 146, The mutating repository methods (update, delete_by_id, closure-row
mutations and membership cleanup helpers) lack tenant scoping; add a new
parameter scope: &AccessScope to the signatures for update, delete_by_id,
insert_closure_self_row, insert_ancestor_closure_rows,
delete_ancestor_closure_rows, delete_all_closure_rows, rebuild_subtree_closure,
has_memberships, delete_memberships (and any other write/closure methods you
missed) and update their callers; in each implementation ensure queries are
executed via the repository's security helper (e.g. call
.secure().scope_with(scope) before running DB operations) so tenant isolation is
enforced at the repo boundary.
```

</details>

</blockquote></details>

</blockquote></details>

<details>
<summary>🧹 Nitpick comments (11)</summary><blockquote>

<details>
<summary>.claude/commands/update-docs.md (2)</summary><blockquote>

`55-56`: **Consider generalizing the marker format examples.**

The marker examples use `cpt-rg-algo-{name}` which appears specific to the resource-group module (rg). Since this command is located in `.claude/commands/` (project-wide) rather than module-specific documentation, it may be used across different modules. Consider either:

1. Using more generic examples like `cpt-{module}-algo-{name}`, or  
2. Clarifying that these examples are for resource-group and adjusting the pattern for other modules



<details>
<summary>📖 More generic example format</summary>

```diff
   - Function/impl level: `// `@cpt-algo`:cpt-rg-algo-{name}:p{N}`
   - Block level: `// `@cpt-begin`:cpt-rg-algo-{name}:p{N}:inst-{step}` / `// `@cpt-end`:...`
+
+  (Note: For other modules, replace `rg` with the module abbreviation, e.g., `cpt-tr-algo-{name}` for types-registry)
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In @.claude/commands/update-docs.md around lines 55 - 56, The examples for
markers are too rg-specific; update the example marker patterns (e.g., the lines
starting with // `@cpt-algo`:cpt-rg-algo-{name}:p{N} and //
`@cpt-begin`:cpt-rg-algo-{name}:p{N}:inst-{step} / // `@cpt-end`:...) to a generic,
reusable format such as // `@cpt-algo`:cpt-{module}-algo-{name}:p{N} and //
`@cpt-begin`:cpt-{module}-algo-{name}:p{N}:inst-{step} / // `@cpt-end`:... (or
alternatively add a short note that the shown pattern is specific to
resource-group and show the module-specific variant), so the command docs are
clearly module-agnostic.
```

</details>

---

`91-103`: **Add language identifier to fenced code block.**

The output template code block is missing a language identifier. This triggers a linter warning and reduces clarity.



<details>
<summary>♻️ Proposed fix</summary>

```diff
-```
+```markdown
 ## /update-docs summary
 
 ### Documentation updated:
```
</details>

As per static analysis hint: MD040 (fenced-code-language) - fenced code blocks should have a language specified.

<details>
<summary>🤖 Prompt for AI Agents</summary>

````
Verify each finding against the current code and only fix it if needed.

In @.claude/commands/update-docs.md around lines 91 - 103, The fenced code block
in .claude/commands/update-docs.md lacks a language identifier causing linter
MD040; update the opening triple-backtick for the summary block to include a
language (e.g., change ``` to ```markdown) so the block is properly labeled—edit
the fenced block starting before "## /update-docs summary" and ensure the
closing triple-backticks remain unchanged.
````

</details>

</blockquote></details>
<details>
<summary>libs/modkit/tests/odata_select.rs (1)</summary><blockquote>

`474-475`: **Consider asserting `has_next_page` / `has_previous_page` in these tests.**

Lines 474 and 526 validate constructor adoption, but adding checks for derived flags would better lock in `PageInfo::new(...)` behavior.

<details>
<summary>Proposed test assertion additions</summary>

```diff
@@
     assert_eq!(result.page_info.next_cursor, Some("abc123".to_owned()));
     assert_eq!(result.page_info.limit, 10);
+    assert!(result.page_info.has_next_page);
+    assert!(!result.page_info.has_previous_page);
@@
     assert_eq!(result.page_info.limit, 20);
+    assert!(!result.page_info.has_next_page);
+    assert!(!result.page_info.has_previous_page);
```
</details>


Also applies to: 526-527

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@libs/modkit/tests/odata_select.rs` around lines 474 - 475, Add assertions in
the tests around the PageInfo construction to verify the derived boolean flags;
locate the test cases that call PageInfo::new(Some("abc123".to_owned()), None,
10) (and the similar call at lines ~526) and assert the expected values for
page_info.has_next_page and page_info.has_previous_page (or the corresponding
getter methods) to lock in behavior when a cursor is present/absent and when
page_size is set to 10. Ensure you reference the PageInfo instance used in those
tests (page_info) and add the appropriate has_next_page/has_previous_page
assertions consistent with the constructor inputs.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/odata_mapper_tests.rs (1)</summary><blockquote>

`49-55`: **Placeholder mapping may need future attention.**

The `HierarchyDepth` → `GroupColumn::Id` mapping with the "(placeholder)" comment suggests this is temporary. When the actual depth-based filtering is implemented, ensure this mapping is updated to target the appropriate closure-table column (e.g., a `depth` column from `resource_group_closure`).

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In
`@modules/system/resource-group/resource-group/src/infra/storage/odata_mapper_tests.rs`
around lines 49 - 55, The test currently asserts
HierarchyFilterField::HierarchyDepth maps to GroupColumn::Id as a placeholder;
update the mapping in HierarchyODataMapper::map_field so that when real
depth-based filtering is implemented it returns the proper closure-table column
(e.g., a GroupColumn variant representing the depth column from
resource_group_closure) instead of GroupColumn::Id, and adjust the test
expectation accordingly to reference that new GroupColumn variant tied to the
closure-table depth field.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/domain/mod.rs (1)</summary><blockquote>

`2-2`: **Consider tracking the lint suppression as technical debt.**

The `#![allow(unknown_lints, de0301_no_infra_in_domain)]` suppression suggests the domain layer has dependencies on infrastructure code, which breaks clean architecture boundaries. While pragmatic for initial implementation, consider tracking this for future refactoring to properly invert the dependency.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/mod.rs` at line 2,
The crate-level lint suppression `#![allow(unknown_lints,
de0301_no_infra_in_domain)]` masks a domain→infrastructure dependency that
should be tracked as technical debt; add a short TODO/FIXME comment immediately
above that attribute stating this is temporary, reference a new issue ID (create
an issue like "DOMAIN-XXX: invert domain dependency from infra") and include the
issue ID in the comment so reviewers can find the task; also add the same issue
ID to the PR description or changelog entry so the debt is recorded outside
code.
```

</details>

</blockquote></details>
<details>
<summary>libs/modkit-odata/src/page.rs (1)</summary><blockquote>

`118-127`: **Prefer reusing `PageInfo::new` in `Page::empty`.**

This avoids duplicating pagination-flag initialization logic.

<details>
<summary>♻️ Suggested refactor</summary>

```diff
     pub fn empty(limit: u64) -> Self {
         Self {
             items: Vec::new(),
-            page_info: PageInfo {
-                next_cursor: None,
-                prev_cursor: None,
-                limit,
-                has_next_page: false,
-                has_previous_page: false,
-            },
+            page_info: PageInfo::new(None, None, limit),
         }
     }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@libs/modkit-odata/src/page.rs` around lines 118 - 127, The Page::empty
implementation duplicates PageInfo field initialization; update Page::empty to
construct its page_info by calling PageInfo::new(limit) instead of manually
setting next_cursor/prev_cursor/has_next_page/has_previous_page so the
pagination-flag logic is reused from PageInfo::new (locate the Page::empty
function and replace the inline PageInfo {...} with a call to
PageInfo::new(limit)).
```

</details>

</blockquote></details>
<details>
<summary>libs/modkit-db/src/secure/cond.rs (1)</summary><blockquote>

`292-340`: **Strengthen new filter tests with SQL-shape assertions.**

Current tests only check “not deny-all.” Consider additionally asserting key subquery fragments (membership/closure table columns) so regressions in query construction are caught earlier.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@libs/modkit-db/src/secure/cond.rs` around lines 292 - 340, Add stronger
assertions to the three tests (test_in_group_filter_produces_subquery_condition,
test_in_group_subtree_filter_produces_subquery_condition,
test_tenant_plus_in_group_produces_and_condition) that validate the SQL-shaped
subquery fragments produced by build_scope_condition rather than only checking
for "not deny-all"; specifically, assert the condition debug string (cond_str)
contains expected subquery tokens such as "EXISTS" or "SELECT", the
membership/closure table column names used by ScopeFilter::in_group and
ScopeFilter::in_group_subtree (e.g., "member", "group", "ancestor", "descendant"
or whatever column identifiers your schema uses), and any join/where fragments
(e.g., comparison to the provided UUID placeholders) so regressions in subquery
construction are caught early.
```

</details>

</blockquote></details>
<details>
<summary>dylint_lints/de11_testing/de1101_tests_in_separate_files/src/lib.rs (2)</summary><blockquote>

`148-150`: **Potential false positive: `is_cfg_test_attr` may match non-test attributes.**

The pattern `line.contains("test")` could match strings like `#[cfg(feature = "contest")]` or `#[cfg(attestation)]`. Consider using a more precise pattern.


<details>
<summary>♻️ More precise test detection</summary>

```diff
 fn is_cfg_test_attr(line: &str) -> bool {
-    line.starts_with("#[cfg(") && line.contains("test")
+    line.starts_with("#[cfg(") && (line.contains("test)") || line.contains("test,") || line.contains("test "))
 }
```

Alternatively, use a regex like `\btest\b` for word-boundary matching.
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@dylint_lints/de11_testing/de1101_tests_in_separate_files/src/lib.rs` around
lines 148 - 150, The is_cfg_test_attr function is too permissive because
line.contains("test") will match substrings like "contest"; update
is_cfg_test_attr to check for a whole-word "test" inside the cfg(...) instead of
raw substring matching — for example, apply a word-boundary regex (e.g.
r"\btest\b") or parse the contents between "#[cfg(" and the matching ")" and
check tokens for the exact identifier "test"; ensure you reference the
is_cfg_test_attr function when making this change so the detection only returns
true for actual test cfg attributes.
```

</details>

---

`152-164`: **Simplify `extract_simulated_dir` logic.**

The `for` loop with `.take(1)` only iterates once, making the early-exit check on lines 158-160 unreachable. Consider simplifying to use `lines().next()`.


<details>
<summary>♻️ Simplified implementation</summary>

```diff
 fn extract_simulated_dir(source: &str) -> Option<&str> {
-    for line in source.lines().take(1) {
-        let trimmed = line.trim();
-        if trimmed.starts_with("// simulated_dir=") {
-            return trimmed.strip_prefix("// simulated_dir=");
-        }
-        if !trimmed.is_empty() && !trimmed.starts_with("//") && !trimmed.starts_with("#!") {
-            break;
-        }
-    }
-
-    None
+    let first_line = source.lines().next()?.trim();
+    first_line.strip_prefix("// simulated_dir=")
 }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@dylint_lints/de11_testing/de1101_tests_in_separate_files/src/lib.rs` around
lines 152 - 164, The function extract_simulated_dir uses a for loop with
.take(1) which makes the "early-exit" check unreachable; replace the loop with
lines().next(), trim that single line, then perform the existing checks: if it
starts_with "// simulated_dir=" return strip_prefix, otherwise if non-empty and
not starting with "//" or "#!" return None, finally return None by default;
update function extract_simulated_dir accordingly to remove the unnecessary loop
and keep the same prefix and comment checks.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/infra/storage/group_repo.rs (1)</summary><blockquote>

`46-49`: **Document the security implications of `system_scope()`.**

`AccessScope::allow_all()` bypasses tenant filtering. While necessary for internal closure table operations (which have no tenant column), consider adding a brief comment explaining why this is safe here to help future maintainers understand the security boundary.


<details>
<summary>📝 Suggested documentation</summary>

```diff
 /// System-level access scope (no tenant/resource filtering).
+/// 
+/// Used for internal operations on tenant-agnostic tables (closure table)
+/// or when the caller has already validated tenant access and needs to
+/// perform cross-cutting operations (e.g., type resolution).
 fn system_scope() -> AccessScope {
     AccessScope::allow_all()
 }
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/infra/storage/group_repo.rs`
around lines 46 - 49, Add a short explanatory comment above the system_scope
function clarifying that it intentionally returns AccessScope::allow_all()
because these internal closure-table operations (referenced by system_scope)
have no tenant/resource column and must bypass tenant filtering; explicitly
state the security boundary and that this scope is only used for internal
closure-table logic to reassure future maintainers that the bypass is
intentional and safe.
```

</details>

</blockquote></details>
<details>
<summary>modules/system/resource-group/resource-group/src/api/rest/auth.rs (1)</summary><blockquote>

`176-179`: **Avoid hardcoding a mount-specific path in the default MTLS allowlist.**

This literal already drifted from other route references in the PR, and `determine_auth_mode()` does exact path matching. A prefix mismatch here will silently disable MTLS for the hierarchy endpoint. Please derive this from the same route constant/config used for route registration instead of duplicating the full path string.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/api/rest/auth.rs` around
lines 176 - 179, The allowlist currently hardcodes the full hierarchy path in
allowed_endpoints which can drift from the registered route and break
determine_auth_mode() exact matching; change the AllowedEndpoint.path_pattern to
be derived from the same route constant/config used when registering the
hierarchy route (e.g., use the ROUTE_* or CONFIG constant that names the
hierarchy path or compose it from the shared base route constant), keeping the
http::Method::GET and the "{group_id}" placeholder intact so
determine_auth_mode() continues to match exactly.
```

</details>

</blockquote></details>

</blockquote></details>

il10241024:
**coderabbitai[bot]** reviewed `modules/system/resource-group/resource-group/src/domain/group_service.rs` line 1054 on 2026-04-01T13:58:56Z:

---

_⚠️ Potential issue_ | _🔴 Critical_

**Reparenting needs the same tenant-invariant check as creation.**

`create_group_inner` rejects parent/child tenant mismatches, but `move_group_internal_impl` only validates parent type. `update_group`, `patch_group`, and `move_group` can therefore place a group under a parent from another tenant while leaving the child's `tenant_id` unchanged.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/group_service.rs`
around lines 1041 - 1054, The move_group_internal_impl path only validates
parent type but misses the tenant invariant enforced in create_group_inner,
allowing cross-tenant moves; update_group, patch_group and move_group should
therefore perform the same tenant check: after loading the parent via
group_repo.find_model_by_id and resolving parent_type_path (as in the shown
diff), compare parent.tenant_id against the moving child's tenant_id and return
a DomainError (similar to DomainError::invalid_parent_type or a new
DomainError::invalid_parent_tenant) when they differ; ensure the check is added
in move_group_internal_impl (and any helper it calls) so
update_group/patch_group/move_group cannot reparent across tenants.
```

</details>

il10241024:
**coderabbitai[bot]** reviewed `modules/system/resource-group/resource-group/src/domain/seeding.rs` line 100 on 2026-04-01T13:58:56Z:

---

_⚠️ Potential issue_ | _🔴 Critical_

**`GroupSeedDef.id` never reaches persistence.**

`seed_groups` uses `seed.id` as the stable lookup key, but on create it calls `GroupService::create_group`, which generates a fresh UUID internally. The first run therefore stores a random ID, the second run misses `seed.id` again and creates a duplicate, and child seeds that reference `parent_id = seed.id` cannot resolve the parent. This needs an explicit-ID create path (or an ID remapping layer) for seeding to be idempotent.



Also applies to: 136-147

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/seeding.rs` around
lines 84 - 100, The seed pipeline fails to persist the provided stable ID
because seed_groups passes GroupSeedDef.id into seed logic but
GroupService::create_group always generates a new UUID; update the seeding flow
to use an explicit-ID creation path: add an overload or new method on
GroupService (e.g., create_group_with_id or accept an Option<Uuid> id parameter
in create_group) that will persist the supplied id, or implement an ID
remapping/lookup that checks for an existing group by the stable seed id before
creating; then change seed_groups to call that new explicit-ID path (using
GroupSeedDef.id) so parent_id references resolve and seeding is idempotent.
```

</details>

il10241024:
**coderabbitai[bot]** reviewed `modules/system/resource-group/resource-group/src/domain/type_service.rs` line 35 on 2026-04-01T13:58:56Z:

---

_⚠️ Potential issue_ | _🔴 Critical_

**Guard type-registry operations with AuthZ.**

This service is the only public domain service here that never receives a `SecurityContext` or `PolicyEnforcer`, yet it backs the type-management entry points. As written, create/get/list/update/delete of global type definitions run without any authorization check.



Also applies to: 39-42

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@modules/system/resource-group/resource-group/src/domain/type_service.rs`
around lines 25 - 35, The TypeService currently lacks any authorization context:
update the TypeService API to require and use authorization (either accept a
SecurityContext or a PolicyEnforcer) and enforce checks in all type-registry
operations (create/get/list/update/delete). Concretely, add a SecurityContext or
PolicyEnforcer parameter to the TypeService constructor or to each public method
(e.g., create_type, get_type, list_types, update_type, delete_type) and call the
appropriate enforcement method before performing repository actions via
type_repo; ensure failure returns/propagates an authorization error. Locate and
modify the TypeService struct and its public methods (TypeService::new,
create/get/list/update/delete) to wire in and invoke the authorization guard.
```

</details>

il10241024:
**striped-zebra-dev** reviewed `libs/modkit-canonical-errors/src/error.rs` line 581 on 2026-04-01T15:34:51Z:

---

The PR is 26K already, help me to understand, how is this change related to the resource group feature?

il10241024:
**striped-zebra-dev** reviewed `libs/modkit-canonical-errors/src/error.rs` line 581 on 2026-04-01T15:42:21Z:

---

Apart of performance degradation, why exactly we change that?

PR #4039, opened 2026-06-08
refactor: complete Module -> Gear term renaming everywhere

Artifizer:
refactor: complete Module -> Gear term renaming everywhere
test: recover CI, dylint, e2e and coverage


Artifizer:
Merging, as Dylint failure is not related to the patch changes and need to be fixed separately

PR #4067, opened 2026-06-10
docs(license-resolver): prepare design documents

nonameffh:
## Summary by CodeRabbit

* **Documentation**
  * Added PRD and full technical design for a read-only, plugin-delegating License Resolver exposing is_licensed(request).
  * Added ADR-0001: subject/resource identity uses required domain type + optional instance id (rejects combined/free-form ids); telemetry by domain type.
  * Added ADR-0002: resolver delegates to plugins, owns no grant store, and is fail-closed when no plugin/backends are available.
  * Added ADR-0003: typed, registry-derived licensing contracts with engine-side validation and versioning guidance.

### DIFF
--- a/gears/system/license-resolver/docs/PRD.md
+++ b/gears/system/license-resolver/docs/PRD.md
@@ -1,0 +1,391 @@
+<!-- cpt:
+version: 1.0.0
+status: draft
+module: license-resolver
+system: cf
+-->
+
+# PRD — License Resolver
+
+<!-- toc -->
+
+- [1. Overview](#1-overview)
+  - [1.1 Purpose](#11-purpose)
+  - [1.2 Background / Problem Statement](#12-background--problem-statement)
+  - [1.3 Goals (Business Outcomes)](#13-goals-business-outcomes)
+  - [1.4 Glossary](#14-glossary)
+- [2. Actors](#2-actors)
+  - [2.1 Human Actors](#21-human-actors)
+  - [2.2 System Actors](#22-system-actors)
+- [3. Operational Concept & Environment](#3-operational-concept--environment)
+- [4. Scope](#4-scope)
+  - [4.1 In Scope](#41-in-scope)
+  - [4.2 Out of Scope](#42-out-of-scope)
+- [5. Functional Requirements](#5-functional-requirements)
+  - [5.1 License Resolution](#51-license-resolution)
+- [6. Non-Functional Requirements](#6-non-functional-requirements)
+  - [6.1 Module-Specific NFRs](#61-module-specific-nfrs)
+  - [6.2 NFR Exclusions](#62-nfr-exclusions)
+- [7. Public Library Interfaces](#7-public-library-interfaces)
+  - [7.1 Public API Surface](#71-public-api-surface)
+  - [7.2 External Integration Contracts](#72-external-integration-contracts)
+- [8. Use Cases](#8-use-cases)
+- [9. Acceptance Criteria](#9-acceptance-criteria)
+- [10. Dependencies](#10-dependencies)
+- [11. Assumptions](#11-assumptions)
+- [12. Risks](#12-risks)
+- [13. Open Questions](#13-open-questions)
+- [14. Traceability](#14-traceability)
+
+<!-- /toc -->
+
+## 1. Overview
+
+### 1.1 Purpose
+
+License Resolver is a read-only CF/Gears system module that answers a single question: *is a specific resource
+licensed (granted) to a specific subject right now?* Callers ask `is_licensed(request)` — a single `LicenseCheckRequest`
+carrying the subject, the resource, optional evaluation `metadata`, and the tenant context — and receive a yes/no
+decision plus structured, non-authoritative diagnostics (debug information about how the decision was reached). It is
+the authoritative point-in-time license check used by other modules to gate access.
+
+License Resolver owns no grant data; it delegates the lookup to a pluggable backend selected at runtime, mirroring
+authz-resolver and tenant-resolver. This keeps licensing storage, issuance, and billing concerns out of the resolver and
+behind a stable contract.
+
+### 1.2 Background / Problem Statement
+
+Multiple CF modules need to check whether a subject may use a licensable resource (a feature, a content item, a
+capability) before granting access. The subject is whoever the license is granted to — a tenant, a user, or any future
+subject type — identified by subject type + id, exactly as the authz-resolver and quota-enforcement subject models do.
+Today no such check exists: CF modules simply do not contain license-resolution logic, so there is no shared contract
+for it, no common GTS-typed resource identity, and no fail-closed guarantee to rely on when gating access.
+
+A dedicated resolver consolidates the check behind one contract keyed by subject (subject type + id) with consistent
+resource identity (GTS-typed) and consistent deny semantics. Because grant facts live in heterogeneous backends owned by
+different vendors, the resolver must delegate the lookup rather than own a store — matching the proven authz-resolver /
+tenant-resolver delegation model. (Tenancy enters only as the isolation scope, carried in the request context that the
+caller derives from its `SecurityContext`.)
+
+### 1.3 Goals (Business Outcomes)
+
+- Single check contract: one `is_licensed` operation reused by all callers, giving modules a license check they do not
+  have today and preventing future per-module divergence.
+- Fail-closed guarantee: a license is never granted by default when the backend is unavailable; deny is the safe outcome
+  in 100% of unavailable-backend cases.
+- Backend independence: licensing backends can be swapped or added via plugin discovery with zero caller code changes.
+
+### 1.4 Glossary
+
+| Term                  | Definition                                                                                                                                                                            |
+|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `LicenseCheckRequest` | A single object bundling the inputs of one check: subject, resource, optional `metadata`, and tenant context. The contract's growth surface (new inputs are added as request fields). |
+| Subject               | The "someone" a license is checked for, identified by subject type + id. Polymorphic — e.g. a tenant, a user, or any future subject type (the licensee is not restricted to tenants). |
+| Resource              | The licensable thing, identified by a single GTS instance id (`GtsInstanceId`) — named (§2.3) or UUID-based (§2.4).                                                                   |
+| Grant                 | A backend fact that a resource is licensed to a subject.                                                                                                                              |
+| Plugin                | A backend implementation discovered via the GTS types registry that answers the check.                                                                                                |
+| Metadata              | Caller-supplied JSON on a check, opaque to the resolver core and forwarded to the backend; the extension point for attribute/constraint-based licensing (region, country, …).         |
+
+## 2. Actors
+
+> **Note**: Stakeholder needs are managed at project/task level by steering committee. Document **actors** (users,
+> systems) that interact with this module.
+
+### 2.1 Human Actors
+
+This module exposes no direct human interface; all actors are systems.
+
+### 2.2 System Actors
+
+#### Consuming Module
+
+**ID**: `cpt-cf-license-resolver-actor-consuming-module`
+
+- **Role**: Any CF/Gears module that must gate access to a licensable resource. Calls
+  `is_licensed(request)` and enforces the returned decision.
+
+#### License Backend Plugin
+
+**ID**: `cpt-cf-license-resolver-actor-backend-plugin`
+
+- **Role**: A vendor-supplied backend implementation that holds grant facts and answers the delegated check. Discovered
+  and selected at runtime via the GTS types registry by vendor + priority.
+
+## 3. Operational Concept & Environment
+
+This module introduces no environment constraints beyond project defaults. Runtime, OS, lifecycle, and integration
+patterns are inherited from the root PRD.
+
+## 4. Scope
+
+### 4.1 In Scope
+
+- Point-in-time check of whether a single resource is licensed to a single subject.
+- Tenant-scoped resolution via the request's tenant context (derived from the caller's `SecurityContext`).
+- GTS-typed resource identity (named/well-known and opaque resources), referencing externally-owned resource types.
+- Plugin-delegated backend selection via the GTS types registry.
+
+### 4.2 Out of Scope
+
+- **Listing / enumeration of granted resources** — answering "everything licensed to a subject" is a catalog/query
+  concern, not a resolver concern; no list operation and no pagination.
+- License issuance and revocation — grant lifecycle is owned by issuing/management modules.
+- Billing and usage metering — owned by the billing/usage domain.
+- Grant storage and management — the resolver owns no grant store; backends do.
+- Defining resource types — resource types are owned by their respective modules and only referenced here by GTS type
+  path.
+
+## 5. Functional Requirements
+
+> **Testing strategy**: All requirements verified via automated tests (unit, integration, e2e) targeting 90%+ code
+> coverage unless otherwise specified.
+
+### 5.1 License Resolution
+
+#### License Check
+
+- [ ] `p1` - **ID**: `cpt-cf-license-resolver-fr-is-licensed-check`
+
+The system **MUST** provide a single check operation `is_licensed(request)` taking a `LicenseCheckRequest` — which
+bundles the subject (whom), the resource (what), optional evaluation `metadata`, and the tenant context (the caller
+derives it from its `SecurityContext`) — and returning a decision indicating whether the resource is licensed to the
+subject at the time of the call, together with structured, non-authoritative **diagnostics** (a string-keyed map of
+debug information about how the decision was reached — e.g. which backend answered, matched grant, denial cause).
+Diagnostics are advisory only and **MUST NOT** be required for the caller to interpret the boolean outcome. The single
+request object is the contract's growth surface — new inputs are added as request fields, not as new parameters or
+method-signature changes. (See `cpt-cf-license-resolver-fr-evaluation-metadata` for the `metadata` field.)
+
+- **Rationale**: A single shared check is the module's reason to exist: it provides license-resolution logic that no
+  module has today, and gives one consistent contract instead of each module growing its own.
+- **Actors**: `cpt-cf-license-resolver-actor-consuming-module`
+
+#### Subject Identity
+
+- [ ] `p1` - **ID**: `cpt-cf-license-resolver-fr-subject-identity`
+
+The system **MUST** identify the subject of a check by a subject type plus an id, consistent with the authz-resolver
+subject model. The subject type **MUST** be open-ended — a license may be granted to a tenant, a user, or any future
+subject type — and the resolver **MUST NOT** assume the subject is a tenant.
+
+- **Rationale**: Consistent, polymorphic subject identity lets callers reuse existing subject references (as
+  authz-resolver and quota-enforcement do), supports licensing any subject kind, and aligns deny semantics across
+  resolvers.
+- **Actors**: `cpt-cf-license-resolver-actor-consuming-module`
+
+#### Resource Identity
+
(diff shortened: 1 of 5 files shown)


nonameffh:
Open question — how callers source metadata, especially system-wide context (e.g. region/environment)

Per `cpt-cf-license-resolver-fr-evaluation-metadata`, metadata is caller-supplied, opaque to the resolver core, and forwarded to the backend for attribute/constraint-based licensing. Writing up an unresolved design point this surfaces.

There are really two kinds of attributes that can end up in metadata, with very different sourcing:

1. Caller/domain-specific ("secret knowledge"). Attributes the calling module naturally holds and passes per call (e.g. a request-specific flag, a content category). This is straightforward — the module just puts them in metadata. No issue.
2. System-wide / deployment context — e.g. region, environment, cluster. The calling module usually doesn't know and shouldn't have to plumb these: they're ambient, fixed at deploy time, and identical for every call in a given deployment. Requiring every caller to populate them on every `is_licensed` call is repetitive, error-prone, and leaks deployment concerns into every module that gates access.

Proposed direction (SDK-level; contract unchanged): in an SDK implementation, provide a thin wrapper around LicenseResolverClient configured with a static key-value map (sourced from deployment/system config). On each `is_licensed`, the wrapper merges its configured map into the caller-supplied metadata before delegating. Callers pass only what they know; deployment-wide attributes (region, environment, …) come from config — set once per deployment, not per call.

Points to settle if we go this way:
- Merge precedence: caller-provided keys win over config (override), or config wins (authoritative deployment context)? I'd lean caller-wins for flexibility, config as defaults — but for things like region you may want config to be authoritative so a caller can't spoof it. May need per-key policy or just a documented rule.
- Shallow vs deep merge of the JSON.

gears/system/license-resolver/docs/PRD.md
```diff
@@ -0,0 +1,392 @@
...
+- Fail-closed guarantee: a license is never granted by default when the backend is unavailable; deny is the safe outcome
+  in 100% of unavailable-backend cases.
+- Backend independence: licensing backends can be swapped or added via plugin discovery with zero caller code changes.
+
+### 1.4 Glossary
+
+| Term                  | Definition                                                                                                                                                                            |
+|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `LicenseCheckRequest` | A single object bundling the inputs of one check: subject, resource, optional `metadata`, and tenant context. The contract's growth surface (new inputs are added as request fields). |
+| Subject               | The "someone" a license is checked for, identified by subject type + id. Polymorphic — e.g. a tenant, a user, or any future subject type (the licensee is not restricted to tenants). |
```
Artifizer: Is Subject Type also a GTS type? I'd make it GTS so that we'd have both subject and resource of GTS type, with optional id (named or UUID)

gears/system/license-resolver/docs/PRD.md
```diff
@@ -0,0 +1,392 @@
...
+
+**ID**: `cpt-cf-license-resolver-actor-backend-plugin`
+
+- **Role**: A vendor-supplied backend implementation that holds grant facts and answers the delegated check. Discovered
+  and selected at runtime via the GTS types registry by vendor + priority.
+
+## 3. Operational Concept & Environment
+
+This module introduces no environment constraints beyond project defaults. Runtime, OS, lifecycle, and integration
+patterns are inherited from the root PRD.
```
Artifizer: what is root PRD?

Artifizer:
I have some proposal, please find it below:

# Proposal: Typed, Registry-Observable Licensing Contracts

# Part 1 — PRD Proposal

## 1. Background and rationale

Licensing is tricky because **enforcement and decision happen on different levels**.

Gears define *what* they may want to check, and perform the actual check by calling the licensing
API — while only the platform vendor (and the licensing-plugin vendor) know what actually *needs*
to be checked. Moreover, there can be a single Gear but multiple platform vendors with different
ideas about which Subject/Resource combinations and which of their properties need to be
licensed.

For example: the LLM gateway itself does not care whether model X is licensed or not. It is the
platform vendor who decides whether, in a given platform, we want to license these or those
models. Platform vendor A may be fine with access to any model, while vendor B may want to
maintain different licenses for different models. The Gear cannot anticipate either policy; the
platform vendor cannot reach into Gear internals to express it.

This fact opens a dilemma: **who defines WHICH Subject/Resource pairs and WHICH of their
properties are passed to the licensing Gear from a regular Gear implementing enforcement?**

In my opinion, the current contract resolves this dilemma in the wrong direction: it makes the check *callable*
but not *governable*. The request metadata is an opaque bag whose keys exist only in Gear source
code, so a platform vendor writing licensing rules has no inventory of the checks, no schema of
the available properties, and no change-control point when a Gear silently adds, renames, or
removes a key.

It follows that **license checks must become a stable contract**, ideally observable by platform
vendors, so they can easily see:

1. **What license checks exist in the Gears?**
2. **What are the Subject/Resource pairs eligible for the license checks?**
3. **What exact Subject/Resource properties are available for the licensing plugin to check?**

The PRD must therefore require that the licensing engine **allow platform vendors to control,
review, and approve licensing contracts**. The requirements below make that concrete.

## 2. New actor: Platform Vendor

The PRD currently models two actors: the consuming module and the backend plugin. The rationale
above shows a **third actor exists and is unserved**.

**Platform Vendor** — composes a concrete platform out of Gears and a licensing backend, and
defines the licensing rules. Does not call the check and does not implement the plugin; instead,
**reviews, approves, and authors rules against the licensing contracts** that Gears expose. The
platform vendor needs to enumerate and inspect those contracts without reading Gear source code.
In fact, the Constructor Fabric steering committee can play a role of this Platform Vendor

## 3. Functional requirements

### FR: Licensing contract registration

Every Subject and Resource shape used in a license check **MUST** be a registered, versioned
contract in the type registry — never an ad-hoc payload assembled at the call site. A Gear that
performs license checks **MUST** declare, as registered types, the Subject/Resource pairs it
checks and the exact schema of the properties it supplies with each.

- **Rationale**: Platform vendors author licensing rules against contracts, not against source
  code. A contract that exists only in a Gear's internals cannot be reviewed, approved, or relied
  upon.

### FR: Contract discoverability and review

Platform vendors **MUST** be able to enumerate, for a given environment:

- which licensing contracts (Subject/Resource type pairs) exist;
- which Subject types each Resource type admits;
- the exact, versioned property schema of each Subject and Resource type.

This enumeration **MUST** be possible through the type registry alone, independently of Gear
source code, and **MUST** allow licensing contracts to be reviewed in isolation from the rest of
the type registry (i.e. licensing contracts are identifiable as such, not mixed indistinguishably
with all other registered types).

- **Rationale**: The licensing engine must allow platform vendors to control, review, and approve
  licensing contracts. Discoverability is what turns "the Gear sends some fields" into "the
  platform exposes a licensing surface."

### FR: Request validation against the registered contract

The licensing engine **MUST** validate every check request against the registered contracts
*before* delegating to the backend plugin, and **MUST** reject non-conforming requests with an
error (fail-closed), distinct from a "not granted" decision. At minimum:

- Subject or Resource properties do not match the registered schema of their type → **ERROR**;
- the Subject type is not admitted by the Resource type's declared subject constraint → **ERROR**.

- **Rationale**: An invalid check silently evaluated is an unauditable check. Validation in the
  engine gives every plugin vendor the same guarantee — a request that reaches the plugin conforms
  to a published contract — instead of each backend re-implementing (or skipping) validation
  differently.

### FR: Contract stability and compatibility rules

Licensing contracts **MUST** follow explicit compatibility rules:

- adding an **optional** property to a Subject/Resource schema is non-breaking — plugins and
  platform vendors ignore properties they do not use;
- removing or renaming a property, changing its type, or narrowing the admitted Subject types is
  breaking and **MUST** be published as a new contract version.

- **Rationale**: "Stable contract" must be a property, not an aspiration. The ignore-unknown-fields
  rule is also the mechanism that makes the contracts flexible: a Gear may expose rich,
  combination-heavy metadata, and each platform vendor consumes only the slice relevant to its
  licensing model.

## 4. Relation to existing PRD requirements

- `fr-evaluation-metadata` is **amended**, not removed: metadata remains the contract's extension
  point, but becomes *schematized* — still opaque to the engine's business logic (the engine never
  interprets what `model_name` means), yet validated against the registered schema. New properties
  arrive as new (optional, non-breaking) schema fields rather than undocumented keys.
- The existing "no enumeration" scope exclusion is untouched: it excludes enumeration of *grants*.
  This proposal requires enumeration of *contract types* via the type registry — a different
  concern, already native to the registry.

---

# Part 2 — DESIGN Proposal

## 1. The licensing projection layer

We already split responsibility across object kinds: DTO (transport), Domain (business logic),
Model (persistence). This proposal adds one more deliberate representation: the **licensing
projection** — a license-level representation of system entities, owned by the licensing domain.

A Gear does not send its domain `User` or `Model` object into a license check. It *projects* the
licensing-relevant slice of those objects into a registered licensing Subject/Resource type. The
copy is the point, not an accident: domain types evolve freely with the Gear's internals, while
the licensing projection changes only deliberately, under contract review, with its own version
lifecycle.

## 2. Base types

The licensing Gear defines two base GTS types from which all licensing contracts derive.

### `gts.cf.core.lic.subj.v1~` — base Subject type

| Field      | Type                  | Notes                                                          |
|------------|-----------------------|----------------------------------------------------------------|
| `tenant`   | optional              | Tenant scope                                                   |
| `user`     | optional              | User reference                                                 |
| `type`     | GTS type ID, optional | Subject's domain type                                          |
| `id`       | UUID or named GTS, optional | At least one of `type` or `id` **MUST** be set           |
| `metadata` | object                | Subject properties, conforming to the derived type's schema    |

### `gts.cf.core.lic.res.v1~` — base Resource type

| Field      | Type                  | Notes                                                          |
|------------|-----------------------|----------------------------------------------------------------|
| `type`     | GTS type ID, optional | Resource's domain type                                         |
| `id`       | UUID or named GTS, optional | At least one of `type` or `id` **MUST** be set           |
| `metadata` | object                | Resource properties, conforming to the derived type's schema   |
| *trait*    | admitted Subject types | The set of `gts.cf.core.lic.subj.v1~*` types this Resource may be checked against |

The "at least one of `type` or `id`" constraint is a validation rule enforced by the engine, not a
convention.

> Namespace note: examples use `gts.cf.core.lic.*`; if the licensing Gear's types live under
> `gts.cf.bss.licensing.*` (as in the current PR), the same scheme applies under that root. One
> namespace should be picked before merge.

## 3. Derived types — Gears define their licensing surface

Each Gear derives concrete Subject/Resource types from the base types. This derivation *is* the
Gear's published licensing contract.

Example — the LLM gateway:

**Subject** `gts.cf.core.lic.subj.v1~cf.genai.llm_gateway.user.v1~`

```yaml
metadata:
  category: string        # e.g. "free" | "pro" | "enterprise"
```

**Resource** `gts.cf.core.lic.res.v1~cf.genai.llm_gateway.model_usage.v1~`

```yaml
metadata:
  model_vendor: string    # e.g. "openai"
  model_name: string      # e.g. "gpt-5"
trait:
  admitted_subjects:
    - gts.cf.core.lic.subj.v1~cf.genai.llm_gateway.user.v1~*
```

Because both types descend from the licensing base types, a platform vendor can query the type
registry for everything derived from `gts.cf.core.lic.subj.v1~` / `gts.cf.core.lic.res.v1~` and
obtain the **complete licensing surface of the platform** — every check that exists, every
Subject/Resource pair, every property available for rules — with none of the surrounding type
"noise".

## 4. The check request

The check carries one Subject object and one Resource object, each an instance of a derived
licensing type:

```text
LicensingRequest:
  subject:  instance of gts.cf.core.lic.subj.v1~…   (derived Subject type)
  resource: instance of gts.cf.core.lic.res.v1~…    (derived Resource type)
```

This is a stable Rust contract. On an LLM gateway request, the gateway projects the caller's user
and the requested model into its derived types and submits:

```json
{
  "subject": {
    "_type": "gts.cf.core.lic.subj.v1~cf.genai.llm_gateway.user.v1~",
    "tenant": "1234...",
    "id": "8f7c…",
    "metadata": { "category": "pro" }
  },
  "resource": {
    "_type": "gts.cf.core.lic.res.v1~cf.genai.llm_gateway.model_usage.v1~",
    "metadata": { "model_vendor": "openai", "model_name": "gpt-5" }
  }
}
```

A platform vendor that licenses per-model writes rules over `model_vendor`/`model_name`; a vendor
that does not, ignores those fields entirely. Same Gear, same contract, different licensing
models.

## 5. Validation pipeline

The licensing engine validates each request **before** plugin delegation:

1. **Structural** — `subject.metadata` and `resource.metadata` conform to the schemas of their
   declared derived types; the base-type invariants hold (at least one of `type`/`id`).
2. **Compatibility** — the Subject's type matches the Resource type's `admitted_subjects` trait.

Both failures are **errors** (fail-closed), never "not granted" decisions — an invalid request is
a contract violation, not a licensing outcome. Validation uses the existing high-performance
schema-validation library with schemas resolved from the registry and cached, so the engine-side
cost is negligible against the check-latency budget.

The plugin therefore receives only requests that conform to a published, versioned contract — a
guarantee no per-backend validation can provide uniformly.

## 6. Trade-offs

### Upsides

1. **Independent contract lifecycle.** Every licensable Subject and Resource becomes a contract
   with its own review and versioning. Domain types (user, model, …) change independently of the
   licensing types; nothing leaks into the licensing surface by accident.
2. **Enumerability and reviewability.** Platform vendors observe exactly which Subject/Resource
   pairs exist and define their licensing rules accordingly — and they review licensing contracts
   in isolation, separated from all other registered types.
3. **Flexibility.** Subject and Resource metadata are open schemas: any Gear can expose whatever
   properties reflect its potentially complicated licensing combinations, and platform vendors
   simply ignore the fields they do not need. With the additive-is-non-breaking rule, richness
   costs nothing.

### Downsides (accepted)

1. **Projection copying.** A Gear must initialize the licensing Subject/Resource from its domain
   objects (e.g. copy fields from `User` into `…llm_gateway.user.v1~`). This is the price of
   decoupling — and the same price we already pay, deliberately, at the DTO and persistence
   boundaries.
2. **Governance overhead.** Derived types must be registered (and reviewed) before a Gear can ship
   a new licensable surface. This is the feature wearing its cost: the review point is exactly
   what platform vendors are asking for.
3. **Registry growth.** Each licensing-aware Gear adds a small number of derived types. Bounded by
   the number of distinct licensable surfaces, which is inherently small per Gear.
4. **The mapping work moves; it does not disappear.** Plugin vendors still must understand each
   Gear's derived types to write rules. What this design changes is that those contracts are now
   discoverable, schematized, and versioned — instead of reverse-engineered from source.

### Prerequisite

The `admitted_subjects` constraint must be expressible and validatable as a GTS trait (or an
equivalent attribute on the derived Resource type). If GTS does not support this today, it is a
prerequisite work item for this design.

---

# Part 3 — ADR: Typed, Registry-Observable Licensing Contracts vs. Opaque Metadata

**Status**: proposed
**Relation to existing ADRs**: supersedes the resource-identity choice of ADR-0001 (single
`GtsInstanceId`) for the check payload, and amends the opaque-metadata principle backing
`fr-evaluation-metadata`. ADR-0002 (plugin delegation, no grant store, fail-closed) is untouched
and fully compatible.

## Context and Problem Statement

The license check carries a Subject and a Resource. Enforcement (the Gear calling the check) and
decision (the platform vendor's licensing rules, evaluated by the plugin) happen at different
levels: the Gear knows where a check belongs, but only the platform vendor knows what must be
checked, and different platform vendors apply different licensing models to the same Gear. How
should the check payload represent Subject, Resource, and their properties so that platform
vendors can discover, review, and rely on what Gears pass into licensing?

## Decision Drivers

* Platform vendors must be able to enumerate the platform's licensing surface (checks,
  Subject/Resource pairs, available properties) without reading Gear source code.
* Licensing contracts need their own review and version lifecycle, decoupled from Gear domain
  types.
* The plugin must be able to trust the shape of what it receives; validation must be uniform, not
  re-implemented per backend.
* Gears must stay free to expose arbitrary, Gear-specific properties (flexibility), and platform
  vendors must be free to ignore what they do not need.
* Fail-closed semantics: a malformed check must be an error, never a silent evaluation.

## Considered Options

* **(a) Opaque metadata, backend-owned validation** (the current PR): Subject = type + id,
  Resource = a single `GtsInstanceId`, `metadata` = an uninterpreted JSON bag forwarded to the
  plugin, which owns all validation.
* **(b) Typed derived GTS contracts validated by the engine** (this proposal): licensing base
  types `lic.subj.v1~` / `lic.res.v1~`; Gears derive concrete Subject/Resource types with
  metadata schemas and an admitted-subjects trait; the engine validates every request against the
  registered contracts before delegation.
* **(c) Registered schemas for documentation only**: keep the request shape of (a), but require
  Gears to register metadata schemas in the type registry for discoverability — without engine
  validation.

## Decision Outcome

Chosen option: **(b)**. Only (b) makes the licensing surface a real contract: discoverable through
the registry, reviewable in isolation, versioned independently of domain types, and *enforced* —
the engine rejects (as a fail-closed error) any request whose Subject/Resource does not match its
registered schema or whose Subject type is not admitted by the Resource type.

Option (a) was rejected because it makes the check callable but not governable: the keys a plugin
can rely on exist only in Gear source code, every backend validates (or doesn't) differently, and
a Gear can silently change the payload with no review point. Option (c) was rejected because
unvalidated contracts are not contracts: the registered schema and the actual behavior drift
apart, and the platform vendor is back to trusting source code — (c) buys the documentation
upside of (b) while keeping the integrity downside of (a). Schema-validation performance is not a
differentiator: validation uses an existing high-performance library with schemas cached from the
registry, negligible against the check-latency budget.

### Consequences

* The licensing Gear owns the base types; every licensing-aware Gear registers derived
  Subject/Resource types — its published licensing contract.
* Gears pay a projection cost (copying licensing-relevant fields from domain objects into the
  licensing types); this is the same deliberate boundary cost as DTO and persistence mappings,
  and it is what decouples domain evolution from contract evolution.
* The engine gains a validation pipeline (structural + subject-compatibility) ahead of plugin
  delegation; validation failures are errors, never "not granted" decisions.
* Compatibility rule: additive optional properties are non-breaking (consumers ignore unknown
  fields); removals, renames, type changes, and narrowing of admitted subjects require a new
  contract version.
* Prerequisite: GTS must support expressing and validating the admitted-subjects constraint
  (trait or equivalent attribute).

## Pros and Cons of the Options

### (a) Opaque metadata, backend-owned validation

* Good, because zero coupling and zero engine-side work; maximal short-term flexibility.
* Bad, because the licensing surface is invisible to platform vendors — no inventory, no schema,
  no change control.
* Bad, because every backend re-implements validation differently, or skips it.

### (b) Typed derived GTS contracts validated by the engine — **chosen**

* Good, because the full licensing surface is enumerable from the registry, reviewable in
  isolation from other types, and versioned.
* Good, because validation is uniform and fail-closed; plugins receive only conforming requests.
* Good, because subject/resource compatibility (`admitted_subjects`) is expressible and checked.
* Good, because metadata stays flexible: Gears expose what they need, platform vendors ignore the
  rest.
* Bad, because of projection boilerplate, registry governance overhead, and the GTS-trait
  prerequisite.

### (c) Registered schemas, documentation only

* Good, because cheap: discoverability without engine changes.
* Bad, because nothing keeps schema and behavior in sync — the contract can lie, which is worse
  than no contract.

gears/system/license-resolver/docs/PRD.md
nonameffh: Yes, missed that in the first iteration - fixed now

gears/system/license-resolver/docs/PRD.md
nonameffh: Good catch, thank you, fixed with the source of this hallucination

nonameffh:
> # Proposal: Typed, Registry-Observable Licensing Contracts

That's helped a lot and resolved some of my concerns I wanted to address later. Adopted with minor changes

Artifizer (approved):

entropyshift (approved):

PR #4069, opened 2026-06-10
refactor(type-registry): adopt SDK canonical-projection pattern

asmith987:
## Summary by CodeRabbit

* **Refactor**
  * Standardized error handling across services and the types-registry to a unified canonical error model for consistent classification and mappings.
* **New Features**
  * Types-registry SDK publishes canonical error vocabulary, projection helpers, and related modules to enable consistent error semantics and richer error information.
* **Tests**
  * Updated tests and mocks across the codebase to use the canonical error helpers and validate the new projection and mapping behavior.


gears/system/account-management/account-management/src/domain/gts_validation_tests.rs
```diff
@@ -103,82 +104,80 @@ fn tenant_schema_with_name_max(max_chars: usize) -> GtsTypeSchema {
...
-    TypesRegistryError::ServiceUnavailable {
-        message: "registry transport down (test stub)".to_owned(),
-        retry_after: std::time::Duration::from_secs(0),
-    }
+fn unavailable() -> CanonicalError {
+    // Any non-`NotFound` canonical error drives the helper's
+    // `ServiceUnavailable` arm (the production code only short-circuits
+    // `NotFound`). `internal` is the cleanest stand-in for a transport
+    // failure here.
+    types_registry_sdk::testing::internal("registry transport down (test stub)")
```
Artifizer: **MEDIUM**

`unavailable()` returns an internal canonical error while the tests and stub documentation model a temporarily unavailable registry.

Both categories currently hit the same catch-all arm, but future code that discriminates on `ServiceUnavailable` for retry behavior would regress without these tests catching it.

Return `CanonicalError::service_unavailable().with_detail("registry transport down (test stub)").create()` so the stub matches the scenario under test.

gears/system/types-registry/types-registry-sdk/src/error.rs
```diff
@@ -1,219 +1,294 @@
...
+            CanonicalError::InvalidArgument { ctx, .. } => Self::Validation {
+                issues: project_field_issues(ctx),
+            },
+
+            CanonicalError::NotFound {
+                resource_type,
+                resource_name,
+                ..
+            } => Self::NotFound {
+                resource_type: resource_type.unwrap_or_default(),
```
Artifizer: **MEDIUM**

`resource_type.unwrap_or_default()` silently converts a malformed `NotFound` canonical envelope into `TypesRegistryError::NotFound { resource_type: "", .. }`.

Callers that dispatch on the documented `TYPE_RESOURCE_TYPE` value get an empty string instead of a signal that the canonical error was missing required resource metadata.

Fall through to `Other` when `resource_type` is `None`, or explicitly document this silent-degradation contract and add a test for the empty-string projection case.

gears/system/resource-group/resource-group/tests/common/mod.rs
```diff
@@ -43,48 +43,46 @@ impl types_registry_sdk::TypesRegistryClient for StubTypesRegistry {
     async fn register(
         &self,
         _entities: Vec<serde_json::Value>,
-    ) -> Result<Vec<types_registry_sdk::RegisterResult>, types_registry_sdk::TypesRegistryError>
+    ) -> Result<Vec<types_registry_sdk::RegisterResult>, toolkit_canonical_errors::CanonicalError>
```
Artifizer, line 46: **MEDIUM**

This test module uses `toolkit_canonical_errors::CanonicalError` without declaring `toolkit-canonical-errors` as a direct dev-dependency.

The test binary compiles only because another dependency currently brings the crate into the build graph, so a future SDK dependency refactor can break these tests accidentally.

Add `toolkit-canonical-errors = { workspace = true }` to `[dev-dependencies]` in `resource-group/Cargo.toml`.

gears/system/account-management/account-management/src/domain/gts_validation_tests.rs
asmith987: Fixed. `unavailable()` returns `CanonicalError::service_unavailable()...` to ensure tests are consistent

gears/system/types-registry/types-registry-sdk/src/error.rs
asmith987: Fixed. Falls through to `Other` now.

gears/system/resource-group/resource-group/tests/common/mod.rs
asmith987: Fixed. Added `toolkit-canonical-errors` to dev dependencies.

PR #4070, opened 2026-06-10
refactor(resource-group): adopt SDK canonical-projection pattern

asmith987:
## Summary by CodeRabbit

* **Refactor**
  * Standardized error handling across resource-group features to a unified canonical error model, producing more consistent, typed error responses for type, group, and membership operations.
  * Improved error context (validation fields, precondition subjects, reason tokens) for clearer, machine- and human-readable failures.
* **Documentation**
  * Expanded crate-level docs and exported vocab constants to describe the canonical error vocabulary and mappings.
* **Tests**
  * Added and expanded tests to validate canonical error projection, round-trip wire mappings, and idempotent/error-handling behaviors.

### DIFF
--- a/gears/system/resource-group/resource-group-sdk/src/error.rs
+++ b/gears/system/resource-group/resource-group-sdk/src/error.rs
@@ -1,158 +1,302 @@
 // Created: 2026-04-16 by Constructor Tech
 // @cpt-dod:cpt-cf-resource-group-dod-sdk-foundation-sdk-errors:p1
-//! Public error types for the resource-group gear.
+//! Resource Group SDK error surface — typed projection of
+//! [`CanonicalError`].
 //!
-//! These errors are safe to expose to other gears and consumers.
+//! # Opt-in convenience, not the contract
+//!
+//! Per [ADR 0005][adr] the [`ResourceGroupClient`] /
+//! [`ResourceGroupReadHierarchy`] trait boundaries are
+//! `Result<_, CanonicalError>`. [`ResourceGroupError`] is an **opt-in**
+//! typed view over that envelope, shipped for consumers that want flat
+//! dispatch on the categories resource-group emits. It is *not* part of
+//! the trait contract: adding a variant is non-breaking, and the single
+//! authoritative AIP-193 classification lives in the impl crate's one
+//! `From<DomainError> for CanonicalError` ladder (`api::rest::error`) —
+//! this projection only reads the finished `CanonicalError`.
+//!
+//! The conversion is infallible (`From<CanonicalError>`). Canonical
+//! categories resource-group does not emit fall through to
+//! [`ResourceGroupError::Other`], which preserves the full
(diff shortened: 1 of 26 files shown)


gears/system/resource-group/resource-group-sdk/src/error.rs
```diff
@@ -1,158 +1,281 @@
...
+        // Borrow the canonical detail before consuming `err`; the borrow
+        // ends here so each arm below can move its fields out (no clones).
+        let detail = err.detail().to_owned();
+        match err {
+            CanonicalError::NotFound {
+                resource_type,
+                resource_name,
+                ..
+            } => Self::NotFound {
+                resource_type: resource_type.unwrap_or_default(),
```
Artifizer: **MEDIUM**

`resource_type.unwrap_or_default()` silently converts a malformed `NotFound` canonical envelope into `ResourceGroupError::NotFound { resource_type: "", .. }`.

Callers that dispatch on the documented resource-type constants receive an empty string instead of a signal that the canonical error was missing required resource metadata.

Fall through to `Other` when `resource_type` is `None`, or explicitly document this silent-degradation contract and add a test for the empty-string projection case.

gears/system/resource-group/resource-group-sdk/src/error.rs
```diff
@@ -1,158 +1,281 @@
...
-    }
+            CanonicalError::InvalidArgument { ctx, .. } => project_invalid_argument(ctx, detail),
+
+            // Resource-group emits exactly one PreconditionViolation per
+            // FailedPrecondition error; the meaningful message lives in
+            // the violation `description`, so surface it as `detail`.
+            CanonicalError::FailedPrecondition { ctx, .. } => {
+                ctx.violations.into_iter().next().map_or_else(
+                    || Self::FailedPrecondition {
+                        subject: Subject::Unknown(String::new()),
```
Artifizer: **MEDIUM**

An empty `FailedPrecondition` violation list is projected into `ResourceGroupError::FailedPrecondition` with empty `subject` and `type_` values.

That makes a malformed canonical envelope look like a modeled resource-group precondition, so callers matching on `Subject` or violation type cannot distinguish bad context from a real domain failure.

Project empty violation lists to `Other { canonical }`, or document the empty-field degradation contract and add a regression test for it.

gears/system/resource-group/resource-group-sdk/src/error.rs
asmith987: Fixed. Went with the fallthrough option

gears/system/resource-group/resource-group-sdk/src/error.rs
asmith987: Fixed. Added a guard so an empty violation falls through to `Other`. As long as the `FailedPrecondition` error is created via the builder, this should never be reachable. It should only be reachable if an arbitrary `Problem` is constructed over the wire (which shouldn't happen)

PR #4071, opened 2026-06-10
feat(toolkit-http): OpenTelemetry HTTP client metrics with ClassifyFn

MikeFalcon77:
Implement MetricsLayer for automatic http.client.request.duration recording:
- Duration histogram (seconds) tagged with method, route, server, and status
- ClassifyFn-based route classification to prevent cardinality explosion
- RequestType extension support for per-request logical operation labels
- Error type tracking (timeout, transport, tls, deadline_exceeded)

Add builder methods: with_metrics() (default classifier), with_metrics_by() (custom)
Include example: toolkit-auth/examples/oauth2_metrics.rs

## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry HTTP client metrics (feature-flagged) with builder integration.
  * Configurable route classification to keep metric labels low-cardinality.
  * Per-request metrics labeling via an opt-in request label.
  * Example demonstrating an OAuth2 client instrumented with these metrics.

* **Documentation**
  * Example docs describe required OTel startup and histogram tag behavior.

* **Tests**
  * Unit tests validate metric attributes for success, failures, and classification.

### DIFF
--- a/libs/toolkit-http/src/layers/metrics.rs
+++ b/libs/toolkit-http/src/layers/metrics.rs
@@ -1,0 +1,448 @@
+//! Tower layer that records OpenTelemetry metrics for outbound HTTP requests.
+//!
+//! Emits a single instrument following [OpenTelemetry HTTP client semantic
+//! conventions][semconv]:
+//! - `http.client.request.duration` — histogram (seconds)
+//!
+//! Attributes: `http.request.method`, `http.route`, `server.address`,
+//! `server.port` (when the URI carries an explicit port), and
+//! `http.response.status_code` (on success) or `error.type` (on failure).
+//!
+//! Modeled after go-appkit's `MetricsRoundTripper`: one duration histogram plus
+//! a build-time request classifier that produces the bounded `http.route`
+//! label, preventing cardinality explosion from raw paths. Like the Go version,
+//! this layer sits outside the retry loop, so it observes one logical request
+//! regardless of transport-level retries.
+//!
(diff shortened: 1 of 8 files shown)


libs/toolkit-http/src/layers/metrics.rs
```diff
@@ -0,0 +1,448 @@
...
+
+/// Default classifier producing `"METHOD host"` (mirrors go-appkit's default
+/// `summary`). Never returns a raw path, so it cannot blow up cardinality.
+#[must_use]
+pub fn default_classify(req: &Request<Full<Bytes>>) -> Cow<'static, str> {
+    let host = req.uri().host().unwrap_or("unknown");
+    // Use the normalized method (`_OTHER` for unknown verbs) so the route label
+    // stays consistent with the `http.request.method` attribute and cannot be
+    // widened by arbitrary method strings.
+    Cow::Owned(format!("{} {}", normalize_method(req.method()), host))
```
Bechma, line 62: I see there are a lot of use of Cow, can you explain the motivation of using that instead of String?

MikeFalcon77: Cow pattern for APIs that are almost always called with &'static str to avoid an allocations on every call vs requiring to_string() with String.

Could you point to a specific usage you find problematic? I wouldn't call 7 places across 7 files "a lot"

Bechma: Sorry, by a lot I meant in the PR based on the public API. I just wanted to know the motivation, save allocations for 'static looks good to me. 👍

PR #4083, opened 2026-06-12
docs(event-broker): module design + SDK

striped-zebra-dev:
## Summary by CodeRabbit

* **Documentation**
  * Added a new **Event Broker** section to the GEARS guide, describing multi-consumer, partitioned, append-only typed event streaming, delivery semantics, and consumer cursor tracking.
  * Documented current rollout status for the Event Broker SDK and added links to the PRD, signed-off design, and SDK materials.

* **Chores**
  * Updated workspace configuration to include the new Event Broker SDK crate and adjusted build tooling ignore patterns for UI/trybuild tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ members = [
     "libs/toolkit-security",
     "libs/toolkit-odata",
     "libs/toolkit-node-info",
+    "libs/toolkit-stable-hash",
     "libs/toolkit-transport-grpc",
     "libs/toolkit-utils",
     "libs/rustls-corecrypto-provider",
@@ -108,6 +109,7 @@ members = [
     "gears/bss/ledger/ledger",
     "gears/bss/ledger/ledger-sdk",
     "gears/bss/libs/coord",
+    "gears/system/event-broker/event-broker-sdk",
 ]
 exclude = ["tools/fuzz"]
 resolver = "3"
@@ -256,6 +258,7 @@ toolkit-odata = { package = "cf-gears-toolkit-odata", version = "0.8.3", path = 
 toolkit-odata-macros = { package = "cf-gears-toolkit-odata-macros", version = "0.6.8", path = "libs/toolkit-odata-macros" }
 toolkit-sdk = { package = "cf-gears-toolkit-sdk", version = "0.7.1", path = "libs/toolkit-sdk" }
 toolkit-security = { package = "cf-gears-toolkit-security", version = "0.7.3", path = "libs/toolkit-security" }
+toolkit-stable-hash = { package = "cf-gears-toolkit-stable-hash", version = "0.1.0", path = "libs/toolkit-stable-hash" }
 toolkit-transport-grpc = { package = "cf-gears-toolkit-transport-grpc", version = "0.7.0", path = "libs/toolkit-transport-grpc" }
 toolkit-utils = { package = "cf-gears-toolkit-utils", version = "0.6.4", path = "libs/toolkit-utils" }
 rustls-corecrypto-provider = { package = "cf-gears-rustls-corecrypto-provider", version = "0.1.1", path = "libs/rustls-corecrypto-provider" }
@@ -602,7 +605,7 @@ gts = "0.11.0"
 gts-macros = "0.11.0"
 
 [workspace.metadata.cargo-shear]
-ignored-paths = ["**/tests/ui/**"]
+ignored-paths = ["**/tests/ui/**", "**/tests/trybuild/**"]
 # Gear crate deps only consumed by macro-generated re-exports in #[toolkit::gear(deps = [...])].
 # cargo-shear cannot trace usage through proc-macro expansion, so these appear unused.
 ignored = [
--- a/docs/GEARS.md
+++ b/docs/GEARS.md
@@ -699,6 +699,19 @@ Introduces an abstraction layer behind the real Outbound API Gateway. The main g
 - [API](../gears/system/oagw/oagw/README.md)
 - [SDK](../gears/system/oagw/oagw-sdk/README.md)
 
+### Event Broker
+
+Multi-consumer, partitioned, append-only event streaming for Cyber Ware modules.
+Typed events, at-least-once delivery, idempotent producers (chained/monotonic/stateless),
+pluggable storage backends, consumer-group cursor tracking.
+
+**Status**: SDK landed (`cyberware-event-broker-sdk`) — impl crate TODO.
+
+#### More details
+- [PRD](../gears/system/event-broker/docs/PRD.md)
+- [Design](../gears/system/event-broker/docs/DESIGN.md) (R4 signed off)
+- [SDK](../gears/system/event-broker/event-broker-sdk/README.md)
+
 ## Core Platform Services
 
 Core Platform Services are authoritative, enterprise-level services that may exist outside of Gears and act as systems of record for critical governance domains such as accounts, identity, access policies, licensing, credentials, and outbound egress control. These components typically belong to an organization’s broader platform or SaaS ecosystem and may already be deployed, certified, and governed independently of Gears.
--- a/gears/system/event-broker/docs/ADR/0002-partition-selection.md
+++ b/gears/system/event-broker/docs/ADR/0002-partition-selection.md
@@ -1,0 +1,265 @@
+---
+status: proposed
+date: 2026-05-12
+decision-makers: Event Broker Team
+revision-history:
+  - 2026-05-06 — initial draft (key-hash with explicit producer override + subject fallback)
+  - 2026-05-12 — revised (drop explicit `partition` override; broker re-hashes and is authoritative; `partition_key` is the producer-facing input selector)
+  - 2026-06-07 — default partition key changed from `subject` to `tenant` (PR #1978 review): a tenant's events are totally ordered by default; per-subject grouping is opt-in via explicit `partition_key`
+---
+
+# Partition Selection — Broker-Authoritative `partition_key` / `tenant_id` Input
+
+<!-- toc -->
+
+- [Context and Problem Statement](#context-and-problem-statement)
+- [Decision Drivers](#decision-drivers)
+- [Considered Options](#considered-options)
+- [Decision Outcome](#decision-outcome)
+  - [Default Algorithm](#default-algorithm)
+  - [Partition-Key Source](#partition-key-source)
+  - [Hash Location](#hash-location)
+  - [Broker-Side Validation](#broker-side-validation)
+  - [Encoding](#encoding)
+  - [Consequences](#consequences)
+  - [Confirmation](#confirmation)
+- [Pros and Cons of the Options](#pros-and-cons-of-the-options)
+  - [Broker-Authoritative `partition_key` / `tenant_id` Input (chosen)](#broker-authoritative-partition_key--tenant_id-input-chosen)
+  - [Earlier Draft — Add Explicit `partition` Producer Override (rejected on revision)](#earlier-draft--add-explicit-partition-producer-override-rejected-on-revision)
+  - [Always Derive Partition From `event.subject`](#always-derive-partition-from-eventsubject)
+  - [Round-Robin With No Key Affinity](#round-robin-with-no-key-affinity)
+  - [Custom Pluggable Partitioner Trait in SDK (MVP)](#custom-pluggable-partitioner-trait-in-sdk-mvp)
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-cf-evbk-adr-partition-selection`
+
+## Context and Problem Statement
+
+A topic in the Gears event broker is divided into a fixed number of partitions declared at topic registration time (`Topic.partitions`, see DESIGN §3.1). Every `Event` is bound to a partition; sequence assignment, ordering, idempotent-producer state, and consumer cursors are all scoped to `(topic, partition)`. The broker therefore needs a contract for **how partition assignment is computed** before the event is enqueued in the producer outbox and ultimately landed in `backend.persist`.
+
+The existing design imposes hard constraints on this decision:
+
+- `Topic.partitions` is **fixed at topic creation** and cannot be grown or shrunk on a live topic. Re-partitioning would break per-key ordering for every key already published; the migration path is "create new topic, dual-write, cut consumers over." Partition selection MUST therefore behave deterministically across the topic's full lifetime.
+- The broker assigns the consumer-visible `event.sequence` per `(topic, partition)`. Producers never set `sequence`; they only carry chain state for ingest-side dedup in `meta.previous` and `meta.sequence` (see [ADR-0003 Event Schema](0002-event-schema.md)).
+- `Event` carries `subject` (a per-event entity identifier) and `subject_type`. A dedicated `partition_key` field exists as an optional producer-supplied grouping input.
+- Idempotent-producer state is keyed by `(producer_id, topic, partition)` (`evbk_producer_state`). The same producer publishing "the same logical event" on a retry MUST land on the same partition, otherwise the chain check fires against a state row that does not contain the previous attempt and the duplicate is admitted.
+
+The **initial 2026-05-06 draft** of this ADR allowed three paths to a partition value: `partition_key` (hashed), `subject` fallback (hashed), and an explicit `event.partition` producer override. The override path turned out to be a silent-ordering-bug surface — a producer refactor that switches a code-path from "set `partition_key`" to "set `partition` directly" quietly breaks per-key ordering on a live topic, and the SDK has no way to tell whether the producer *meant* to bypass the hash. Re-iteration during pre-implementation review removed the override path. This revision records the corrected contract.
+
+## Decision Drivers
+
+* Per-key order: events sharing a stable partition key MUST land on the same partition for the lifetime of the topic so consumers observe them in publish order
+* Even distribution under unbiased keys: the chosen partition distribution SHOULD be approximately uniform across `[0, topic.partitions)`
+* Idempotent retry determinism: a retry of the same logical publish MUST resolve to the same `partition`, otherwise idempotent-producer dedup degrades to "best effort"
+* First-party SDK / broker parity: when an SDK computes a local partition hint for outbox routing, it must use the same input field as the broker and produce a value the broker can validate
+* Operational simplicity: the broker stays a thin validator; the partition is computed deterministically from inputs already on the event
+* No silent freedom: a producer should not be able to bypass the hash with an explicit partition number — this turned out to be a refactor-induced bug surface in the initial draft
+* Schema extensibility: support legitimate use cases where the partition key differs from the event subject (per-tenant audit, system events with no business-domain subject, deliberate fan-out)
+
+## Considered Options
+
+* Broker-authoritative partition selection from `partition_key` by default, falling back to `tenant_id`; SDK-local partitioning is only a hint for first-party outbox routing (chosen — this revision)
+* Earlier draft: key-hash with explicit `event.partition` producer override (rejected on revision)
+* Always derive partition from `event.subject` (no `partition_key` field at all)
+* Round-robin with no key affinity
+* Custom pluggable partitioner trait exposed in the SDK at MVP
+
+## Decision Outcome
+
+Adopt **broker-authoritative partition selection from `partition_key` when present, otherwise `tenant_id`**. `partition_key` is the producer-facing way to choose a grouping input other than the default `tenant_id`. There is no explicit `partition` field on publish input. The broker computes the final topic partition; any producer SDK partition computation is an internal/local hint for outbox routing, not a native Kafka producer compatibility contract.
+
+Defaulting to `tenant` gives **per-tenant total ordering** out of the box — every event a tenant emits to a topic lands on one partition and is observed in publish order (the property the audit pipeline needs). Producers that want finer-grained grouping (per-subject, per-region, fan-out) set `partition_key` explicitly.
+
+### Default Algorithm
+
+The partition is computed deterministically from a single input:
+
+```text
+partition_input = partition_key  if partition_key is present
+                  tenant_id      otherwise
+partition       = local_derivation(ascii_bytes(partition_input)) % partition_count
+```
+
+- Current first-party SDK/broker implementation: **MurmurHash3 (32-bit, x86 variant)** with a fixed seed of `0x00000000`, masked with `& 0x7FFFFFFF` before modulo. This pins first-party SDK hints to broker validation but is not a native Kafka producer compatibility promise.
+- The mask `& 0x7FFFFFFF` strips the sign bit so the modulo operates on a non-negative `u31` value and avoids the negative-modulo edge case in languages with signed `%`.
+- The bytes hashed MUST be the **ASCII byte representation** of the input. Per the platform convention recorded in [ADR-0003 Event Schema § Event Field Encoding](0002-event-schema.md#event-field-encoding-ascii-only), all event string fields (`partition_key`, `subject`, etc.) are ASCII; UTF-8 is permitted only inside `data`. This eliminates UTF-8-vs-ASCII determinism concerns in the hash path.
+- Producers MUST NOT provide a top-level topic `partition`. First-party SDKs that send an internal `meta.partition_hint` MUST compute it with the broker-supported local derivation for that broker version.
+
+### Partition-Key Source
+
+The `Event` schema carries an **optional, body-level** field:
+
+- **`partition_key: Option<String>`** — producer-supplied, opaque (ASCII, ≤ 1024 bytes), used only for partition selection. Not validated against any GTS schema. Preserved on the read projection so consumers can see which grouping key the producer chose.
+
+Resolution rules:
+
+1. If `event.partition_key` is `Some(s)`, hash `s`.
+2. Else, hash `event.tenant_id` (which is always present per the event schema contract).
+
+Since `tenant_id` is required on every event, there is no third case — no missing-input failure mode, no silent fallback. The default path is always defined.
+
+`tenant_id`, `subject`, and `partition_key` are conceptually different:
+
+- `tenant_id` identifies the tenant the event belongs to — the default co-location key, giving per-tenant ordering.
+- `subject` identifies the entity the event is *about*.
+- `partition_key` is the explicit grouping input controlling *co-location* when tenant-wide grouping is not the desired behavior.
+
+The tenant default fits the common platform case (audit, notifications, per-tenant streams) where a tenant's events should be totally ordered. Producers needing per-subject ordering set `partition_key = subject`; system events with no natural tenant grouping or deliberate fan-out for non-causal high-volume events set `partition_key` explicitly (e.g., `partition_key = uuid_v4()`).
+
+### Hash Location
+
+Partition selection happens in **both** the producer SDK and the broker:
+
+- **Producer SDK** computes the partition locally before calling `outbox.enqueue()`, so the `toolkit-db` outbox can route the row to the correct per-`(topic, partition)` outbox shard and preserve order.
+- **Broker** re-computes the partition on ingest from `partition_key` / `tenant_id` in the received event. The broker's value is authoritative; if persisted at all, the SDK-computed value is treated as a hint only.
+
+This is a deliberate change from the initial draft (which had the SDK as the single computer of `partition`). The trade-offs:
+
+- Adds one Murmur3-32 hash to the ingest path (~ns-scale; negligible against the DB write that follows).
+- Eliminates the producer-supplied partition surface (no explicit `partition` field that producers can stamp).
+- Adds defense-in-depth against SDK bugs: if the SDK stamps an internal `meta.partition_hint` for outbox routing, the broker validates equality and returns `400 PartitionHashMismatch` on drift.
+
+The SDK may fetch `topic.partitions` once at startup (or on first publish to a topic) via `GET /v1/topics/{id}` when it needs a local topic-partition hint. The count is immutable per `Topic.partitions` design, so that cache does not go stale. This does not require producer-local outbox partitions, broker topic partitions, and ingest-service shards to have the same count; each partition domain derives its own local partition from the agreed input field and its own partition count.
+
+Example:
+
+```text
+partition_input = partition_key if present else tenant_id
+
+producer local/outbox partition = local_derivation(partition_input) % producer_outbox_partitions
+broker topic partition          = broker_derivation(partition_input) % topic.partitions
+ingest service shard            = ingest_derivation(partition_input or topic/partition) % ingest_shard_count
+```
+
+These counts can legitimately differ, such as 16 producer outbox partitions, 64 broker topic partitions, and 8 ingest shards.
+
+### Broker-Side Validation
+
+On `POST /v1/events` and `POST /v1/events:batch`, the ingest service MUST:
+
+- Compute `partition` from `partition_key` (if present) or `tenant_id` using the broker-supported derivation for this broker version.
+- Use the computed value as the authoritative partition for sequence assignment, `evbk_producer_state` lookup, and storage routing.
+- Reject any publish carrying a top-level `partition` field with `400 BadRequest` (RFC 9457 problem type `gts.cf.core.errors.err.v1~cf.core.partition.forbidden.v1`). The wire contract DOES NOT accept a producer-supplied partition.
+- If the publish carries `meta.partition_hint` (an internal SDK-stamped optimization), validate it equals the broker-computed value; reject mismatches with `400 PartitionHashMismatch` (RFC 9457 problem type `gts.cf.core.errors.err.v1~cf.core.partition.hash.mismatch.v1`).
+
+### Encoding
+
+All inputs to the hash are ASCII per [ADR-0003 § Event Field Encoding](0002-event-schema.md#event-field-encoding-ascii-only). The broker rejects publishes with non-ASCII bytes in `partition_key` or `tenant_id` with `400 InvalidEventFieldEncoding` before partition computation is attempted. `partition_key` is additionally length-capped at 1024 bytes (`400 EventFieldTooLong` on overflow).
+
+### Consequences
+
+- Good, because per-`tenant` ordering holds by default, with zero producer configuration — a tenant's events on a topic are totally ordered, which is the common platform need (audit, notifications, per-tenant streams).
+- Good, because producers needing a different grouping (per-subject, per-region, fan-out) opt in by setting `partition_key` explicitly — no SDK fork, no custom partitioner.
+- Good, because idempotent retries are deterministic: the same `(producer_id, partition_key | tenant_id)` resolves to the same `partition` and therefore to the same `evbk_producer_state` row, so chain dedup works as designed.
+- Good, because future first-party SDKs can validate their local hints against broker fixtures without making native Kafka producer compatibility part of the publish contract.
+- Good, because the broker is the authority on partition assignment, removing the silent-ordering-bug class that a producer-set `partition` override created in the initial draft.
+- Good, because the broker stays a thin validator: the partition computation is a single line of code (hash + mask + mod), not a policy engine.
+- Bad / accepted limitation, because **re-partitioning is not supported**. Once a topic is created with N partitions, the only way to change is the dual-write migration path. Deliberate match to Kafka semantics; consumers depend on stable key-to-partition mapping.
+- Bad / accepted limitation, because **no per-topic partitioner choice in MVP**. Every topic uses the same Murmur3 algorithm.
+- Bad / accepted limitation, because **hash collisions are accepted**. Two distinct partition keys can map to the same partition; intrinsic to modulo-hash partitioning.
+- Bad / accepted limitation, because **a large tenant hot-spots its partition** under the tenant default — all of one tenant's events route to a single partition, so a high-volume tenant gets no intra-tenant parallelism and can become a noisy neighbour. Accepted in exchange for per-tenant ordering; the escape hatch is an explicit `partition_key` (e.g., `partition_key = subject`) for topics where a tenant's volume outweighs its ordering need.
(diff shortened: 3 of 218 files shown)


gears/system/event-broker/docs/ADR/0006-offset-authority.md
```diff
@@ -0,0 +1,64 @@
...
+The broker resolves the timestamp to the offset of the first event whose `occurred_at ≥ timestamp` and sets the cursor in one step. The response returns the resolved integer offset per partition, which the consumer may persist for future re-seeks. No separate endpoint is introduced.
+
+## Consequences
+
+**Removed:**
+- `POST /v1/subscriptions/{id}/ack` endpoint (and all associated broker-side cursor durability logic)
+- `cursor.acked` position (renamed to `cursor.offset` — the ephemeral session cursor set by SEEK)
+- `PartitionNotAssigned` error type (was only raised by ACK)
+- `acknowledge()` method from `DeliveryService`
+- Any future `/v1/groups/{group}/offsets` KV endpoint
```
nonameffh: `DESIGN.md:478` treats that same endpoint as a reserved future extension:

Consumers that need stronger durability use the future `/v1/groups/{group}/offsets` KV endpoint (§4.8) or track offsets in their own DB and seek on JOIN via `POST /v1/subscriptions/{id}:seek`.

gears/system/event-broker/docs/ADR/0002-partition-selection.md
```diff
@@ -0,0 +1,255 @@
...
+
+The partition is computed deterministically from a single input:
+
+```text
+partition_input = partition_key  if partition_key is present
+                  tenant_id      otherwise
+partition       = (murmur3_32(ascii_bytes(partition_input)) & 0x7FFFFFFF) % topic.partitions
+```
+
+- Hash function: **MurmurHash3 (32-bit, x86 variant)** with a fixed seed of `0x00000000`. Matches Kafka's `DefaultPartitioner` / `BuiltInPartitioner` choice and is the de-facto standard for partition selection in log-structured brokers.
```
nonameffh: Kafka's DefaultPartitioner uses MurmurHash2. For services that use only event broker this is harmless. But if a native Kafka producer ever writes to the same topic, the same key routes to a different partition, splitting per-key ordering

gears/system/event-broker/docs/DESIGN.md
```diff
@@ -0,0 +1,2584 @@
...
+
+##### Where each sequence is assigned
+
+**Producer chain (`meta.previous`, `meta.sequence`)** — assigned by the producer client at publish time (chained / monotonic modes; `meta` absent in stateless). Carried inside the publish-input `meta` block. Validated by the IngestService's chain check against `evbk_producer_state.last_sequence`. Storage MAY retain `meta` for audit; the consumer-visible read projection strips it.
+
+**Outbox sequence** — assigned by `modkit-db`'s outbox sequencer when an event is enqueued in the ingest-side outbox. Guarantees events for the same outbox partition are delivered to `backend.persist` in enqueue order. Stays internal to the outbox; not exposed in the API; not stored by the backend.
+
+**Backend `sequence`** — assigned by the backend during `persist`:
+- **DB backend**: native DB sequence (e.g., Postgres `BIGINT GENERATED ALWAYS AS IDENTITY` per `(topic, partition)` partition, or per-row sequence within segment); backend's storage layout dictates the implementation.
+- **Kafka backend**: Kafka's own offset. Native, atomic. The wire-level `sequence` IS the Kafka offset.
```
nonameffh: Kafka offsets start at 0 and are non-dense (compaction, retention head-trim, transaction markers), whereas ADR-0001 mandates sequence ≥ 1, monotonic, last-processed semantics. Equating the consumer-visible sequence with the raw Kafka offset (or even offset + 1) isn't fully correct. 

Suggestion: replace by "Kafka backend: the broker assigns its own 1-based sequence at persist and indexes it against the Kafka offset (Kafka offsets are 0-based and non-dense, so they cannot be used directly as the wire sequence)"

gears/system/event-broker/event-broker-sdk/src/producer/sync_producer.rs
```diff
@@ -0,0 +1,471 @@
...
+                    let mut guard = self.resolved_type_ids.write().await;
+                    guard.insert(type_id.to_owned());
+                }
+                self.schema_cache.validate(type_id, &data).await?;
+            }
+        }
+
+        // Step 4: Compute partition.
+        let partition_count = self.get_partition_count(topic).await;
+        let key = partition_key_str.as_deref().unwrap_or(subject_str.as_ref());
```
Artifizer: I think there’s still a contract mismatch around partition fallback that’s worth clearing up. In the design docs, the fallback reads as `tenant_id` when `partition_key` is missing, but here the SDK falls back to `subject`. That changes which events land on the same partition, so it would be good to make docs and code converge on one rule.

gears/system/event-broker/docs/DESIGN.md
```diff
@@ -0,0 +1,2584 @@
...
+- future ADR (idempotent-producers) — Idempotent producers via Producer-Id + per-`(producer_id, topic, partition)` sequence tracking. No epoch fencing — sequence ordering itself acts as the fence.
+- future ADR (consumer-lifecycle) — Subscription-based consumers with session timeout and offset tracking
+- future ADR (outbox-ingest) — Outbox-based producer library and ingest pipeline built on `modkit-db` outbox
+- future ADR (cluster-capabilities) — Platform-level cluster coordination abstraction (pub/sub, leader election, locks, discovery) with pluggable providers
+- future ADR (long-polling) — Notification-based long-poll via ClusterCapabilities pub/sub
+- future ADR (topic-sharding) — Topic-sharded ingest and consumer-group-sharded delivery in cluster mode
+- future ADR (dispatcher) — Stateless HTTP router with topic-based and consumer-group sticky-session routing
+- `cpt-cf-evbk-adr-partition-selection` — Partition selection algorithm: MurmurHash3-32 over `partition_key` with producer override and broker-side range validation. See [`ADR/0002-partition-selection.md`](ADR/0002-partition-selection.md).
```
Artifizer: This still reads like there is a producer override path for partition selection, while other parts of the PR say the broker is fully authoritative and explicit producer-set partitioning is gone. I think the intended direction is probably fine, but the wording here still feels like a mix of old and new contracts.

gears/system/event-broker/docs/DESIGN.md
```diff
@@ -0,0 +1,2584 @@
...
+- **DB-restore resync tooling**: a CLI to recover from a broker DB restore by realigning `evbk_producer_state.last_sequence` per `(producer_id, topic, partition)` to match the storage backend's actual state — **producer-cursor desync**. (Consumer-cursor desync is no longer a DB-level concern: consumer cursors live in the ClusterCapabilities cache, not in SQL — see §3.6 Cursor.) Until the tooling exists, manual operator intervention (`UPDATE evbk_producer_state SET last_sequence = …`; consumer-side seek or cache flush as needed) is the only recovery path.
+- **Server-side persistent offset KV**: a stateless offset-storage endpoint at `GET / PUT /v1/groups/{group}/offsets`, keyed by `consumer_group`, decoupled from subscription lifecycle. Consumers that need offset durability stronger than what the ClusterCapabilities cache provides (e.g., standalone mode where the cache is in-process) can opt in by writing offsets here. Reserved namespace; not in MVP.
```
Artifizer: I think the offset ownership story is still slightly out of sync here. `ADR-0006` says broker-managed offset durability is intentionally dropped, but this section still reserves a future server-side offset KV endpoint. For someone new to the feature, that makes it hard to tell what the final model actually is.

gears/system/event-broker/event-broker-sdk/src/consumer/builder.rs
```diff
@@ -0,0 +1,927 @@
...
+// -- NoDlq + WithTx (db feature) --
+
+#[cfg(feature = "db")]
+impl<M, H> ConsumerReady<NoDlq, WithTx<M>, H>
+where
+    M: CommitOffsetInTx + 'static,
+    H: EventHandler<TxCommitHandle<M>, HandlerOutcome> + 'static,
+{
+    pub async fn start(self) -> Result<ConsumerHandle, EventBrokerError> {
+        todo!("NoDlq + WithTx start - task 19")
```
Artifizer: This looks like a public builder/runtime path that users could reasonably assume is supported, but it still ends in `todo!()`. That can be pretty confusing from the outside because the type surface says ‘available’ while runtime behavior says ‘not ready yet’. I’d either finish this path or gate/hide it for now.

gears/system/event-broker/event-broker-sdk/src/internal/chain_state.rs
```diff
@@ -0,0 +1,50 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
```
Artifizer: Small consistency thing: this uses `std::sync::RwLock`, while the repo guidance generally prefers `parking_lot` for normal synchronous locking. Not a huge issue by itself, but aligning with the local convention would make the locking style more predictable unless there’s a specific reason for `std` here.

gears/system/event-broker/event-broker-sdk/README.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+---
+
+## Producer quick-start
+
+### Sync (direct to broker)
+
+```rust
+use event_broker_sdk::{ChainMode, EventBroker, ProducerBuilder, TypedEvent};
```
Artifizer: This example still uses `ChainMode`, while the current code seems to expose `ProducerMode`. Small thing, but when examples and public names drift like this it gets harder for readers to tell which terminology is current and which is leftover from an earlier draft.

gears/system/event-broker/docs/ADR/0006-offset-authority.md
striped-zebra-dev: Fixed. I removed this because the endpoint should not exist in the design at all.

The current event-broker docs/code no longer mention `/v1/groups/{group}/offsets` or `evbk_group_offsets`. Durable progress remains consumer-owned; the broker only keeps runtime/session cursor state.

gears/system/event-broker/docs/DESIGN.md
striped-zebra-dev: Fixed together with the other offset-KV thread.

I dropped the reserved/post-MVP server-side group offset KV wording instead of keeping it as a future extension. That keeps ADR-0006 consistent: durable consumer progress is owned by the consumer/application side, not by a broker-managed persistent offset store.

gears/system/event-broker/docs/ADR/0002-partition-selection.md
striped-zebra-dev: Fixed by narrowing the contract rather than trying to match native Kafka producers.

We do not have native Kafka producers in this design, and the public contract is not "same hash as Kafka." The contract is field-based and broker-authoritative: use `partition_key` when present, otherwise `tenant_id`; the broker derives the final topic partition. I removed the Kafka `DefaultPartitioner` / `BuiltInPartitioner` compatibility wording.

gears/system/event-broker/event-broker-sdk/src/producer/sync_producer.rs
striped-zebra-dev: Good catch. I kept the design contract as `partition_key` when present, otherwise `tenant_id`.

The current live SDK path goes through `producer/event_factory.rs` + `producer/partitioning.rs`, and that path already uses `tenant_id` fallback. I added a deterministic direct-producer regression test for that behavior and cleaned the stale docs that still mentioned `subject` or read like a producer partition override.

gears/system/event-broker/docs/DESIGN.md
striped-zebra-dev: Fixed. I kept the public sequence contract from ADR-0001 and stopped describing Kafka native offsets as the wire-level consumer sequence.

Kafka offsets are backend-internal now. At the backend boundary we map them to broker-logical values: `logical_sequence = kafka_offset + 1`, and `kafka_offset = logical_sequence - 1` for backend reads/seeks.

gears/system/event-broker/docs/DESIGN.md
striped-zebra-dev: Agreed, the wording was too easy to read as producer authority over the broker topic partition.

I tightened this to the intended contract: `partition_key` only changes the partition input, the broker remains authoritative for the final topic partition, and producer outbox partitions / broker topic partitions / ingest shards are separate partition domains. Their counts do not have to match; each domain just has to be deterministic inside its own partition count.

I also made the event `partition` field explicit as consumer-facing/read-side only: producers cannot set it, publish rejects it as read-only, and ingestion does not use a supplied top-level `partition` value for routing.

I also added a regression test with a 4-partition producer outbox and a 16-partition broker topic, so we pin that the envelope carries the correct broker partition while the row lands in the correct local outbox partition.

gears/system/event-broker/event-broker-sdk/src/consumer/builder.rs
striped-zebra-dev: I rechecked the current consumer builder path and this one is stale now.

There is no `todo!()` / `unimplemented!()` under `event-broker-sdk/src/consumer`, and the public terminal builder paths now return `Result<ConsumerHandle, EventBrokerError>` through implemented constructors. I am resolving this as already fixed in the current diff.

gears/system/event-broker/event-broker-sdk/src/internal/chain_state.rs
striped-zebra-dev: I do not want to make `parking_lot` a universal rule here.

For this SDK/framework state, poisoning is useful as an invariant tripwire: if a panic happens while mutating sequence/cursor state, silently continuing through a non-poisoning lock can hide corrupted state. So I am keeping `std::sync` at these sites unless we have a local reason to trade that semantic for `parking_lot` on a measured hot path.

gears/system/event-broker/event-broker-sdk/README.md
striped-zebra-dev: This one is stale in the current diff.

The README now imports and uses `ProducerMode`, and a targeted scan of the event-broker tree found no `ChainMode` token. The public API defines and re-exports `ProducerMode`, so there is no README/API change left for this thread.

gears/system/event-broker/event-broker-sdk/src/consumer/builder.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    pub async fn run_blocking(
+        self,
+        ctx: &SecurityContext,
+        cancel: tokio_util::sync::CancellationToken,
+    ) -> Result<(), EventBrokerError> {
+        let _ = (ctx, cancel);
+        todo!("WithDlq + WithTx run_blocking — task 19")
+    }
```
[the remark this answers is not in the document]
striped-zebra-dev: Stale. The current consumer builder has no `todo!()` / `unimplemented!()` path.

Cargo.toml
```diff
@@ -91,6 +91,7 @@ members = [
     "gears/system/oagw/oagw-sdk",
     "gears/mini-chat/mini-chat-sdk",
     "gears/mini-chat/mini-chat",
+    "modules/system/event-broker/event-broker-sdk",
 ]
```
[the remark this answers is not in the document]
striped-zebra-dev: Stale. The current SDK manifest has no `modkit` workspace dependency, and `cargo metadata --no-deps --format-version 1` passes.

gears/system/event-broker/event-broker-sdk/src/internal/schema_cache.rs
```diff
@@ -0,0 +1,180 @@
...
+            Err(errors)
+        }
+    }
+}
+
+/// Schema cache backed by `types-registry-sdk`.
+pub struct SchemaCache {
+    #[doc(hidden)]
+    pub inner: RwLock<HashMap<String, Arc<ResolvedSchema>>>,
+    client: Arc<dyn TypesRegistryClient>,
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `SchemaCache::inner` is now crate-visible, so it no longer leaks the crate-private `ResolvedSchema` type through the public API.

gears/system/event-broker/event-broker-sdk/src/api.rs
```diff
@@ -0,0 +1,584 @@
...
+/// let broker = hub.get::<dyn EventBroker>()?;
+/// ```
+///
+/// Implemented by: the in-process direct backend, the remote HTTP backend,
+/// and the mock (`--features test-util`). This boundary is **transport-agnostic** - no
+/// HTTP types leak through it, and method docs describe operations, not wire paths.
+/// HTTP verbs/paths (and their renames) live solely in `openapi.yaml` and the HTTP
+/// backend, the single source of truth.
+#[async_trait]
+pub trait EventBroker: Send + Sync {
```
fluiderson, line 408: `DESIGN.md` says about split `IngestApi`/`DeliveryApi` and unified `EventBrokerApi`. The SDK has only the latter. I think we should have no single everything-trait and leave only `IngestApi`/`DeliveryApi` in `DESIGN.md` and the SDK.

gears/system/event-broker/docs/ADR/0005-subscription-filter-typing.md
```diff
@@ -0,0 +1,388 @@
...
+- Apache Kafka — topic-anchored subscription model, referenced as the prior art for "topic is the partition / rebalance / authz unit."
+
+## Traceability
+
+- **PRD**: [PRD.md](../PRD.md)
+  - `cpt-cf-evbk-fr-subscription-join` — JOIN body shape (this ADR pins the interests[] structure).
+  - `cpt-cf-evbk-fr-filter-expression` — new FR for typed filter engines + plugin pattern.
+  - `cpt-cf-evbk-nfr-tenant-rate-caps` — extended to cover `POST /v1/subscriptions` per-tenant rate cap.
+- **DESIGN**: [DESIGN.md](../DESIGN.md)
+  - §3.1 Subscription Schema — updated to reference the new interests[] body shape.
```
fluiderson, line 376: It seems `DESIGN.md` is outdated on this matter. This ADR describes it with `expression_type` + `expression`. `DESIGN.md` uses `filter { engine, expression }`.

gears/system/event-broker/docs/openapi.yaml
```diff
@@ -0,0 +1,778 @@
...
+          application/json:
+            schema:
+              type: object
+              required: [mode]
+              additionalProperties: false
+              properties:
+                mode:
+                  type: string
+                  enum: [chained, monotonic]
+                  description: Dedup protocol mode for this producer.
```
fluiderson, line 217: Where's requred `client_agent` from `0004-idempotent-producer-protocol.md`?

gears/system/event-broker/docs/PRD.md
```diff
@@ -0,0 +1,722 @@
...
+- **Rationale**: The two checks cover different concerns — capability vs correctness — and cannot substitute for each other.
+- **Actors**: `cpt-cf-evbk-actor-authz-resolver`
+
+### 5.7 Error Codes
+
+#### RFC 9457 Problem Details
+
+- [ ] `p1` - **ID**: `cpt-cf-evbk-fr-rfc9457-errors`
+
+All error responses MUST follow RFC 9457 Problem Details (`application/problem+json`) with stable GTS type identifiers under the `gts.cf.core.errors.err.v1~cf.core.<namespace>.<name>.v1` namespace. Each error MUST carry `type`, `title`, `status`, `detail`, `instance`, optionally `trace_id` and `retry_after_seconds`.
```
fluiderson, line 476: Is this implied in `openapi.yaml`? If not, should we explicitly add this?

docs/GEARS.md
```diff
@@ -699,6 +699,19 @@ Introduces an abstraction layer behind the real Outbound API Gateway. The main g
 - [API](../gears/system/oagw/oagw/README.md)
 - [SDK](../gears/system/oagw/oagw-sdk/README.md)
 
+### Event Broker
+
+Multi-consumer, partitioned, append-only event streaming for Cyber Ware modules.
+Typed events, at-least-once delivery, idempotent producers (chained/monotonic/stateless),
+pluggable storage backends, consumer-group cursor tracking.
+
+**Status**: SDK landed (`cyberware-event-broker-sdk`) — impl crate TODO.
```
fluiderson, line 708:
```suggestion
**Status**: SDK landed (`cf-gears-event-broker-sdk`) — impl crate TODO.
```

gears/system/event-broker/docs/ADR/0002-partition-selection.md
```diff
@@ -0,0 +1,265 @@
...
+
+A topic in the Gears event broker is divided into a fixed number of partitions declared at topic registration time (`Topic.partitions`, see DESIGN §3.1). Every `Event` is bound to a partition; sequence assignment, ordering, idempotent-producer state, and consumer cursors are all scoped to `(topic, partition)`. The broker therefore needs a contract for **how partition assignment is computed** before the event is enqueued in the producer outbox and ultimately landed in `backend.persist`.
+
+The existing design imposes hard constraints on this decision:
+
+- `Topic.partitions` is **fixed at topic creation** and cannot be grown or shrunk on a live topic. Re-partitioning would break per-key ordering for every key already published; the migration path is "create new topic, dual-write, cut consumers over." Partition selection MUST therefore behave deterministically across the topic's full lifetime.
+- The broker assigns the consumer-visible `event.sequence` per `(topic, partition)`. Producers never set `sequence`; they only carry chain state for ingest-side dedup in `meta.previous` and `meta.sequence` (see [ADR-0003 Event Schema](0002-event-schema.md)).
```
fluiderson, line 46:
```suggestion
- The broker assigns the consumer-visible `event.sequence` per `(topic, partition)`. Producers never set `sequence`; they only carry chain state for ingest-side dedup in `meta.previous` and `meta.sequence` (see [ADR-0003 Event Schema](0003-event-schema.md)).
```

There's more stale links like that. Please go over all references and ensure that they point at correct files.

gears/system/event-broker/scenarios/consumer/flows/1.01-flow-two-consumer-rebalance.md
```diff
@@ -0,0 +1,191 @@
...
+    "gts.cf.core.events.topic.v1~acme.orders.v1:3": "earliest"
+  }
+}
+```
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{ "partition_positions": { "gts.cf.core.events.topic.v1~acme.orders.v1:0": 99, "gts.cf.core.events.topic.v1~acme.orders.v1:1": -1, "gts.cf.core.events.topic.v1~acme.orders.v1:2": -1, "gts.cf.core.events.topic.v1~acme.orders.v1:3": -1 } }
```
fluiderson, line 64:
```suggestion
{ "partition_positions": { "gts.cf.core.events.topic.v1~acme.orders.v1:0": 99, "gts.cf.core.events.topic.v1~acme.orders.v1:1": 0, "gts.cf.core.events.topic.v1~acme.orders.v1:2": 0, "gts.cf.core.events.topic.v1~acme.orders.v1:3": 0 } }
```

Shouldn't `partition_positions` start from 0 as was described in `ADR/0001-offset-semantics.md`?

gears/system/event-broker/scenarios/INDEX.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Integration Journey
+
+To publish and consume events through the broker, follow these steps. Each step links to the scenario(s) with full coverage.
+
+#### Step 1 — Establish a consumer group
+
+Anonymous groups are broker-minted via REST. Named groups are registered via `types_registry` at startup — no REST call needed.
+
+- [Create anonymous group](consumer/groups/1.01-positive-create-anonymous-group.md) — `POST /v1/consumer_groups` → broker mints GTS id; distribute it to your consumer fleet out-of-band.
```
fluiderson, line 22:
```suggestion
- [Create anonymous group](consumer/groups/1.01-positive-create-anonymous-group.md) — `POST /v1/consumer-groups` → broker mints GTS id; distribute it to your consumer fleet out-of-band.
```

Or we can change `openapi.yaml`. I suppose there are more places where this endpoint has different casings. They should be the same.

gears/system/event-broker/scenarios/topics/1.04-positive-list-event-types.md
```diff
@@ -0,0 +1,45 @@
+# List event types
+
+Returns a paged list of event types registered in `types_registry`. Producers use this to discover which event types are available on a given topic; consumers use it to understand the data schemas they'll receive.
+
+## Request
+
+```http
+GET /v1/event_types?topic=gts.cf.core.events.topic.v1~acme.orders.v1&limit=20 HTTP/1.1
```
fluiderson, line 8:
```suggestion
GET /v1/event-types?topic=gts.cf.core.events.topic.v1~acme.orders.v1&limit=20 HTTP/1.1
```

Ditto.

gears/system/event-broker/event-broker-sdk/src/producer/direct.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    pub async fn publish_batch<E: TypedEvent>(
+        &self,
+        events: Vec<E>,
+    ) -> Result<Vec<IngestOutcome>, EventBrokerError> {
+        let mut prepared = Vec::with_capacity(events.len());
+        let mut batch_sequence_state = self
+            .sequence_state
+            .read()
+            .expect("producer sequence state lock poisoned")
```
fluiderson, line 290: Should we simply panic here? Can we use `EventBrokerError::Internal` or similar?

gears/system/event-broker/event-broker-sdk/src/consumer/runtime.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    /// Current subscription ids (one per active parallelism slot).
+    pub fn subscription_ids(&self) -> Vec<SubscriptionId> {
+        self.consumer.subscription_ids()
+    }
+}
+
+impl Consumer {
+    pub fn new(parallelism: u32) -> Self {
+        let _ = parallelism;
```
fluiderson, line 235: Why do we need this argument?

gears/system/event-broker/event-broker-sdk/src/consumer/types.rs
```diff
@@ -0,0 +1,829 @@
...
+pub enum BatchHandlerOutcome {
+    Success,
+    AdvanceThrough { offset: i64 },
+    Retry { reason: String },
+}
+
+/// Batch cursor over events from one topic partition.
+pub struct EventBatch<'a> {
+    events: &'a [RawEvent],
+    cursor: usize,
```
fluiderson, line 56: It looks like it's used as a constant that equals 0. Is it updated somewhere?

gears/system/event-broker/event-broker-sdk/src/consumer/runtime.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+fn gts_pattern_matches(pattern: &str, value: &str) -> bool {
+    if pattern == "*" {
+        return true;
+    }
+    pattern
+        .strip_suffix('*')
+        .map_or(pattern == value, |prefix| value.starts_with(prefix))
+}
```
fluiderson, line 215: https://github.com/striped-zebra-dev/gears-rust/blob/8fcc371dd3f0c8e6278ce18eed9d68798d9702b4/gears/system/event-broker/event-broker-sdk/src/producer/schema_cache.rs#L236-L242

Should they match? And why not use [the `gts` crate](https://docs.rs/gts) here?

fluiderson (changes requested):
In overall, there are a lot of minor mismatches between documents and the code. Please fix them first.

gears/system/event-broker/event-broker-sdk/src/api.rs
striped-zebra-dev: We'll go with a single unified trait, but adopt the name `EventBrokerApi` (the one DESIGN.md already uses). The SDK trait `EventBroker` is renamed to `EventBrokerApi`, and the `IngestApi`/`DeliveryApi` split - which only ever existed in DESIGN.md, never in code - is dropped from the docs. So one trait, one name, docs and code aligned.

Done in this PR: renamed the trait and all references crate-wide (`api.rs`, all `dyn`/`impl`/`use` sites, `lib.rs` re-export), updated `DESIGN.md` (dropped the split) and `README.md`; `cargo build` + test targets compile.

gears/system/event-broker/docs/ADR/0005-subscription-filter-typing.md
striped-zebra-dev: Good catch on the mismatch - though the fix direction is the reverse of the comment. DESIGN.md §3.1 and the authoritative `docs/schemas/subscription.v1.schema.json` both already use the nested `filter { engine, expression }` shape, and DESIGN explicitly states it *replaces* the flat `expression_type` + `expression` fields. So DESIGN is the current one; ADR-0005 (and the openapi prose) were the stale artifacts.

Fixed in this PR by making the nested shape canonical everywhere: rewrote ADR-0005 (JSON example, body table, paired-optional rule → filter-object rule, JOIN validation steps 5e/5g/5h/5i, delivery pseudocode, and the consequences/confirmation wording), the openapi prose (`400`/`UnknownFilterEngine` now key off `filter.engine`/`filter.expression`), PRD `cpt-cf-evbk-fr-subscription-join` + `-filter-expression`, feature 0002 (JSON + AC-9/10/11 + test matrix), and scenario 1.03. Semantics: `filter` is optional; when present both `engine` and `expression` are required (a `filter` missing either → `400 BadRequest`); absent means match by topic+tenant+types. Also reconciled scenario 1.03's engine id `cf.events.expression.cel.v1` → `cf.core.expression.cel.v1` to match the ADR.

gears/system/event-broker/docs/openapi.yaml
striped-zebra-dev: Right - `client_agent` was missing from the OpenAPI spec. Worth noting it was a spec-only gap: the SDK request model (`String`, not `Option`), the mock validation (ASCII, 1-256 bytes), the DB column (`NOT NULL`), and ADR-0004 all already treat it as required + immutable.

Fixed in this PR: `POST /v1/producers` request schema now has `client_agent` (RFC 9110 User-Agent grammar, ASCII, 1-256 bytes) with `required: [mode, client_agent]`, keeping `additionalProperties: false`; the 201 response now echoes `client_agent` and `mode` (was `id`-only). No code change needed - the implementation already enforces it.

gears/system/event-broker/docs/PRD.md
striped-zebra-dev: Yes - modeled it explicitly. The RFC 9457 representation itself stays owned by `toolkit-canonical-errors` (EB inherits it: no EB-specific wire format, canonical `gts://` `type`), but there's no shared OpenAPI schema to `$ref`, and the peer gears (account-management, usage-collector) each declare a local `Problem` schema + `application/problem+json` responses inline. EB was the outlier (prose-only).

Fixed in this PR:
- Added `components/schemas/Problem` (RFC 9457 envelope, `required: [type, title, status, detail]` + `instance`/`trace_id`/`context`; `type` is the canonical `gts://gts.cf.core.errors.err.v1~cf.core.err.<category>.v1~` URI; described as produced by `toolkit-canonical-errors`), mirroring account-management.
- Added `components/responses/Problem` (`application/problem+json`).
- Applied `application/problem+json` -> `Problem` to all 41 4xx/5xx responses, keeping each response's existing prose (InvalidMode, SequenceViolation, UnknownFilterEngine, …).
- `retry_after_seconds` lives in `Problem.context`; the `Retry-After` header stays the canonical-errors layer's concern (503). PRD `cpt-cf-evbk-fr-rfc9457-errors` reworded accordingly.

docs/GEARS.md
striped-zebra-dev: Fixed - `cf-gears-event-broker-sdk` per `Cargo.toml`. This was the only stale reference in the tree; every other doc already uses the correct name (verified no `cyberware-` references remain).

gears/system/event-broker/docs/ADR/0002-partition-selection.md
striped-zebra-dev: Fixed all of them - it was a systematic off-by-one from a file renumbering: 19 relative links across the ADR folder (`0002-event-schema.md` → `0003-event-schema.md` ×14, `0003-idempotent-producer-protocol.md` → `0004-idempotent-producer-protocol.md` ×5, targets + backtick labels). All resolve now.

Root cause: the repo-root `lychee` check (`make lychee`) only scanned `docs examples tools guidelines`, never `gears/**`, so module docs were unchecked. I've extended it to `gears/system/event-broker/docs`. (A full `gears` scan surfaces ~34 pre-existing broken links across other gears, so broadening beyond event-broker is deferred to a separate cleanup rather than dragged into this PR.)

Per "go over all references", the now-in-scope check also caught two more broken links in the module, both fixed:
- `PRD.md` `guidelines/` link (`../guidelines/` → `../../../../guidelines/`).
- `DESIGN.md` referenced a `REVIEW.md` appendix that was never produced - removed the two dangling references.

`lychee --offline gears/system/event-broker/docs` is now clean (350 links, 0 errors).

gears/system/event-broker/scenarios/consumer/flows/1.01-flow-two-consumer-rebalance.md
striped-zebra-dev: Correct, and thanks. Per ADR-0001, `"earliest"` resolves to `cursor = RF - 1` with `RF >= 1`, so a resolved position is never negative - a fresh partition (RF = 1) resolves to `0`. Here partition 0 has a raised retention floor (RF = 100 -> `99`, unchanged), while partitions 1/2/3 are fresh -> `0`.

Fixed at line 64 (p1/p2/p3 `-1` -> `0`) and also at line 158, where the same `-1` appeared for p2/p3 in Exchange 5.

gears/system/event-broker/scenarios/INDEX.md
striped-zebra-dev: Fixed - the scenarios/docs now use `/v1/consumer-groups` (hyphen). The OpenAPI spec is canonical and already uses the hyphen, and hyphen-case for multi-word URL path segments is the unanimous convention across every gear (resource-group `types-registry`, usage-collector `usage-types`, account-management `child-conversions`; none use underscores in paths).

Underscore stays for the non-URL identifiers, which the edit deliberately left intact on shared lines: the GTS type `consumer_group` (singular), the DB table `evbk_consumer_group`, JSON fields, and the Rust filename `consumer_groups.rs`.

gears/system/event-broker/scenarios/topics/1.04-positive-list-event-types.md
striped-zebra-dev: Same fix - the 3 URL-path occurrences now use `/v1/event-types` (`scenarios/topics/1.04`, `scenarios/INDEX.md`, and a comment in `mock/tests/topics.rs`). The OpenAPI spec was already correct.

The Rust module file `event_types.rs`, the `list_event_types()` method, and the `event_types` struct/map fields keep snake_case - only literal URL path segments follow the hyphen convention.

gears/system/event-broker/event-broker-sdk/src/producer/direct.rs
striped-zebra-dev: Good question, and it turned into more than a one-liner - so it's being split into its own change/PR rather than folded in here.

Investigation: the state behind those locks is a rebuildable cache (the direct producer re-primes from `get_producer_cursors`; the outbox producer refreshes from the DB), and the critical sections are single infallible `HashMap` ops never held across `.await` - so they can't leave torn state. That means mapping to `EventBrokerError::Internal` would actually be the wrong fix: with `std::sync` the lock stays poisoned for its lifetime, permanently bricking the producer.

Chosen approach (separate PR): keep `std::sync`, and on a poisoned lock recover explicitly - `into_inner()` + `clear_poison()`, re-prime from the authority, and fall back to `Internal` only if that recovery itself fails. Since it's a behavior/reliability change deserving its own tests, these 9 sites are untouched in this PR; tracked in `event-broker-producer-lock-poisoning`.

gears/system/event-broker/event-broker-sdk/src/consumer/runtime.rs
striped-zebra-dev: Removed. `parallelism` legitimately comes from the `ConsumerBuilder` (`.parallelism(n)`, default 1) and is consumed by the real slot-building constructors (`new_with_batch_slots_with_cancel` / `new_with_tx_batch_slots`). `Consumer::new` is a test-only shell that discarded the arg and couldn't honor it anyway (it has none of the builder context needed to build slots), so the parameter was pure dead weight.

Done: dropped the param (and the `let _ = parallelism;`), updated the one test caller `Consumer::new(2)` → `Consumer::new()`, and added `impl Default for Consumer` since the now-argless public `new()` tripped clippy's `new_without_default`. `cargo clippy` (lib) is clean and `cargo test --lib` passes (65).

gears/system/event-broker/event-broker-sdk/src/consumer/types.rs
striped-zebra-dev: Confirmed - `cursor` was a constant `0` and never advanced. We first covered
`EventBatch` with `chunk → error → retry → advance` combination tests, which show progress is
driven entirely by the handler outcome offset + the committed frontier: on `Retry` the
frontier isn't advanced and a fresh `EventBatch` re-anchors the redelivered tail from the
front, so the batch cursor plays no part. Given that, removed the dead `cursor` field -
`EventBatch` is now an explicit stateless view (`next_event` = front, `next_chunk(n)` = first
`n`). Names kept; all tests green.

gears/system/event-broker/event-broker-sdk/src/consumer/runtime.rs
striped-zebra-dev: We're keeping both for now. Consumer and producer are separate packages with no shared code by design, so extracting a shared helper would introduce coupling we don't want, and pulling in the `gts` crate here would be premature - a types-registry SDK is expected to own GTS wildcard semantics, and that's the natural place to consolidate this (rather than hand-rolling a third variant now).

You're right that the two implementations diverge (the producer's `strip(".*")` can over-match relative to GTS §10 segment-boundary rules, e.g. `foo.*` vs `foobar`). We've captured that as a follow-up to resolve when the types-registry SDK lands. No code change in this PR.

PR #4084, opened 2026-06-12
feat(toolkit-contract): add contract stack for REST and gRPC projections

MikeFalcon77:
Introduce the Toolkit contract system: a Rust trait is the single source of truth, projected onto REST, gRPC, and in-process transports, with consumers resolving a binding-agnostic `Arc<dyn Contract>` through `ClientHub`.

- add `toolkit-contract`, `toolkit-contract-macros`, and `toolkit-contract-protogen`
- add contract descriptors, IR validation, OpenAPI generation, and runtime support
- add REST and gRPC contract/codegen support with proto lockfile handling
- extend canonical error wire-compatibility support
- add macro UI tests and contract integration tests
- add `examples/toolkit/api-contracts` sample gear and SDK

## Correctness fixes

Four bugs that lived in code paths the test suite never exercised. Each is now covered by a regression test that was verified to fail without its fix.

- **Path parameters could be silently swapped.** The generated server bound multiple `{param}` segments into an axum `Path<(T0, T1)>` tuple in *trait declaration* order, but axum fills that tuple *positionally* from the URL. With two same-typed path parameters declared out of template order, the client (which substitutes by name) sent a correct request and the handler decoded it wrong — no type error, no 400. The tuple is now ordered by the template, and a placeholder with no matching parameter is a `compile_error!`.
- **The generated gRPC client panicked on peer data.** `#[derive(ProtoBridge)]` emits an infallible `From<Proto>` that `.expect()`s on `#[proto_bridge(via_string)]` fields; the client used it on all three inbound response paths, so one malformed UUID from a peer took the process down. Responses now decode through a new fallible `TryFromProto` trait, implemented for both structs and enums so codegen can convert any response type uniformly.
- **The "provider not up yet" branch was dead.** `DirectoryNotFound` was defined but never constructed anywhere, so `DirectoryEndpointResolver` could not return `Ok(None)` and every unregistered provider looked like a directory outage — `warn!` on every readiness tick *and* on every business call through a resolving client, unbounded. Both directory clients now construct the sentinel (the gRPC one from `Code::NotFound`), and the orchestrator emits that code only for a genuine not-found.
- **A remote proxy was indistinguishable from a local impl.** `ClientHub` is keyed by type alone, so with two consumers of one contract in a process the second found the first's *proxy*, reported `WireOutcome::Local`, and marked the shared dependency readiness-resolved — `/readyz` went 200 with no provider up. `ClientHub` now records provenance (`register_remote_proxy` / `try_get_local`).

## Query parameters

Client and server never shared a codec: the client flattened the parameter to JSON and hand-rolled a query string, the server decoded via `serde_urlencoded` through `axum::extract::Query`, and the spec was inferred separately. Every non-trivial shape failed — a `Vec<T>` went out as repeated keys that `serde_urlencoded` cannot collect into a sequence, and even a bare scalar decoded as a map at the top level (both guaranteed 400s, undetected because the only scalar query parameter in the suite was `#[server_manual]`). Flat-struct query parameters were absent from the OpenAPI document entirely, since the macro cannot see a struct's fields.

Both ends now share `serde_html_form`, and a query parameter must be a struct deriving `QueryParams`:

- `Vec<T>` fields round-trip as repeated keys, documented as `style: form, explode: true` arrays;
- the derive emits the OpenAPI parameter list the route registers, so the spec and the wire format come from one declaration;
- nested structs and bare scalars are rejected at compile time rather than 400ing at runtime.

## Known limitation

**A resolved dependency is not verified to be a conformant one.** Readiness flips as soon as the directory returns any endpoint for the gear name — nothing checks path, schema, or major version. A consumer of `PaymentApiV2` will go Ready against a provider serving only `/v1/**` and 404 on its first call. There is no contract identity on the wire; schema evolution rests on `oasdiff` in CI. Accepted deliberately and documented in ADR-0004; closing it needs a wire-protocol addition on both sides and belongs in its own ADR.

## Documentation

- **ADR-0008** (new) — the gRPC transport projection, which had shipped while the PRD and DESIGN still called it future work and an open question. Records the code-first-with-committed-`.proto` decision, the `TryFromProto` rule, streaming scope, and two deliberate asymmetries with REST.
- **ADR-0002** — amended: repeated query parameters are now in scope; nested query objects stay out.
- **ADR-0007** — §2 contradicted the section below it on version spelling; the code follows the latter.
- **PRD / DESIGN** — corrected requirements the implementation does not meet (the `#[path]`/`#[query]`/`#[header]` annotations were never built; REST-only projection methods are rejected by ADR-0003, not merely unbuilt; `schemars`/`utoipa` are not feature-gated), and reframed "base trait purity" as *zero transport annotations* — a base trait does carry `#[toolkit::contract]` and `#[idempotency]`.

Also bumps `lru` to 0.18.2 for RUSTSEC-2026-0253, which was blocking every PR rather than just this one.

## Summary by CodeRabbit

* **New Features**
  * Added contract and API binding framework supporting REST and gRPC transports with automatic client/server code generation
  * Introduced payment service example demonstrating contract framework usage
  * Added REST endpoint discovery to the directory service for dynamic service location

* **Documentation**
  * Added architecture decision records documenting contract binding design, server-side route generation, and consumer wiring patterns
  * Published typed-error wire format compatibility guide

* **Improvements**
  * Enhanced error handling with typed error categories and RFC 9457 problem format support
  * Improved protocol buffer generation with stable field-number assignment via lockfile mechanism


docs/arch/toolkit-contract-binding/ADR/0003-cpt-cf-binding-adr-projection-server-gen.md
```diff
@@ -0,0 +1,355 @@
...
+3. [Considered Options](#considered-options)
+4. [Decision Outcome](#decision-outcome)
+5. [Pros and Cons of the Options](#pros-and-cons-of-the-options)
+6. [More Information](#more-information)
+
+## Context and Problem Statement
+
+ADR-0001 introduced a two-layer architecture: a clean **base trait** (zero transport annotations)
+and a **projection trait** (`*Rest`, `*Grpc`) that carries transport annotations and is processed
+by `#[modkit_rest_contract]`. The macro currently generates a REST client struct from the
```
striped-zebra-dev: `#[modkit_rest_contract]` -> `#[toolkit_rest_contract]`

docs/arch/toolkit-contract-binding/ADR/0003-cpt-cf-binding-adr-projection-server-gen.md
```diff
@@ -0,0 +1,355 @@
...
+projection. Two gaps remain unfixed.
+
+**Gap 1 — no parity enforcement.** There is no compile-time guarantee that the projection trait
+covers the same methods as the base with matching signatures. A rename or new parameter on the
+base that is not propagated to the projection silently creates a stale client — the old method
+signature is still compiled, the mismatch is invisible until a runtime call fails.
+
+**Gap 2 — two independent OpenAPI sources.** Server-side handler code (`routes.rs`,
```
striped-zebra-dev, line 44: 1. Since the SDK is compiled into the service module, API compatibility is largely the responsibility of the service developer.
2. The real question is whether we can guarantee, through automation, that the SDK stays compatible with the server API. While we can introduce automated validation, we should not assume a one-to-one mapping between server and SDK APIs. SDK authors may intentionally expose only part of the server API or combine multiple server operations into a single SDK method.

docs/arch/toolkit-contract-binding/ADR/0003-cpt-cf-binding-adr-projection-server-gen.md
```diff
@@ -0,0 +1,355 @@
...
+This ADR decides how to close both gaps while keeping the two-layer design intact.
+
+## Decision Drivers
+
+* **Keep base trait clean** — the base trait is the domain interface; it must carry zero transport
+  annotations. Authors and IDE users reading the base trait see only the domain contract.
+* **Scale across multiple transports** — a contract may have both a REST and a gRPC projection;
+  if annotations for all transports were placed on the base, every method would accumulate a stack
+  of per-transport attributes. Each projection must remain a self-contained, transport-specific file.
+* **Single OpenAPI source of truth** — the same IR that drives REST client generation must also
```
striped-zebra-dev, line 59: As I mentioned above, I'd like to challenge this statement.

docs/arch/toolkit-contract-binding/ADR/0003-cpt-cf-binding-adr-projection-server-gen.md
```diff
@@ -0,0 +1,355 @@
...
+* **Zero migration cost** — existing projection traits must continue to compile unchanged; new
+  features are additive.
+* **Preserve the escape hatch** — ADR-0001 and ADR-0002 promise manual implementation as an
+  opt-out; that promise is kept.
+* **Preserve ADR-0002 scope** — the macro covers the same narrow REST subset; the annotation
+  vocabulary grows only by the additions already anticipated in ADR-0002 § Phase 2.
+
+## Considered Options
+
+* **Option A**: Extend the `#[modkit::rest_contract]` projection macro to (a) enforce method-set
```
striped-zebra-dev: I guess `modkit` is all over the place.

docs/arch/toolkit-contract-binding/ADR/0004-cpt-cf-binding-adr-consumer-wiring.md
```diff
@@ -0,0 +1,305 @@
...
+ADR-0001 defines how a module *provides* a contract to its consumers via `#[modkit::provides]`. There
+is no symmetric consumer-side counterpart. In the current PoC (`c280de1`), a consumer that needs a
+remote implementation must either:
+
+1. Load the **provider's implementation crate** as a stub compiled with `transport = rest` — coupling
+   the consumer's binary to the provider's internal types and build artifacts, or
+2. Hard-code a static endpoint string in `ClientWiring::Rest { endpoint }` and call `wire_*()`
+   manually in `init()` — with no integration with service discovery and no readiness gating.
+
+The `DirectoryService` was extended in `c280de1` with `resolve_rest_service(name)`, which maps a
```
striped-zebra-dev, line 30: `DirectoryService` - let's use a single name for this entity everywhere, I believe it should be `FlightControl`

docs/arch/toolkit-contract-binding/ADR/0004-cpt-cf-binding-adr-consumer-wiring.md
```diff
@@ -0,0 +1,305 @@
...
+There is no static mapping — a module may provide multiple contracts — and a convention-based
+name→`TypeId` lookup would require either fragile string matching or a global side-table populated
+by provider-side inventory items, reintroducing provider-crate linkage. Option C is the status quo;
+it does not satisfy the SDK-only or discovery requirements.
+
+### Macro shape
+
+```rust
+#[modkit::module(name = "orders")]
+#[modkit::consumes(contract = billing_sdk::BillingApi, from = "billing")]
```
striped-zebra-dev: What is `from`?
Why it is hardcoded,  and not configured?
In my understanding, the FlightControl shall have metadata-based resolution capability, and the operator might want to change the provider for some APIs.

docs/arch/toolkit-contract-binding/ADR/0004-cpt-cf-binding-adr-consumer-wiring.md
```diff
@@ -0,0 +1,305 @@
...
+registered it returns immediately without constructing an HTTP client or making a discovery call. No
+polling task is spawned for Profile 1 builds.
+
+### Static endpoint override (escape hatch)
+
+For local development and integration tests a static endpoint overrides discovery:
+
+```toml
+# config.yaml (development / test only)
+modules.orders.wiring.billing = "http://localhost:8081"
```
striped-zebra-dev: Probably, this setting could be extended to be used as a main routing mechanics;
If we add metadata-based resolution here, then we might have something like that:
```yaml
-- EB ingest instance A
serve_topics:
  - cf.core.*

modules.wiring.eb_backend = "serve_topic: cf.core.*" -- some meta filtering
 ```

File: docs/typed-errors-wire-compatibility.md [hunk 8]
```diff
@@ -0,0 +1,302 @@
+# Typed Errors Wire Compatibility — modkit-contract vs PRD #1536
```
striped-zebra-dev: If I understand this intend correctly, then Typed Errors - Canonical Errors, and we have thought about eveloping other types of errors.

libs/toolkit/src/directory.rs
```diff
@@ -182,4 +227,112 @@ mod tests {
...
+        dir.register_instance(inst);
+
+        let instances = api.list_instances("rest-only").await.unwrap();
+        assert_eq!(instances.len(), 1);
+        assert!(instances[0].grpc_services.is_empty());
+        assert_eq!(
+            instances[0].rest_endpoint.as_ref().unwrap().uri,
+            "http://127.0.0.1:8088"
+        );
+    }
```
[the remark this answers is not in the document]
MikeFalcon77: resolved

libs/toolkit/src/directory.rs
```diff
@@ -182,4 +227,112 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_register_instance_with_rest_endpoint() {
+        let dir = Arc::new(ModuleManager::new());
+        let api = LocalDirectoryClient::new(dir.clone());
+
+        let instance_id = Uuid::new_v4();
+        let register_info = RegisterInstanceInfo {
+            module: "billing".to_owned(),
+            instance_id: instance_id.to_string(),
```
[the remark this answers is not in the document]
MikeFalcon77: resolved

libs/toolkit/src/directory.rs
```diff
@@ -182,4 +227,112 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn test_resolve_rest_service_not_found() {
+        let dir = Arc::new(ModuleManager::new());
+        let api = LocalDirectoryClient::new(dir);
+
+        let result = api.resolve_rest_service("nonexistent_module").await;
+        assert!(result.is_err());
+    }
```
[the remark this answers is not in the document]
MikeFalcon77: resolved

libs/toolkit/src/readiness.rs
mattgarmon, line 1: There is already existing readiness check here: https://github.com/constructorfabric/gears-rust/blob/main/libs/toolkit/src/runtime/readiness.rs

We should merge onto the existing `runtime::readiness::ReadinessState` (add `register_dep` + `unresolved_deps()` so consumer-wiring can feed it)

libs/toolkit/src/discovery.rs
mattgarmon, line 1: This duplicates the existing OoP resolution path at https://github.com/constructorfabric/gears-rust/blob/main/libs/toolkit/src/runtime/oop_registration.rs

 `DirectoryEndpointResolver` and `resolve_one_dep` (oop_registration.rs) both resolve via `resolve_rest_service`. Let's feed the existing `runtime::readiness::ReadinessState` from `ConsumerRegistration.wire` and retire `resolve_deps/ResolvedRestEndpoints` rather than run two resolution loops side by side.

docs/arch/toolkit-contract-binding/ADR/0004-cpt-cf-binding-adr-consumer-wiring.md
MikeFalcon77: DirectoryService is a gear. But FlightControl is deployable service which includes DS, apigw(?) and etc

libs/toolkit/src/readiness.rs
MikeFalcon77: refactored

libs/toolkit/src/discovery.rs
MikeFalcon77: refactored

docs/typed-errors-wire-compatibility.md
MikeFalcon77: this is decsision file. removed

examples/toolkit/api-contracts/api-contracts-consumer/src/domain.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// whatever the backend `charge` returns.
+    pub async fn place_charge(
+        &self,
+        ctx: SecurityContext,
+        req: ChargeRequest,
+    ) -> Result<ChargeResponse, CanonicalError> {
+        let api = self
+            .hub
+            .get::<dyn PaymentApi>()
+            .map_err(|_| CanonicalError::service_unavailable().create())?;
```
MikeFalcon77: `map_err(|_| ...)` drops the `ClientHubError`, which distinguishes `NotFound { type_key }` from `TypeMismatch { type_key }`. Both collapse into a bare 503 whose detail is the default `"Service temporarily unavailable"`, so an operator can't tell a missing wiring from a wrong-type registration, nor which dependency failed.

The `ServiceUnavailableBuilder::with_detail` doc states the caller contract explicitly: preserve the precise reason "rather than collapsing every 503 into the same opaque message". `ClientHubError` renders a type name, not secrets, so it is safe for the public `Problem` body.

```rust
.map_err(|err| {
    CanonicalError::service_unavailable()
        .with_detail(format!("PaymentApi not wired: {err}"))
        .create()
})?;
```

This file is a template others will copy, so the pattern propagates.

MikeFalcon77: done

PR #4097, opened 2026-06-16
docs(file-storage): split into control plane + signed-URL sidecar (ADR-0003)

ffedoroff:
## Overview

Reworks the FileStorage design to a **two-plane architecture** (new **ADR-0003**, supersedes ADR-0001).

- **Control plane** = FileStorage **API / SDK**. Owns metadata, authorization, versioning, and conditional-request semantics. Its HTTP REST surface **never accepts or returns file content** — it issues short-lived **signed URLs** instead. The only path where control-side code touches bytes is the in-process SDK proxy mode, which streams inside the *consumer gear's* process.
- **Data-plane sidecar** = the only component that moves user bytes. Own domain/URL, connected to the storage backends, verifies signed-URL signatures **and** platform auth tokens, and reaches the control plane through the FS SDK (direct-DB or REST).

The presigned URL always points at **our sidecar**, never at the raw backend — so backend opacity, centralized per-byte metering, and uniform audit/policy coverage are all preserved, while the bandwidth-bound data plane scales independently of the control plane. Every content operation is therefore **at least two requests**: a control request to mint a signed URL, then one or more data requests against the sidecar.

---

## Identity & version types

A file's identity is separated from its content. Content is an **immutable blob per version**; the file is a small mutable record holding a pointer to the current version.

| Field          | Type           | Lives on        | Meaning / lifecycle |
|----------------|----------------|-----------------|---------------------|
| `file_id`      | `uuid`         | `files` (PK)    | The stable logical identity of a file. Never changes for the life of the file. Everything else hangs off it. |
| `version_id`   | `uuid`         | `file_versions` | Identity of **one immutable content blob**. Assigned by the control plane (not the backend) at pre-register. The backend object lives at `/{file_id}/{version_id}` and is **written once, never mutated**. |
| `content_id`   | `uuid` (nullable) | `files`      | The **content pointer**: the `version_id` currently bound as the file's live content. `NULL` until the first bind. Changing content is a *pointer swap* (`content_id := new version_id`), not an in-place mutation. |
| `meta_version` | `bigint`       | `files`         | Monotonic counter, bumped **only on metadata-only writes** (custom-metadata PATCH). Independent of content. |

Derived, not stored:

- **Full snapshot** = `(file_id, content_id, meta_version)` — uniquely pins the complete observable state (content axis × metadata axis, orthogonal).
- **ETag** (content-only) = opaque, derived from `(file_id, content_id)`:
  ```text
  etag_payload = file_id_bytes (16) || content_id_bytes (16)   # both UUIDs, 32 bytes
  etag_header  = '"' || base64url(etag_payload) || '"'
  ```
  It changes **exactly** when content is (re)bound; a metadata-only write does not touch it (so a CDN does not invalidate cached bytes on a metadata change). It is **never** equal to the content hash (a separate concern, ADR-0002). Restoring a prior version re-binds its `content_id`, so the ETag returns to that version's value (content-addressed).

**Write semantics.** A content write is: upload a new immutable version → **bind** it. Bind is an **optimistic CAS** on `content_id` guarded by `If-Match`. On conflict the bind returns `412` **and the `version_id`**; the client re-reads the current ETag and replays the bind with the fresh `If-Match` — **no byte re-upload**, because the version already exists durably. `meta_version` gives metadata its own validator via the optional `If-Match-Metadata: <u64>` header (absent → last-write-wins).

`content_id` replaces the old `content_revision` counter: it *is* the pointer (no separate counter to maintain), at the cost of monotonicity (use `created_at` for ordering).

---

## Signed URLs & token validation

Signed URLs are how the control plane delegates a single content operation to the sidecar. They are **stateless** (à-la S3 presigned: verifiable with no DB lookup).

### Signing (Ed25519, control-minted)

- **Asymmetric Ed25519.** The control plane signs with the **private key** and is the **sole minter**; the sidecar verifies with the **public key** and can never forge a URL. P1 ships **one static keypair** distributed by config (private in control, public in sidecar) — no rotation, no per-URL revocation; key rotation + keyset is P2. Distributing the public key (and the backend registry) is the only control→sidecar \"sync\".
- **Parameters** (query, `X-FS-*`): `X-FS-Algorithm=Ed25519`, `X-FS-KeyId`, `X-FS-Expires=<unix>`, `X-FS-Op=GET|PUT|part`, resource (`X-FS-FileId`, `X-FS-ContentId`/`X-FS-VersionId`, `X-FS-BackendId`), constraints (`X-FS-Ip`, `X-FS-Tok-<claim>`), baked response headers (`X-FS-Rh-<name>`), and finally `X-FS-Signature`.
- **Canonical string to sign** = `method` + `host` + `path` + **all** `X-FS-*` params except `X-FS-Signature`, sorted by key and percent-encoded (RFC 3986), joined `k=v&…`.
  - `host` is signed → a URL cannot be replayed against a different sidecar.
  - Signing *all* `X-FS-*` (minus the signature) means any added, removed, or altered parameter breaks verification — so there is **no separate `SignedParams` list**, and a client cannot strip or weaken a constraint.
- **Deliberately NOT signed:** the `Range` header (so one signed download URL serves many ranges — random access), conditional headers, and the `PUT` body (byte integrity is verified by the hash at bind, not the URL). Consequence: a signed `PUT` URL can be replayed with different bytes until `exp` → that only creates an orphan version/blob (swept by the P2 cleanup engine).

### Constraints (AND-combined)

A signed URL carries any number of constraints, all of which must hold:

- `exp` — expiry (required).
- `ip` / CIDR — optional client-address restriction.
- **Token-claim predicates** — optional `X-FS-Tok-<claim>=<value>` (e.g. `typ=user`, `sub=<id>`, `tenant_id=<id>`, any claim).

\"Available to everyone for 5 minutes\" = only `exp` (no token needed). \"Only a `user` token with `sub=123…`\" = `exp` + `X-FS-Tok-typ=user` + `X-FS-Tok-sub=123…`.

### Verification (at the sidecar)

On each request the sidecar:

1. Recomputes the signature over the canonical string using the public key named by `X-FS-KeyId`; rejects (`403`) on mismatch.
2. Checks `exp`; checks `ip`/CIDR if present.
3. **Token validation** — if (and only if) the URL carries one or more `X-FS-Tok-<claim>` predicates, the request **must** present a valid platform JWT. The sidecar validates that token the **standard platform way** and then checks each predicate (`claim == value`), all AND-combined. Any missing/invalid token or unmatched claim → `403`. A URL with no token predicate is served on signature + `exp`/`ip` alone.
4. Echoes the baked `X-FS-Rh-*` response headers verbatim (e.g. `Content-Disposition`, `Content-Type` override, `Cache-Control`) — no control round-trip.

This keeps **authorization decisions on the control plane** (made at presign, encoded into the URL) and the **sidecar a pure enforcer** of what the URL + token assert.

---

## Flows

**Upload (single-shot, P1):** `POST /files` (authz) → signed `PUT` url → client `PUT`s to the sidecar with `If-Match` → sidecar **pre-registers** a `pending` version via SDK (checks `If-Match` as a free early-fail, before any bytes) → streams to the backend computing size + SHA-256 → **auto-binds** under app-token + on-behalf-of user. `412` → sidecar returns `version_id` → client rebinds, no re-upload.

**Download (P1):** `GET /files/{id}/download-url` (authz) → signed `GET` url pinning the current `content_id` → client fetches from the sidecar (Range/conditional handled there).

**Multipart (P2):** server-authoritative parts plan (control returns exact part sizes/offsets + a signed URL per part); native backend multipart or offset-writes into the new-version object; per-part **BLAKE3 subtree** hashes persisted in `multipart_upload_parts.part_hash` and combined into the root at `complete`.

---

## Versioning & cleanup

- **Versioning is P1**, FileStorage-level and **backend-agnostic** (each version is a distinct immutable object + the `content_id` pointer) — works on any backend; the `versioning_native` capability is **removed**.
- No automatic cleanup in P1 (versions accumulate — the indefinite-retention default). The **cleanup engine** (layered version retention `≤ X` / age `T` + orphan reconciliation) is **P2**, a single internal control-plane background process.

---

## DB schema

- `files` slimmed to identity + pointer: `file_id, tenant_id, owner_kind, owner_id, name, gts_file_type, content_id, meta_version, timestamps` (no bytes, no per-content fields).
- New P1 `file_versions`: `(file_id, version_id, mime_type, size, hash_algorithm, hash_value, status pending|available, is_current, backend_id, backend_path, created_at)`.
- P2 hash-allow-list widening and P3 encryption columns now target `file_versions`.

---

## Files & validation

- `ADR/0003-…sidecar-data-plane.md` (new); `ADR/0001` → superseded; `ADR/0002` references repointed to the sidecar.
- `PRD.md`, `DESIGN.md`, `api.md`, `migration.sql` reworked.
- `cpt validate`: **0 errors**, code coverage 104/104.

> Draft — pending review.

## Summary by CodeRabbit

* **New Features**
  * Introduced a two-plane FileStorage flow: a control plane issues short-lived signed URLs, and a sidecar exclusively handles upload/download bytes, including sidecar-served `Range` behavior.
  * Added opaque PASETO (`v4.public`) signed URL tokens delivered via query (`fs-token`) or `X-FS-Token`, with strict expiry and claim validation.
* **Documentation**
  * Rewrote PRD/API docs to reflect signed-URL semantics, conditional headers, and pointer-based immutable versioning.
  * Added ADRs for the sidecar data plane and signed-URL transport; removed the deprecated proxy-traffic ADR.
* **Refactor**
  * Updated database migration to support immutable, versioned content and moved encryption metadata to per-version storage.


gears/file-storage/docs/DESIGN.md
```diff
@@ -1119,18 +1330,381 @@ Concurrency caps:
...
+signing step itself. Two cases:
+
+- **SDK in-process (private key local to the caller's control instance):** signing is a direct local call, so a caller
+  can mint **many URLs at once essentially for free** — e.g. 100 presigned URLs in single-digit milliseconds, no
+  round-trips. Useful for listing/galleries that need a signed download URL per item.
+- **Control plane as an external service (SDK over REST):** each presign is a control request, so to mint many URLs
+  the caller **batches** them into a single request (the control plane signs them all in one in-memory pass and
+  returns the set). That keeps bulk presigning to one round-trip — still fast.
+
+### 4.6 Worked example (LMS image upload and display)
```
ffedoroff, line 1493: @Artifizer please check, here is detailed example

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,182 @@
+---
+status: accepted
+date: 2026-06-17
+---
+
+# ADR-0004: Signed-URL Field & Signature Transport
```
ffedoroff: @Artifizer please check

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,182 @@
...
+## Pros and Cons of the Options
+
+### A. Discrete query parameters (chosen)
+
+* Good, because the result is a **bare, shareable URL** — embeddable in HTML, openable in a browser, fetchable by
+  `curl`, seekable by media players, with no headers required
+* Good, because it is the **canonical S3 SigV4 scheme** — ~20 years in production, universally understood, well-tooled
+* Good, because discrete fields are **debuggable** (visible/greppable in logs) and verified without an extra decode step
+* Bad, because re-presigning changes the URL → CDN cache-key churn unless normalized
+* Bad, because the signature appears in URLs (logs/Referer/history) — accepted, same as S3
```
Artifizer: Bad, because browser/frontend/proxy may log everything (maybe it's OK for some cases, but not OK for others), imagine some of the fields need to be hiddent (e.g. signature expiration)
Bad, because if I have a batch request I'll need to pass and parse parameters for every file

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,182 @@
...
+
+### B. Single opaque query token
+
+* Good, because it is compact and hides internal structure
+* Bad, because it must be base64-decoded and parsed before verification — extra ceremony, easy to get wrong
```
Artifizer: Does it mean 'query' parameters will not be in base64 and will not parse it? 
I think we need to have base64-decode, parsing and verification regardless of the URL vs header vs other options.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,182 @@
...
+* Good, because the signature stays out of the URL (not in access logs / Referer / history)
+* Bad, because it is **not a bare URL** — cannot be embedded as `<img src>` or opened in a browser; every consumer must
+  attach headers, which fails the primary download use case
+* Bad, because it diverges from the canonical query-signing mental model
+* Neutral — a reasonable **optional** transport for header-capable SDK/S2S callers; deferred
+
+### D. Single opaque HTTP-header token
+
+* Good, because URL is stable and the secret is out of the URL
+* Bad, because it combines the worst of B and C: opaque blob (no debuggability, decode ceremony, no benefit) **and**
```
Artifizer: I see contradiction of what is Bad and what is Good:

Option A says:
Bad, because the signature appears in URLs (logs/Referer/history) <<< I think is correct 'bad'

Option D says:
Bad, because ... no debuggability

What is our debuggability pattern?

I think it should be like - whatever logging is intentional:
- no information leak on rpoxy / routers / browser / etc
- whatever logging is intentional and sanitized (e.g. log tenant ID, but do not log token expiration)

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
Artifizer: ## PROPOSAL Summary

ADR-0004 debates **where** to put authentication credentials (query params vs. headers), treating transport as a format decision. This creates a false choice: discrete query parameters are optimal for embeddable bare URLs but terrible for batch/programmatic access. The solution is to **decouple the signed token from its envelope**.

**One signed claim-set. Two envelopes chosen by access intent.**

This unifies the best of Options A (embeddability) and C (debuggability, caching, batch efficiency) without requiring a single "best" transport — making each access pattern optimal without compromise.

---

## Problem with Current ADR-0004

The ADR frames the decision as "which transport is best for all access patterns?" and settles on Option A (discrete query parameters), which generates these objections:

| Concern | Impact |
|---------|--------|
| Query params cannot be cached by browsers/CDNs when re-issued | Embeddable URLs lose cache efficiency |
| Signatures leak into access logs, `Referer`, browser history | Security exposure, albeit short-lived |
| Per-file query params in batch requests require parsing per URL | Batch operations become awkward |
| "Debuggability" claimed for visible URLs contradicts security stance against visibility | Logical inconsistency in pros/cons |
| Each field as its own query param implies per-field signatures | Unclear if composite or per-field; if per-field, design is broken |

**Root cause:** The decision is asked in the wrong frame. Transport shouldn't be "one size fits all"; it should be "transport follows intent."

---

## Core Insight: Intent-Based Envelopes

The two real access patterns have *opposite* constraints:

### Public / Long-term Embeddable Access
- **Example:** ``
- **Constraint:** Caller cannot set HTTP headers (browser, media player, `` tag)
- **Tradeoff acceptable:** Token visible in logs/history because it's (a) short-lived (XXd max), (b) narrowly scoped, (c) bearer credential only, (d) endpoint is verified
- **Optimization goal:** Bare URL works everywhere

### Private / Short-term Programmatic Access
- **Example:** SDK call, server-to-server fetch, batch operation
- **Constraint:** Caller is code; can set headers, parse responses, handle retries
- **Tradeoff preferred:** Token hidden from logs/history, stable URL for caching, bulk operations grouped efficiently
- **Optimization goal:** Efficiency, auditability, debuggability via structured logging

**These are not the same requirement.** Forcing one transport for both loses on both fronts.

---

## Proposed Solution: Transport-Agnostic Token + Intent-Picked Envelope

### 1. Single Signed Token Format (PASETO v4.public)

The token is **one opaque claim-set**, signed once with Ed25519. Transport (query param or header) does not change the signature or the constraints — it only changes where the bytes ride.

```
claims = {
  v:    1,                          // format version
  kid:  "fs-2026-06",               // key rotation id
  iss:  "filestorage-control",
  aud:  "fs-sidecar:tenant-42",     // binds to sidecar + tenant
  sub:  "file-123/ver-9",           // immutable blob: {file_id}/{version_id}
  op:   "GET",                      // HTTP method: GET | HEAD | PUT
  exp:  1718700000,                 // REQUIRED; capped at max_url_ttl
  nbf:  1718699000,                 // optional; claim not-before
  jti:  "01H2XCEJQTF2K08ZZJ29V0YQA", // optional; for revocation / single-use
  
  // Optional constraints, ALL AND-combined, ALL inside the signature
  ip:   "203.0.113.0/24",           // IP/CIDR restriction
  put:  {                           // PUT-specific constraints
    max_size:      50_000_000,
    expected_hash: "sha256:abc123..."
  }
}

token = paseto.sign(ed25519_priv[kid], claims)
// Output: "v4.public.<base64-payload>.<base64-sig>"
```

**Why PASETO?**
- Spec'd for exactly this use case (cryptographic credentials, no algorithm confusion)
- One signature over canonical claims (not per-field)
- Immutable, opaque on the wire
- Lightweight (no JWT bloat)

---

### 2. Two Envelopes, Chosen by Access Intent

#### Envelope A: Query Parameter (Public / Embeddable)

```
GET /d/file-123/download?fs=<token> HTTP/1.1
```

| Property | Value | Rationale |
|----------|-------|-----------|
| **Default TTL** | XX days | Short enough for incidental logging to be acceptable |
| **Use case** | Bare URLs: ``, `<video>`, `curl`, shared links | Can't set headers in these contexts |
| **SDK method** | `client.get_embeddable_url(file_id)` | Returns a string ready to embed |
| **Leakage** | Token in logs, `Referer`, history | Accepted risk; token is short-lived, scoped, bearer-only |

**Cache behavior:**
- Static blob path `/d/file-123` (from ADR-0003: immutable per version) is the cache key
- `fs` token is stripped from cache key by CDN, still forwarded to verifier
- Content cached for 1 year; token validated per-request
- Re-presigning (new token) produces same cache entry; zero cache fragmentation

#### Envelope B: Authorization Header (Private / Programmatic)

```
GET /d/file-123/download HTTP/1.1
Authorization: FS-PASETO <token>
```

| Property | Value | Rationale |
|----------|-------|-----------|
| **Default TTL** | 1 hour | Longer acceptable because token is opaque to intermediaries |
| **Use case** | SDK calls, server-to-server, batch, internal tools | All callers can set headers |
| **SDK method** | `client.get_signed_request(file_id)` or `client.get_headers(file_id)` | Returns HTTP request or header dict |
| **Leakage** | Token not in logs, not in `Referer`, not in history | Header-based, sidecar logs intentionally sanitized |

**Cache behavior:**
- Static blob path `/d/file-123` is the cache key
- `Authorization` header stripped from cache key, forwarded to verifier
- Content cached; token validated per-request
- Full cache hit rate; stable URL across token re-issue

**Batch operations:**
```
// Programmatic: N files, single auth
POST /batch/download HTTP/1.1
Authorization: FS-PASETO <token>

{
  "files": [
    {"path": "/d/file-123", "range": "0-1000"},
    {"path": "/d/file-456", "range": "*"}
  ]
}
```

The token authenticates the batch operation; each file reference is just a path. No per-file parameter parsing, no awkwardness.

---

### 3. Unified Sidecar Verification Flow

Verification is **identical** regardless of envelope; only extraction differs.

```
1. Extract token:
   - Check Authorization: FS-PASETO header first
   - Else check ?fs query param
   - Else reject (401)

2. Decode and verify signature:
   - Extract kid from token header
   - Fetch public key from control-plane JWKS endpoint (cached)
   - paseto.verify(public_key[kid], token)
   - On error → 401

3. Validate claims (all must pass; AND-combined):
   - aud matches this sidecar:tenant
   - nbf ≤ now < exp
   - op matches HTTP method (GET token against HEAD request → OK; PUT token against GET → 403)
   - sub matches requested resource ({file_id}/{version_id})

4. Enforce constraints:
   - If ip in claims: verify request source IP ∈ CIDR
   - If put constraints: stream-enforce max_size, validate hash (ADR-0002)
   - If jti in claims: check not-in denylist (revocation check via shared metadata)

5. Serve content:
   - Forward request to backend storage
   - Stream response; emit per-byte usage counter to metrics
   - Log sanitized record: (timestamp, tenant_id, file_id, op, bytes, status)
     ❌ Never log: token, IP, hash, or any bearer credential
     ✅ Safe to log: tenant, file, operation, size, outcome
```

---

## Key Advantages Over Current ADR-0004

| Problem (ADR-0004) | Solution (This Proposal) | Benefit |
|---|---|---|
| "Query params can't be cached" | Immutable-blob addressing from ADR-0003 means content path never changes; token is stripped from cache key. Re-presigning ≠ cache miss. | Full cache efficiency; no fragmentation. |
| "Signatures leak" | Query tokens are short-lived (XXd); logging policy is intentional + sanitized regardless of transport. Header tokens never leak. | Leakage is time-bounded & intent-driven; no contradiction. |
| "Batch is awkward with per-file params" | Header-based batch groups files under one auth; no per-file parameter sprawl. | Batch operations are clean & efficient. |
| "Per-field signatures unclear" | One PASETO token, one composite signature over canonical claims. Unambiguous. | No ambiguity; signature model is proven (PASETO spec). |
| "Option A vs D: visible=bad XOR invisible=bad" | Visibility isn't inherently good/bad; it's *context-dependent*. Public embeddable = acceptable leakage; private = header-hidden. Debuggability via sidecar logs, not URL inspection. | Logical consistency; each envelope optimized for its access pattern. |

---

## Design Details: Support Systems

### Key Rotation

**Multi-key operation:** Sidecar caches public keys from control-plane JWKS endpoint.

```
1. Control plane holds private key(s) in KMS/HSM
2. Control plane publishes public keys + kid + rotation schedule:
   {
     "keys": [
       {
         "kid": "fs-2026-06",
         "kty": "OKP",
         "crv": "Ed25519",
         "x": "<base64-public-key>",
         "use": "sig",
         "valid_from": 1718500000,
         "valid_until": 1750036000
       },
       {
         "kid": "fs-2026-07",
         "kty": "OKP",
         "crv": "Ed25519",
         "x": "<base64-public-key>",
         "use": "sig",
         "valid_from": 1750036000,
         "valid_until": 1781572000
       }
     ]
   }
3. Rotate with overlap: new key signs, both verify, retire old key after max_url_ttl elapses
4. Sidecar refreshes JWKS hourly (or on 401 → fetch fresh JWKS and retry)
```

### Revocation

Short expiry (`exp`) is the primary revocation mechanism. For emergencies:

```
// Shared metadata store
user:42:revoke_epoch = 5  // if revoke_epoch ≤ token.epoch, token is rejected

// Or, per-token denylist (for single-use uploads via jti)
revoked_jtis = {
  "01H2XCEJQTF2K08ZZJ29V0YQA": 1718700100,  // revoked_at timestamp
  // pruned when timestamp + max_url_ttl < now
}
```

**Latency:** Same for both envelopes — not a differentiator.

### Upload: Single-Use Tokens

Uploads use `PUT` tokens with `jti` and `put` constraints:

```
claims = {
  ...
  op:   "PUT",
  jti:  "01H2XCEJQTF2K08ZZJ29V0YQA",  // unique per upload attempt
  put:  {
    max_size:      10_000_000,
    expected_hash: "sha256:abc123..."
  }
}
```

Workflow:
1. Client requests PUT token from control plane
2. Control plane generates token with `jti`, `max_size`, optional `expected_hash`
3. Client streams to sidecar; sidecar enforces constraints
4. On success: sidecar writes immutable blob, returns blob path
5. Client calls control plane CAS: swap file pointer from old blob → new blob (optimistic)
6. Control plane invalidates the `jti` (adds to denylist or marks used)

**Upload with retry:** On network interrupt, client can retry the same `jti` (idempotent); sidecar detects replay and re-serves the same blob path.

---

## Verification Strategy

### Code Review Checkpoints

1. **Control plane isolation:** Private key never leaves KMS; public key only used in sidecar
2. **Signature model:** One canonical PASETO payload, one Ed25519 verify per request
3. **Constraint enforcement:** All AND-combined constraints checked before serving bytes
4. **Logging sanitization:** No bearer tokens, keys, or sensitive fields in logs

### Integration Tests

- Bare-URL embeddable flow: GET with `?fs=token` works in browser, curl, `` tag
- Programmatic flow: Authorization header works; stable URL across re-presign
- Batch flow: POST with manifest; each file verified independently
- Constraint enforcement: ip/put/exp violations return 403 before byte 0
- Token expiry: expired token → 401
- Single-use uploads: second PUT with same `jti` works (idempotent); wrong `jti` → 403
- Cache validation: repeated GET with same token or different token hits cache (same content path)

### Performance Benchmarks

- Per-request overhead: token extraction + PASETO verify + constraint check < 2ms (p99)
- Key rotation: JWKS fetch + retry on 401 < 50ms
- Cache hit rate: programmatic header-based access achieves 95%+ CDN hits

---

## Migration: From ADR-0004 Discrete Query Params to Unified Transport

**If current implementation already uses `X-FS-*` discrete query parameters:**

1. **Phase 1:** Keep discrete params; add PASETO enum to SDKs (opt-in, feature-flagged)
2. **Phase 2:** SDKs dual-emit (both discrete + PASETO) for 1 release cycle
3. **Phase 3:** Sidecar accepts both; logs which form is used
4. **Phase 4:** Deprecate discrete params; SDKs emit PASETO only
5. **Phase 5:** Sidecar removes discrete-param code path

**If starting fresh:**
- Implement PASETO transport immediately; no migration burden
- Discrete query params never enter the codebase

---

## When to Use Each Envelope: Decision Tree

```
Does the caller need to embed the URL in HTML or share it as a bare link?
├─ YES  → Query parameter (?fs=token)
│        - Use: <img>, <video>, media players, shared links
│        - TTL: 7 days max
│        - SDKcall: get_embeddable_url(file_id)
│
└─ NO → Authorization header (FS-PASETO token)
         - Use: SDK, server-to-server, batch, internal tools
         - TTL: 15 min to 24 hours
         - SDK call: get_signed_request() or get_headers()
```

**If unsure:** Default to header (more secure, better caching, cleaner logging).

---

## Rationale: Why This Is Better Than Alternatives

### vs. "Ship Query Params Only" (ADR-0004 as-is)
- ❌ Batch operations are verbose and awkward
- ❌ Cache fragmentation on re-presign
- ❌ Logging concern (visibility) contradicts security stance
- ❌ All access patterns forced into one model

### vs. "Ship Headers Only"
- ❌ Breaks bare-URL embeddability (real requirement if `` is used)
- ❌ Shared links require round-trip to control plane to refresh auth
- ✅ Would work if embeddability is not actually required (should be tested)

### vs. Current ADR-0004 "Pick A" (Discrete params)
- ❌ Per-param parsing in batch; per-field signature ambiguity
- ❌ No escape hatch for programmatic access that wants headers
- ❌ Cache model requires complex normalization

### This Proposal
- ✅ One token; two envelopes; each optimal for its intent
- ✅ Batch operations are natural (header-based)
- ✅ Bare URLs work (query-based)
- ✅ Caching is clean (immutable-blob path is cache key, auth is orthogonal)
- ✅ Logging is consistent and sanitized (independent of transport)
- ✅ No ambiguity (PASETO spec, composite signature)

---

## Questions to Resolve Before Implementation

1. **Is bare-URL embeddability a real requirement?**
   - Who actually embeds ``?
   - If honest answer is "nobody" or "only public media," validate that first.
   - If nobody embeds, ship header-only (simpler system).

2. **PASETO vs. Custom Token Format?**
   - PASETO v4.public is recommended (spec'd, audited, no foot-guns).
   - Custom CBOR would be smaller but yours to secure.

3. **Edge Verification for Cache / Auth Decoupling?**
   - For full cache/auth decoupling at CDN scale, edge needs to verify tokens.
   - Either CDN calls a sidecar verify endpoint, or you deploy verification at the edge.
   - If neither is viable, caching stays at the sidecar level (still works, just not edge).

4. **KMS for Key Management?**
   - Strongly recommended to store Ed25519 private key in an HSM or cloud KMS (AWS KMS, Google Cloud KMS, HashiCorp Vault).
   - Control plane signs; never rotates key material locally.

---

## What Changes in the Code

### Control Plane

- Replace discrete `X-FS-*` query-param generation with PASETO token generation
- Expose two methods: `get_embeddable_url()` (query param) and `get_signed_request()`/`get_headers()` (header)
- Both methods call the same underlying `sign_claims()` function

### Sidecar

- Replace discrete-param parsing with PASETO decoding
- Try Authorization header first; fall back to `?fs` query param
- Single verify path regardless of envelope
- Logging stays intentionally sanitized (no tokens, no keys)

### SDK

- Add enum: `TransportMode::EmbeddableQuery | TransportMode::ProgrammaticHeader` (or just pick based on method)
- `get_embeddable_url()` → TransportMode::EmbeddableQuery
- `get_signed_request()` → TransportMode::ProgrammaticHeader

### ADR Updates

- **Replace ADR-0004** with this proposal (different model, not incremental patch)
- **Update ADR-0003** to reference unified token transport (link forward to new ADR-0004)
- **Create ADR-0005** (if needed) for key rotation and revocation strategy

---

## Appendix: PASETO Token Example

```
// Control plane generates

claims = {
  v:    1,
  kid:  "fs-2026-06",
  iss:  "filestorage-control",
  aud:  "fs-sidecar:tenant-1a2b3c",
  sub:  "file-xyz/ver-001",
  op:   "GET",
  exp:  1718872800,  // 7 days from now
  nbf:  1718700000,  // now
  ip:   "203.0.113.0/24"
}

token = paseto.sign(ed25519_priv["fs-2026-06"], claims)
// → "v4.public.MjAyNi0wNi1maWxlc3RvcmFnZQ.z1x2c3v4b5..."

// Client uses embeddable
GET /d/file-xyz?fs=v4.public.MjAyNi0wNi1maWxlc3RvcmFnZQ.z1x2c3v4b5...

// Or programmatic
GET /d/file-xyz
Authorization: FS-PASETO v4.public.MjAyNi0wNi1maWxlc3RvcmFnZQ.z1x2c3v4b5...

// Sidecar verifies (same code path)
token = extract_token(req)  // from header or query
claims = paseto.verify(public_key[kid], token)
assert(aud, op, sub, exp, ip, ...)
serve_bytes()
```

---

## Summary: Next Steps

1. **Validate the embeddability requirement:** Who embeds bare URLs? Is it a real use case or assumed?
2. **Choose token format:** PASETO v4.public (recommended) vs. custom format?
3. **Decide KMS approach:** Where does the Ed25519 private key live? Cloud KMS, local HSM, or in-process?
4. **Update ADRs:** Replace 0004, forward-reference in 0003, add 0005 if needed.
5. **Implement:** Control plane signing + sidecar verification, with integration tests for both envelopes.
6. **Test:** Bare-URL works in browsers; headers work in SDK; batch is efficient; caching is clean.

---

**End of Proposal**

Artifizer (changes requested):

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Addressed in the reworked ADR-0004. Fields like `exp` aren't secrets — the only secret is the signature, which is in the URL regardless, so hiding the other fields buys no security. If the credential must stay out of URLs/logs, that's the new **header envelope** (for programmatic/SDK/batch), which also removes the per-file query-param parsing for batch. The **query** envelope stays only for bare, embeddable URLs (browser/``).

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Correct — parsing + Ed25519 verification always happen, in any envelope. I dropped the misleading 'verify without a decode step' pro. The point is only that with **discrete fields** there is no opaque token to base64-decode (named params + one composite signature); decode/parse isn't an argument for either side. We kept discrete fields.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Fixed — the contradictory 'debuggability via visible URL' pro is removed. Debuggability is now defined as intentional, **sanitized server-side structured logging** (log tenant/file ids + outcome, never the signature or raw `exp`), independent of the envelope. URL visibility is treated only as the leak it is, not as a feature.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Adopted the dual-envelope idea (query for embeddable + header for programmatic/batch) — thanks, that's the right split and matches SigV4's two forms. But kept **discrete fields** and rejected the **opaque token (PASETO)**:
- Ed25519 stays — control is the sole minter, the sidecar only verifies and can't forge; PASETO `v4.local` would need the sidecar to hold a symmetric secret, breaking that.
- 'one signature, not per-field' is already true for discrete fields — composite SigV4-style signing over all params.
- opacity gives no security: `v4.public` is signed-not-encrypted, so `exp` is still readable after base64-decode; fields aren't secrets and the signature is in the URL regardless.
- discrete keeps edge observability (CDN/WAF/logs read `exp`/scope without decoding) and is REST-native/per-field-versionable.

Reworked ADR-0004 pushed (discrete fields, dual envelope; options A–D + comparison table updated).

gears/file-storage/docs/ADR/0001-cpt-cf-file-storage-adr-proxy-content-traffic.md
```diff
@@ -1,10 +1,20 @@
 ---
-status: accepted
+status: superseded
```
Artifizer: Can you please remove the ADRs that are superseded? it makes no much sense to keep it as it can poison LLMs/humans and those who want to know the history may always use `git log`

pr0tey:
Quota enforcement: the data plane should enforce the cap directly at upload, with the allowance carried in the signed URL

  Context. DESIGN §4.5 ("No policy or quota in the data plane", lines 1427–1434) and PRD cpt-cf-file-storage-fr-storage-quota currently place all quota logic in the control plane, evaluated only at presign: if an
  upload would violate a limit, the control plane either refuses to mint the URL or "bakes a tighter constraint, e.g. MaxSize, into it." The sidecar holds none of this state and makes no tenant/user/backend decision.

  We discussed this and believe presign-only enforcement leaves two gaps that the control/data-plane split (ADR-0003) introduces and that the monolith (ADR-0001) did not have, because in the monolith the quota check
  and the byte-write were a single act:

  1. Presign cannot know the real size. The control plane gates the quota before any bytes exist. Unless the client declares X-FS-ExactSize, the presign-time check operates on an unknown quantity. Today nothing
  requires ExactSize/MaxSize on an upload URL when a quota is active, so the check can be vacuous.
  2. TOCTOU across concurrent presigns. Actual consumption is known only at bind (the sidecar reports the size), and usage reporting is asynchronous. With a presign→use window of up to ~`max_url_ttl(recommended 7
  days), N concurrent presigns can each pass the quota check yet collectively blow past the owner's limit. There is no reservation or re-check onbind`.

  Proposed direction (what we want the P2 design to adopt). Make the data plane enforce the quota cap directly during the upload stream, and carry the maximum allowance inside the signed URL / token — not just a
  per-file MaxSize, but an explicit quota allowance for that operation.

  Concretely:

  - At presign, the control plane computes the owner's remaining quota and bakes the allowed byte ceiling into the signed fields (e.g. an X-FS-MaxSize-style quota allowance). The limit travels with the credential,
  tamper-evident under the same composite signature.
  - The sidecar already streams the body and counts bytes (it enforces MaxSize/ExactSize and computes the hash). It therefore aborts mid-stream the moment the signed allowance is exceeded (413), reusing the exact
  machinery that already exists — no new state, no cross-request coordination. The cap is per-URL, so the data plane stays stateless: it enforces what the signature asserts, nothing more.
  - This makes the upload-time limit authoritative at the point where the bytes actually arrive, closing gap (1): the cap no longer depends on the client truthfully declaring its size — the sidecar terminates the
  transfer at the signed ceiling regardless.

  Reconciling with the "policy-free data plane" principle. This does not ask the sidecar to read tenant/owner quota state or make a policy decision — the decision stays in the control plane at presign; the sidecar
  only enforces a per-URL number it was handed, exactly as it already does for MaxSize. So §4.5's statelessness invariant holds.

  What remains a control-plane responsibility is the aggregate owner quota across many outstanding URLs (gap 2). The clean complement to the above is a reserve-at-presign / commit-at-bind model: the control plane
  reserves the baked allowance against the owner's quota when it mints the URL, the sidecar reports the actual bytes at bind, and the control plane commits the real size and releases the unused reservation (and
  reclaims reservations from URLs that expire unused). The per-URL ceiling enforced by the sidecar is what makes the reservation bounded and safe.

  Asks for the P2 FEATURE (cpt-cf-file-storage-fr-storage-quota):
  - Make a signed quota allowance (or ExactSize/MaxSize) mandatory on upload URLs when a quota is active; specify the sidecar's mid-stream 413 on breach.
  - Specify the control-plane reserve/commit/release lifecycle keyed to presign → bind → expiry, so concurrent presigns cannot collectively exceed the owner's quota.
  - Note explicitly in DESIGN §4.5 that the per-URL quota allowance is the one quota-related value the data plane enforces directly, and why that does not violate the stateless/policy-free invariant.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: **Adopted — thank you.** After discussing with the team we reversed to your proposal: ADR-0004 now specifies a single **opaque PASETO `v4.public` token** (Ed25519, asymmetric — control mints, sidecar verifies, can't forge), carried in the `fs-token` query param **or** the `X-FS-Token` header by access intent. The decisive driver is exactly yours: we're not S3-wire-compatible, so discrete-field readability buys nothing externally and only couples intermediaries — an atomic, opaque token is simpler and lets the claim-set/crypto evolve freely. Added a hard **Token Opacity Contract** (only control + sidecar parse it; everyone else treats it as opaque bytes; format may change). ADR-0004 reworked + propagated to PRD/DESIGN/api.md; pushed in `60bed982`.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Resolved by the token model. The credential is now an opaque token; to keep it out of URLs/logs entirely, programmatic/batch callers use the `X-FS-Token` **header** (the `fs-token` query form is only for bare embeddable URLs). One opaque token per file also removes the per-file param parsing in batch. (Note: PASETO `v4.public` is signed-not-encrypted, so `exp` is technically decodable — the opacity is an interface/evolvability boundary, not secrecy; `exp` isn't a secret anyway.)

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Agreed, and that is exactly the model now: the sidecar base64-decodes + PASETO-verifies the token in whichever carrier it arrives (query or header). The earlier 'verify without a decode step' framing is gone.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Fixed. Debuggability is defined as sanitized **server-side** structured logging by control/sidecar (tenant/file ids + outcome; never the token or `exp`), independent of carrier. With the opaque token the edge intentionally cannot read fields — observability is server-side only. The contradictory 'debuggability via visible URL' pro is removed.

gears/file-storage/docs/ADR/0001-cpt-cf-file-storage-adr-proxy-content-traffic.md
ffedoroff: Done. Removed the superseded **ADR-0001** (proxy-all monolith) and scrubbed every reference to it — ADR-0003/ADR-0002/DESIGN no longer link it, and the dangling cpt-id is gone. ADR-0003 is now self-contained (it describes the prior proxy-all design in-place). History remains in `git log` as you suggested. Pushed in `dc0489fb`.

gears/file-storage/docs/api.md
```diff
@@ -3,132 +3,217 @@
...
+   `pending` version (checking `If-Match` as an early fail — if the file already moved on, it errors **before** the
+   bytes are uploaded), streams to the backend computing size + SHA-256, then **binds** `content_id := version_id`
+   under `If-Match`.
+3. On a **bind conflict** the sidecar returns `412` **and the `version_id`**; the client re-reads the current ETag and
+   replays `POST /files/{id}/bind` with that `version_id` and the fresh `If-Match` — **no byte re-upload**, because the
+   version already exists.
+
+`If-Match` is therefore checked twice (opportunistically at pre-register, authoritatively at bind). Backend content is
+never mutated in place; a replacement is always a new version + a pointer swap.
+
```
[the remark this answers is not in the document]
ffedoroff: Clarified in api.md. Added: on a `412`, **re-bind, don't re-presign or re-upload** — rebinding is a control-plane call (`POST /files/{id}/bind`) **independent of the signed upload URL**, so the URL's `exp` is irrelevant and the bytes persist. Re-presigning is **not** idempotent: a fresh `POST /files/{id}/versions` + upload creates a **new sibling `version_id`** (the unbound one is swept by the P2 cleanup engine). Clients should rebind the returned `version_id`.

gears/file-storage/docs/PRD.md
```diff
@@ -186,39 +211,57 @@ roles) from the platform authentication middleware.
...
+asks the control plane for a **signed upload URL** (`cpt-cf-file-storage-fr-signed-urls`), then transfers the bytes to
+the **sidecar** at that URL; the control-plane REST surface never receives the content itself. Each upload writes a
+new **immutable** content blob to the backend object `/{file_id}/{version_id}`; the file's live content is a pointer
+(`content_id`) that is **bound** to that version (`cpt-cf-file-storage-fr-file-versioning`). Backend content is
+**never mutated in place** — replacing content writes a new version and swaps the pointer; partial-byte mutation is
+**not** supported.
+
+Binding the pointer is an optimistic compare-and-swap guarded by `If-Match` (`cpt-cf-file-storage-fr-conditional-requests`):
+if the file's content changed concurrently the bind returns `412`, and the client **MUST** be able to retry the bind
+against the already-uploaded `version_id` **without re-uploading the bytes**.
```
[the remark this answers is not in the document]
ffedoroff: Clarified in the PRD upload FR. Rebind is a **control-plane** op independent of the signed upload URL, so the URL's expiry does **not** affect the retry and no re-presign is needed; the `412` returns the current content ETag for the fresh `If-Match`. (Re-presigning would instead upload a new sibling version, reconciled by orphan-reconciliation.)

ffedoroff:
@pr0tey good idea — and it already fits the model: the sidecar enforces a per-URL size cap carried in the token (the `max_size` claim) and aborts mid-stream with `413`, which is exactly the "data plane enforces the signed allowance at upload" you describe — no new state, no cross-request coordination.

One contract decision, though: **`max_size` stays optional**, not mandatory-when-quota-active. If the claim is present the data plane **MUST** enforce it; if it is absent, **only the backend's own default size limits apply** (no FS-imposed per-URL cap). So quota policy decides whether to bake a `max_size` into the token; absent that, backend defaults govern.

On the aggregate-owner-quota / TOCTOU gap: the **reserve-at-presign / commit-at-bind / release-on-expiry** lifecycle is the right direction and we've recorded it as a **P2 control-plane** concern in the `cpt-cf-file-storage-fr-storage-quota` FEATURE. We are not making the per-URL allowance mandatory, but that reservation model (control-side) is what bounds concurrent presigns.

Captured in DESIGN §4.5 (the size claim is optional; when present the sidecar MUST enforce it; the per-URL `max_size` is the one quota-derived value the data plane enforces; reserve/commit/release deferred to the P2 quota FEATURE).

ffedoroff:
Pushed `08c52ba0` addressing the latest review:

- **CodeRabbit (api.md / PRD §5.1):** clarified the `412` retry — the client **re-binds** the returned `version_id` via `POST /files/{id}/bind`, a control-plane call **independent of the signed upload URL** (its `exp` is irrelevant; no re-presign, no re-upload). The `412` carries the current content ETag for the fresh `If-Match`. Re-presigning is **not** idempotent (creates a sibling version, swept by the P2 cleanup engine).
- **@pr0tey (quota):** the token's `max_size` is **optional** — present ⇒ the data plane MUST enforce it (mid-stream `413`); absent ⇒ only the backend's default limits apply. The per-URL `max_size` baked from a quota decision is the one quota-derived value the data plane enforces (a signed number, not a policy lookup). Reserve-at-presign / commit-at-bind / release-on-expiry recorded as a **P2** control-plane concern in `cpt-cf-file-storage-fr-storage-quota`.

(Inline replies left on each thread.)

gears/file-storage/docs/PRD.md
```diff
@@ -59,8 +59,22 @@ FileStorage is a universal file storage and management service for the Gears mid
...
 
-The service supports pluggable storage backends, multiple access protocols (REST, S3-compatible, WebDAV), tenant-scoped
-access control with an ownership model, and policy-driven governance for file types and sizes.
+FileStorage is split into two cooperating planes (see
+[ADR-0003](./ADR/0003-cpt-cf-file-storage-adr-sidecar-data-plane.md)):
+
+- a **control plane** (the FileStorage API/SDK) that owns metadata, authorization, versioning, and
+  conditional-request semantics, and whose REST surface **never carries file content** — it issues short-lived
+  **signed URLs** instead;
+- a **data-plane sidecar** that is the only component to move user bytes, is connected to the storage backends, and
```
MikeFalcon77, line 68: This looks like a new Gears pattern, not a normal gear REST route. Can we call it out explicitly? Today the docs say API Gateway owns REST hosting, so the sidecar should be documented as a data-plane exception, not just another custom HTTP server.

gears/file-storage/docs/DESIGN.md
```diff
@@ -314,40 +403,125 @@ graph TB
...
+##### Why this component exists
+
+The **sidecar's** HTTP entry point on its own domain. Verifies the signed URL and (when a token-claim predicate is
+present) the platform token, then drives the byte path. The only component clients hit for content.
+
+##### Responsibility scope
+
+- Verify the **PASETO `v4.public`** token (Ed25519) with the control-distributed public key; check `exp`, `ip`, and token-claim predicates
+  (validating a real platform JWT when a predicate references a claim); reject with `403` on any failure
+- Parse the `Range` header (to `ByteRange`) and conditional headers; serve `200`/`206`/`304`/`412`/`416` for downloads
```
MikeFalcon77, line 509: Can we spell out who validates the platform JWT here? If the sidecar validates it itself, that is an exception to the normal gateway-auth flow. If API Gateway fronts the sidecar, please say how SecurityContext, request id, and rate limits are handled.

gears/file-storage/docs/DESIGN.md
```diff
@@ -1119,18 +1332,389 @@ Concurrency caps:
...
+| Exact size | `exact_size` | no | P1 | upload | `413` over / `400` under |
+| Expected hash | `expected_hash` (`<alg>:<hex>`) | no | P1 | upload | `422` |
+| Max rate | `max_rate` (bytes/s) | no | **P2** | up/down | throttled to ≤ rate |
+| Max connections | `max_conns` | no | **P2** | up/down | `429` |
+
+- **`exp` is mandatory and capped.** It MUST NOT be further out than the configured `max_url_ttl` (recommended **7
+  days**); the cap is enforced by the **control plane at signing** (it refuses to mint a longer token). The sidecar
+  rejects when `now > exp`. "Available to everyone for 5 minutes" = only `exp`, no token-claim predicate.
+- **A size claim is optional.** If `max_size` (or `exact_size`) is present, the sidecar **MUST** enforce it (mid-stream
+  `413`); if **neither** is present, the sidecar imposes **no FS-level size cap** — only the backend's own default size
```
MikeFalcon77, line 1424: This creates a stale-permission window: authz is checked when the URL is minted, but permissions can change before `exp`. Is 7 days really safe for private content? I would prefer a much shorter default TTL and an explicit note that this is an accepted trade-off.

MikeFalcon77 (commented):
Inline architecture notes.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,207 @@
...
+## Decision Outcome
+
+* **Encoding = one opaque, atomically-signed token.**
+* **Token format = PASETO `v4.public`** (Ed25519, asymmetric; **not** JWT). It carries the full claim-set — `op`,
+  resource (`file_id`, `content_id`/`version_id`), `exp` (required, capped at `max_url_ttl`, recommended 7 days), the
+  constraints (`ip`, token-claim predicates, upload `max_size`/`exact_size`/`expected_hash`; P2 `max_rate`/`max_conns`),
+  and the baked response-header set — with **one signature over the whole set**. The PASETO **footer** carries a key id
+  (`kid`) for P2 rotation.
+* **Transport = both query and header**, chosen by access intent; the **token bytes are identical** either way:
+  * **query** — `?fs-token=<token>` — for bare, embeddable URLs (browser, `<img>`/`<video>`, `curl`, media `Range`; the caller
```
MikeFalcon77, line 78: For query tokens, leaks through logs, browser history, or `Referer` are a real concern. Since there is no per-token revocation, can we state that bare URLs get very short TTLs, and long-lived sharing belongs to the future FileShare/P3 surface?

gears/file-storage/docs/DESIGN.md
```diff
@@ -587,24 +768,25 @@ To keep the P1 surface forward-compatible without binding the P2 shape, only the
...
-  prevents silent content mutation when a client accidentally forwards a `content` part (which `If-Match` would not
-  catch, since the current content ETag still matches at request time). Metadata-only `PATCH` needs no flag
-- **Conditional headers**: `If-Match` required on `PATCH`/`DELETE`; `If-Match` and `If-None-Match` optional on
-  `GET`/`HEAD`. ETag is `(file_id, content_revision)`-derived and content-only. `If-Match-Metadata: <u64>` is an
-  optional metadata-concurrency validator on metadata-only `PATCH`, matched against `metadata_revision` (mismatch →
-  `412`); absent → last-write-wins (see `cpt-cf-file-storage-principle-content-only-etag`)
-- **Range**: full `bytes=` syntax supported; `Accept-Ranges: bytes` advertised on every download response; `HEAD`
-  ignores `Range` and returns full-file headers. See §4.1 for backend translation rules
+- **Technology**: control-plane REST (axum + OperationBuilder for JSON), no GraphQL, no gRPC; the sidecar serves a
+  signed-URL HTTP content surface. Versioned per `cpt-cf-file-storage-interface-rest-api`
```
MikeFalcon77, line 786: How will this sidecar surface be documented for clients? Control routes can use OperationBuilder/OpenAPI, but this binary `GET`/`PUT`/`HEAD` API is outside the usual generated OpenAPI flow unless we add a separate contract for it.

gears/file-storage/docs/migration.sql
```diff
@@ -126,10 +99,61 @@ CREATE INDEX files_owner_listing_idx
...
+    -- Content hash. P1 allow-list locked to SHA-256 per ADR-0002; widened in P2.
+    hash_algorithm   text         NOT NULL  DEFAULT 'SHA-256'
+                                  CHECK (hash_algorithm = 'SHA-256'),
+    hash_value       bytea        NOT NULL  CHECK (octet_length(hash_value) = 32),
+
+    -- Lifecycle: 'pending' from pre-register until bind, then 'available'.
+    status           text         NOT NULL  DEFAULT 'pending'
+                                  CHECK (status IN ('pending', 'available')),
+    -- True for the file's current version (matches files.content_id).
+    is_current       boolean      NOT NULL  DEFAULT false,
```
MikeFalcon77, line 154: This duplicates `files.content_id`. If we keep both, bind needs to update `content_id`, old `is_current = false`, and new `is_current = true` in one transaction, with tests for split-brain cases. Otherwise it may be simpler to derive current from `files.content_id` and drop this flag.

gears/file-storage/docs/DESIGN.md
```diff
@@ -872,40 +1059,67 @@ sequenceDiagram
...
+| `size`            | `bigint`                              | Content size in bytes                                                        |
+| `hash_algorithm`  | `text`                                | P1: `'SHA-256'` only                                                          |
+| `hash_value`      | `bytea`                               | Content digest (32 bytes for SHA-256)                                        |
+| `status`          | `text` (`'pending'` \| `'available'`) | `'pending'` from pre-register until bind, then `'available'`                 |
+| `is_current`      | `boolean`                             | Whether this version is the file's current content (matches `files.content_id`) |
+| `backend_id`      | `text`                                | `BackendConfig` that holds the bytes (TOML in P1)                            |
+| `backend_path`    | `text`                                | Opaque per-driver path (`/{file_id}/{version_id}` convention)                |
+| `created_at`      | `timestamptz`                         | Version creation time                                                        |
```
MikeFalcon77, line 1129: Small mismatch with the SQL: the migration creates the pending cleanup index on `created_at`, while this says `(file_id)`. For cleanup, `created_at WHERE status = 'pending'` sounds right; let's align the wording.

gears/file-storage/docs/ADR/0003-cpt-cf-file-storage-adr-sidecar-data-plane.md
```diff
@@ -0,0 +1,213 @@
...
+* **Control plane** — the FileStorage **API / SDK**. Owns metadata, authorization, versioning, and
+  conditional-request semantics. **Its HTTP REST surface never accepts or returns file content.**
+  It issues short-lived **signed URLs** that point at the sidecar. The only path where control-side
+  code touches bytes is the in-process **SDK proxy mode**, which streams inside the *consumer
+  gear's* process — never through the control-plane service.
+* **Data plane** — the **sidecar**. It has its own domain and URL and is the only component that
+  moves user bytes. It is connected to N storage backends, validates platform auth tokens (the
+  standard way) **and** the signed-URL signature, and reaches the control plane through the FS SDK
+  (direct-DB or REST). It is effectively a full FileStorage instance over the shared metadata DB.
+
```
MikeFalcon77, line 90: This is a bit ambiguous: is the sidecar using an internal SDK over direct DB, or calling the control plane over REST? Those have very different latency, failure, and security properties. Can we pick the P1 path and name it explicitly?

gears/file-storage/docs/DESIGN.md
```diff
@@ -1119,18 +1332,389 @@ Concurrency caps:
...
+5. **token-claim predicates** — *if and only if* the token carries one or more `tok.<claim>` predicates, requires a valid
+   platform JWT, validates it the standard way, and matches each `claim == value` (all AND); `403` on any miss;
+6. **resolves the target object** from the version row — download: `(file_id from path, content_id)` → backend, path,
+   size; upload: the backend/path returned by control at pre-register (`404` if the version is gone);
+7. **on upload**, enforces the content claims on the streaming pass (alongside SHA-256 + magic-bytes): `max_size` → abort
+   `413` the moment the cap is exceeded; `exact_size` → abort `413` if exceeded, or `400` if the final length is short;
+   `expected_hash` → compare the computed digest at end-of-stream, `422` on mismatch. Any failure aborts before bind and
+   best-effort deletes the partial object.
+
+Verifying the token is DB-free; only object resolution (step 6) touches the DB. Authorization decisions stay on the
```
MikeFalcon77, line 1454: If the sidecar skips fresh AuthZ, let's state that the signed token is the delegated authorization artifact for this resource and operation until `exp`. Otherwise this looks like sensitive DB/object access without a PDP decision at request time.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
```diff
@@ -0,0 +1,207 @@
...
+This supersedes the earlier "discrete fields" outcome. It adopts the PASETO proposal raised in PR review while keeping
+the dual-envelope (query + header) and the asymmetric, sidecar-cannot-mint property.
+
+### Consequences
+
+* `cpt-cf-file-storage-design-signed-urls` (DESIGN §4.5), api.md, and the worked examples (§4.6/§4.7) change: the
+  discrete `X-FS-*` parameters are replaced by a **single token** carried as `?fs-token=<token>` (query) or an
+  `X-FS-Token` header. SigV4-style canonical-string signing is replaced by **PASETO mint (control) / verify
+  (sidecar)**; all claims live inside the token.
+* **New dependency:** a PASETO v4 library — control plane signs (`v4.public`), sidecar verifies. Ed25519 keys as before
```
MikeFalcon77, line 102: Before we lock in PASETO, can we confirm the FIPS story? Ed25519/PASETO may be fine cryptographically, but if this must run in FIPS mode we need a compliant dependency/provider plan, or a fallback, before implementation.

gears/file-storage/docs/PRD.md
ffedoroff: Done — PRD now calls the sidecar out explicitly as a data-plane platform component (a documented exception to "API Gateway owns REST hosting"), not a normal gear REST route.

gears/file-storage/docs/DESIGN.md
ffedoroff: Spelled out in DESIGN §4.5: the sidecar validates the platform JWT itself when a token-claim predicate is present — an explicit, documented exception to the gateway-auth flow.

gears/file-storage/docs/DESIGN.md
ffedoroff: Agreed. Default URL TTL is now short (15 min); 7 days is only the hard ceiling (max_url_ttl). The stale-permission window is documented and long-lived sharing is opt-in.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Stated: bare query URLs take the short default TTL; long-lived/programmatic sharing uses the X-FS-Token header (out of logs/Referer/history). Per-token revocation is deferred to P2.

gears/file-storage/docs/DESIGN.md
ffedoroff: Addressed: the sidecar GET/PUT/HEAD surface is documented as a hand-authored contract in api.md (outside the generated OpenAPI flow), with a note explaining why it is separate.

gears/file-storage/docs/migration.sql
ffedoroff: Kept both and made bind atomic: content_id CAS + clear old is_current + set new in one transaction. Added a split-brain test. Invariant documented in migration.sql.

gears/file-storage/docs/DESIGN.md
ffedoroff: Aligned with the SQL: the pending-cleanup index is created_at WHERE status='pending'. Removed the stale (file_id) wording in DESIGN.

gears/file-storage/docs/ADR/0003-cpt-cf-file-storage-adr-sidecar-data-plane.md
ffedoroff: Picked one: ADR-0003 now states the sidecar calls the control plane over the FileStorage SDK (s2s REST) in P1 — never direct DB access.

gears/file-storage/docs/DESIGN.md
ffedoroff: Stated explicitly in DESIGN §4.5: the signed token IS the delegated authorization artifact for that resource+operation until exp; the sidecar performs no fresh AuthZ.

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Confirmed in ADR-0004: documented the FIPS posture — Ed25519/PASETO is the choice, with a noted fallback plan if a FIPS-mode provider is mandated (tracked for P3).

Artifizer (approved):

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
MikeFalcon77: What I mean that we shouldn't bring new rust crates which are using non-FIPS algorithms and we will unable to replace thse algs

MikeFalcon77 (approved):
Approved with note: please be sure that we don't use non-FIPS compatible crates

gears/file-storage/docs/ADR/0004-cpt-cf-file-storage-adr-signed-url-transport.md
ffedoroff: Understood — the constraint is on dependency selection, not a deferred fallback. Reworked ADR-0004 "FIPS posture" (`cc63f02f`) accordingly:

- **MUST NOT** add any crate that hard-wires a non-FIPS algorithm or a self-contained crypto impl we cannot swap.
- Sign/verify sits behind a thin in-house `SignatureProvider` abstraction (`sign`/`verify`); that abstraction — not the crate — binds to the concrete algorithm + module.
- FIPS deployments back the provider with a validated module (`rustls-corecrypto-provider`); otherwise swap to a FIPS-approved alternative (ECDSA P-256) with **no codec/claim-set change** (opaque token).
- The candidate PASETO crate is evaluated for this replaceability **before** adoption; one that statically links a non-replaceable non-FIPS impl is rejected.

DESIGN.md and api.md carry the same rule. The replaceability requirement gates dependency selection and is no longer deferrable.

PR #4098, opened 2026-06-16
feat(cluster): add cluster gear wiring and sdk implementation

asmith987:
## Summary by CodeRabbit

* **New Features**
  * Added a new cluster coordination SDK introducing cache (with versioned compare-and-swap), leader election, distributed locks, and service discovery primitives, each with watch-based updates and automatic watch restart on retryable failures.
  * Added typed profile scoping plus a polling-based prefix-watch polyfill for backends lacking native support, along with SDK default backends and registration/deregistration helpers.
  * Introduced the `cluster` gear wiring and coordinated shutdown lifecycle, plus an in-process standalone backend plugin.

* **Documentation**
  * Added comprehensive ADR/design docs, an observability naming contract, and new SDK examples.


gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,695 @@
...
+        // available.
+        let (receiver, handle) = ServiceHandle::channel(instance_id, COMMAND_BUFFER);
+        let task = HeartbeatTask {
+            cache: Arc::clone(&self.cache),
+            key,
+            ttl: self.ttl,
+            renewal: self.renewal,
+            record,
+        };
+        tokio::spawn(task.run(receiver));
```
MikeFalcon77: **MEDIUM**

The spawned `HeartbeatTask` carries no `CancellationToken` and is neither tracked nor awaited by `ShutdownRevoke::revoke`, so on graceful cluster shutdown it keeps running its renewal loop (re-calling `cache.put`, which returns `Err(Shutdown)` and is discarded) until the consumer happens to drop the `ServiceHandle`.

A background task that ignores the shutdown signal delays clean teardown and burns a renewal tick of work after the cluster is supposed to be stopped.

Give `HeartbeatTask` the backend's shutdown `CancellationToken` (as `WatchTranslator` has), `select!` on `shutdown.cancelled()` in its loop to exit, and track its `JoinHandle` so `revoke` awaits it.

gears/system/cluster/cluster-sdk/src/discovery/handle.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// NOT a health observation (ADR-008): this declares intent; liveness is
+    /// signaled by heartbeat/TTL renewal.
+    ///
+    /// # Errors
+    /// - [`ClusterError::Provider`] (`ConnectionLost`) when the backend channel
+    ///   is already gone, or when the backend accepted the request but dropped
+    ///   the reply without responding — the flip is unconfirmed and surfaces
+    ///   rather than being masked.
+    /// - Any other [`ClusterError`] the backend returns for the flip.
+    pub async fn set_enabled(&self, enabled: bool) -> Result<(), ClusterError> {
```
MikeFalcon77: **MEDIUM**

`ServiceHandle::set_enabled` takes a bare `bool` for serving intent even though the domain already models that state space as the `InstanceState` enum (`Enabled`/`Disabled`).

A boolean parameter is not self-documenting at call sites (`set_enabled(false)` reads ambiguously) and cannot be extended if a third serving state is ever needed.

Change the public signature to `set_state(&self, state: InstanceState)` (or have `set_enabled` accept `InstanceState`) and carry `InstanceState` through the command/request path.

gears/system/cluster/cluster-sdk/src/discovery/handle.rs
```diff
@@ -0,0 +1,316 @@
...
+        /// The replacement metadata map.
+        metadata: HashMap<String, String>,
+        /// The reply side the backend completes with the update outcome.
+        responder: ServiceResponder,
+    },
+    /// Flip the instance's serving intent. The backend applies it and replies;
+    /// watchers observe a `Change(Updated)`.
+    SetEnabled {
+        /// The requested serving intent (`true` = enabled).
+        enabled: bool,
```
MikeFalcon77: **LOW**

`ServiceRequest::SetEnabled` exposes the serving-intent transition to backends as a raw `bool` field instead of the existing `InstanceState` enum.

Every backend must re-map `true`/`false` to `InstanceState` by hand, duplicating that mapping and inviting inconsistency.

Replace the `enabled: bool` field with `state: InstanceState` so backends receive the typed intent directly.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -0,0 +1,348 @@
...
+pub struct ClusterWiringBuilder {
+    hub: Arc<ClientHub>,
+    profiles: Vec<(String, ProfileBackends)>,
+    stop_hooks: Vec<StopHook>,
+}
+
+impl ClusterWiringBuilder {
+    /// Binds `backends` to the typed profile `P`. Only [`ClusterProfile::NAME`]
+    /// is read — the profile string is never re-typed at this call site.
+    pub fn profile<P: ClusterProfile>(mut self, backends: ProfileBackends) -> Self {
```
MikeFalcon77: **LOW**

`ClusterWiringBuilder::profile` binds the typed profile via a turbofish-only generic (`P::NAME` with no argument), whereas every SDK resolver's `profile<P>(self, _marker: P)` takes the marker as a value argument.

The two builder families that consumers use side by side have inconsistent profile-binding ergonomics, which is a needless source of confusion.

Make them consistent: either accept a `_marker: P` value in `ClusterWiringBuilder::profile`, or drop the discarded `_marker` parameter from the resolver builders' `profile` method.

gears/system/cluster/cluster-sdk/src/leader/types.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    #[test]
+    fn features_constructor_sets_flag() {
+        assert!(LeaderElectionFeatures::new(true).linearizable);
+        assert!(!LeaderElectionFeatures::new(false).linearizable);
+    }
+
+    #[test]
+    fn leader_status_is_copy_and_comparable() {
```
MikeFalcon77: **LOW**

This test only exercises the derived `Copy` and `PartialEq`/`Eq` on `LeaderStatus` (copy a value, compare it to itself, compare two distinct variants), verifying compiler-generated impls rather than any domain logic.

Tests of derives add maintenance cost without guarding against any regression the compiler doesn't already prevent.

Delete this test, or replace it with an assertion on behavior that depends on `LeaderStatus` comparison.

gears/system/cluster/cluster-sdk/src/cache/types.rs
```diff
@@ -0,0 +1,130 @@
...
+        assert_eq!(
+            CacheEvent::Expired {
+                key: "e".to_owned()
+            }
+            .key(),
+            "e"
+        );
+    }
+
+    #[test]
```
MikeFalcon77: **LOW**

`entry_versions_are_comparable` asserts only on the derived `PartialEq` and `Clone` of `CacheEntry` (two entries differ, an entry equals its clone), testing compiler-generated impls rather than cache logic.

The derive is guaranteed by the compiler, so the test adds maintenance cost without catching any real regression.

Remove it, or replace it with an assertion on behavior that depends on version comparison (e.g. that compare-and-swap rejects a stale version).

gears/system/cluster/cluster-sdk/src/defaults/guard.rs
```diff
@@ -0,0 +1,94 @@
...
+            Err(ClusterError::InvalidConfig { .. })
+        ));
+    }
+
+    #[test]
+    fn accepts_linearizable() {
+        assert!(reject_weak_consistency(CacheConsistency::Linearizable, "TestBackend").is_ok());
+    }
+
+    #[test]
```
MikeFalcon77, line 91: **LOW**

`warn_is_infallible_for_both_consistencies` calls `warn_weak_consistency` twice with no assertion and no captured side effect, so it passes regardless of whether the warning is emitted or its fields are correct.

A test with no assertion only proves the function does not panic, which gives false confidence that the warning behavior is covered.

Capture the emitted event via a `tracing` test subscriber and assert the warning is logged with the expected `weak_consistency` field, or drop the test.

gears/system/cluster/cluster-sdk/src/lock/types.rs
```diff
@@ -0,0 +1,47 @@
...
+        Self { linearizable }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::LockFeatures;
+
+    #[test]
+    fn features_constructor_sets_flag() {
```
MikeFalcon77: **LOW**

`features_constructor_sets_flag` constructs `LockFeatures::new(true)`/`(false)` and reads back the same `linearizable` field, verifying only that the constructor stores its argument.

A constructor-echo test asserts nothing about behavior and breaks only if the field is renamed.

Drop it and rely on the resolver/capability tests that exercise `LockFeatures` through real capability validation.

gears/system/cluster/cluster-sdk/src/leader/types.rs
```diff
@@ -0,0 +1,223 @@
...
+        // 1ns / (255 + 1) == 0ns — a non-zero ttl/budget that still derives a
+        // zero renewal interval must be rejected.
+        assert!(matches!(
+            ElectionConfig::new(Duration::from_nanos(1), 255),
+            Err(ClusterError::InvalidConfig { .. })
+        ));
+    }
+
+    #[test]
+    fn features_constructor_sets_flag() {
```
MikeFalcon77: **LOW**

`features_constructor_sets_flag` constructs `LeaderElectionFeatures::new(true)`/`(false)` and immediately reads back the `linearizable` field, exercising no logic beyond field storage.

A constructor-echo test gives no coverage of the behavior that matters and only adds maintenance surface.

Remove it; the leader resolver tests already verify the `linearizable` flag drives capability validation.

gears/system/cluster/cluster-sdk/src/cache/watch.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// the returned [`CacheWatchSender`]; the [`CacheWatch`] is handed to the
+    /// consumer. `capacity` bounds the in-flight buffer.
+    ///
+    /// The watch carries no resubscribe seam, so [`auto_restart`](Self::auto_restart)
+    /// propagates a retryable close unchanged rather than reconnecting; the
+    /// facade installs the seam on watches it hands to consumers.
+    #[must_use]
+    pub fn channel(capacity: usize) -> (CacheWatchSender, Self) {
+        let (tx, rx) = mpsc::channel(capacity);
+        (
```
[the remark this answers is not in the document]
asmith987: Resolved by documenting the panic. `mpsc::channel(0)` is a programmer-error input; returning `Result` would churn the `#[must_use]` tuple-returning API, and clamping with `max(_, 1)` would silently mask the bug. The idiomatic fix is a `# Panics` doc section (tokio already panics with a clear message). Added `# Panics` here and to the sibling `*::channel` constructors in `discovery/watch.rs`, `discovery/handle.rs`, `lock/guard.rs`, and `leader/watch.rs`.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,695 @@
...
+                CacheWatchEvent::Lagged { dropped } => Some(ServiceWatchEvent::Lagged { dropped }),
+                CacheWatchEvent::Reset => Some(ServiceWatchEvent::Reset),
+                CacheWatchEvent::Closed(err) => {
+                    let _closed = sender.send(ServiceWatchEvent::Closed(err)).await;
+                    break;
+                }
+            };
+            if let Some(service_event) = outcome
+                && sender.send(service_event).await.is_err()
+            {
```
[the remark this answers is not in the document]
asmith987: Confirmed and fixed. `CacheBasedServiceDiscoveryBackend::revoke` cancels the shutdown token and then awaits each translator `JoinHandle`, so a `sender.send(...).await` blocked on a backpressured-but-still-live consumer would hang `ClusterHandle::stop`. The `WatchTranslator` run loop now routes every outgoing send through a `tokio::select!` against `shutdown.cancelled()`; on shutdown it best-effort `try_send`s a terminal `Closed(Shutdown)` and exits, mirroring the existing recv-arm pattern. Normal-path backpressure semantics are unchanged.

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
```diff
@@ -0,0 +1,493 @@
...
+        false
+    }
+
+    /// Deletes the claim only if this participant still holds it (so a foreign
+    /// holder is never released).
+    async fn release_if_holder(&self) -> Result<(), ClusterError> {
+        match self.cache.get(&self.key).await {
+            Ok(Some(entry)) if entry.value.as_slice() == self.node_id.as_bytes() => {
+                self.cache.delete(&self.key).await.map(|_| ())
+            }
```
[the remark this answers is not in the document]
asmith987: Acknowledged, but this is an accepted, documented design tradeoff rather than a bug fixable here. The cache backend contract exposes no compare-and-delete (CAS matches on *version*, not value), so an atomic check-and-delete cannot be expressed without extending the frozen cache contract. Per ADR-002 the no-remote-in-critical-section rule plus the TTL bound is the chosen mitigation, and the identical window is documented exhaustively on `LockGuard::release`. `release_if_holder` deliberately mirrors that release path; changing only this site would diverge from the lock primitive and the ADR. No code change.

gears/system/cluster/cluster-sdk/src/discovery/handle.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// for the registered `instance_id`.
+    ///
+    /// `buffer` bounds the in-flight command buffer. A buffer of `1` suffices
+    /// when the consumer awaits each command before issuing the next; size it
+    /// larger only if a handle is shared across tasks issuing concurrent
+    /// updates.
+    #[must_use]
+    pub fn channel(instance_id: String, buffer: usize) -> (ServiceCommandReceiver, Self) {
+        let (tx, rx) = mpsc::channel(buffer);
+        let handle = Self {
```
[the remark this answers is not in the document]
asmith987: Resolved by documenting the panic — added a `# Panics` section. See the `cache/watch.rs` thread for the rationale on choosing docs over `Result`/clamping for this programmer-error input.

gears/system/cluster/cluster-sdk/src/leader/watch.rs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    /// Emits a non-status event (`Lagged` / `Reset` / `Closed`). These do not
+    /// change the cached leadership status.
+    ///
+    /// # Errors
+    /// Returns the unsent event back if the consumer has dropped the watch.
+    pub async fn send(&self, event: LeaderWatchEvent) -> Result<(), LeaderWatchEvent> {
+        self.events.send(event).await.map_err(|err| err.0)
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed. `send()` now `debug_assert!`s that the event is not `Status(_)`, with a message directing callers to `send_status` (which updates the snapshot and emits the matching event together). Returning `Err(event)` was rejected because it would conflate misuse with the existing consumer-dropped error and change the return contract; a debug assertion enforces the invariant at the boundary without altering release behavior. Internal callers already route `Status` only through `send_status`/`revoke_for_shutdown`.

gears/system/cluster/cluster-sdk/src/lock/guard.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// held lock `name`.
+    ///
+    /// `buffer` bounds the in-flight command buffer. A buffer of `1` suffices
+    /// when the consumer awaits each [`extend`](Self::extend) before issuing the
+    /// next; size it larger only if a guard is shared across tasks that may
+    /// issue concurrent extensions.
+    #[must_use]
+    pub fn channel(name: String, buffer: usize) -> (LockCommandReceiver, Self) {
+        let (tx, rx) = mpsc::channel(buffer);
+        let guard = Self { name, commands: tx };
```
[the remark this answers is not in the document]
asmith987: Resolved by documenting the panic — added a `# Panics` section. See the `cache/watch.rs` thread for the rationale on choosing docs over `Result`/clamping.

gears/system/cluster/cluster-sdk/tests/watch_lifecycle.rs
```diff
@@ -0,0 +1,221 @@
...
+    assert!(matches!(
+        watch.recv().await,
+        Some(CacheWatchEvent::Closed(ClusterError::Shutdown))
+    ));
+    // @cpt-end:cpt-cf-clst-flow-cache-primitive-watch-recover:p1:inst-w-closed
+    // Every event was consumed exactly once; the stream then ends.
+    // @cpt-begin:cpt-cf-clst-flow-cache-primitive-watch-recover:p1:inst-w-stop
+    drop(tx);
+    assert!(watch.recv().await.is_none());
+    // @cpt-end:cpt-cf-clst-flow-cache-primitive-watch-recover:p1:inst-w-stop
```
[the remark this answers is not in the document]
asmith987: Wont change. In every flagged case all events are `send`-ed into the bounded channel before any `recv` is awaited, so each `recv` resolves immediately from the buffer (and the `drop(tx)` cases resolve to `None`/synthesized `Closed` immediately). There is no path where these awaits block indefinitely under the current API — a regression would surface as a wrong or `None` value (assertion failure), not a hang. Wrapping the ~12 calls in `tokio::time::timeout` adds churn without removing a real hang risk.

gears/system/cluster/cluster/src/gear.rs
```diff
@@ -0,0 +1,114 @@
...
+                Self::MODULE_NAME
+            )
+        })?;
+
+        // Backends (and their background tasks) come up here and are registered
+        // under `cluster:{profile}`; the handle owns each plugin's shutdown hook.
+        let handle =
+            ClusterWiring::from_config(Arc::clone(hub), config, &Self::provider_registry())?;
+        *self.handle.lock().unwrap_or_else(PoisonError::into_inner) = Some(handle);
+        Ok(())
```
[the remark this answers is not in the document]
asmith987: Wont change. The `RunnableCapability` lifecycle invokes `start` once per gear instance, and `init` uses `OnceLock` for hub/config so a second init already errors — there is no supported path that calls `start` twice on a live gear. The suggested stop-the-old-handle-first guards against a state the framework does not produce.

gears/system/cluster/cluster/src/gear.rs
```diff
@@ -0,0 +1,114 @@
...
+        let handle = self
+            .handle
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .take();
+        if let Some(handle) = handle {
+            tokio::select! {
+                () = handle.stop() => {}           // graceful: deregister + compose plugin stops
+                () = deadline.cancelled() => {}    // framework deadline elapsed
+            }
```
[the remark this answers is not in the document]
asmith987: This is intentional. The deadline branch exists precisely because `handle.stop()` is taking too long; unconditionally calling `stop()` again after the deadline elapsed would re-block on the same hung teardown and defeat the deadline. When the deadline wins, the `stop()` future is dropped and the already-`take()`n handle is dropped — `Drop` does no I/O and remaining resources lapse via backend TTL, exactly as DESIGN §3.13 / `cpt-cf-clst-fr-shutdown-ttl-cleanup` specify. No change.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -0,0 +1,348 @@
...
+                        "profile `{name}`: unknown cache provider `{}`",
+                        profile.cache.provider
+                    ),
+                })?;
+            let (cache, stop) = provider.build_cache(&profile.cache.options)?;
+            builder = builder
+                .profile_named(name.clone(), ProfileBackends::new(cache))
+                .on_stop(move || async move { stop().await });
+        }
+        builder.build_and_start()
```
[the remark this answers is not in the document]
asmith987: Not a permanent leak under the current design. The standalone cache sweeper holds only a `Weak` to the cache and `StandaloneClusterHandle` does no I/O on `Drop`; on the error unwind the builder drops both the cache `Arc` and the captured handle, so the last strong ref goes away and the sweeper self-terminates on its next tick. Running the async stop hooks from the synchronous `from_config`/`build_and_start` error path would require blocking or spawning for marginal benefit on a cold-start error path. Leaving as-is.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -0,0 +1,348 @@
...
+            register_cache_backend(&self.hub, &profile.name, profile.cache)?;
+            register_leader_election_backend(&self.hub, &profile.name, profile.leader_election)?;
+            register_lock_backend(&self.hub, &profile.name, profile.lock)?;
+            register_service_discovery_backend(
+                &self.hub,
+                &profile.name,
+                profile.service_discovery,
+            )?;
+            registered.push(profile.name);
+        }
```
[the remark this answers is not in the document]
asmith987: Fixed. The per-profile registration loop is now wrapped so any failure rolls back the just-attempted profile and every previously-registered one (via a new `deregister_profile` helper, also reused by `ClusterHandle::stop`) before propagating the error, leaving the hub all-or-nothing.

gears/system/cluster/docs/DESIGN.md
```diff
@@ -474,6 +526,8 @@ Multiple resolutions of the same primitive on the same profile are cheap (`Arc`-
...
+
```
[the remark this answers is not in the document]
asmith987: Fixed. Rewrote the §3.7 body and the code-example header so the section presents the single implemented model: the `cluster` gear (`cf-gears-cluster`) is the `RunnableCapability` that owns the handle and also exports the builder/handle as library API. Removed the stale separate-`cf-cluster`-crate / parent-host-gear (follow-up) narrative that contradicted the 2026-06-16 amendment.

gears/system/cluster/plugins/standalone-cluster-plugin/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+It implements the **cache** primitive natively over an in-memory store
+(`HashMap` + monotonic per-key versions, lazy TTL, a background sweeper emitting
+`Expired`, and exact/prefix watches) and derives **leader election**,
+**distributed lock**, and **service discovery** from the SDK's cache-based
+default backends — the "implement cache only, get all four" guarantee. The cache
+declares `Linearizable` consistency, which the default leader/lock backends
+require.
+
+## Lifecycle
```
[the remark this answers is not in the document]
asmith987: Wont add. Repo convention is that READMEs do not carry a TOC — neither the cluster gears three READMEs nor sampled sibling gear READMEs (`api-gateway`, `grpc-hub`, `nodes-registry`, `authn-resolver`). TOCs (`<!-- toc -->`) are used only in the `docs/` spec files (DESIGN/DECOMPOSITION/PRD/OBSERVABILITY), and there is no guideline mandating README TOCs. For consistency I did add a TOC to `docs/TRACEABILITY-AUDIT.md`, which is a docs/ spec artifact.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
asmith987: Fixed.

gears/system/cluster/cluster-sdk/src/discovery/handle.rs
asmith987: Fixed. Public API is now `set_state(&self, state: InstanceState)`

gears/system/cluster/cluster-sdk/src/discovery/handle.rs
asmith987: Fixed. Renamed to `SetState { state: InstanceState }`

gears/system/cluster/cluster/src/wiring.rs
asmith987: Fixed. `ClusterWiringBuilder::profile<P>(_marker: P, backends) `now takes the marker by value

gears/system/cluster/cluster-sdk/src/leader/types.rs
asmith987: Fixed. Deleted.

gears/system/cluster/cluster-sdk/src/cache/types.rs
asmith987: Fixed. Deleted.

gears/system/cluster/cluster-sdk/src/defaults/guard.rs
asmith987: Fixed

gears/system/cluster/cluster-sdk/src/lock/types.rs
asmith987: Fixed. Deleted; the linearizable flag is covered

gears/system/cluster/cluster-sdk/src/leader/types.rs
asmith987: Fixed. Deleted the test

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
```diff
@@ -0,0 +1,493 @@
...
+        let _closed = self.sender.send(LeaderWatchEvent::Closed(err)).await;
+        false
+    }
+
+    /// Deletes the claim only if this participant still holds it (so a foreign
+    /// holder is never released).
+    async fn release_if_holder(&self) -> Result<(), ClusterError> {
+        match self.cache.get(&self.key).await {
+            Ok(Some(entry)) if entry.value.as_slice() == self.node_id.as_bytes() => {
+                self.cache.delete(&self.key).await.map(|_| ())
```
MikeFalcon77: Release is a `get` + unguarded `delete`: between reading my own value and deleting, my TTL can lapse and another node can claim the key — this `delete` then wipes the successor's fresh claim. No split-brain (the victim re-claims on its next renewal), but it causes a spurious leadership flap on the new leader.

The k8s elector already solves this — see `release_lease` in `gears/mini-chat/mini-chat/src/infra/leader/k8s_lease.rs`: it re-checks the holder and does a resourceVersion-guarded replace (409 = someone else moved on, no-op).

Delete only if the version still matches: add a version-guarded delete (CAS-delete using `entry.version` from the same `get`), or skip the read and rely on TTL expiry.

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
```diff
@@ -0,0 +1,493 @@
...
+        // best-effort release and stop.
+        if self.sender.send_status(initial).await.is_err() {
+            let _release = self.release_if_holder().await;
+            return;
+        }
+        // @cpt-begin:cpt-cf-clst-algo-leader-election-renewal:p1:inst-rn-interval
+        let interval = self.config.renewal_interval();
+        // @cpt-end:cpt-cf-clst-algo-leader-election-renewal:p1:inst-rn-interval
+        let mut renewal =
+            tokio::time::interval_at(tokio::time::Instant::now() + interval, interval);
```
MikeFalcon77: Renewal/reclaim runs on a fixed interval with no jitter. When many participants join at once (cluster startup, or all followers after a leader drops), their `put_if_absent` reclaim attempts line up on the same cadence — a synchronized stampede on the election key every tick. Not a correctness bug (`put_if_absent` is atomic), just contention.

The k8s elector desyncs competing pods with equal-jitter backoff — see `Backoff`/`next_delay` in `gears/mini-chat/mini-chat/src/infra/leader/k8s_lease.rs` (delay = base/2 + rand(0..base/2)).

Add a small random jitter to the follower reclaim cadence so contenders spread out across the interval.

gears/system/cluster/cluster-sdk/src/leader/facade.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.inner.features()
+    }
+
+    /// Joins the named election with the default [`ElectionConfig`], returning a
+    /// [`LeaderWatch`]. The claim is renewed automatically and the watch
+    /// auto-reenrolls on a transient leadership loss.
+    ///
+    /// # Errors
+    /// Propagates any [`ClusterError`] from the backend.
+    pub async fn elect(&self, name: &str) -> Result<LeaderWatch, ClusterError> {
```
MikeFalcon77, line 94: The API is observe-only: every consumer that just wants "run this singleton while I'm leader" has to hand-write the spawn-work / cancel-on-`Status(Lost)` / abort-if-unresponsive loop themselves. That's the common case and it'll be re-implemented (inconsistently) per gear.

The k8s elector ships this as one call — `LeaderElector::run_role(role, cancel, work)` in `gears/mini-chat/mini-chat/src/infra/leader/`, including a force-abort of work that ignores its cancel token after a timeout (`stop_with_timeout`).

Add a `run_role`-style convenience on top of `LeaderWatch` (spawn work on `Leader`, cancel its token on `Lost`, abort after a stop timeout) so consumers don't each rebuild it.

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
asmith987: Fixed. `release_if_holder` now does an atomic value-guarded delete (`compare_and_delete(key, node_id)`) instead of an unconditional delete. It guards on the `node_id`. The new function `compare_and_delete(...)` should remove the read to delete window.

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
asmith987: Fixed. Added a small jitter to followers.

gears/system/cluster/cluster-sdk/src/leader/facade.rs
asmith987: Fixed

gears/system/cluster/cluster-sdk/src/defaults/leader_tests.rs
```diff
@@ -0,0 +1,277 @@
...
+    ));
+    // The intruder leaves; the watch auto-reenrolls back to leader.
+    assert!(cache.delete("election/primary").await.is_ok());
+    loop {
+        match watch.changed().await {
+            LeaderWatchEvent::Status(LeaderStatus::Leader) => break,
+            LeaderWatchEvent::Status(_) | LeaderWatchEvent::Reset => {}
+            other => panic!("unexpected event while reenrolling: {other:?}"),
+        }
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed. Wrapped the re-enrollment loop in tokio::time::timeout(5s, …) so a re-election regression fails fast with a clear panic instead of hanging CI. Duration was already imported; no new deps.

gears/system/cluster/cluster-sdk/src/defaults/leader.rs
```diff
@@ -0,0 +1,536 @@
...
+/// compare-and-swap operations (DESIGN §3.11, ADR-001).
+///
+/// Candidacy is a `put_if_absent(election_key, node_id, ttl)`; the claim is
+/// renewed on [`ElectionConfig::renewal_interval`] via a version
+/// `compare_and_swap`, so a renewal that races a foreign takeover is detected
+/// as a [`ClusterError::CasConflict`] and surfaces as
+/// [`LeaderStatus::Lost`] followed by auto-reenrollment. A `watch` on the
+/// election key reconciles status reactively (it never writes — only the
+/// renewal timer renews, so a renewal's own change event cannot re-trigger a
+/// renewal).
```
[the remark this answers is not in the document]
asmith987: Fixed.  Corrected both doc sites to state it issues no renewal compare_and_swap but may claim a vacant key, while preserving the true invariant (a renewal's own change event reconciles to a no-op, so it can't re-trigger a renewal).

gears/system/cluster/docs/OBSERVABILITY.md
```diff
@@ -0,0 +1,112 @@
+# Cluster Observability Contract — v1
```
MikeFalcon77, line 1: I haven't check all observability features but do we have :
1. Mandatory set of metrics
2. Nice-to-have (optional) extended set
?
Because it's too expensive to ask backend developers to implement 40+ metrics in MVP.

asmith987: We currently have 8 metrics defined in the SDK, which are emitted for every backend. All metrics flow through the trait contract below, so if a plugin needs to control how those metrics are emitted, it can supply its own implementation of the trait. The SDK provides an OTel adapter `OtelClusterMetrics` behind the `otel` feature. By default, the metric emission is a noop using `NoopMetrics` until opted in. For now, we leave any custom metrics emission outside of the cluster contract, but backends are free to output as they please.

```rust
/// The metrics sink the SDK instrumentation emits through.
///
/// OTel-agnostic output port so the SDK core carries no OpenTelemetry dependency
/// and the emission sites (the [`InstrumentedCache`](crate::InstrumentedCache)
/// decorator and the SDK default backends) stay testable with a recording
/// double. The concrete OpenTelemetry adapter
/// [`OtelClusterMetrics`](crate::observability::otel::OtelClusterMetrics) lives
/// behind the `otel` crate feature.
///
/// The `provider` label is fixed when a concrete sink is constructed (one sink
/// per provider in a deployment), so it is not a per-call argument. Every method
/// takes only the bounded, allowlisted label values from this module — never a
/// high-cardinality key, lock name, or instance id (the cardinality rule).
///
/// Spans and log events are emitted directly via `tracing` at the instrumentation
/// sites (they need no sink); only metrics flow through this port.
pub trait ClusterMetrics: Send + Sync {
    /// Records one cache operation. `op` is a facade op (e.g. `get`); `result` is
    /// a [`result`](self::result) value. Backs [`metrics::CACHE_OPS_TOTAL`].
    fn cache_op(&self, op: &str, result: &str);
    /// Records a cache operation's latency. Backs
    /// [`metrics::CACHE_OP_DURATION_SECONDS`].
    fn cache_op_duration(&self, op: &str, seconds: f64);
    /// Records one lock operation. Backs [`metrics::LOCK_OPS_TOTAL`].
    fn lock_op(&self, op: &str, result: &str);
    /// Records a lock operation's latency. Backs
    /// [`metrics::LOCK_OP_DURATION_SECONDS`].
    fn lock_op_duration(&self, op: &str, seconds: f64);
    /// Records a leadership transition. `transition` is a
    /// [`transition`](self::transition) value. Backs
    /// [`metrics::LEADER_TRANSITIONS_TOTAL`].
    fn leader_transition(&self, transition: &str);
    /// Records one service-discovery operation. Backs
    /// [`metrics::DISCOVERY_OPS_TOTAL`].
    fn discovery_op(&self, op: &str, result: &str);
    /// Records a watch reset/resubscription for `primitive` (a
    /// [`primitive`](self::primitive) value). Backs
    /// [`metrics::WATCH_RESETS_TOTAL`].
    fn watch_reset(&self, primitive: &str);
    /// Records a backend/provider error of bounded `kind` (a [`kind`](self::kind)
    /// value). Backs [`metrics::PROVIDER_ERRORS_TOTAL`].
    fn provider_error(&self, kind: &str);
}
```

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,845 @@
...
+        tasks.retain(|handle| !handle.is_finished());
+        tasks.push(handle);
+    }
+
+    fn prefix(name: &str) -> String {
+        format!("{name}/")
+    }
+
+    fn instance_key(name: &str, instance_id: &str) -> String {
+        format!("{name}/{instance_id}")
```
MikeFalcon77, line 167: HIGH (ADR-001). The SD default key scheme is `{name}/{instance_id}` with no `svc/` prefix, while leader uses `election/{name}` (leader.rs:126) and lock uses `lock/{name}` (lock.rs:155). ADR-001 specifies `svc/{name}/{instance_id}` and `watch_prefix(svc/{name}/)` precisely to namespace SD. In an omit-primitive profile the wiring auto-wrap clones one cache Arc into all four defaults, so they share a keyspace. An unprefixed SD key then collides with `election/`/`lock/`: registering a service named `election` does a plain `put` over `election/{instance_id}` and can overwrite a live leader claim; `discover("election", ..)` surfaces leader claims as phantom instances via `watch_prefix("election/")`/`scan_prefix`. Restore the `svc/` prefix, ideally via a single shared set of per-primitive key prefixes.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -0,0 +1,375 @@
...
+            stop_hooks: self.stop_hooks,
+            revokers,
+        })
+    }
+}
+
+/// The running cluster wiring. Backends are registered in the hub; consumers
+/// resolve them with the SDK resolvers (e.g.
+/// `ClusterCacheV1::resolver(handle.hub())`). Owns the wired plugins' shutdown.
+pub struct ClusterHandle {
```
MikeFalcon77, line 312: HIGH (ADR-006). `ClusterHandle` has no `Drop` impl. ADR-006 mandates a panic-safe diagnostic guard (debug-build panic / release-build warn, gated by `std::thread::panicking()`) plus two named confirmation tests. Without it, dropping the handle without calling `stop()` (e.g. a panic during gear `start()` after `build_and_start()`) leaves backends registered in the hub and never runs the plugin stop-hooks, leaking the standalone sweeper and SDK-default election tasks with no diagnostic. Add the guard: no async I/O, warn/panic when `registered` is non-empty.

gears/system/cluster/cluster-sdk/src/restart.rs
```diff
@@ -0,0 +1,444 @@
...
+    ///   cap, or a retryable close on a seam-less watch — then the watch is
+    ///   terminal;
+    /// - `None` once terminal, or if the underlying stream ended without a
+    ///   `Closed`.
+    pub async fn recv(&mut self) -> Option<W::Event> {
+        if self.terminated {
+            return None;
+        }
+        // @cpt-begin:cpt-cf-clst-flow-watch-auto-restart-wrap:p1:inst-ar-next
+        let Some(event) = self.inner.next_event().await else {
```
MikeFalcon77, line 232: MEDIUM (ADR-003). When the underlying cache/service watch drops its sender without a terminal `Closed`, `next_event()` returns `None` and the combinator marks itself terminated here without reconnecting or synthesizing `Closed`, even when a resubscribe seam exists. `LeaderWatch::changed()` is hardened (synthesizes `Closed(Shutdown)` on `None`), but cache/service watches are not. A remote backend whose connection task drops the sender on a transient fault (the canonical reconnect trigger) yields a permanently dead, never-reconnected watch. Treat `None` on a resubscribable watch as a retryable close: reconnect, or synthesize `Closed(ConnectionLost)`.

gears/system/cluster/cluster-sdk/src/leader/watch.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// buffer cannot stall shutdown: the snapshot guard and `changed()`'s
+    /// synthesized close keep the contract even if the events are dropped.
+    pub(crate) fn revoke_for_shutdown(&self, was_leader: bool) {
+        if was_leader {
+            // Latches even with no receiver; a racing `status()` cannot observe
+            // stale leadership after this returns.
+            self.snapshot.send_replace(LeaderStatus::Lost);
+            let _lost = self
+                .events
+                .try_send(LeaderWatchEvent::Status(LeaderStatus::Lost));
```
MikeFalcon77, line 471: MEDIUM (ADR-003). The shutdown two-step `Status(Lost)` then `Closed(Shutdown)` is delivered via `try_send` with the result discarded. Under a full event buffer both are dropped, and a consumer reading via `changed()` then observes only the synthesized `Closed(Shutdown)`, never `Status(Lost)`. ADR-003 requires the two events to be distinct so event-stream consumers revoke leader confidence before the stream ends. Gate-pattern consumers are covered by the `status()`->`Lost` guard, but pure event-stream consumers lose the guarantee. Reserve buffer headroom for terminal events or deliver them on a non-droppable path.

gears/system/cluster/cluster-sdk/src/defaults/lock.rs
```diff
@@ -0,0 +1,509 @@
...
+    ///
+    /// Non-atomic window (accepted tradeoff, ADR-002): this is a `get`-then-CAS,
+    /// and the CAS matches on *version* only. If the lease lapses between the
+    /// `get` and the CAS and a new holder re-acquires via `put_if_absent` (a
+    /// fresh entry at version `1`), a holder whose last-seen version was also `1`
+    /// can have its CAS match and overwrite the foreign entry — a lock steal,
+    /// strictly worse than the release window's delete. Bounded by the same
+    /// critical-section-shorter-than-TTL rule as [`release`](Self::release); see
+    /// [`LockGuard::extend`](crate::lock::LockGuard::extend) for the consumer note.
+    async fn extend(&self, additional_ttl: Duration) -> Result<(), ClusterError> {
```
MikeFalcon77, line 440: MEDIUM. `extend(additional_ttl)` sets the lease to `additional_ttl` (CAS overwrite) rather than adding to the remaining TTL. Passing a value smaller than the time currently left silently shortens the lock, which contradicts the method name. Rename to `renew`/`set_ttl`, or compute against the remaining TTL.

gears/system/cluster/cluster-sdk/src/profile.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/// Maps a profile name to its stable cluster lookup [`ClientScope`]
+/// (`cluster:{profile}`), the scope under which every primitive resolves and
+/// registers its backend for that profile.
+///
+/// # Errors
+/// Returns [`ClusterError::InvalidName`] if `name` violates
+/// [`CLUSTER_NAME_RULE`], before any backend lookup is attempted.
+pub fn profile_scope(name: &str) -> Result<ClientScope, ClusterError> {
```
MikeFalcon77, line 78: LOW (ADR-007). `profile_scope` (and the name-validation helpers) are `pub` and re-exported from the crate root (lib.rs:83); ADR-007 describes `profile_scope` as an internal-only helper. No string-based profile resolution path is exposed to consumers, so this is surface hygiene only. Consider `pub(crate)` to keep the typed `ClusterProfile` marker the sole consumer-facing profile path.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,845 @@
...
+        );
+    }
+}
+
+/// Default registration TTL — an instance that stops heartbeating disappears
+/// within this window.
+const DEFAULT_TTL: Duration = Duration::from_secs(30);
+
+/// Default heartbeat cadence (`ttl / 3`, mirroring the leader-election default).
+const DEFAULT_RENEWAL: Duration = Duration::from_secs(10);
```
MikeFalcon77, line 51: LOW. SD heartbeat cadence is the hardcoded constant `ttl/3` rather than the derived `ttl/(max_missed_renewals+1)` used by leader election (SD config has no `max_missed_renewals`). Numerically equal to the leader default; minor divergence from the documented formula. Acceptable, but a shared helper or a comment linking the two would prevent drift.

gears/system/cluster/cluster-sdk/src/defaults/lock.rs
```diff
@@ -0,0 +1,509 @@
...
+        let (receiver, guard) = LockGuard::channel(name.to_owned(), COMMAND_BUFFER);
+        let task = GuardTask {
+            cache: Arc::clone(&self.cache),
+            name: name.to_owned(),
+            key,
+            holder,
+            provider: self.provider,
+            metrics: Arc::clone(&self.metrics),
+        };
+        tokio::spawn(task.run(receiver));
```
MikeFalcon77, line 169: LOW (ADR-006). `GuardTask` is a detached spawn not tied to backend `revoke()`; on graceful shutdown it is not cancelled or joined and self-terminates only when the consumer drops the guard. Bounded and documented, but it outlives `revoke()` — worth a CancellationToken or at least an explicit note that this is intentional.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,845 @@
...
+                return Err(err);
+            }
+        };
+        let task = StoreMaintainer {
+            cache: Arc::clone(&self.cache),
+            shared: Arc::downgrade(&self.shared),
+            name: name.to_owned(),
+            prefix,
+        };
+        tokio::spawn(task.run(watch));
```
MikeFalcon77, line 204: LOW (ADR-006). `StoreMaintainer` is spawned detached and not tracked/cancelled by `revoke()`, unlike the tracked tasks at lines 281/351. It self-terminates via `Weak` upgrade failure on its next tick rather than being joined on shutdown. Bounded, but it outlives graceful shutdown; consider tracking it like the others for deterministic teardown.

MikeFalcon77 (commented):
Architecture-conformance review against ADRs 001-010. Findings posted inline below: 2 high (H1 key-collision, H2 missing Drop guard), 3 medium, 4 low. Severity rationale is in each comment.

asmith987:
Remaining comments were resolved here: https://github.com/constructorfabric/gears-rust/pull/4123#issuecomment-4774206449

PR #4121, opened 2026-06-18
feat(security): add two-plane auth primitives and HTTP/gRPC middleware

### LINKED ISSUE #4103: [OoP-1] SecurityContext HTTP Propagation (JWT Re-validation)
**Parent**: [OoP](https://github.com/constructorfabric/gears-rust/issues/4102)

### Description

Implement SecurityContext propagation over HTTP using JWT re-validation. JWT is **always re-validated** at every hop — no
trusted-peer fast path.

OoP-1, OoP-2, and OoP-3 have no hard dependencies and can be worked in parallel. They converge at OoP-4 — the OoP runtime installs the OoP-1/OoP-2 middleware and consumes OoP-3's REST registration.

**PRD**: `cpt-cf-fr-secctx-propagation`, `cpt-cf-interface-secctx-http`
**Design**: `cpt-cf-component-secctx-http`
**ADR**: [0008-cpt-cf-adr-two-plane-auth](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md)

Header contract (shared with OoP-2):

| Scenario | `Authorization` | `X-ToolKit-Internal-Token` |
|----------|-----------------|---------------------------|
| User request (gateway → OoP) | `Bearer <jwt>` | — |
| User-propagated (gear → gear) | `Bearer <jwt>` (forwarded from inbound) | — |
| Platform / non-tenant (registration, global GTS) | — | `<sa-jwt>` |
| Tenant-scoped system, no user (e.g. event for a tenant) | `Bearer <s2s-jwt>` | — |

The plane is chosen by **tenant-scoped vs non-tenant**, not user-vs-system: a system actor doing tenant-scoped work
uses the tenant plane (S2S JWT + AuthZ); only genuinely non-tenant work uses the platform plane. A request carries one
plane, not both. The tenant-plane JWT is re-validated at every hop
([ADR-0008](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md)). Gear identity on user-propagated calls
is not currently established; mTLS+SPIFFE would provide it in the next phase (see
[ADR-0008](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md) — platform plane).

**SA token scope**: SA tokens target infrastructure endpoints only (DirectoryService, GTS registration, health
probes) — no `SecurityContext` or `PolicyEnforcer`. For business endpoints without a user context, gears use
S2S client-credentials (`AuthNResolverClient::exchange_client_credentials()`) — already implemented and in
production. See [ADR-0008](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md) (tenant vs platform plane).

### Tasks

- [x] **1.1** Implement SecurityContext HTTP helpers in `libs/toolkit-http/src/security.rs`:
  - `SecCtxHttpError` error enum (missing auth header, invalid auth header format, empty token)
  - `attach_bearer_http<B>(request: &mut http::Request<B>, secctx: &SecurityContext)` — sets `Authorization: Bearer <jwt>` from `secctx.bearer_token()`; expose the secret only at this boundary and never log it
  - `extract_bearer_http(headers: &HeaderMap) -> Result<String, SecCtxHttpError>` — extracts the raw bearer token from `Authorization`; match the `Bearer` scheme **case-insensitively**, trim surrounding whitespace, and reject an empty token
- [x] **1.2** Define a minimal `BearerAuthenticator` trait (+ neutral `AuthnError`) in `toolkit-security` (`async fn authenticate(&self, token: &str) -> Result<SecurityContext, AuthnError>`); add an `axum-middleware` feature to `toolkit-http` and implement `secctx_middleware` generic over it:
  - Extract the token via `extract_bearer_http`, then always call `BearerAuthenticator::authenticate(token)` — no fast path (zero-trust). The concrete `AuthNResolverClient` adapter is injected at the gear/bootstrap layer (OoP-4 § 4.1) via Axum state
  - Insert the resulting `SecurityContext` into request extensions
  - Derive whether a route requires a user context per-route from `OperationBuilder` (`.authenticated()` / `.public()` → `OperationSpec.is_public`), surfaced as a route extension/marker:
    - protected route + no/invalid `Authorization` → **401**
    - public / system-only route + no `Authorization` → pass through
  - Return 401 on invalid token or auth failure
- [x] **1.3** Add deps to `toolkit-http/Cargo.toml`: `secrecy`; `toolkit-security` (for `SecurityContext` + the `BearerAuthenticator` trait / `AuthnError`); optional `axum` behind the `axum-middleware` feature. **Do NOT add `authn-resolver-sdk`** — it pulls the whole `toolkit` framework into the transport crate and inverts the layering. The concrete `AuthNResolverClient` is adapted to `BearerAuthenticator` at the gear/bootstrap layer instead.
- [x] **1.4** Export public API from `toolkit-http/src/lib.rs`
- [x] **1.5** Unit tests: attach→extract round-trip, error cases, pass-through for system-only calls

> **Note (1.2):** `BearerAuthenticator` lives in `toolkit-security` (not `toolkit-http`) so it stays transport-agnostic and reusable by the gRPC path — it returns `SecurityContext`, which `toolkit-security` already owns, and `toolkit-security` has no dependency on `toolkit-http`.

### Acceptance Criteria

- `cargo test -p toolkit-http --all-features` passes
- `secctx_middleware` always re-validates the JWT via the injected `BearerAuthenticator`
- A public / system-only route without a user JWT passes through `secctx_middleware` without 401
- A route that requires user context returns 401 when no `Authorization` header is present
- Forged JWT is rejected with 401
- `toolkit-http` has no dependency on `authn-resolver-sdk` (verify with `cargo tree`)

### LINKED ISSUE #4104: [OoP-2] Platform-Plane Authentication (K8s SA Tokens)
**Parent**: [OoP](https://github.com/constructorfabric/gears-rust/issues/4102)

### Description

Implement platform-plane authentication using K8s SA tokens (Profile 3) and the
`InternalAuthMiddleware` that sets `PeerAuthenticated { gear }` for workload policy. SA tokens always use
`X-ToolKit-Internal-Token`, never `Authorization` (avoids collision with user JWT). The `InternalCredential`
enum also **reserves** struct-only variants for Profile 2 bootstrap tokens and mTLS, but their
validation/wiring is **out of scope for P1** (deferred) — the variants exist only to freeze the API surface.

SA token scope: **infrastructure endpoints only** (DirectoryService registration/heartbeats, GTS entity
registration, health probes). These endpoints do not require `SecurityContext` or `PolicyEnforcer`. SA tokens cannot
produce a meaningful `SecurityContext`, so `PolicyEnforcer` would deny business endpoint calls.

For business endpoints without a user context, gears use S2S client-credentials
(`AuthNResolverClient::exchange_client_credentials()`) — already implemented and used by mini-chat for OAGW
provisioning. This is orthogonal to OoP-2.

OoP-1, OoP-2, and OoP-3 have no hard dependencies and can be worked in parallel. They converge at OoP-4 — the OoP runtime installs the OoP-1/OoP-2 middleware and consumes OoP-3's REST registration.

**PRD**: `cpt-cf-fr-internal-auth`
**Design**: `cpt-cf-component-platform-plane-auth`
**ADR**: [0006-cpt-cf-adr-platform-plane-auth](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md)

### Tasks

- [x] **2.1** Define `InternalCredential` enum in `libs/toolkit-security/src/internal_auth.rs`:
  - `None` (Profile 1 — no header)
  - `BootstrapToken(SecretString)` (Profile 2 — struct only; validation/wiring deferred to P2)
  - `KubeServiceAccountToken { token_path, audience }` (Profile 3 — struct + impl)
  - `MtlsIdentity { cert, key, ca }` (struct only — mTLS deferred, see ADR-0006)
- [x] **2.2** Implement `attach_internal_auth` — reads credential, sets `X-ToolKit-Internal-Token` header (HTTP) or gRPC metadata. SA tokens always go in `X-ToolKit-Internal-Token`, never `Authorization` (avoids collision with user JWT)
- [ ] **2.3** Implement `validate_internal_auth` — dispatches by an explicit token-shape discriminator (three dot-separated base64url segments that decode to a JWT header ⇒ JWT; otherwise opaque):
  - JWT-shaped → validate via K8s TokenReview API (cached, configurable TTL)
  - Opaque (bootstrap token) → Profile 2, **deferred to P2**; not wired in P1
- [x] **2.4** `InternalAuthMiddleware` for Axum:
  - Extracts `X-ToolKit-Internal-Token` header
  - Validates credential → sets `PeerAuthenticated { gear }` in request extensions (workload policy only — does NOT skip JWT validation)
  - Permissive on user-only endpoints (no SA token required)
  - Must run BEFORE `secctx_middleware` (DESIGN § 3.2 "Middleware order")
- [x] **2.5** `InternalAuthInterceptor` for tonic — gRPC counterpart for system calls to DirectoryService. (The tenant-plane gRPC hot-path is P2; when added, it reuses OoP-1's `BearerAuthenticator` from `toolkit-security` rather than defining a new abstraction.)
- [ ] **2.6** `KubeServiceAccountToken` reader — watches projected volume, reloads on rotation
- [x] **2.7** Define `PlatformSecurityContext` in `libs/toolkit-security` — a type **distinct** from the tenant `SecurityContext` (no tenant subject; never accepted by the tenant `PolicyEnforcer`). First-class from phase 1, populated from the validated SA identity now, from the SPIFFE `PlatformIdentity` once mTLS lands. See [ADR-0008](https://github.com/mattgarmon/gears-rust/blob/359eea89abcdc9bb53328d4f692ffedf52496332/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md) (platform plane)
- [x] **2.8** `InternalAuthMiddleware` builds a `PlatformSecurityContext` from the validated credential and inserts it into request extensions (alongside `PeerAuthenticated`)
- [x] **2.9** Export public API from `toolkit-security/src/lib.rs`
- [x] **2.10** Unit tests: attach→validate round-trip (SA token), rejection of bad tokens, validation order, `PeerAuthenticated` marker propagation, and a compile-fail test that `PlatformSecurityContext` cannot be passed where a tenant `SecurityContext` is expected

> **Note (2.8):** No P1 *reader* exists yet (consumers are in the gear-code follow-up), but populating `PlatformSecurityContext` alongside `PeerAuthenticated` now freezes the middleware API and keeps the platform plane symmetric with the tenant plane — avoiding a later change to security-critical middleware.

### Acceptance Criteria

- `cargo test -p toolkit-security --all-features` passes
- `InternalAuthMiddleware` validates SA tokens, sets `PeerAuthenticated`, and builds a `PlatformSecurityContext`
- `PlatformSecurityContext` is type-distinct from the tenant `SecurityContext` and never reaches the tenant `PolicyEnforcer`
- `PeerAuthenticated` does NOT skip JWT validation (verified by OoP-1 tests when both middlewares are stacked)
- Requests with invalid SA token are rejected before SecurityContext is processed

mattgarmon:
Closes https://github.com/constructorfabric/gears-rust/issues/4103
Closes https://github.com/constructorfabric/gears-rust/issues/4104

Implements the library-layer foundation for both authentication planes defined
in the two-plane auth model. Nothing is wired to a running gear yet - that comes
later (OoP-4), along with the concrete K8s `TokenReview` validator and the SA
token file-watcher.
 
Adds a new crate, **`toolkit-http-middleware`**, as the shared home for
server-side inbound auth layers - kept separate from the outbound HTTP client.
Rejections render as canonical RFC 9457 `application/problem+json`.
 
## OoP-1 - Tenant-plane JWT re-validation
 
The user's JWT is **always** re-validated at every hop via an injected
authenticator - there is no trusted-peer fast path (zero-trust). Outbound calls
forward the user's token from the security context; inbound middleware
re-validates it and makes the reconstructed security context available to
handlers. Routes that carry no JWT (health probes, platform-only handlers) can
be marked so they pass through instead of returning `401`.
 
## OoP-2 - Platform-plane authentication primitives
 
System-initiated calls carry a dedicated internal-token header, never
`Authorization`, so they never collide with the tenant JWT - the two planes are
mutually exclusive per request. This adds the platform credential, identity, and
security-context types, a neutral validation error, and the validator trait the
transport layers depend on. The credential set is frozen up front; only the
first-phase variants are validated, with the bootstrap-token and mTLS variants
present but deferred.
 
Inbound HTTP middleware validates the internal token and exposes the platform
identity (plus a peer-authenticated marker) to handlers - permissive when the
header is absent, `401` when it is present but invalid. The gRPC path gets
matching attach/extract helpers and a rotation-aware client interceptor
(rotating, static, or disabled credential modes).
 
A compile-fail test enforces that a platform security context can never be passed
where a tenant security context is expected.

## Summary by CodeRabbit

* **New Features**
  * Added two-plane authentication support for HTTP and gRPC requests.
  * Introduced middleware and helpers for handling bearer and internal tokens, plus public-route bypass support.
  * Added support for authenticated platform/internal requests and request-level identity propagation.

* **Bug Fixes**
  * Improved token handling by trimming whitespace, rejecting invalid formats, and avoiding accidental exposure in logs.
  * Added clearer error responses for missing, invalid, or unavailable authentication services.

* **Documentation**
  * Added architecture and crate documentation for the new authentication flow.

### DIFF
--- a/libs/toolkit-http-middleware/src/auth.rs
+++ b/libs/toolkit-http-middleware/src/auth.rs
@@ -1,0 +1,262 @@
+//! Axum middleware for two-plane authentication.
+//!
+//! Two complementary middlewares:
+//!
+//! - [`security_context_middleware`] (**tenant plane**) extracts the bearer token from
+//!   the incoming `Authorization` header and **always** re-validates it via an
+//!   injected [`BearerAuthenticator`] — there is no trusted-peer fast path
+//!   (zero-trust). On success the reconstructed [`SecurityContext`](toolkit_security::SecurityContext)
+//!   is inserted into the request extensions for downstream handlers and the
+//!   `AuthZ` resolver.
+//! - [`internal_auth_middleware`] (**platform plane**) extracts the
+//!   `X-ToolKit-Internal-Token` header and, if present, validates it via an
+//!   injected [`InternalAuthenticator`], inserting [`PeerAuthenticated`] and a
+//!   [`PlatformSecurityContext`] for workload-policy / platform handlers.
+//!
+//! **Middleware order:** when both are installed, [`internal_auth_middleware`]
+//! runs **before** [`security_context_middleware`] (DESIGN § 3.2). The two middlewares are
+//! independent: each handles its own plane and the planes are mutually exclusive
+//! per request — system calls carry `X-ToolKit-Internal-Token` (no JWT); user
+//! calls carry `Authorization: Bearer` (no internal token). [`PeerAuthenticated`]
+//! is never a prerequisite for JWT validation; [`security_context_middleware`] does not
+//! consult it.
+//!
+//! Routes that carry no tenant JWT (probes, platform-plane-only handlers) are
+//! marked with the [`PublicRoute`] request extension by the gear/bootstrap layer;
+//! note this is distinct from `OperationSpec.is_public`, which controls gateway
+//! registration. The concrete authenticator adapters are injected via Axum state
+//! at the same layer.
+
+use std::sync::Arc;
+
+use axum::{
+    extract::{Request, State},
+    middleware::Next,
+    response::{IntoResponse, Response},
+};
+use toolkit_canonical_errors::CanonicalError;
+use toolkit_security::{
+    AuthNError, BearerAuthenticator, InternalAuthNError, InternalAuthenticator, PeerAuthenticated,
+    PlatformSecurityContext,
+};
+
+use crate::security::{
+    InternalTokenHttpError, SecurityContextHttpError, extract_bearer_http,
+    extract_internal_token_http,
+};
+use secrecy::ExposeSecret;
+
+/// Retry hint (seconds) advertised when the authentication backend is
+/// temporarily unavailable, mirroring `api-gateway`'s authn middleware.
+const AUTH_RETRY_AFTER_SECONDS: u64 = 5;
+
+/// Public detail for an unexpected authentication-infrastructure failure,
+/// mirroring `api-gateway`'s authn middleware. Carries no diagnostic specifics.
+const AUTH_INFRA_FAILURE_DETAIL: &str = "authentication infrastructure failure";
+
+/// Per-route marker indicating the route carries **no tenant JWT** and must not
+/// require a [`SecurityContext`](toolkit_security::SecurityContext).
+///
+/// Inserted by the gear/bootstrap layer for routes that never carry an
+/// `Authorization: Bearer` header — framework probe endpoints (`/healthz`,
+/// `/readyz`), and platform-plane-only handlers that authenticate via
+/// `X-ToolKit-Internal-Token` instead. When present, [`security_context_middleware`] lets
+/// a request without an `Authorization` header pass through instead of
+/// returning `401`.
+///
+/// **Not the same as `OperationSpec.is_public`.** `is_public` controls whether
+/// a route is registered in the gateway for external access; this marker
+/// controls whether [`security_context_middleware`] requires a JWT. The two are
+/// independent: most gateway-exposed routes DO carry a JWT and do NOT need this
+/// marker; most probe routes are NOT gateway-exposed but DO need it.
+#[derive(Clone, Copy, Debug)]
+pub struct PublicRoute;
+
+/// Tenant-plane `SecurityContext` middleware.
+///
+/// Behaviour:
+/// - A bearer token, if present, is **always** re-validated via the injected
+///   [`BearerAuthenticator`]; on success the [`SecurityContext`](toolkit_security::SecurityContext) is inserted
+///   into request extensions.
+/// - A protected route (no [`PublicRoute`] marker) with a missing or invalid
+///   `Authorization` header is rejected with `401`.
+/// - A public / system-only route (carrying the [`PublicRoute`] marker) with no
+///   `Authorization` header passes through.
+/// - A rejected token is `401`; an unreachable backend is `503`; any other
+///   unexpected authentication failure is `500`.
+///
+/// Rejections are rendered as canonical RFC 9457 `application/problem+json`
+/// responses (via [`CanonicalError`]) so they match the platform-wide error
+/// contract; `instance` / `trace_id` enrichment is left to the outer canonical
+/// error middleware installed at the gear/bootstrap layer.
+///
+/// The handler is generic over `A`; the concrete authenticator is supplied via
+/// Axum state as `Arc<A>` at the gear/bootstrap layer.
+///
+/// TODO: Rework to align with the gateway's `authn_middleware`
+/// (`gears/system/api-gateway/src/middleware/auth.rs`), the more mature
+/// implementation: route-policy-driven auth requirements (vs. the binary
+/// `PublicRoute` marker), CORS-preflight handling, anonymous-`SecurityContext`
+/// insertion for public routes, and RFC 6750 `WWW-Authenticate` Bearer
+/// challenges. Part of consolidating all gateway middlewares into this crate.
+pub async fn security_context_middleware<A>(
+    State(authenticator): State<Arc<A>>,
+    mut request: Request,
+    next: Next,
+) -> Response
+where
+    A: BearerAuthenticator + 'static,
+{
+    let is_public = request.extensions().get::<PublicRoute>().is_some();
+
+    match extract_bearer_http(request.headers()) {
+        Ok(token) => match authenticator.authenticate(token.expose_secret()).await {
+            Ok(secctx) => {
+                request.extensions_mut().insert(secctx);
+                next.run(request).await
+            }
+            Err(err) => authn_error_to_response(&err),
+        },
+        // No credential presented: allow through only for public/system-only
+        // routes; protected routes require a user context.
+        Err(SecurityContextHttpError::MissingAuthHeader) if is_public => next.run(request).await,
+        Err(SecurityContextHttpError::MissingAuthHeader) => unauthenticated("MISSING_BEARER"),
+        Err(SecurityContextHttpError::InvalidAuthHeader | SecurityContextHttpError::EmptyToken) => {
+            unauthenticated("INVALID_BEARER")
+        }
+    }
+}
+
+/// Map a neutral [`AuthNError`] to a canonical `problem+json` response.
+///
+/// The token and any provider-specific detail are never surfaced on the wire.
+fn authn_error_to_response(err: &AuthNError) -> Response {
+    match err {
+        // A reachable backend that rejected the token: the caller's credential
+        // is bad (401).
+        AuthNError::InvalidToken => {
+            tracing::warn!("bearer token rejected: invalid or expired");
+            unauthenticated("AUTHN_FAILED")
+        }
+        // The backend could not be reached: surface 503 with a retry hint so
+        // callers can distinguish "try later" from "your token is bad".
+        AuthNError::Unavailable => {
+            tracing::warn!("bearer token validation: authentication backend unavailable");
+            CanonicalError::service_unavailable()
+                .with_retry_after_seconds(AUTH_RETRY_AFTER_SECONDS)
+                .create()
+                .into_response()
+        }
+        // `Other` (and, defensively, any future neutral variant) is an
+        // unexpected authentication-infrastructure failure, not a bad
+        // credential — surface 500 rather than blaming the caller. The
+        // diagnostic detail is redacted on the wire by `CanonicalError`.
+        // `AuthNError` is `#[non_exhaustive]`, so the wildcard is required.
+        _ => {
+            tracing::error!("bearer token validation: unexpected infrastructure failure");
+            CanonicalError::internal(AUTH_INFRA_FAILURE_DETAIL)
+                .create()
+                .into_response()
+        }
+    }
+}
+
+/// Build a canonical `Unauthenticated` (`401`) `problem+json` response with the
+/// given machine-readable reason.
+fn unauthenticated(reason: &str) -> Response {
+    CanonicalError::unauthenticated()
+        .with_reason(reason)
+        .create()
+        .into_response()
+}
+
+/// Platform-plane internal-auth middleware.
+///
+/// Behaviour:
+/// - When an `X-ToolKit-Internal-Token` header is present, it is validated via
+///   the injected [`InternalAuthenticator`]; on success [`PeerAuthenticated`]
+///   and a [`PlatformSecurityContext`] are inserted into request extensions.
+/// - When the header is **absent**, the request passes through unchanged
+///   (permissive): user-only endpoints do not require a system credential, and
+///   the tenant plane is enforced independently by [`security_context_middleware`].
+/// - When the header is present but **invalid/empty**, or validation fails, the
+///   request is **rejected** — so an invalid SA token is turned away before
+///   [`security_context_middleware`] (and any handler) runs.
+///
+/// This sets workload-policy state only; it **never** skips or substitutes for
+/// tenant-plane JWT validation. Install this layer so it runs **before**
+/// [`security_context_middleware`] (DESIGN § 3.2).
+///
+/// Rejections are rendered as canonical RFC 9457 `application/problem+json`:
+/// an invalid credential is `401`, an unreachable validation backend is `503`,
+/// and any other unexpected failure is `500`.
+///
+/// The handler is generic over `A`; the concrete validator (K8s `TokenReview`
+/// in the first phase) is supplied via Axum state as `Arc<A>` at the
+/// gear/bootstrap layer.
+pub async fn internal_auth_middleware<A>(
+    State(authenticator): State<Arc<A>>,
+    mut request: Request,
+    next: Next,
+) -> Response
+where
+    A: InternalAuthenticator + 'static,
+{
+    match extract_internal_token_http(request.headers()) {
+        Ok(token) => match authenticator.authenticate(token.expose_secret()).await {
+            Ok(identity) => {
+                request.extensions_mut().insert(PeerAuthenticated {
+                    name: identity.peer_name().to_owned(),
+                });
+                request
+                    .extensions_mut()
+                    .insert(PlatformSecurityContext::new(identity));
+                next.run(request).await
+            }
+            Err(err) => internal_authn_error_to_response(&err),
+        },
+        // No system credential presented: permissive — user-only endpoints do
+        // not require one, and the tenant plane is enforced separately.
+        Err(InternalTokenHttpError::MissingHeader) => next.run(request).await,
+        // A credential was presented but is malformed: reject before the
+        // tenant plane runs.
+        Err(InternalTokenHttpError::InvalidHeader | InternalTokenHttpError::EmptyToken) => {
+            unauthenticated("INVALID_INTERNAL_TOKEN")
+        }
+    }
+}
+
+/// Map a neutral [`InternalAuthNError`] to a canonical `problem+json` response.
+///
+/// The token and any provider-specific detail are never surfaced on the wire.
+fn internal_authn_error_to_response(err: &InternalAuthNError) -> Response {
+    match err {
+        // A reachable backend that rejected the credential: it is bad (401).
+        InternalAuthNError::InvalidToken => {
+            tracing::warn!("internal token rejected: invalid or expired credential");
+            unauthenticated("INTERNAL_AUTH_FAILED")
+        }
+        // The validation backend (e.g. K8s TokenReview) was unreachable: 503.
+        InternalAuthNError::Unavailable => {
+            tracing::warn!("internal token validation: authentication backend unavailable");
+            CanonicalError::service_unavailable()
+                .with_retry_after_seconds(AUTH_RETRY_AFTER_SECONDS)
+                .create()
+                .into_response()
+        }
+        // `Other` (and, defensively, any future neutral variant) is an
+        // unexpected infrastructure failure — surface 500. `InternalAuthNError`
+        // is `#[non_exhaustive]`, so the wildcard is required.
+        _ => {
+            tracing::error!("internal token validation: unexpected infrastructure failure");
+            CanonicalError::internal(AUTH_INFRA_FAILURE_DETAIL)
+                .create()
+                .into_response()
+        }
+    }
+}
+
+#[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
+#[path = "auth_tests.rs"]
+mod tests;
--- a/libs/toolkit-http-middleware/src/lib.rs
+++ b/libs/toolkit-http-middleware/src/lib.rs
@@ -1,0 +1,31 @@
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![warn(warnings)]
+
+//! Server-side HTTP middleware for `ToolKit`.
+//!
+//! This crate provides the inbound request-handling layers a gear's HTTP server
+//! installs at its edge. It does **not** run an HTTP server itself — the server
+//! is owned by the `OoP` bootstrap (`toolkit::bootstrap::oop`) and by the
+//! `api-gateway` gear's `rest_host`, which install these layers onto their
+//! router.
+//!
+//! - [`auth`] — Axum middleware for two-plane authentication. It turns
+//!   inbound credentials into a [`toolkit_security::SecurityContext`] (tenant
+//!   plane) or a [`toolkit_security::PlatformSecurityContext`] (platform plane),
+//!   rejecting invalid credentials with canonical RFC 9457 `problem+json`
+//!   responses.
+//! - [`security`] — the supporting extractors that pull the tenant-plane bearer
+//!   token and the platform-plane internal token out of inbound request headers.
+//!
+//! Keeping these out of `toolkit-http` (the outbound HTTP client) keeps the
+//! client crate free of `axum` and the canonical-error stack, and gives every
+//! gear a single place to depend on for the server-side auth planes.
+
+pub mod auth;
+pub mod security;
+
+pub use auth::{PublicRoute, internal_auth_middleware, security_context_middleware};
+pub use security::{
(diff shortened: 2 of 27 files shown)


libs/toolkit-security/Cargo.toml
```diff
@@ -27,3 +27,4 @@ thiserror = { workspace = true }
 
 [dev-dependencies]
 serde_json = { workspace = true }
+trybuild = { workspace = true }
```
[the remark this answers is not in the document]
mattgarmon: Not a Cargo manifest field. cf-studio-path = ".cf-studio" is an AI-agent config hint in AGENTS.md pointing to the Constructor Studio workspace directory - no Cargo.toml in the repo carries it, and the project has never treated it as a per-crate directive.

libs/toolkit-http-middleware/src/middleware.rs
```diff
@@ -0,0 +1,589 @@
...
+///   unexpected authentication failure is `500`.
+///
+/// Rejections are rendered as canonical RFC 9457 `application/problem+json`
+/// responses (via [`CanonicalError`]) so they match the platform-wide error
+/// contract; `instance` / `trace_id` enrichment is left to the outer canonical
+/// error middleware installed at the gear/bootstrap layer.
+///
+/// The handler is generic over `A`; the concrete authenticator is supplied via
+/// Axum state as `Arc<A>` at the gear/bootstrap layer.
+pub async fn secctx_middleware<A>(
```
aviator5: We already have a bunch of middlewares in `gears/system/api-gateway/src/middleware` including `auth.rs`. Perhaps it worth to extend API Gateway's middlewares instead adding a new middleware for platform-level authentication there.

Another thing that I would like to discuss is extensibility. In my view, we need to be vendor agnostic for platform-level authentication as well. Seems like we may consider an option to add a new `authentication_platform()` method to AuthNResolver and call it for platform-level authN.

@mattgarmon, @MikeFalcon77 what do you think?

mattgarmon: Hi @aviator5 

On the first point, just wanted to clarify that these middlewares are intended to be used by every oop gear. It makes more sense to have them in a shared library like `toolkit-http`. I do agree there is some duplication that can be cleaned up (bearer extraction, error mapping, etc.). `api-gateway` could use some functionality from this middleware for it.

For the extensibility, you are right that a `authenticate_platform() -> PlatformSecurityContext` method could keep the planes separated. Being vendor agnostic is already also supported by [InternalAuthenticator](https://github.com/constructorfabric/gears-rust/pull/4121/changes#diff-9e1cbb6843021d879b9c4e62507a5f816bc7941067ef878f7615af29b39c23b3R194-R216), just in a different way than `AuthNResolver`. I am still thinking about whether to fold into `AuthNResolver`, or keep separate.

@MikeFalcon77 Any thoughts?

MikeFalcon77: Thanks @aviator5, both points were settled when we accepted the two-plane ADRs:

1. Location. Per-hop re-validation means every OoP gear runs this, not just the edge (ADR-0008, NFR per-hop-revalidation). api-gateway is one gear, and ADR-0007 wants it kept minimal - so the middleware belongs in shared toolkit-http, generic over BearerAuthenticator. The real point is the duplication you spotted: worth a follow-up to have api-gateway reuse the shared helpers + expose a BearerAuthenticator over its AuthNResolverClient. Not in this PR. @mattgarmon please create an issue to re-use helpers in apigw.

2. Platform authN. Let's keep InternalAuthenticator separate. Folding authenticate_platform() into AuthNResolver is Option E in ADR-0006/0008 — rejected as default: separation of mechanisms is the core decision, and the trust roots differ (JWKS/OIDC vs k8s SA token/mTLS). It's already vendor-agnostic via the trait + #[non_exhaustive] PlatformIdentity. The "one backend" case is the opt-in internal_auth.mode = "jwt" adapter (ADR-0006), not a new method on the tenant trait.

Neither blocks this PR - it's the library layer. We ned to file the two follow-ups only.

aviator5: @mattgarmon, @MikeFalcon77 thank you for the explanation.

1. Yes, I believe we need to have a separate crate for all shared HTTP middlewares and all gears (including `system/api-gateway`) should be able to use them. I agree that it may be `tookit-http`, but we will need to think about code organization there - currently `toolkit-http` contains only client-related code (tower layers).

2. I also think that separation is a right option (we discussed this @mattgarmon previously), but it could be implemented in different ways and as I can see `https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md` ADR  doesn't cover them. Vendor can bring its own `AuthNResolverPluginClient` implementation via our plugin system. How vendor-agnostic platform-level authN plane will be customized? Via plugin system as well?

libs/toolkit-security/src/authenticator.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Any other authentication failure. The message must not contain the
+    /// token or other sensitive material.
+    #[error("authentication failed: {0}")]
+    Other(String),
+}
+
+/// Re-validates a raw bearer token and reconstructs a [`SecurityContext`].
+///
+/// Implementations perform a full validation on every call — there is no
+/// trusted-peer fast path (zero-trust; see ADR-0008). The transport layer
```
aviator5: Suggest specifying a full path to ADR document or (better) `cpt-*` identifiers.

libs/toolkit-http-middleware/src/middleware.rs
striped-zebra-dev: One more on location: I don't think `toolkit-http` is the right home.
Its README literally says *"HTTP client library for ToolKit"* with a "What it does NOT do" list - it's the outbound client.
This PR adds inbound Axum server middleware and pulls `axum` into it.

The fact that we gate it ("so the bare HTTP client does not pull in axum") kind of proves it doesn't belong here.

libs/toolkit-security/src/authenticator.rs
striped-zebra-dev: If we use doc links, we should use the full relative path.

libs/toolkit-security/src/internal_auth.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+impl PlatformIdentity {
+    /// The caller's gear name, distilled for workload-policy decisions.
+    ///
+    /// For a [`PlatformIdentity::ServiceAccount`] this is the `ServiceAccount`
+    /// name; for [`PlatformIdentity::Spiffe`] it is the gear component of the
+    /// SPIFFE ID.
+    #[must_use]
+    pub fn peer_gear(&self) -> &str {
```
striped-zebra-dev: I'd avoid using `_gear` in API for practical reasons.

libs/toolkit-http-middleware/src/middleware.rs
```diff
@@ -0,0 +1,589 @@
...
+        },
+        // No system credential presented: permissive — user-only endpoints do
+        // not require one, and the tenant plane is enforced separately.
+        Err(InternalTokenHttpError::MissingHeader) => next.run(request).await,
+        // A credential was presented but is malformed: reject before the
+        // tenant plane runs.
+        Err(InternalTokenHttpError::InvalidHeader | InternalTokenHttpError::EmptyToken) => {
+            unauthenticated("INVALID_INTERNAL_TOKEN")
+        }
+    }
```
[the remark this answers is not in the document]
mattgarmon: The platform plane must never bypass tenant
JWT validation: accepting a valid internal token to skip tenant auth would let a
forged user JWT + valid SA token through (see [peer_authenticated_does_not_skip_jwt_validation](cci:1://file:///home/matt/code/rust/gears-rust/libs/toolkit-http-middleware/src/middleware.rs:531:4-544:5)).
The two contexts coexisting isn't an escalation — [PlatformSecurityContext](cci:2://file:///home/matt/code/rust/gears-rust/libs/toolkit-security/src/internal_auth.rs:134:0-136:1) is a
distinct type the tenant `PolicyEnforcer` never sees. "One plane per request" is
caller convention; each plane is validated independently. Leaving as-is.

libs/toolkit-http-middleware/src/security.rs
```diff
@@ -0,0 +1,247 @@
...
+/// trimmed, and an empty token is rejected.
+///
+/// # Errors
+///
+/// Returns [`SecCtxHttpError`] when the header is absent, not a valid `Bearer`
+/// credential, or carries an empty token.
+pub fn extract_bearer_http(headers: &HeaderMap) -> Result<SecretString, SecCtxHttpError> {
+    let header = headers
+        .get(AUTHORIZATION)
+        .ok_or(SecCtxHttpError::MissingAuthHeader)?;
```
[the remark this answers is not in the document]
mattgarmon: Good catch — fixed. Both extractors now reject duplicate `Authorization` /
`X-ToolKit-Internal-Token` headers via a shared single-value helper, returning
the existing invalid-header (401) path. Added regression tests for both.

libs/toolkit-http/src/security.rs
```diff
@@ -1,8 +1,127 @@
...
+    // Expose the secret only here, at the transport boundary, and never log it.
+    let Ok(mut value) = HeaderValue::from_str(&format!("Bearer {}", token.expose_secret())) else {
+        tracing::warn!(
+            "bearer token contains invalid HTTP header characters; outgoing request sent without Authorization header"
+        );
+        return;
+    };
+    value.set_sensitive(true);
+    request.headers_mut().insert(AUTHORIZATION, value);
+}
```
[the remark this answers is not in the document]
mattgarmon: Skipping — same rationale as the server-side "both planes present" thread.
`insert` already replaces same-header duplicates, so the only way both headers
appear is a caller invoking both attachers. Coexistence isn't a security issue:
the receiving middleware validates each plane independently and [PlatformSecurityContext](cci:2://file:///home/matt/code/rust/gears-rust/libs/toolkit-security/src/internal_auth.rs:134:0-136:1)
is a distinct type the tenant PolicyEnforcer never evaluates. Having a single-purpose
`attach_*` helper silently delete an unrelated auth header is a surprising,
order-dependent side effect that would mask caller bugs. "One plane per request"
is caller convention; if we want to hard-enforce it, that belongs at the call
site choosing the credential, not in these helpers.

libs/toolkit-http-middleware/src/middleware.rs
mattgarmon: @aviator5 Thanks for the feedback, I've updated the PR.

1. Agreed, I implemented as a new crate, `cf-gears-toolkit-http-middleware`, not inside `toolkit-http` (which stays the client / tower layers). It's the shared home for middleware now.
 
2. The platform plane is vendor-extensible too, but **bootstrap/config-selected and linked locally, not GTS-discovered**. Platform auth secures the Directory/GTS discovery itself, so discovering it over the network would be a bootstrapping cycle. Covered in a new Extensibility section in ADR-0008.

striped-zebra-dev (approved):

libs/toolkit-http-middleware/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+//! Keeping these out of `toolkit-http` (the outbound HTTP client) keeps the
+//! client crate free of `axum` and the canonical-error stack, and gives every
+//! gear a single place to depend on for the server-side auth planes.
+
+pub mod auth;
+pub mod security;
+
+pub use auth::{PublicRoute, internal_auth_middleware, secctx_middleware};
+pub use security::{
+    InternalTokenHttpError, SecCtxHttpError, extract_bearer_http, extract_internal_token_http,
```
aviator5: Suggest being consistent in public API and avoid using short forms like `SecCtx` - we already have `SecurityContext` in toolkit-security.

libs/toolkit-http-middleware/src/auth.rs
```diff
@@ -0,0 +1,254 @@
...
+///   unexpected authentication failure is `500`.
+///
+/// Rejections are rendered as canonical RFC 9457 `application/problem+json`
+/// responses (via [`CanonicalError`]) so they match the platform-wide error
+/// contract; `instance` / `trace_id` enrichment is left to the outer canonical
+/// error middleware installed at the gear/bootstrap layer.
+///
+/// The handler is generic over `A`; the concrete authenticator is supplied via
+/// Axum state as `Arc<A>` at the gear/bootstrap layer.
+pub async fn secctx_middleware<A>(
```
aviator5: @mattgarmon, as we discussed in https://github.com/constructorfabric/gears-rust/pull/4121#discussion_r3482060888, suggest creating a github issue for moving all middlewares from `gears/system/api-gateway/src/middleware` to `libs/toolkit-http-middleware`. Additionally, let's add a `TODO` comment somewhere here to rework this `secctx_middleware` - it should be aligned with the already existing one from `gears/system/api-gateway/src/middleware/auth.rs`.

libs/toolkit-security/src/authenticator.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::context::SecurityContext;
+
+/// Neutral authentication error returned by a [`BearerAuthenticator`].
+///
+/// Intentionally coarse-grained and transport-agnostic: it never carries the
+/// token or any provider-specific detail so it is safe to surface at a trust
+/// boundary. Concrete adapters map their own error types into these variants.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum AuthnError {
```
aviator5: Suggest renaming: `AuthnError` -> `AuthNError`. Reason: consistency with already existing `AuthNResolver`

libs/toolkit-security/src/internal_auth.rs
```diff
@@ -0,0 +1,263 @@
...
+/// handlers consume a [`PlatformSecurityContext`] without branching on it. New
+/// authentication methods add a variant — hence `#[non_exhaustive]` — without
+/// changing [`PlatformSecurityContext`]'s shape.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[non_exhaustive]
+#[serde(tag = "type")]
+pub enum PlatformIdentity {
+    /// First phase: a validated K8s `ServiceAccount` (from a projected SA token
+    /// verified via the `TokenReview` API).
+    ServiceAccount {
```
aviator5: `ServiceAccount` -> `KubernetesServiceAccount`?
We already use k8s-specific here like `pod` field.

libs/toolkit-security/src/internal_auth.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/// Neutral platform-plane authentication error.
+///
+/// Intentionally coarse-grained and transport-agnostic: it never carries the
+/// token or provider-specific detail, so it is safe to surface at a trust
+/// boundary. Concrete validators map their own errors into these variants.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum InternalAuthError {
```
aviator5: `InternalAuthError` -> `InternalAuthNError`?
Reasons:
a) consistency (`AuthNError`)
b) more explicit - we are doing authentication here

libs/toolkit-security/src/internal_auth.rs
```diff
@@ -0,0 +1,263 @@
...
+/// middleware on a multi-threaded runtime.
+pub trait InternalAuthenticator: Send + Sync {
+    /// Validate the raw `X-ToolKit-Internal-Token` value and resolve the
+    /// caller's [`PlatformIdentity`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`InternalAuthError`] if the credential is invalid, the backend
+    /// is unavailable, or validation otherwise fails.
+    fn validate(
```
aviator5: Sorry for probably nitpicking, but why validate() and not `authenticate()` or `authenticate_internal()`?

aviator5 (approved):

libs/toolkit-http-middleware/src/middleware.rs
mattgarmon: > One more on location: I don't think `toolkit-http` is the right home. Its README literally says _"HTTP client library for ToolKit"_ with a "What it does NOT do" list - it's the outbound client. This PR adds inbound Axum server middleware and pulls `axum` into it.
> 
> The fact that we gate it ("so the bare HTTP client does not pull in axum") kind of proves it doesn't belong here.

Moved to `toolkit-http-middleware`

libs/toolkit-security/src/internal_auth.rs
mattgarmon: Agreed, updated to `peer_name/name`

libs/toolkit-http-middleware/src/lib.rs
mattgarmon: Agreed, updated all public API surfaces to use `SecurityContext`/`security_context`

libs/toolkit-http-middleware/src/auth.rs
mattgarmon: Created issue: https://github.com/constructorfabric/gears-rust/issues/4169

Also added TODO that is referenced from the issue

libs/toolkit-security/src/internal_auth.rs
mattgarmon: Agreed, renamed

libs/toolkit-security/src/internal_auth.rs
mattgarmon: Good point, renamed to `authenticate` for symmetry

PR #4123, opened 2026-06-18
feat(cluster): cluster docs and cluster-sdk crate

asmith987:
Introduce the cluster module documentation (PRD, DESIGN, DECOMPOSITION, OBSERVABILITY, ADRs, per-feature specs, traceability audit) and the `cf-gears-cluster-sdk` crate: unified error model, typed profiles and scoping, cache/leader/lock/discovery primitives with facade+backend traits, SDK default in-memory backends, watch auto-restart, registration and OpenTelemetry observability, plus unit and integration tests.

Wires the crate into the workspace and adds the cf-studio "Nested SDK Tests" artifact autodetect rule. Examples ship with the wiring PR; the standalone backend plugin and the cluster wiring crate follow in their own PRs.

## Summary by CodeRabbit

* **New Features**
  * Added the cluster coordination SDK with a cache (CAS + watches), leader election, distributed locking, and service discovery primitives.
  * Introduced capability-validated typed resolvers with scoped namespaces and per-primitive watch auto-restart on retryable disconnects.
  * Added polling-based prefix-watch support when native prefix watching isn’t available.
  * Included cache/leader/lock/service discovery default backends, plus registration helpers.
* **Bug Fixes**
  * Improved consistent watch termination/shutdown behavior with explicit terminal shutdown signaling.
* **Documentation**
  * Added detailed design/feature specifications and ADR updates for cache prefix enumeration, scoping, and observability.
* **Chores**
  * Updated workspace wiring, traceability ignore rules, autodetection, and pinned SDK tooling version.


gears/system/cluster/cluster-sdk/Cargo.toml
```diff
@@ -0,0 +1,72 @@
...
+readme = "README.md"
+keywords = ["constructorfabric", "cf-gears"]
+categories = ["development-tools"]
+metadata.docs.rs.all-features = true
+
+[lib]
+name = "cluster_sdk"
+
+[lints]
+workspace = true
```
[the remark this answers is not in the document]
asmith987: Not applying this one. No sibling SDK crate carries `cf-studio-path` in `[package]` — checked `account-management-sdk`, `resource-group-sdk`, and `usage-collector-sdk`, and `grep -r cf-studio-path gears/system/*/*/Cargo.toml` is empty. The `cf-studio-path = ".cf-studio"` in the repo-root `CLAUDE.md` is cf-studio tooling config, not a required Cargo manifest key; adding it here would diverge from every other crate in the workspace.

gears/system/cluster/cluster-sdk/src/cache/backend.rs
```diff
@@ -0,0 +1,167 @@
...
+    async fn compare_and_delete(
+        &self,
+        key: &str,
+        expected_value: &[u8],
+    ) -> Result<bool, ClusterError> {
+        match self.get(key).await? {
+            Some(entry) if entry.value.as_slice() == expected_value => self.delete(key).await,
+            _ => Ok(false),
+        }
+    }
```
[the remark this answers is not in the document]
asmith987: Intentional and documented. The default's own rustdoc states it is a best-effort, non-atomic `get`→`delete` that "narrows but does not close the read-to-delete window" and that a backend with an atomic store **should override** it (the linearizable default does exactly that). Failing closed in the default would break best-effort backends that legitimately rely on it. We did tighten the *dependent guarantee's* wording — see the `004-distributed-lock.md` release-guarantee qualifier in 14d7ef4b.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    /// Returns a sub-namespaced view of this cache: every key (and the prefix of
+    /// `watch_prefix`/`scan_prefix`) is auto-prefixed with `prefix + "/"` on the
+    /// write path and stripped on the read path (DESIGN §3.8). Scoping composes —
+    /// `cache.scoped("a")?.scoped("b")?` makes the backend observe `"a/b/<key>"`.
+    ///
+    /// # Errors
+    /// Returns [`ClusterError::InvalidName`] if `prefix` violates the scope-prefix
+    /// rule (`[a-zA-Z0-9_/-]+`).
+    pub fn scoped(&self, prefix: &str) -> Result<Self, ClusterError> {
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The doc now states the real rule — slash-separated `[a-zA-Z0-9_-]` segments with no leading, trailing, or doubled-slash (empty) segments, max 255 chars — matching `validated_prefix` / `SCOPE_PREFIX_RULE`.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// recommendation to prefer a native-prefix-watch backend at scale. Dropping
+    /// the returned [`CacheWatch`] stops the polling task. Pair with
+    /// [`watch_prefix`](Self::watch_prefix) (native) when the backend supports it.
+    ///
+    /// A zero `interval` does not panic: the returned watch yields a single
+    /// terminal [`CacheWatchEvent`](crate::cache::CacheWatchEvent)`::Closed`
+    /// carrying [`ClusterError::InvalidConfig`] (non-retryable) — see
+    /// [`PollingPrefixWatch::spawn`]. Disappeared keys are reported as
+    /// [`CacheEvent`](crate::cache::CacheEvent)`::Deleted`, never `Expired`.
+    #[must_use]
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The links now use full variant paths — `[CacheWatchEvent::Closed]` and `[CacheEvent::Deleted]` — and resolve cleanly (verified with `cargo doc`; no broken-intra-doc-link warning).

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
```diff
@@ -0,0 +1,200 @@
...
+///
+/// # Lifecycle
+///
+/// Dropping the returned [`CacheWatch`] stops the polling task on its next tick
+/// (it checks `is_closed()` each tick, so even a quiescent keyspace with nothing
+/// to emit still terminates). A backend error is surfaced as
+/// [`CacheWatchEvent::Closed`] and ends the task — the polyfill does **not**
+/// retry internally, so wrap it with the watch auto-restart combinator
+/// (`CacheWatch::auto_restart`, DECOMPOSITION §2.8) to recover transparently from
+/// a retryable [`ProviderErrorKind`](crate::error::ProviderErrorKind).
```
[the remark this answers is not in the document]
asmith987: Not a defect. `PollingPrefixWatch::spawn` is a building block that returns a bare watch by design; the consumer-facing path `CacheFacade::watch_prefix_polling` installs the resubscribe seam via `install_polling_watch_seam` **before** the watch is handed out, so `auto_restart` reconnects as advertised. The seam re-runs the polling `watch` on the bound backend on each retryable close. No code change.

gears/system/cluster/cluster-sdk/src/cache/scoped.rs
```diff
@@ -0,0 +1,185 @@
...
+                    other => other,
+                };
+                if tx.send(forwarded).await.is_err() {
+                    // The consumer dropped the watch: stop forwarding.
+                    return;
+                }
+            }
+        });
+        watch
+    }
```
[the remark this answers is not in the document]
asmith987: Two sub-issues here:

**Idle-stream task/subscription leak — fixed in 14d7ef4b.** The forwarding task now uses `tokio::select!` racing `tx.closed()` against `inner.recv()`, so it tears down promptly on consumer drop even when the stream is quiescent (added a `pub(crate) CacheWatchSender::closed`). Both futures are cancel-safe.

**Lost resubscribe seam / reset-observability — left as-is (by design).** At the backend layer these are intentionally absent: the resubscribe seam and observability context are installed by the *facade* layer, and on reconnect the facade seam re-invokes the (scoped) backend's `watch`/`watch_prefix`, which re-creates the forwarding wrapper — so scoping under the facade keeps restartability and reset metrics. Re-stamping them at the backend-forwarding layer would duplicate facade responsibility.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,845 @@
...
+            }
+        };
+        let task = StoreMaintainer {
+            cache: Arc::clone(&self.cache),
+            shared: Arc::downgrade(&self.shared),
+            name: name.to_owned(),
+            prefix,
+        };
+        tokio::spawn(task.run(watch));
+        Ok(())
```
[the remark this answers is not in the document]
asmith987: `ensure_maintainer` awaits `self.cache.watch_prefix(&prefix).await?` and only returns `Ok(())` **after** the subscription is established — the `tokio::spawn` at the end merely drives event consumption, it isn't what the subscription depends on. A concurrent `discover` during warmup sees an empty-but-converging view, which is within the discovery consistency contract (ADR-008: discovery state is *intent*, eventually consistent — not health). Leaving as-is.

gears/system/cluster/cluster/src/defaults/leader.rs
```diff
@@ -0,0 +1,661 @@
...
+                shutdown: self.shutdown.clone(),
+                provider: self.provider,
+                metrics: Arc::clone(&self.metrics),
+            };
+            self.track(tokio::spawn(task.run(
+                initial,
+                Some(cache_watch),
+                resign_rx,
+            )));
+            Ok(watch)
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. `join()` now returns `Err(ClusterError::Shutdown)` when `self.shutdown.is_cancelled()`, so no election task is enrolled after `revoke()` starts draining the tracked handles. This reuses the existing shutdown token and mirrors the guard already in `lock()`.

gears/system/cluster/cluster/src/defaults/leader.rs
```diff
@@ -0,0 +1,661 @@
...
+    /// (`cpt-cf-clst-fr-shutdown-ttl-cleanup`).
+    async fn revoke(&self) {
+        self.shutdown.cancel();
+        let handles = {
+            let mut tasks = self.tasks.lock().unwrap_or_else(PoisonError::into_inner);
+            std::mem::take(&mut *tasks)
+        };
+        for handle in handles {
+            let _joined = handle.await;
+        }
```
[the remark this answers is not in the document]
asmith987: `revoke()` awaiting the tasks **is** the documented revocation guarantee — "a current leader has observed loss before this returns". The tasks are cancellation-aware (the wait loops `select!` on `shutdown.cancelled()`), so they unwind promptly once the token is cancelled. Adding a timeout to the join would weaken that guarantee; per ADR-002 the individual backend round-trips are expected prompt / bounded by the caller. Left as-is by design.

gears/system/cluster/cluster-sdk/src/defaults/lock.rs
```diff
@@ -0,0 +1,509 @@
...
+        let op_started = std::time::Instant::now();
+        let out = async {
+            let key = Self::lock_key(name);
+            let holder = identity::fresh_id();
+            let started = tokio::time::Instant::now();
+            // Subscribe before the first attempt so a release between a failed claim
+            // and the wait cannot be missed.
+            // @cpt-begin:cpt-cf-clst-algo-sdk-default-backends-cas-lock:p1:inst-cl-watch
+            let mut watch = self.cache.watch(&key).await?;
+            // @cpt-end:cpt-cf-clst-algo-sdk-default-backends-cas-lock:p1:inst-cl-watch
```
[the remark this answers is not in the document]
asmith987: The `timeout` bounds `watch.recv()` — the part that blocks unboundedly waiting for a release. The discrete backend round-trips (initial `watch`, `put_if_absent`, re-subscribe) aren't separately wrapped; per ADR-002 those are expected prompt and bounded by the caller's outer timeout, and the loop already fails out via the elapsed-budget check. Bounding every round-trip would change the acquisition loop's error semantics, so I'm flagging rather than silently rewriting — happy to wrap them in the remaining budget if `timeout` is meant as a hard ceiling on *total* acquisition.

gears/system/cluster/cluster-sdk/src/discovery/watch.rs
```diff
@@ -0,0 +1,231 @@
...
+    ///
+    /// # Errors
+    /// Returns the unsent event back if the consumer has dropped the watch.
+    pub async fn send(&self, event: ServiceWatchEvent) -> Result<(), ServiceWatchEvent> {
+        self.tx.send(event).await.map_err(|err| err.0)
+    }
+
+    /// Tries to deliver `event` without ever blocking the caller, returning the
+    /// unsent event if the buffer is full or the consumer dropped the watch.
+    ///
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The doc now states `try_send` returns a `ServiceWatchTrySendError` (`Full` / `Closed`), not the unsent event.

gears/system/cluster/cluster-sdk/src/leader/watch.rs
```diff
@@ -0,0 +1,520 @@
...
+    Status(LeaderStatus),
+    /// The watcher fell behind and `dropped` events were lost; wait for the
+    /// next `Status` event before resuming leader-only work.
+    Lagged {
+        /// The number of events dropped.
+        dropped: u64,
+    },
+    /// The subscription was re-established (reconnect or compaction); wait for
+    /// the next `Status` event before resuming leader-only work.
+    Reset,
```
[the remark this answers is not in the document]
asmith987: `run_while_leader` is explicitly advisory (documented: "not mutual exclusion" — two participants can transiently run `work`; workloads needing exclusion gate writes on `try_lock` / `compare_and_swap`). A transient `Lagged`/`Reset` isn't a leadership change, so work is intentionally left running and reconciled on the next `Status`. The `LeaderWatchEvent` re-read guidance is about treating watched state as stale, which the snapshot reconciliation already honors. No change.

gears/system/cluster/cluster-sdk/src/leader/watch.rs
```diff
@@ -0,0 +1,520 @@
...
+    ///
+    /// # Errors
+    /// Returns the unsent event back if the consumer has dropped the watch.
+    pub async fn send(&self, event: LeaderWatchEvent) -> Result<(), LeaderWatchEvent> {
+        debug_assert!(
+            !matches!(event, LeaderWatchEvent::Status(_)),
+            "LeaderWatchSender::send must not carry Status events; use send_status so the \
+             snapshot stays coherent with the event stream",
+        );
+        self.events.send(event).await.map_err(|err| err.0)
```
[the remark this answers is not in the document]
asmith987: This is a documented, intentional *debug-only* invariant guarding SDK-internal misuse: the only callers of `send` are the SDK's own default backends (covered by tests), and the doc states it is "asserted in debug builds". We kept `debug_assert!` to avoid a release-build panic in a library for a purely-internal invariant. That said it's a one-line change to promote to `assert!` if you'd prefer release enforcement — let us know and we'll switch it.

gears/system/cluster/cluster-sdk/src/observability.rs
```diff
@@ -0,0 +1,434 @@
...
+    fields::label::KIND,
+    fields::label::PRIMITIVE,
+];
+// @cpt-end:cpt-cf-clst-algo-registration-observability-emit-names:p1:inst-en-labels
+
+/// Bounded, enum-like `result` label values (see [`fields::label::RESULT`]).
+///
+/// These are the only values that may be attached as the `result` metric label;
+/// they are derived from a [`ClusterError`] by [`result_label`], never from a
+/// free-form string. Keeping them bounded preserves the cardinality rule.
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The doc now references the actual functions `label` / `from_error`; the stale `result_label` reference is gone.

gears/system/cluster/cluster-sdk/src/restart.rs
```diff
@@ -0,0 +1,444 @@
...
+            tokio::time::sleep(delay).await;
+            // @cpt-begin:cpt-cf-clst-algo-watch-auto-restart-backoff:p1:inst-bo-reset
+            if let Ok(fresh) = self.inner.resubscribe().await {
+                self.inner = fresh;
+                self.attempts = 0;
+                self.emit_reset();
+                return true;
+            }
+            // @cpt-end:cpt-cf-clst-algo-watch-auto-restart-backoff:p1:inst-bo-reset
+        }
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. `reconnect()` now returns `false` on a non-retryable resubscribe error (`Err(err) if !err.is_retryable()`), so permanent failures (shutdown/auth/unsupported) no longer spin forever when `max_retries` is `None`. Retryable errors still fall through and back off toward the cap.

gears/system/cluster/docs/features/004-distributed-lock.md
```diff
@@ -0,0 +1,180 @@
...
+**Input**: A release request carrying the holder's identity
+
+**Output**: Release of the entry only if the requester still holds it
+
+**Steps**:
+1. [x] - `p1` - Compare the requester's holder identity against the current lock entry - `inst-rh-compare`
+2. [x] - `p1` - **IF** the requester is not the current holder - `inst-rh-foreign`
+   1. [x] - `p1` - **RETURN** without releasing another holder's lock - `inst-rh-skip`
+3. [x] - `p1` - Release the entry conditionally so a foreign holder cannot release it - `inst-rh-release`
+
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. Both the `inst-rh-release` step and the matching DoD now qualify the guarantee with the unexpired-TTL window and note the get-then-delete cannot fully close the post-expiry gap, so extension before TTL expiry is required for safety.

gears/system/cluster/docs/features/005-service-discovery.md
```diff
@@ -0,0 +1,250 @@
...
+
+**Error Scenarios**:
+- Heartbeat stops — the instance disappears from discovery within its TTL window.
+
+**Steps**:
+1. [x] - `p1` - Consumer submits a registration with a service name, address, and metadata - `inst-rg-submit`
+2. [x] - `p1` - **IF** no instance id is provided - `inst-rg-noid`
+   1. [x] - `p1` - SDK assigns an instance id - `inst-rg-assign`
+3. [x] - `p1` - SDK registers the instance as enabled with a TTL-bounded heartbeat - `inst-rg-register`
+4. [x] - `p1` - **RETURN** a registration handle - `inst-rg-return`
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The `inst-rg-assign` step now reads "The backend assigns an instance id", matching the contract in `discovery/backend.rs`, `discovery/types.rs`, and ADR-008.

gears/system/cluster/docs/TRACEABILITY-AUDIT.md
```diff
@@ -0,0 +1,143 @@
...
+## 6. Conclusion
+
+- Every requirement maps to a realizing DESIGN section or ADR and to a feature
+  (§2). Principles and constraints likewise (§3).
+- Code traceability markers are wired for every in-scope feature (§4).
+- Both open questions are resolved/recorded (§5).
+- **No traceability gaps** for this change. The five `follow-up`-scoped
+  requirements are intentionally deferred (PRD §4.1) to the wiring crate and
+  parent host gear; their realizing ADR (ADR-006) and DESIGN sections exist, so
+  the follow-up changes build against a frozen, fully-traced contract.
```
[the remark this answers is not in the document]
asmith987: Fixed in 14d7ef4b. The conclusion now reads "the one `follow-up`-scoped requirement", consistent with line 82 and the requirements table (exactly one `follow-up` row: `cpt-cf-clst-fr-routing-per-primitive`).

asmith987:
Addressed @MikeFalcon77's architecture-conformance review from #4098 ([review thread](https://github.com/constructorfabric/gears-rust/pull/4098#pullrequestreview-4534135054)). All findings that land in the `cluster-sdk` crate are fixed in this PR (commit bb00d4b). Each row links back to the original inline comment.

| Finding | Original comment | Fix applied |
|---|---|---|
| **H1** — SD key collision (ADR-001) | [#discussion_r3443573268](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573268) | Restored the `svc/` namespace via shared per-primitive key-prefix constants (`SVC_/ELECTION_/LOCK_KEY_PREFIX`); SD keys are now `svc/{name}/{instance_id}`, so a service named `election` can no longer collide with a leader/lock claim on a shared cache. |
| **M1** — watch `None` not reconnected (ADR-003) | [#discussion_r3443573273](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573273) | `RestartingWatch::recv` now treats an underlying `None` as a retryable close when the watch can resubscribe — it reconnects with backoff and synthesizes `Reset`; seam-less watches still terminate. |
| **M2** — terminal events droppable (ADR-003) | [#discussion_r3443573275](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573275) | `LeaderWatch::channel` reserves two `OwnedPermit`s of terminal-event headroom, so `revoke_for_shutdown` always delivers the distinct `Status(Lost)` → `Closed(Shutdown)` two-step even against a full event buffer. |
| **M3** — `extend` resets the lease (ADR-002) | [#discussion_r3443573279](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573279) | Renamed `LockGuard::extend` → `renew` (param `additional_ttl` → `new_ttl`) across the consumer API, `LockRequest`/`LockCommand` variants, the span (`cluster.lock.renew`) and metric label, matching the reset-to-TTL semantics. Computing against the remaining TTL isn't possible (the cache has no remaining-TTL read), so the rename is the fix. |
| **L1** — `profile_scope` public (ADR-007) | [#discussion_r3443573281](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573281) | `profile_scope` is now `pub(crate)` and removed from the crate re-export; the typed `ClusterProfile` marker is the sole consumer-facing profile path. Name validators stay public. |
| **L2** — hardcoded heartbeat cadence | [#discussion_r3443573283](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573283) | `DEFAULT_RENEWAL` is now derived from `DEFAULT_TTL` (`ttl / 3`) with a cross-reference to the leader-election renewal formula so the two can't drift. |
| **L3** — `GuardTask` detached spawn (ADR-006) | [#discussion_r3443573286](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573286) | Documented that `GuardTask`'s lifetime is intentionally bound to the consumer-held guard, not to `revoke()` — cancelling it on shutdown would yank a lease out from under an in-flight critical section; the held lease lapses via TTL instead. |
| **L4** — `StoreMaintainer` untracked (ADR-006) | [#discussion_r3443573291](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573291) | `StoreMaintainer` now carries the shutdown token, `select!`s on it in its run loop, and is spawned via `track(...)`, so `revoke()` tears it down deterministically instead of leaving it to lapse via `Weak` upgrade failure. |

Each fix ships with a targeted regression test (e.g. the `svc/` non-collision, the `None`-reconnect both branches, the full-buffer terminal two-step, the renamed `renew_*` tests, and the token-driven maintainer teardown). Full suite + clippy `--all-targets` + `fmt --check` are green.

**Not in this PR:** the **H2** Drop-guard finding ([#discussion_r3443573270](https://github.com/constructorfabric/gears-rust/pull/4098#discussion_r3443573270)) is in the `cluster` gear wiring (`cluster/src/wiring.rs`), which lives in the gear PR (#4098), not this SDK crate — the `ClusterHandle` Drop guard and its confirmation tests are already in place there.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
```diff
@@ -0,0 +1,870 @@
...
+            shared: Arc::downgrade(&self.shared),
+            name: name.to_owned(),
+            prefix,
+            shutdown: self.shutdown.clone(),
+        };
+        // Tracked like the watch-translator / heartbeat tasks so `revoke` tears it
+        // down deterministically rather than leaving it to lapse via TTL.
+        self.track(tokio::spawn(task.run(watch)));
+        Ok(())
+    }
```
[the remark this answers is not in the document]
asmith987: Confirmed and fixed — this is a real race in degraded mode (cache without native prefix-watch), reproducible with two concurrent `register("name")` calls:

1. Caller A marks `name` maintained, then awaits `watch_prefix` (which will fail).
2. Caller B observes A's optimistic mark, returns `Ok` immediately, and pre-inserts its instance.
3. A's `watch_prefix` fails, rolls back the mark, returns `Err`.

End state: no maintainer running, but B's instance is in the view with nothing to reap it — violating the degraded-mode "best-effort empty" invariant.

Rather than moving the insert to *after* `watch_prefix` succeeds (which reintroduces a double-spawn race — two healthy-path callers both pass the `contains` check and each spawn a maintainer), the slot is now tri-state:

- `maintained` is `HashMap<String, MaintainerState>` (`Pending` → `Active`).
- `ensure_maintainer` returns `MaintainerStatus::{Live, Pending}`; a name is only `Active` *after* `watch_prefix` succeeds.
- A concurrent caller observing `Pending` gets `Pending` back and **skips** its pre-insert. Its registration reconverges via the live maintainer's initial `scan_prefix` reconcile.

This keeps the optimistic insert for de-dup (no double-spawn) while ensuring a pre-insert only happens behind a confirmed-live maintainer (no leak).

Added `concurrent_degraded_registers_do_not_leak_an_unreapable_local_instance`, made deterministic via a test-cache variant whose `watch_prefix` yields once before returning `Unsupported`. Verified it fails against the old behavior and passes with the fix.

gears/system/cluster/docs/DESIGN.md
```diff
@@ -658,7 +714,7 @@ The SDK ships three default backend implementations built on `Arc<dyn ClusterCac
...
 - `CasBasedDistributedLockBackend` — `put_if_absent(lock_key, holder_id, ttl)` for `try_lock`, `watch(lock_key)` to notify blocked waiters on release, background TTL reaper. Release via delete-if-still-holder using CAS (a foreign holder cannot release another's lock). No fencing tokens (the no-remote-in-critical-section rule eliminates the stale-writer scenario). `features()` returns `LockFeatures { linearizable: cache.consistency() == Linearizable }`.
-- `CacheBasedServiceDiscoveryBackend` — `put(svc/{name}/{instance_id}, metadata, ttl)` for registration, `watch_prefix(svc/{name}/)` for topology change events, background TTL renewal for heartbeat. Metadata filtering is client-side; `features()` returns `ServiceDiscoveryFeatures { metadata_pushdown: false }`.
+- `CacheBasedServiceDiscoveryBackend` — `put({name}/{instance_id}, metadata, ttl)` for registration, `watch_prefix({name}/)` for topology change events, background TTL renewal for heartbeat. Metadata filtering is client-side; `features()` returns `ServiceDiscoveryFeatures { metadata_pushdown: false }`.
```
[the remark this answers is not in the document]
asmith987: Fixed in `b75a6dc7`. Restored the `svc/` namespace in the documented key path — registration now reads `put(svc/{name}/{instance_id}, ...)` and the subscription `watch_prefix(svc/{name}/)`, with a note that the `svc/` prefix keeps service-discovery keys collision-free against the `election/` and `lock/` prefixes. This matches the implementation: `SVC_KEY_PREFIX = "svc/"` (`defaults/mod.rs:92`) and the key builders at `defaults/discovery.rs:175,179`.

gears/system/cluster/docs/DECOMPOSITION.md
```diff
@@ -0,0 +1,550 @@
...
+- **Component `cpt-cf-clst-component-plugins`** — the in-process **standalone** plugin is implemented in this change; the remaining per-backend plugin crates (Postgres, K8s, Redis, NATS, etcd) are still separate follow-up changes.
+- **FRs realized by the wiring / operator config / lifecycle owner**: `cpt-cf-clst-fr-routing-omit-default` (wiring auto-wrap), `cpt-cf-clst-fr-lifecycle-owner` (gear crate owns `ClusterHandle`), `cpt-cf-clst-fr-shutdown-revoke` (leader revocation on `stop`) and `cpt-cf-clst-fr-shutdown-ttl-cleanup` (no remote cleanup on `stop`) are **implemented in this change**. `cpt-cf-clst-fr-routing-per-primitive` remains a follow-up: binding a native non-cache backend per primitive is rejected at config time until those providers ship.
+- **Sequences `cpt-cf-clst-seq-lifecycle-startup` and `cpt-cf-clst-seq-shutdown`** — now realized by the gear crate's `ClusterHandle` start/stop (leader revocation included); `cpt-cf-clst-seq-per-primitive-resolution` is realized by the SDK resolvers.
```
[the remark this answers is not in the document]
asmith987: Fixed in `b75a6dc7`. The scope note now states this change delivers the SDK contract only (consistent with the in-scope list), and the `cpt-cf-clst-component-wiring` / `cpt-cf-clst-component-plugins` bullets plus the related FR/sequence lines describe `ClusterWiring::builder(...).build_and_start() -> ClusterHandle`, `ClusterHandle`, and the standalone plugin as pending follow-up work rather than delivered scope. Confirmed against the code: the repo contains only the `cluster-sdk` crate — no `cf-gears-cluster` gear crate, `ClusterWiring`/`ClusterHandle`, `RunnableCapability`, or standalone plugin exist outside the docs. The matching DESIGN.md §3.7 amendment (the duplicate of this comment) was reworded the same way.

gears/system/cluster/cluster/src/defaults/test_cache.rs
```diff
@@ -0,0 +1,536 @@
...
+        let mut dead = Vec::new();
+        for (id, sender) in targets {
+            if sender
+                .send(CacheWatchEvent::Event(event.clone()))
+                .await
+                .is_err()
+            {
+                dead.push(id);
+            }
+        }
```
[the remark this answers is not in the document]
asmith987: Not applying this one. `test_cache.rs` is a test double, not a production backend, and the blocking `send().await` fan-out is deliberate: `WATCH_CAPACITY = 256` is documented at `test_cache.rs:42-44` as "Generous so a renewal/heartbeat storm in a test never drops events as `Lagged`" — the fixture intentionally guarantees deterministic, lossless delivery and does not model `Lagged`/lagged-receiver pruning. The non-blocking `try_send` + `Lagged` contract the comment describes is correctly implemented where it matters, in the production discovery backend (`defaults/discovery.rs`). Switching the fixture to `try_send` would contradict its stated purpose and make tests non-deterministic.

gears/system/cluster/docs/DESIGN.md
```diff
@@ -331,6 +383,8 @@ Each plugin (Postgres, K8s, Redis, NATS, etcd, standalone) exposes a builder/han
...
 
+> **Backend-trait-only — `compare_and_delete`.** `ClusterCacheBackend` additionally declares `async fn compare_and_delete(&self, key: &str, expected_value: &[u8]) -> Result<bool, ClusterError>`: an atomic value-guarded delete that removes `key` only if its current value equals `expected_value`. A value mismatch or an absent key returns `Ok(false)`, never an error. It is deliberately **not** exposed on `ClusterCacheV1` — the public CAS contract is version-based (`compare_and_swap`), while this is the value/owner-token-guarded counterpart used internally by SDK-default coordination backends (e.g. the leader elector's guarded release, which must survive a key's version resetting to 1 on delete+recreate, where a version guard would alias a successor's fresh claim). The trait's default impl is a best-effort, non-atomic `get`-then-`delete`; backends with an atomic store override it for a genuine compare-and-delete.
+
```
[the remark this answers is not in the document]
asmith987: Not applying — already covered by the existing contract text. The `compare_and_delete` doc on `ClusterCacheBackend` already states the default impl is "a best-effort, non-atomic `get`-then-`delete`" and that "backends with an atomic store override it for a genuine compare-and-delete." It is also already documented as trait-only (deliberately not on `ClusterCacheV1`) and used internally by the SDK-default coordination backends for owner-token-guarded release. So the distinction the comment asks for — that only atomic-overriding backends should rely on it for leader-elector/owner-token release — is present. No code change needed.

MikeFalcon77 (approved):
I approve lt's await Oleg's review

gears/system/cluster/cluster-sdk/src/cache/backend.rs
```diff
@@ -0,0 +1,167 @@
...
+    async fn contains(&self, key: &str) -> Result<bool, ClusterError>;
+
+    /// Atomically creates `key` only if absent: `Some(entry)` when created,
+    /// `None` when the key already existed.
+    ///
+    /// # Errors
+    /// Returns [`ClusterError`] if the backend operation fails.
+    async fn put_if_absent(
+        &self,
+        key: &str,
```
striped-zebra-dev: I see a pattern key+value+ttl, why don't we use:
```
#[derive(Debug, Clone)]
pub struct PutRequest<'a> {
    pub key: &'a str,
    pub value: &'a [u8],
    pub ttl: CacheTtl,
}

```

gears/system/cluster/cluster-sdk/src/cache/backend.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// # Errors
+    /// Returns [`ClusterError::CasConflict`] (carrying the current entry when
+    /// cheaply obtainable) when `expected_version` no longer matches, or
+    /// another [`ClusterError`] if the backend operation fails.
+    async fn compare_and_swap(
+        &self,
+        key: &str,
+        expected_version: u64,
+        new_value: &[u8],
+        ttl: Option<Duration>,
```
striped-zebra-dev: Even if that is ok to use one Option in param, still feels that that good.
I'd think of an enum such as:

```rust
enum TTL {
  Of(Duration)
  Indefinite
}
```

asmith987: I agree. Better to be more explicit. I added this across the API

gears/system/cluster/cluster-sdk/src/cache/backend.rs
asmith987: I added this suggestion.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+/// # Capability
+///
+/// Intended for backends declaring `features().prefix_watch == false`; the caller
+/// is responsible for choosing the polyfill only when native `watch_prefix` is
+/// unavailable (this constructor does not inspect `features()`).
+pub struct PollingPrefixWatch;
+
+impl PollingPrefixWatch {
```
striped-zebra-dev, line 67: Note for myslef:
Check with the design, we supposed to support notify-based watcher.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
```diff
@@ -0,0 +1,183 @@
...
+/// (it checks `is_closed()` each tick, so even a quiescent keyspace with nothing
+/// to emit still terminates). A backend error is surfaced as
+/// [`CacheWatchEvent::Closed`] and ends the task — the polyfill does **not**
+/// retry internally, so wrap it with the watch auto-restart combinator
+/// (`CacheWatch::auto_restart`, DECOMPOSITION §2.8) to recover transparently from
+/// a retryable [`ProviderErrorKind`](crate::error::ProviderErrorKind).
+///
+/// # Capability
+///
+/// Intended for backends declaring `features().prefix_watch == false`; the caller
```
striped-zebra-dev, line 62: Ok, now I see that it is just not wired yet, the `prefix_watch` and the watcher.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Watches an exact key.
+    ///
+    /// The returned watch carries a resubscribe seam, so
+    /// [`CacheWatch::auto_restart`] can transparently re-`watch` this key on a
+    /// retryable terminal close.
+    ///
+    /// # Errors
+    /// Propagates any [`ClusterError`] from the backend.
+    pub async fn watch(&self, key: &str) -> Result<CacheWatch, ClusterError> {
+        let mut watch = self.inner.watch(key).await?;
```
striped-zebra-dev, line 144: The question is: what is our strategy on explicit/implicit poll watch use?

gears/system/cluster/cluster-sdk/src/discovery/backend.rs
```diff
@@ -0,0 +1,124 @@
...
+/// state AND every metadata predicate (`cpt-cf-clst-algo-service-discovery-filter`).
+/// The order of the returned `Vec` is **unspecified** and may differ across
+/// backends and across calls; consumers needing deterministic selection sort
+/// client-side, typically by `instance_id`. A backend declaring
+/// `features().metadata_pushdown == false` MAY return a superset on metadata and
+/// rely on the consumer applying [`DiscoveryFilter::matches`](crate::discovery::DiscoveryFilter::matches)
+/// client-side. `watch` yields an **unfiltered** topology stream; consumers
+/// filter each change client-side.
+#[async_trait]
+pub trait ServiceDiscoveryBackend: Send + Sync {
```
striped-zebra-dev, line 72: There is a plat to have a component called FlightControl in OOP initiative, we should be aligned on the responsibilities and dependencies.

gears/system/cluster/cluster/src/defaults/discovery.rs
```diff
@@ -0,0 +1,897 @@
...
+/// [`ClusterError::Unsupported`] from `watch_prefix`, so
+/// [`watch`](ServiceDiscoveryBackend::watch) surfaces that error and
+/// [`discover`](ServiceDiscoveryBackend::discover) is **best-effort**: with no
+/// maintainer to reap entries on TTL expiry, [`register`](ServiceDiscoveryBackend::register)
+/// deliberately skips its local pre-insert (an unreapable pre-insert would read
+/// as live forever), so degraded-mode `discover` returns an empty set rather
+/// than a stale view. This is a known limitation pending the
+/// `PollingPrefixWatch` polyfill (DECOMPOSITION §2.7), which is out of scope
+/// here.
+pub struct CacheBasedServiceDiscoveryBackend {
```
striped-zebra-dev, line 106: I don't think that any *Backend shall be available to users via -SDK.
In my view: SDK - only client facing API, traits, facades. 

In this particular example if one service takes the `CacheBasedServiceDiscroveryBackend` and uses it to register or resolve and the other service uses some other XXXBackend, they will never find eachother.

As I see it:
`cluster/src` - is where all the backends live. Operator sets up the right backends for the installation.
`cluster-sdk/src` - is where all the clients live. Services just use the facades to communicate to backends. Services never use backends directly.

Let's discuss.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
```diff
@@ -0,0 +1,250 @@
...
+    #[must_use]
+    pub fn features(&self) -> CacheFeatures {
+        self.inner.features()
+    }
+
+    /// Returns the versioned entry for `key`, or `None` if absent.
+    ///
+    /// # Errors
+    /// Propagates any [`ClusterError`] from the backend.
+    pub async fn get(&self, key: &str) -> Result<Option<CacheEntry>, ClusterError> {
```
asmith987, line 71: **HIGH**

Cache-key-accepting methods (`get`, `put`, `delete`, `contains`, `put_if_absent`, `compare_and_swap`, `watch`) pass caller-supplied `key: &str` to the backend without any character-set or length validation.

Unbounded or malformed keys reach the backend store directly, bypassing the name-rule contract used elsewhere in the SDK.

Add a `validate_cache_key` helper enforcing `CLUSTER_NAME_RULE` (or a documented cache-key rule) and call it at the top of each key-accepting public method before delegating to the backend.

gears/system/cluster/cluster-sdk/src/lock/facade.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub fn features(&self) -> LockFeatures {
+        self.inner.features()
+    }
+
+    /// Non-blocking acquisition of `name` with the given `ttl`.
+    ///
+    /// # Errors
+    /// - [`ClusterError::LockContended`] if the lock is already held.
+    /// - Propagates any other [`ClusterError`] from the backend.
+    pub async fn try_lock(&self, name: &str, ttl: Duration) -> Result<LockGuard, ClusterError> {
```
asmith987, line 78: **HIGH**

`try_lock` and `lock` accept a caller-supplied `name: &str` and forward it to the backend without validating it against `CLUSTER_NAME_RULE`.

Arbitrary strings (empty, oversized, or path-traversal-style) reach the backend store with no rejection.

Call `validate_cluster_name(name)?` at the top of both methods before delegating to `self.inner`, mirroring the pattern already used in `scoped` for prefix validation.

gears/system/cluster/cluster-sdk/src/leader/facade.rs
```diff
@@ -0,0 +1,142 @@
...
+        self.inner.features()
+    }
+
+    /// Joins the named election with the default [`ElectionConfig`], returning a
+    /// [`LeaderWatch`]. The claim is renewed automatically and the watch
+    /// auto-reenrolls on a transient leadership loss.
+    ///
+    /// # Errors
+    /// Propagates any [`ClusterError`] from the backend.
+    pub async fn elect(&self, name: &str) -> Result<LeaderWatch, ClusterError> {
```
asmith987, line 92: **HIGH**

`elect` and `elect_with_config` forward the caller-supplied `name: &str` to the backend without validating against `CLUSTER_NAME_RULE`.

Malformed or adversarial election names reach the backend store unchecked.

Call `validate_cluster_name(name)?` at the start of both methods, consistent with how `scoped` validates the prefix.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
```diff
@@ -0,0 +1,183 @@
...
+            let prefix = prefix.to_owned();
+            tokio::spawn(poll_loop(cache, prefix, interval, sender));
+        }
+        watch
+    }
+}
+
+/// The background diff loop. Exits when the consumer drops the watch (a `send`
+/// fails) or a backend call errors (after emitting a terminal `Closed`).
+async fn poll_loop(
```
asmith987, line 107: **MEDIUM**

The `poll_loop` background task accepts no `CancellationToken` and relies solely on the consumer dropping the `CacheWatch` for termination.

The task cannot be stopped by system-wide graceful shutdown before the consumer drops its handle.

Add a `CancellationToken` parameter to `poll_loop` and `PollingPrefixWatch::spawn`, and add a `token.cancelled()` branch to the loop `select!` so the task exits promptly on cancellation.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+        previous = current;
+    }
+}
+
+/// Lists the keys under `prefix` and reads each key's current version, realizing
+/// the documented `N + 1` round-trip cost. A key that vanishes between the scan
+/// and its `get` is simply omitted from the snapshot.
+async fn snapshot(
```
asmith987, line 173: **MEDIUM**

`snapshot` issues one sequential `get` per key returned by `scan_prefix`, resulting in N sequential async round-trips per poll tick.

The polling interval stalls proportionally to keyspace size; a large namespace makes the polyfill unusably slow.

Issue the per-key `get` calls concurrently using `futures::future::join_all` or `tokio::task::JoinSet`, bounding concurrency if the keyspace may be large.

gears/system/cluster/cluster-sdk/src/cache/scoped.rs
```diff
@@ -0,0 +1,196 @@
...
+        Self { inner, prefix }
+    }
+
+    /// Wraps a backend-facing [`CacheWatch`] in a read-path forwarding task that
+    /// strips `prefix` from every event key before handing it to the consumer.
+    /// The task ends — dropping its sender — when the inner watch ends or the
+    /// consumer drops the returned watch.
+    fn strip_watch(prefix: String, mut inner: CacheWatch) -> CacheWatch {
+        let (tx, watch) = CacheWatch::channel(FORWARD_BUFFER);
+        tokio::spawn(async move {
```
asmith987, line 48: **MEDIUM**

The `strip_watch` forwarding task has no `CancellationToken`; it relies on `tx.closed()` and inner stream end for termination.

The task cannot participate in system-level graceful shutdown signaling and will run until the consumer drops its handle.

Pass a `CancellationToken` into `strip_watch` and add a `token.cancelled()` arm to the `select!` block so the task exits on shutdown even if the consumer has not yet dropped its `CacheWatch`.

gears/system/cluster/cluster-sdk/src/observability.rs
```diff
@@ -0,0 +1,426 @@
...
+use crate::error::{ClusterError, ProviderErrorKind};
+
+/// The high-cardinality resource identifier a provider error pertains to. It is
+/// emitted as a **log field** on `cluster.provider.error` (never a metric label,
+/// per the cardinality rule); the variant selects the contracted field name from
+/// OBSERVABILITY.md §6 (`key` / `lock` / `election` / `name`). `tracing` field
+/// names must be static, so the field name is chosen by matching the variant
+/// rather than passed dynamically.
+#[derive(Clone, Copy)]
+pub(crate) enum ResourceId<'a> {
```
asmith987: **MEDIUM**

`ResourceId` is `pub` but is an internal wiring type used only by `emit_provider_error`; general SDK consumers have no use for it.

Exposing it commits this type to the stable public API surface unnecessarily.

Add `#[doc(hidden)]` to `ResourceId` and `emit_provider_error` to signal they are not part of the stable consumer surface, preventing accidental dependency from downstream crates.

gears/system/cluster/cluster-sdk/src/cache/resolver.rs
```diff
@@ -0,0 +1,111 @@
...
+    ///   the profile scope.
+    /// - [`ClusterError::CapabilityNotMet`] if a declared capability is
+    ///   unsupported by the bound backend.
+    pub fn resolve(self) -> Result<ClusterCacheV1, ClusterError> {
+        let profile = self.profile_name.ok_or(ClusterError::ProfileNotSpecified)?;
+        let scope = profile_scope(profile)?;
+        let inner: Arc<dyn ClusterCacheBackend> = self
+            .hub
+            .get_scoped(&scope)
+            .map_err(|_| ClusterError::ProfileNotBound { profile })?;
```
asmith987: **MEDIUM**

`map_err(|_| ...)` discards the error returned by `hub.get_scoped()`, losing all diagnostic detail about why the backend lookup failed.

Operators cannot distinguish a missing profile registration from a type mismatch or hub configuration bug without re-running with debug instrumentation.

Capture the error and embed it in `ClusterError::ProfileNotBound` or log it before mapping, so the original cause is preserved in diagnostics.

gears/system/cluster/cluster-sdk/src/discovery/resolver.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///   registered for the profile scope.
+    /// - [`ClusterError::CapabilityNotMet`] if a declared capability is
+    ///   unsupported by the bound backend.
+    pub fn resolve(self) -> Result<ServiceDiscoveryV1, ClusterError> {
+        let profile = self.profile_name.ok_or(ClusterError::ProfileNotSpecified)?;
+        let scope = profile_scope(profile)?;
+        let inner: Arc<dyn ServiceDiscoveryBackend> = self
+            .hub
+            .get_scoped(&scope)
+            .map_err(|_| ClusterError::ProfileNotBound { profile })?;
```
asmith987: **MEDIUM**

`map_err(|_| ...)` discards the error returned by `hub.get_scoped()`, losing all diagnostic detail about why the backend lookup failed.

Operators cannot distinguish a missing profile registration from a type mismatch or hub configuration bug without re-running with debug instrumentation.

Capture the error and embed it in `ClusterError::ProfileNotBound` or log it before mapping, so the original cause is preserved in diagnostics.

gears/system/cluster/cluster-sdk/src/leader/resolver.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///   registered for the profile scope.
+    /// - [`ClusterError::CapabilityNotMet`] if a declared capability is
+    ///   unsupported by the bound backend.
+    pub fn resolve(self) -> Result<LeaderElectionV1, ClusterError> {
+        let profile = self.profile_name.ok_or(ClusterError::ProfileNotSpecified)?;
+        let scope = profile_scope(profile)?;
+        let inner: Arc<dyn LeaderElectionBackend> = self
+            .hub
+            .get_scoped(&scope)
+            .map_err(|_| ClusterError::ProfileNotBound { profile })?;
```
asmith987: **MEDIUM**

`map_err(|_| ...)` discards the error returned by `hub.get_scoped()`, losing all diagnostic detail about why the backend lookup failed.

Operators cannot distinguish a missing profile registration from a type mismatch or hub configuration bug without re-running with debug instrumentation.

Capture the error and embed it in `ClusterError::ProfileNotBound` or log it before mapping, so the original cause is preserved in diagnostics.

gears/system/cluster/cluster-sdk/src/lock/resolver.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///   registered for the profile scope.
+    /// - [`ClusterError::CapabilityNotMet`] if a declared capability is
+    ///   unsupported by the bound backend.
+    pub fn resolve(self) -> Result<DistributedLockV1, ClusterError> {
+        let profile = self.profile_name.ok_or(ClusterError::ProfileNotSpecified)?;
+        let scope = profile_scope(profile)?;
+        let inner: Arc<dyn DistributedLockBackend> = self
+            .hub
+            .get_scoped(&scope)
+            .map_err(|_| ClusterError::ProfileNotBound { profile })?;
```
asmith987: **MEDIUM**

`map_err(|_| ...)` discards the error returned by `hub.get_scoped()`, losing all diagnostic detail about why the backend lookup failed.

Operators cannot distinguish a missing profile registration from a type mismatch or hub configuration bug without re-running with debug instrumentation.

Capture the error and embed it in `ClusterError::ProfileNotBound` or log it before mapping, so the original cause is preserved in diagnostics.

gears/system/cluster/cluster-sdk/src/observability_tests.rs
```diff
@@ -0,0 +1,114 @@
...
+    fields::attr::KEY,
+    fields::attr::NAME,
+    fields::attr::LOCK,
+    fields::attr::ELECTION,
+    fields::attr::INSTANCE_ID,
+    fields::attr::PROFILE,
+];
+
+#[test]
+fn span_names_follow_dotted_lowercase_convention() {
```
asmith987: **LOW**

`span_names_follow_dotted_lowercase_convention` asserts string-formatting properties of compile-time string constants; the assertions are tautologically true given the current literal values.

The test adds no protection against behavioral regressions and will only fail if a developer edits a constant to violate the convention — at which point the constant itself is the right fix site.

Enforce naming conventions via a linting rule or a `const`-assertion macro at the definition site rather than a runtime test loop over static slices.

gears/system/cluster/cluster-sdk/src/observability_tests.rs
```diff
@@ -0,0 +1,114 @@
...
+        assert_eq!(
+            *name,
+            name.to_lowercase(),
+            "span `{name}` must be lowercase"
+        );
+    }
+}
+
+#[test]
+fn metric_names_follow_underscored_lowercase_convention() {
```
asmith987: **LOW**

`metric_names_follow_underscored_lowercase_convention` asserts formatting properties of compile-time string constants, adding no protection against behavioral regressions.

Same issue as the span-name test above — the assertion is a tautology on the current literal values.

Enforce naming conventions via a linting rule or a `const`-assertion macro at the definition site.

gears/system/cluster/cluster-sdk/src/restart_tests.rs
```diff
@@ -0,0 +1,578 @@
...
+    ));
+    assert_eq!(
+        backend.calls.load(Ordering::SeqCst),
+        3,
+        "initial + two reconnects"
+    );
+}
+
+#[test]
+fn default_policy_matches_design() {
```
asmith987, line 186: **LOW**

`default_policy_matches_design` constructs `RetryPolicy::default()` and asserts each field equals the design-time constant, verifying only that the struct initializer stores the prescribed values.

No retry logic is exercised; the test would pass even if `backoff_for` were completely wrong.

Replace or supplement with a behavioral assertion: pass the default policy through `backoff_for` and verify the resulting schedule falls within the expected bounds.

gears/system/cluster/cluster-sdk/src/observability/instrumented_cache_tests.rs
```diff
@@ -0,0 +1,159 @@
...
+        metrics.cache_ops.lock().unwrap().as_slice(),
+        &[("delete".to_owned(), "error".to_owned())]
+    );
+    assert_eq!(
+        metrics.provider_errors.lock().unwrap().as_slice(),
+        &["other"]
+    );
+}
+
+#[tokio::test]
```
asmith987: **LOW**

`delegates_inner_consistency_and_features` confirms only that `InstrumentedCache` forwards `consistency()` and `features()` calls, exercising no instrumentation behavior.

The test duplicates what the existing operation tests already imply and adds no independent coverage.

Remove this test or merge its assertion into one of the existing operation tests where the instrumented layer's metric recording is already being verified.

gears/system/cluster/cluster-sdk/src/lock/guard_tests.rs
```diff
@@ -0,0 +1,152 @@
...
+            ..
+        })
+    ));
+    assert!(backend.await.is_ok());
+}
+
+#[tokio::test]
+async fn dropping_guard_performs_no_io_and_does_not_release() {
+    let (mut rx, guard) = LockGuard::channel("rate-limit".to_owned(), 1);
+    assert_eq!(guard.name(), "rate-limit");
```
asmith987: **LOW**

`assert_eq!(guard.name(), "rate-limit")` confirms only that the constructor stores the name field, not that the guard behaves correctly.

The meaningful assertion in this test — that dropping the guard sends no request to the backend receiver — is on the next line; the name assertion is noise.

Remove the `assert_eq!(guard.name(), ...)` line and keep only the behavioral assertion.

gears/system/cluster/cluster-sdk/src/discovery/backend_tests.rs
```diff
@@ -0,0 +1,91 @@
...
+        )
+        .await
+    else {
+        panic!("discover must succeed");
+    };
+    assert!(none.is_empty());
+}
+
+#[test]
+fn provider_name_reports_concrete_backend() {
```
asmith987, line 86: **LOW**

`provider_name_reports_concrete_backend` tests that Rust's `std::any::type_name` embeds the struct name in its output — a language-semantics property, not application logic.

The test verifies nothing about the SDK contract and would break only if the Rust compiler changed how it formats type names.

Remove this test or replace it with an assertion on the error message emitted by `validate_*_capabilities` when a capability is unmet, which exercises the actual consumer-visible contract.

gears/system/cluster/cluster-sdk/Cargo.toml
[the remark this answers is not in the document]
asmith987: Verified in current code (33f20e4c). No `cf-studio-path` key appears in any sibling SDK crate. Not applying stands.

gears/system/cluster/cluster-sdk/src/cache/backend.rs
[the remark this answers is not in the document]
asmith987: Confirmed. The default impl's rustdoc states it is a best-effort, non-atomic `get`→`delete` and that backends with an atomic store should override it. DESIGN §3.3 captures the accepted tradeoff. No change needed.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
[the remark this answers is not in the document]
asmith987: Fix confirmed in current code (33f20e4c). The `scoped()` rustdoc now states the full rule: slash-separated `[a-zA-Z0-9_-]` segments, no leading/trailing/doubled-slash segments, max 255 chars — matching `validated_prefix` / `SCOPE_PREFIX_RULE`.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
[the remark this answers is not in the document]
asmith987: Fix confirmed in current code (33f20e4c). Doc links now use full variant paths (`[CacheWatchEvent::Closed](crate::cache::CacheWatchEvent::Closed)`, `[CacheEvent::Deleted](crate::cache::CacheEvent::Deleted)`) and resolve cleanly.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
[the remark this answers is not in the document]
asmith987: Confirmed not a defect. `watch_prefix_polling` installs the resubscribe seam via `install_polling_watch_seam` before handing the watch out, so `auto_restart` reconnects correctly. No code change.

gears/system/cluster/cluster-sdk/src/cache/scoped.rs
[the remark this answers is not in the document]
asmith987: Idle-stream fix confirmed in current code (33f20e4c): the forwarding task `tokio::select!`s on `tx.closed()` and `inner.recv()`, tearing down promptly on consumer drop. The resubscribe seam / reset-observability is intentionally at the facade layer — confirmed by design.

gears/system/cluster/cluster-sdk/src/defaults/discovery.rs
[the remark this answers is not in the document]
asmith987: Confirmed left as-is. `ensure_maintainer` awaits `watch_prefix` before returning, so the subscription is established before `Ok(())`. A concurrent `discover` during warmup sees an empty-but-converging view, within the ADR-008 eventual-consistency contract for discovery state.

gears/system/cluster/cluster/src/defaults/leader.rs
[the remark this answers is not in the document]
asmith987: Fix confirmed in current code (33f20e4c). `join()` checks `self.shutdown.is_cancelled()` and returns `Err(ClusterError::Shutdown)` before enrolling a new election task, mirroring the guard already present in `lock()`.

gears/system/cluster/cluster/src/defaults/leader.rs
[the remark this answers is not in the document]
asmith987: Confirmed left as-is by design. `revoke()` awaiting the tasks IS the revocation guarantee — "a current leader has observed loss before this returns". The tasks `select!` on the shutdown token so they unwind promptly. Adding a timeout would weaken the guarantee per ADR-002.

gears/system/cluster/cluster-sdk/src/defaults/lock.rs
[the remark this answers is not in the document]
asmith987: Fixed. `timeout` is a hard ceiling per DESIGN §3.3 ("blocking up to `timeout`"), so all backend round-trips in `lock()` are now wrapped in the remaining budget:

- Initial `cache.watch(&key)` — wrapped before entering the loop
- `put_if_absent` on each iteration — wrapped with `remaining` before the `watch.recv()` select
- Both re-subscribe `cache.watch(&key)` calls (legitimate rotation and busy-spin arms) — wrapped with `resubscribe_remaining`

All 15 existing lock tests pass. The `watch.recv()` path was already bounded; this closes the residual window on the three round-trip sites.

gears/system/cluster/cluster-sdk/src/discovery/watch.rs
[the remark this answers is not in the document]
asmith987: Fix confirmed in current code (33f20e4c). The `try_send` rustdoc now states it returns `ServiceWatchTrySendError` (`Full` / `Closed`), not the unsent event. The `send()` sibling correctly retains its "returns the unsent event back" doc since it returns `Result<(), ServiceWatchEvent>`.

gears/system/cluster/cluster-sdk/src/leader/watch.rs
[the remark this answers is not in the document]
asmith987: Confirmed no change. `run_while_leader` is explicitly advisory (documented). A transient `Lagged`/`Reset` is not a leadership change, so work is intentionally left running and reconciled on the next `Status` event. Workloads needing exclusion gate writes on `try_lock`/`compare_and_swap` per DESIGN §3.3 Pattern C.

gears/system/cluster/cluster-sdk/src/cache/facade.rs
asmith987: Fixed. Added `validate_cache_key()` (slash-permitted `[a-zA-Z0-9_/-]` segments, max 255 chars — documented as `CACHE_KEY_RULE`) at the top of all 7 key-accepting methods: `get`, `put`, `delete`, `contains`, `put_if_absent`, `compare_and_swap`, `watch`. `put` and `put_if_absent` validate `req.key`. `watch_prefix` and `scan_prefix` are excluded — they take a prefix, not a key, and trailing-slash prefixes like `svc/` are intentional.

gears/system/cluster/cluster-sdk/src/lock/facade.rs
asmith987: Fixed. Added `validate_cluster_name(name)?` at the top of both `try_lock` and `lock` before delegating to `self.inner`.

gears/system/cluster/cluster-sdk/src/leader/facade.rs
asmith987: Fixed. Added `validate_cluster_name(name)?` at the top of both `elect` and `elect_with_config` before any other work.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
asmith987: Not applied. The `sender.is_closed()` check on every tick and backend stream termination (which delivers `Closed(Shutdown)` during cluster shutdown) already handle all termination cases correctly. Adding a `CancellationToken` would be dead infrastructure — nothing in the current codebase can cancel it until the wiring crate (`ClusterHandle`) is built in the follow-up change. Deferring to that PR.

gears/system/cluster/cluster-sdk/src/cache/polyfill.rs
asmith987: Fixed. `snapshot()` now issues all `get` calls concurrently via `tokio::spawn`, bounded to 32 in-flight tasks at a time via `tokio::sync::Semaphore` to prevent unbounded fan-out on large keyspaces. `JoinError` (task panic) is now embedded in the returned `Provider` error message rather than discarded.

gears/system/cluster/cluster-sdk/src/cache/scoped.rs
asmith987: Not applied. `tx.closed()` handles consumer-drop termination and the inner backend watch delivers `Closed(Shutdown)` on cluster shutdown, which propagates through the forwarding task naturally. A self-owned `CancellationToken` created in `ScopedCacheBackend::new()` would have no external owner and be permanently inert — dead code that implies a shutdown mechanism that does not exist yet. DESIGN §3.7 places lifecycle ownership in `ClusterHandle` (follow-up crate); deferring token injection to that PR.

gears/system/cluster/cluster-sdk/src/observability.rs
asmith987: Applied. Added `#[doc(hidden)]` to both `ResourceId` and `emit_provider_error`. They remain `pub` so `cf-gears-cluster`'s default backends can continue using them, but are de-emphasised from the public docs surface.

gears/system/cluster/cluster-sdk/src/cache/resolver.rs
asmith987: Fixed. `map_err` now captures the hub error and logs it via `tracing::debug!(profile, error = %err, "cluster backend lookup failed for profile")` before mapping to `ClusterError::ProfileNotBound`. The original cause is preserved at debug level without changing the public error type.

gears/system/cluster/cluster-sdk/src/discovery/resolver.rs
asmith987: Fixed. `map_err` now captures the hub error and logs it via `tracing::debug!(profile, error = %err, "cluster backend lookup failed for profile")` before mapping to `ClusterError::ProfileNotBound`. The original cause is preserved at debug level without changing the public error type.

gears/system/cluster/cluster-sdk/src/leader/resolver.rs
asmith987: Fixed. `map_err` now captures the hub error and logs it via `tracing::debug!(profile, error = %err, "cluster backend lookup failed for profile")` before mapping to `ClusterError::ProfileNotBound`. The original cause is preserved at debug level without changing the public error type.

gears/system/cluster/cluster-sdk/src/lock/resolver.rs
asmith987: Fixed. `map_err` now captures the hub error and logs it via `tracing::debug!(profile, error = %err, "cluster backend lookup failed for profile")` before mapping to `ClusterError::ProfileNotBound`. The original cause is preserved at debug level without changing the public error type.

gears/system/cluster/cluster-sdk/src/observability_tests.rs
asmith987: Fixed. Removed `span_names_follow_dotted_lowercase_convention`.

gears/system/cluster/cluster-sdk/src/observability_tests.rs
asmith987: Fixed. Removed `metric_names_follow_underscored_lowercase_convention`.

gears/system/cluster/cluster-sdk/src/restart_tests.rs
asmith987: Fixed. Added zero-jitter behavioral assertions alongside the existing field checks: `assert_eq!(zero_jitter.backoff_for(0), Duration::from_secs(1))` and `assert_eq!(zero_jitter.backoff_for(10), Duration::from_secs(30))` pin the exact schedule; the full-jitter variant is verified to stay within bounds.

gears/system/cluster/cluster-sdk/src/observability/instrumented_cache_tests.rs
asmith987: Fixed. Removed `delegates_inner_consistency_and_features`.

gears/system/cluster/cluster-sdk/src/lock/guard_tests.rs
asmith987: Fixed. Removed the `assert_eq!(guard.name(), "rate-limit")` line; the behavioral assertion (no command sent on drop) is kept.

gears/system/cluster/cluster-sdk/src/discovery/backend_tests.rs
asmith987: Fixed. Replaced with a contract-exercising assertion: constructs a `StubBackend { metadata_pushdown: false }`, calls `validate_service_discovery_capabilities` with `MetadataFiltering`, and asserts the resulting `CapabilityNotMet` error's `provider` field contains `"StubBackend"` — verifying the consumer-visible contract, not language semantics.

gears/system/cluster/cluster/src/defaults/discovery.rs
asmith987: I went ahead and moved the defaults into a `cluster/src/defaults` crate where we will wire things up, so the SDK is purely client facing API, traits etc.

striped-zebra-dev (approved):

PR #4127, opened 2026-06-19
fix(api-gateway): emit WWW-Authenticate challenge on rejected bearer …

shaneemberton:
…tokens

Per RFC 6750 §3, a rejected bearer request must carry a WWW-Authenticate challenge. On AuthNResolverError::Unauthorized, append `Bearer error="invalid_token"`; realm and error_description are omitted so the challenge discloses no internal detail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated authentication and authorization failure responses to follow RFC 6750.
  * Unauthorized/invalid-token cases now include the correct `WWW-Authenticate: Bearer` challenge.
  * Missing credentials now return an unauthenticated response with `WWW-Authenticate: Bearer realm="api"`.
  * Insufficient-scope denials now include `WWW-Authenticate: Bearer error="insufficient_scope"`.
  * Infrastructure/authn resolver failures (e.g., 500/503) continue to return consistent responses without a bearer challenge (and still include `Retry-After` where applicable).
* **Tests**
  * Added/extended coverage for `WWW-Authenticate` behavior, including scope enforcement outcomes.

### DIFF
--- a/gears/system/api-gateway/src/middleware/auth.rs
+++ b/gears/system/api-gateway/src/middleware/auth.rs
@@ -1,6 +1,5 @@
 use axum::http::Method;
 use axum::response::IntoResponse;
-use http::HeaderValue;
 use std::{collections::HashMap, sync::Arc};
 
 use crate::middleware::common;
@@ -200,13 +199,19 @@ pub async fn authn_middleware(
         }
         AuthRequirement::Required => {
             let Some(token) = extract_bearer_token(req.headers()) else {
-                let err = CanonicalError::unauthenticated()
-                    .with_reason("MISSING_BEARER")
-                    .create();
                 // `instance` / `trace_id` are filled by the canonical
                 // error middleware (`toolkit::api::canonical_error_middleware`)
                 // on the way out — this middleware sits inside its layer.
-                return err.into_response();
+                let mut response = CanonicalError::unauthenticated()
+                    .with_reason("MISSING_BEARER")
+                    .create()
+                    .into_response();
+                // No bearer credentials were presented (RFC 6750 §3).
+                common::append_bearer_challenge(
+                    &mut response,
+                    common::BearerChallenge::NoCredentials,
+                );
+                return response;
             };
 
             match state.authn_client.authenticate(token).await {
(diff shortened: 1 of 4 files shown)


gears/system/api-gateway/src/middleware/auth.rs
```diff
@@ -226,20 +227,35 @@ pub async fn authn_middleware(
...
-            .with_reason("AUTHN_FAILED")
-            .create(),
+    match err {
+        AuthNResolverError::Unauthorized(_) => {
+            // A token was presented but rejected. RFC 6750 §3 requires a
+            // rejected bearer request to carry a `WWW-Authenticate` challenge,
+            // with an `error` code when a token was supplied — here
+            // `invalid_token`. We deliberately omit `realm` and
+            // `error_description` so the challenge discloses no internal detail.
+            let mut response = CanonicalError::unauthenticated()
```
aviator5, line 238: Returning `WWW-Authenticate` is a great idea. Could you please also check other places in `auth.rs` and `scope_enforcement.rs` files? For example, if token is missing, it's worth to return `WWW-Authenticate: error="invalid_request"`. For invalid token's scope - `insufficient_scope`. Additionally, I would also suggest including `error_description` (it's also part of the https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) to make responses even more descriptive.

aviator5 (changes requested):

gears/system/api-gateway/src/middleware/auth.rs
shaneemberton: I agree. I'll add "invalid_request" and "insufficient_scope" error codes for those cases and will update the code shortly.

Regarding `error_description`: I agree to include it. My question is about the scope—should it map 1:1 to the `error` value, or provide additional details?

For example:
- 1:1: `error="invalid_token", error_description="The token is invalid"`
- Additional info: `error="invalid_token", error_description="The access token expired"`

Which approach would you prefer?

aviator5: Hm, good question. `error_description` like "The token is invalid" is useless. From the other hand, having "The access token expired" may give the attacker a bit more details. I would prefer returning descriptive message in the dev env or for the trusted clients, but since we don't have such distinguishing mechanism and also since we need to be secure-by-default, I would suggest returning only "error" field for now and just using invalid_request/invalid_token/insufficient_scope codes. Sorry, for the confusion.

shaneemberton: I addressed the review comments in the additional commit, except for the case where authentication information is missing.

For that case, I only include the following challenge in the response header:

```http
WWW-Authenticate: Bearer realm="api"
```
This follows RFC 6750:
If the request lacks any authentication information, the resource server SHOULD NOT include an error code or other error information.

The RFC also states that the Bearer scheme must be followed by one or more auth-param values, so realm="api" is still included.

aviator5 (approved):

aviator5:
@shaneemberton could you please also extend gears/system/api-gateway/tests/auth_middleware.rs adding checks for the new `WWW-Authenticate` response header?

shaneemberton:
> @shaneemberton could you please also extend gears/system/api-gateway/tests/auth_middleware.rs adding checks for the new `WWW-Authenticate` response header?

Done.

gears/system/api-gateway/tests/auth_middleware.rs
```diff
@@ -55,6 +57,14 @@ fn content_type(response: &Response) -> &str {
         .unwrap_or("")
 }
 
+fn www_authenticate(response: &Response) -> &str {
+    response
+        .headers()
+        .get(header::WWW_AUTHENTICATE)
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
+}
```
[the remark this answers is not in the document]
shaneemberton: Fixed.

aviator5 (approved):

MikeFalcon77 (commented):
.

MikeFalcon77 (approved):

PR #4136, opened 2026-06-23
feat(model-registry): Model registry SDK types

Qilin101:
SDK for model-registry. P1 features only - model info, parameters, manual model management.  
The rest of the model registry design and implementation are out of scope of this pr.

## Summary by CodeRabbit

* **New Features**
  * Added a new model registry SDK offering tenant-scoped operations to manage models and providers.
  * Included structured request/response types for model lifecycle, approvals, capabilities, defaults, and provider-specific settings.
  * Added a dedicated SDK error type for clearer handling of validation, auth, and registry state issues.
* **Documentation**
  * Updated Model Registry PRD and upstream requirements for revised phased maturity and behaviors.
  * Improved ADR navigation with added table of contents sections.
* **Chores**
  * Adjusted artifact autodetection to include model-registry content and updated release configuration for the new SDK package.


gears/model-registry/model-registry-sdk/src/models/entity.rs
```diff
@@ -0,0 +1,356 @@
...
+#[derive(Debug, Clone, PartialEq)]
+#[non_exhaustive]
+pub struct Provider {
+    pub id: Uuid,
+    /// Human-readable identifier (immutable after creation).
+    /// Format: 1-64 chars, lowercase alphanumeric + hyphen.
+    pub slug: String,
+    pub name: String,
+    /// GTS type identifier for the provider.
+    pub gts_type: String,
```
Artifizer: **HIGH**

`gts_type` should be a typed `gts::GtsTypeId`, not a raw `String`.

Using `String` drops GTS structural validation at the SDK boundary and lets invalid or non-type identifiers flow through the public contract unchecked.

Change both `Provider.gts_type` and the corresponding request/builder fields to `gts::GtsTypeId` so invalid values are rejected by construction.

gears/model-registry/model-registry-sdk/src/models/providers/anthropic.rs
```diff
@@ -0,0 +1,392 @@
...
+    // ── Provider-wire inference defaults ──────────────────────────────
+    /// Sampling temperature (`Anthropic` accepts `0.0..=1.0`; SDK does not
+    /// range-check).
+    pub temperature: Option<f64>,
+    pub top_p: Option<f64>,
+    pub top_k: Option<u32>,
+    /// **Required** by `Anthropic` on every request. The SDK default of
+    /// `0` is treated as "unset" by callers, forcing them to use max
+    /// context size.
+    pub max_tokens: u32,
```
Artifizer: **HIGH**

`max_tokens` uses `0` as an unset sentinel in a public SDK type.

That makes an invalid Anthropic request state representable and serializable, so callers can persist or emit payloads the provider rejects before they patch the field.

Model this as `Option<NonZeroU32>` or a validated newtype so omission and valid values are distinct without relying on a sentinel.

gears/model-registry/model-registry-sdk/src/models/request.rs
```diff
@@ -0,0 +1,399 @@
...
+        assert_eq!(req.slug(), "openai");
+        assert_eq!(req.name(), "OpenAI");
+        assert_eq!(req.oagw_alias(), "openai-prod");
+        assert!(req.discovery_enabled());
+        assert_eq!(req.discovery_interval_seconds(), Some(3600));
+        assert!(!req.managed());
+    }
+
+    #[test]
+    fn create_provider_request_defaults() {
```
Artifizer: **MEDIUM**

This and the adjacent `*_default_is_empty` tests are constructor/default echo tests rather than behavioral contract tests.

They mainly reassert Rust defaults and field assignment, so they will keep passing even if request validation, tri-state PATCH semantics, or serialization behavior regresses.

Replace them with tests that exercise observable behavior such as invalid `gts_type` rejection, tri-state PATCH serialization, and builder invariants.

gears/model-registry/model-registry-sdk/src/errors.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[error("invalid state transition: {detail}")]
+    InvalidTransition { detail: String },
+
+    #[error("validation error: {message}")]
+    Validation { message: String },
+
+    #[error("unauthorized: {detail}")]
+    Unauthorized { detail: String },
+
+    #[error("alias not found: {name}")]
```
Artifizer: **MEDIUM — error enum is wider than the actual API**

Heads up: `AliasNotFound`, `AliasConflict` (here) and `DiscoveryFailed` below don't map to anything on `ModelRegistryClientV1` — there are no alias or discovery methods on the trait. So today these variants can never actually be produced by the public surface.

Two ways to read this: either the trait is missing those operations, or these are placeholders for a future phase. If it's the latter, I'd hold off adding them until the methods land — public error variants are part of the contract, and shipping ones nobody can hit just adds surface we have to keep forever.

While we're here: a single `Unauthorized` will map ambiguously to 401 vs 403 downstream. Cheap to split now, annoying later.

gears/model-registry/model-registry-sdk/src/models/provider_settings.rs
```diff
@@ -0,0 +1,30 @@
...
+// Trait
+// ---------------------------------------------------------------------------
+
+/// Marker trait implemented by every typed provider-settings type shipped
+/// in [`crate::models::providers`] (e.g. `OpenAiSettingsV1`).
+///
+/// Acts purely as documentation that a type is a typed provider-settings
+/// payload — there are no required methods. Provider family identification
+/// is done via `ModelInfoV1::gts_type` (a [`gts::GtsSchemaId`]).
+pub trait ProviderSettings:
```
Artifizer: **MEDIUM — `ProviderSettings` marker trait isn't actually doing anything**

This trait is defined and implemented, but nothing ever *bounds* on it. Everywhere it matters — `Model<P>`, `ModelInfoV1<P>`, `try_into_typed` — the bound is `P: gts::GtsSchema`, not `P: ProviderSettings`. So as written, `Model<SomeRandomGtsSchemaThing>` compiles fine and the marker buys us zero compile-time guarantee; it's basically a doc comment with extra steps.

Pick a lane: either make it earn its keep by bounding the entity/info generics on `P: ProviderSettings`, or drop it and bound on `GtsSchema` directly. Right now it's a dead abstraction in the public API.

gears/model-registry/model-registry-sdk/src/models/info.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Clone, PartialEq)]
+#[non_exhaustive]
+pub struct ModelInfoV1<P: gts::GtsSchema = serde_json::Value> {
+    // ── GTS schema identity ───────────────────────────────────────────
+    /// Full GTS schema chain identifying this model's settings shape — e.g.
+    /// `gts.cf.genai.model.info.v1~cf.genai._.openai.v1~`. Mirrors
+    /// `Provider.gts_type` and is the canonical key for resolving the
+    /// concrete shape of `provider_settings` (which is a raw JSON blob —
+    /// `serde_json::Value` — by default).
+    pub gts_type: gts::GtsTypeId,
```
Artifizer, line 64: **MEDIUM — two sources of truth for the provider type**

For a typed model like `Model<OpenAiSettingsV1>` the provider type is encoded twice: once at runtime in this `gts_type` field, and once at compile time as `OpenAiSettingsV1::TYPE_ID`. Nothing keeps them in sync — you can hand-build a `Model<OpenAiSettingsV1>` whose `gts_type` says Anthropic and it'll happily compile.

`try_into_typed` checks the match on the way *in*, but once you're holding a typed value the invariant isn't preserved by the type anymore. Worth either documenting loudly that `gts_type` is authoritative and the generic is just a view, or deriving `gts_type` from `P` for typed models so the two can't drift.

gears/model-registry/model-registry-sdk/src/models/entity.rs
```diff
@@ -0,0 +1,356 @@
...
+///
+/// Generic over `P: gts::GtsSchema` so that consumers narrowed to a
+/// specific provider can carry `Model<OpenAiSettingsV1>` etc. The default
+/// `P = serde_json::Value` (which implements `gts::GtsSchema` upstream) is
+/// what the public [`crate::api::ModelRegistryClientV1`] trait returns — the
+/// provider settings ride as opaque JSON until the consumer narrows via
+/// [`Model::try_into_typed`], which reads `info.gts_type` for dispatch.
+#[derive(Debug, Clone, PartialEq)]
+#[non_exhaustive]
+pub struct Model<P: gts::GtsSchema = serde_json::Value> {
```
Artifizer: **MEDIUM — versioning is inconsistent and it'll bite on the next revision**

The envelope and payloads are version-suffixed (`ModelInfoV1`, `OpenAiSettingsV1`, `AnthropicSettingsV1`), but the things that *wrap* them — `Model`, `Provider`, the request DTOs, `ModelRegistryError` — are not. And `ModelRegistryClientV1` (a `V1` trait!) returns the unversioned `Model`.

The problem: when the model shape needs to change, you can't cleanly stand up `Model` + `ModelV2` side by side, because `V1` already hard-returns `Model`. So the `V1` suffix on the trait doesn't actually give us the evolution path it implies. Either commit to versioning the wrappers too, or document why these are intentionally unversioned and how v2 is meant to roll out.

Artifizer (commented):
Round 2 — a few deeper **architecture/contract** thoughts on the SDK shape itself (separate from the field-level nits earlier). All MEDIUM. None are blockers on their own, but they're the kind of thing that's painful to change once consumers depend on this crate, so worth settling before merge.

gears/model-registry/model-registry-sdk/src/models/entity.rs
Qilin101: fixed

gears/model-registry/model-registry-sdk/src/models/providers/anthropic.rs
Qilin101: fixed
decided to use `Option<u32>` since `0` is actually a valid anthopic api value (can be used to preheat prompt cache).

gears/model-registry/model-registry-sdk/src/models/request.rs
Qilin101: fixed

gears/model-registry/model-registry-sdk/src/errors.rs
Qilin101: removed unnecessary values

gears/model-registry/model-registry-sdk/src/models/provider_settings.rs
Qilin101: removed this trait

gears/model-registry/model-registry-sdk/src/models/info.rs
Qilin101: yeah, but we need to be able to understand what type this is in runtime. sdk will return generic "model" and caller needs to know what specific settings type is returned. so keeping this for now, updated the comment.

gears/model-registry/model-registry-sdk/src/models/entity.rs
Qilin101: added version suffix to other types as well.

gears/model-registry/model-registry-sdk/src/models/providers/anthropic.rs
```diff
@@ -0,0 +1,401 @@
...
+        budget_tokens: u32,
+        display: Option<AnthropicThinkingDisplay>,
+    },
+    /// Extended thinking is off.
+    Disabled,
+    /// Server adapts the thinking budget; recommended replacement for
+    /// `Enabled` on newer models.
+    Adaptive {
+        display: Option<AnthropicThinkingDisplay>,
+    },
```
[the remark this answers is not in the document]
Qilin101: i don't think that's necessary. this is not provider payload, this is settings for provider plugin. plus i'm not sure how this will influence gts schemas.

PR #4142, opened 2026-06-24
fix(toolkit-config): allow k8s env override of gear config via underscore gear names

### LINKED ISSUE #4115: Gear/plugin config can't be overridden via env vars in Kubernetes (dash in gear name vs C_IDENTIFIER)
AppConfig::load_layered lets any config value be overridden by an env var through the figment Env::prefixed("APP__").split("__") layer (libs/toolkit/src/bootstrap/config/mod.rs:321-325). This works app-wide, including for gear/plugin config under gears.<gear-name>.config.<field>.

It works bad in k8s for any gear whose name contains a dash (which is all of them — static-authz-plugin, static-authn-plugin, etc.). k8s validates env[].name as a C_IDENTIFIER ([A-Za-z_][A-Za-z0-9_]*), so a variable name containing a dash is rejected by the API server. Since figment maps the env var name to the config key verbatim (only lowercasing and splitting __ into nesting levels), the gear-name segment must literally contain the dash to match the gears map key — which k8s won't allow.

**Proposed fix**

In AppConfig::load_layered, customize the Env provider so that, for keys under the gears branch, underscores in the gear-name segment are translated to dashes. Scope this strictly:

Only the segment immediately following gears (the gear name) is remapped _ → -.
All other segments are left untouched — in particular nested config field names that legitimately use underscores (e.g. max_age_days, console_level) must not be altered.
Do not touch other top-level branches (vendor.<name> uses underscore-style keys and must keep them).

``` Rust
.merge(
    Env::prefixed("APP__")
        .split("__")
        .map(|key| {
            let mut parts: Vec<&str> = key.as_str().split('.').collect();
            if parts.first() == Some(&"gears") && parts.len() >= 2 {
                let gear = parts[1].replace('_', "-");
                // rebuild with only the gear-name segment remapped
                ...
            } else {
                key.into()
            }
        })
)
```
This makes APP__GEARS__STATIC_AUTHZ_PLUGIN__CONFIG__VENDOR=acme resolve to gears.static-authz-plugin.config.vendor, which is a valid k8s env var name and overrides the right key.

Assumption to confirm and document: gear names never contain underscores (kebab-case only). If that ever changes, this remap becomes ambiguous.

**Acceptance criteria**

APP__GEARS__STATIC_AUTHZ_PLUGIN__CONFIG__VENDOR overrides gears.static-authz-plugin.config.vendor.
The existing dash form (APP__GEARS__static-authz-plugin__CONFIG__VENDOR) still works (back-compat).
Nested config fields with underscores (e.g. logging max_age_days, a gear config field named some_field) are still overridable and not mangled.
vendor.<name> overrides (e.g. test_vendor_config_env_override) are unaffected.
Tests covering: gear-name remap, untouched field-name underscores, untouched vendor branch.
Out of scope

ExpandVars / ${VAR} substitution — that's a separate, opt-in mechanism for developer-declared placeholders and not a substitute for generic env override.

mattgarmon:
Closes https://github.com/constructorfabric/gears-rust/issues/4115

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment-based configuration parsing so gear names provided with underscores correctly map to kebab-case keys.
  * Preserved underscores in nested configuration field names during env-var ingestion.
  * Ensured `vendor` environment overrides remain unaffected by gear-name remapping.
  * Updated OData cursor datetime formatting for `DateTimeUtc` to use a more reliable RFC3339 format description.
* **Tests**
  * Reduced flakiness by running env-dependent configuration tests serially.
  * Added/expanded coverage for underscore-form gear names and nested field override behavior.
* **Chores**
  * Added `serial_test` as a shared development test utility.

### DIFF
--- a/libs/toolkit/src/bootstrap/config/mod.rs
+++ b/libs/toolkit/src/bootstrap/config/mod.rs
@@ -303,6 +303,29 @@ pub fn default_logging_config() -> LoggingConfig {
     logging
 }
 
+/// Remap a split, `APP__`-prefixed environment key into a figment config path.
+///
+/// k8s env var names must be `C_IDENTIFIERs` (no dashes), but gear names are
+/// kebab-case. Under the `gears` branch we therefore remap `_` -> `-` in the
+/// gear-name segment only (the one right after `gears`). Nested field names
+/// (e.g. `max_age_days`) and other branches (e.g. `vendor`) are left alone; the
+/// dash form is unaffected since remapping `-` is a no-op.
+///
+/// Safe because gear names are guaranteed kebab-case by `validate_kebab_case`
+/// (`#[toolkit::gear]` macro + `make validate-gear-names`).
+pub(crate) fn remap_gear_env_key(key: &str) -> String {
+    // Lowercase so the match is independent of figment's own (later) lowercasing.
+    let lower = key.to_ascii_lowercase();
+    let mut parts: Vec<&str> = lower.split('.').collect();
+    if parts.first() == Some(&"gears") && parts.len() >= 2 {
+        let gear = parts[1].replace('_', "-");
+        parts[1] = gear.as_str();
+        parts.join(".")
+    } else {
+        lower
+    }
+}
+
 impl AppConfig {
     /// Load configuration with layered loading: defaults → YAML file → environment variables.
     /// Also normalizes `server.home_dir` into an absolute path and creates the directory.
@@ -321,8 +344,12 @@ impl AppConfig {
         let figment = Figment::new()
             .merge(Serialized::defaults(AppConfig::default()))
             .merge(StrictYaml::file(config_path))
-            // Example: APP__SERVER__PORT=8087 maps to server.port
-            .merge(Env::prefixed("APP__").split("__"));
+            // Example: APP__SERVER__PORT=8087 maps to server.port.
+            .merge(
+                Env::prefixed("APP__")
+                    .split("__")
+                    .map(|key| remap_gear_env_key(key.as_str()).into()),
+            );
 
         let mut config: AppConfig = figment
             .extract()
@@ -1304,6 +1331,7 @@ pub fn gear_home(app: &AppConfig, gear_name: &str) -> PathBuf {
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
     use super::*;
+    use serial_test::serial;
     use std::fs;
     use temp_env::with_var;
     use tempfile::tempdir;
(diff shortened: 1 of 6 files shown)


libs/toolkit/src/bootstrap/config/mod.rs
```diff
@@ -2892,6 +2917,114 @@ vendor:
...
+            "APP__VENDOR__ENV_TEST_VENDOR__API_TOKEN",
+            Some("from_env"),
+            || {
+                let config = AppConfig::load_layered(&cfg_path).unwrap();
+                let v: TestVendorConfig = config.vendor_config("env_test_vendor").unwrap();
+                assert_eq!(v.api_token, "from_env");
+            },
+        );
+    }
+
```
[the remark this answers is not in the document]
mattgarmon: Added

mattgarmon:
@coderabbitai review

libs/toolkit/src/bootstrap/config/mod.rs
```diff
@@ -321,8 +321,28 @@ impl AppConfig {
...
+            // Example: APP__SERVER__PORT=8087 maps to server.port.
+            //
+            // k8s env var names must be C_IDENTIFIERs (no dashes), but gear names
+            // are kebab-case, so under the `gears` branch we remap `_` -> `-` in the
+            // gear-name segment only (the one right after `gears`). Nested field
+            // names (e.g. `max_age_days`) and other branches (e.g. `vendor`) are left
+            // alone; the dash form is unaffected since remapping `-` is a no-op.
+            // Safe because gear names are guaranteed kebab-case by validate_kebab_case
+            // (#[toolkit::gear] macro + `make validate-gear-names`).
+            .merge(Env::prefixed("APP__").split("__").map(|key| {
```
MikeFalcon77: **MEDIUM**

Correctness-sensitive env-key remapping (the gears kebab-case naming contract) is embedded inline as a Figment `map` closure instead of an extracted helper.

The logic can only be exercised end-to-end through `load_layered` with process-global env vars, so it cannot be unit-tested in isolation and obscures the config-assembly flow on a startup-critical path.

Extract the closure body into `pub(crate) fn remap_gear_env_key(key: &str) -> String`, add table-driven unit tests for it, and reduce the `.map()` call to a one-line reference.

libs/toolkit/src/bootstrap/config/mod.rs
```diff
@@ -2892,6 +2926,114 @@ vendor:
...
+                let config = AppConfig::load_layered(&cfg_path).unwrap();
+                let gear = config.gears.get("static-authz-plugin").unwrap();
+                assert_eq!(gear["config"]["some_field"], serde_json::json!("from_env"));
+            },
+        );
+    }
+
+    #[test]
+    #[serial]
+    fn test_vendor_config_env_override_unaffected_by_gear_remap() {
```
MikeFalcon77, line 3053: **LOW**

This test duplicates the existing `test_vendor_config_env_override` (lines 2906-2927): same env var, same YAML keys, same assertion.

The remap only rewrites the segment after `gears`; this vendor input has no `gears` branch, so the test cannot distinguish remap-affected from unaffected behavior and would never fail if the gear remap leaked into the vendor branch.

Use a vendor name containing an underscore that would break the lookup if wrongly remapped to a dash, so the test actually fails when the isolation invariant is violated.

MikeFalcon77 (changes requested):

libs/toolkit/src/bootstrap/config/mod.rs
mattgarmon: Moved to separate function, and added additional unit tests

libs/toolkit/src/bootstrap/config/mod.rs
mattgarmon: Updated

MikeFalcon77 (approved):

PR #4146, opened 2026-06-26
feat(toolkit-macros): compile-time gear dependency linking via re-exports

maurolacy:
## Problem

The `deps` field in `#[toolkit::gear(deps = [...])]` declares runtime lifecycle
dependencies between gears. At startup, `build_dependency_graph()` validates that
every declared dep exists in the `inventory`-based registry. However, `inventory`
relies on linker-section tricks: only crates explicitly referenced with `use X as _;`
in the generated `main.rs` survive the linker. Transitive Cargo dependencies get
their `inventory::submit!` registrations silently stripped.

This forces users to flatten the entire gear dependency tree in `Gears.toml`. For
example, listing `api-gateway` is not enough. You must also manually list
`grpc-hub`, `authn-resolver`, `types-registry`, and every other transitive dep.
Forgetting one causes a **runtime** `UnknownDependency` error at startup, with no
compile-time signal that anything is wrong.

## Solution

The `#[toolkit::gear]` macro now generates `pub use` re-exports for each declared
dependency:

```rust
#[toolkit::gear(name = "api-gateway", deps = ["grpc-hub", "authn-resolver"])]
```

expands to (among other things):

```rust
#[cfg(not(test))]
#[doc(hidden)]
pub use grpc_hub as _gear_dep_grpc_hub;

#[cfg(not(test))]
#[doc(hidden)]
pub use authn_resolver as _gear_dep_authn_resolver;
```

This forces the linker to keep the dependency gear crates (and their
`inventory::submit!` registrations) alive when a gear is pulled in transitively.
The re-exports chain: if `api-gateway` re-exports `authn-resolver`, and
`authn-resolver` re-exports `types-registry`, then pulling in `api-gateway` alone
is sufficient to register the entire transitive dependency tree.

The re-exports are gated behind `#[cfg(not(test))]`, so that test builds (where
gears are often defined in the same compilation unit) don't require phantom
external crate dependencies.

## Trade-off: Cargo.toml dependencies

For the `pub use` to compile, each gear's `Cargo.toml` must list its dep gear
crates as Cargo dependencies (not just SDKs). This is a new requirement, but
**forgetting a dependency is now a compile-time error** (`no external crate`),
not a silent runtime failure. The SDK dependencies remain necessary for the
actual API types used in gear implementations.

## Naming convention

This change relies on a strict, bidirectional naming convention:

| Gear name | Crate name | Lib name |
|---|---|---|
| `api-gateway` | `cf-gears-api-gateway` | `api_gateway` |
| `types-registry` | `cf-gears-types-registry` | `types_registry` |

The macro derives the lib name from the dep's gear name by replacing `-` with `_`.
This convention is already followed by all gear crates in the repo (24/27 gears;
the 3 exceptions are internal plugins in `mini-chat` which are same-crate and
unaffected).

## Changes

- **`libs/toolkit-macros/src/lib.rs`**: Generate `pub use` re-exports for each
  entry in `deps = [...]`, gated behind `#[cfg(not(test))]`.
- **`Cargo.toml` (root)**: Added 8 workspace-level dependency entries for gear
  crates that were previously only available as SDKs.
- **22 gear `Cargo.toml` files**: Added the full gear crate as a dependency
  (via `{ workspace = true }`) for each declared dep.

## Verification

- `cargo build`: Entire workspace compiles cleanly.
- `cargo test --workspace`: All test suites pass.

## What this enables

With this change, `cargo gears` and future tooling no longer needs to resolve
and flatten the transitive gear dependency tree. Users can list only the gears
they directly use in `Gears.toml`, and the macro-generated re-exports ensure the
full dependency chain is linked.

## Summary by CodeRabbit

- **New Features**
  - Gear dependency declarations now use consistent identifier-based syntax, with clearer validation for incorrect formats.
  - Gear integrations now benefit from automatic dependency handling/re-exports to improve reliability.
  - Expanded workspace support adds additional system crates, enabling broader capabilities across gears and plugins.

- **Documentation**
  - Added guidance on declaring gear dependencies and configuring CI “shear” exclusions.

- **CI Improvements**
  - Coverage and shear jobs now free disk space before running to improve build stability.


aviator5:
Could we avoid deriving the Rust crate identifier from the dependency’s runtime name (`"authn-resolver"` -> `authn_resolver`) and make `deps` use crate identifiers instead?

For example:

```rust
#[toolkit::gear(
    name = "api-gateway",
    deps = [grpc_hub, authn_resolver],
)]
```

Each gear macro expansion could expose its canonical runtime name:

```rust
#[doc(hidden)]
pub const __TOOLKIT_GEAR_NAME: &str = "authn-resolver";
```

Then the `api-gateway` expansion could both force-link and register the dependency without duplicating or deriving its name:

```rust
#[doc(hidden)]
pub use ::authn_resolver as _toolkit_gear_dep_authn_resolver;

builder.register_core_with_meta(
    "api-gateway",
    &[::authn_resolver::__TOOLKIT_GEAR_NAME],
    // ...
);
```

This would keep the link-time behavior from this PR, while removing the implicit `gear name -> lib crate name` convention and allowing Cargo aliases or custom `[lib]` names. The compiler would also validate the actual dependency path directly.

@maurolacy, what do you think about this direction?

maurolacy:
> Could we avoid deriving the Rust crate identifier from the dependency’s runtime name (`"authn-resolver"` -> `authn_resolver`) and make `deps` use crate identifiers instead?

I think this is a great idea, since in that way the compiler can validate those identifiers directly.

Now, introducing a new `__TOOLKIT_GEAR_NAME` const has some problems: Every gear defines its struct in a submodule (`src/gear.rs` or `src/module.rs`), not at crate root, and the `#[gear]` macro generates code at the call site. So the const would end up at e.g. `::authn_resolver::gear::__TOOLKIT_GEAR_NAME`, not at `::authn_resolver::__TOOLKIT_GEAR_NAME`.
For the macro expansion to reference `::authn_resolver::__TOOLKIT_GEAR_NAME`, every gear crate's `lib.rs` would need to add a re-export. That's ~30 gear crates to update, plus a new convention to maintain.

Since the naming convention currently holds, and for ~30 crates, why introduce a mapping at all? In any case, that mapping can be introduced later, if / when needed (i.e. when somebody breaks the naming convention for some (sound) reason).

So, my proposal would be a hybrid of the current changes and your proposal:

- Accept idents: `deps = [authn_resolver]`.
- Use idents directly for `pub use` re-exports (compile-time validation).
- Derive runtime name at macro time: `authn_resolver` → `"authn-resolver"`.
- Pro: Same compile-time validation win, zero `lib.rs` changes, tests work as-is.
- Con: Still has a naming convention (just reversed).

I'll go ahead and implement this variant, and we can discuss it further in any case.

MF: sorry, I edited this comment by mistake. (reverted to original)

MikeFalcon77:
Agreeing with the hybrid @maurolacy described. The `__TOOLKIT_GEAR_NAME` route (as @aviator5 suggested) does need a re-export in each gear crate's `lib.rs` since the macro expands in a submodule (`src/gear.rs` / `module.rs`), so the const lands at `::crate::gear::__TOOLKIT_GEAR_NAME`, not at crate root — i.e. ~30 crates to touch, not a one-liner. Fine to defer until the naming convention actually needs to break.

One thing not yet raised: the `cargo-shear` ignore list is a maintenance footgun. Every new gear dep has to be added to the workspace/package `ignored` list by hand, since the dep is only consumed through the macro-generated re-export and shear can't see it. Forgetting it is a CI break unrelated to the change — might be worth a note in the macro docs / CONTRIBUTING so it isn't discovered the hard way.

maurolacy:
Yeah, I was thinking and analysing this a bit more with AI, and the fact remains: This is a real limitation of proc macros in Rust; they can only emit code at their invocation site. No way to "reach up" to the crate root.

So the options remain:
- Current approach: _ → - derivation, zero manual steps, universally correct today.
- Const approach: accurate names from source, but requires a one-line `pub use` in every gear's `lib.rs` (and 4 gears need `module::` instead of `gear::`).

Regarding `cargo-shear` ignore list, there's not much to do here, except maybe turning these into warnings / non-errors in CI. Will investigate / propose this, as I have some ideas on CI improvements already for another PR.
Adding a note in `CONTRIBUTING.md`.

MikeFalcon77 (approved):

PR #4147, opened 2026-06-26
fix(resource-group): add explicit `[lib] name` to follow gear naming convention

maurolacy:
## Problem

Every gear crate in the repo follows a consistent naming convention:

| Gear name | Package name | Lib name |
|---|---|---|
| `api-gateway` | `cf-gears-api-gateway` | `api_gateway` |
| `types-registry` | `cf-gears-types-registry` | `types_registry` |
| `authn-resolver` | `cf-gears-authn-resolver` | `authn_resolver` |

The `resource-group` crate is the only exception: it has no explicit `[lib]`
section in its `Cargo.toml`, so Cargo defaults the lib name to
`cf_gears_resource_group` (derived from the package name) instead of
`resource_group` (derived from the gear name).

This breaks the bidirectional `gear-name <-> lib-name` mapping.
If this mapping is maintained (or even enforced, at some point), it can serve
as a basis for avoiding explicit / hard-coded mappings from gear and lib
names to and from package names.

## Fix

- Added `[lib] name = "resource_group"` to
  `gears/system/resource-group/resource-group/Cargo.toml`
- Renamed all 68 `cf_gears_resource_group` references to `resource_group`
  across 10 test files

## Verification

- `cargo build`: Full workspace compiles cleanly.
- `cargo test -p cf-gears-resource-group`: All tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Explicitly defined the resource-group library crate name for consistent project integration.
  * Updated resource-group test modules to use the standardized crate namespace.

* **Tests**
  * Preserved existing test behavior and assertions while aligning imports with the updated library name.

### DIFF
--- a/gears/system/resource-group/resource-group/Cargo.toml
+++ b/gears/system/resource-group/resource-group/Cargo.toml
@@ -12,6 +12,9 @@ keywords = ["constructorfabric", "cf-gears", "resource-group"]
 categories = ["web-programming"]
 metadata.docs.rs.all-features = true
 
+[lib]
+name = "resource_group"
+
 [lints]
 workspace = true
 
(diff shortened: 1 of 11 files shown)


ffedoroff (approved):
looks good

PR #4167, opened 2026-06-30
feat(toolkit): add REST healthcheck infrastructure

refur-nfn:
- Concurrent healthcheck execution with timeout protection
- HealthcheckRegistry for registering and running per-gear probes
- Message sanitization to prevent secret leaks in responses
- API Gateway integration for /health and /readyz endpoints
- Comprehensive test coverage with panic/timeout/degradation scenarios

## Summary by CodeRabbit

* **New Features**
  * Added readiness-focused `/health`, `/healthz`, and `/readyz` endpoints with JSON status, RFC3339 timestamp, and per-component details.
  * Introduced a shared readiness healthcheck registry and a new REST readiness hook for services to publish checks.
  * Ensured health endpoints remain reachable at root paths even when the gateway is configured with a non-root prefix.
* **Bug Fixes**
  * Health responses now consistently reflect aggregated healthy/degraded/unhealthy readiness with correct `200` vs `503`.
* **Tests**
  * Added integration tests covering endpoint behavior, aggregation semantics, prefix handling, and readiness hook default/override behavior.

### DIFF
--- a/gears/system/api-gateway/src/gear.rs
+++ b/gears/system/api-gateway/src/gear.rs
@@ -3,7 +3,7 @@
 //! Contains the `ApiGateway` gear struct and its trait implementations.
 
 use async_trait::async_trait;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 
 use arc_swap::ArcSwap;
 use dashmap::DashMap;
@@ -12,7 +12,7 @@ use anyhow::Result;
 use axum::error_handling::HandleErrorLayer;
 use axum::http::Method;
 use axum::middleware::from_fn_with_state;
-use axum::{Router, extract::DefaultBodyLimit, middleware::from_fn, routing::get};
+use axum::{Extension, Router, extract::DefaultBodyLimit, middleware::from_fn, routing::get};
 use parking_lot::Mutex;
 use std::net::SocketAddr;
 use std::time::Duration;
@@ -49,7 +49,7 @@ async fn timeout_to_canonical(err: BoxError) -> axum::response::Response {
 
 use authn_resolver_sdk::AuthNResolverClient;
 
-use crate::config::ApiGatewayConfig;
+use crate::config::{ApiGatewayConfig, HealthServeMode};
 use crate::middleware::auth;
 use toolkit_security::SecurityContext;
 use toolkit_security::constants::{DEFAULT_SUBJECT_ID, DEFAULT_TENANT_ID};
@@ -77,6 +77,12 @@ pub struct ApiGateway {
     pub(crate) final_router: Mutex<Option<axum::Router>>,
     // AuthN Resolver client (resolved during init, None when auth_disabled)
     pub(crate) authn_client: Mutex<Option<Arc<dyn AuthNResolverClient>>>,
+    // Readiness registry, set once from `rest_prepare`; `OnceLock` = lock-free reads.
+    pub(crate) healthcheck_registry: OnceLock<Arc<toolkit::RestHealthcheckRegistry>>,
+    // Built-once standalone health router. Served on the separate health listener in
+    // `separate`/`both` mode and merged onto the main router in `main`/`both` mode. Cached
+    // so repeat `health_router()`/`build_health_router()` calls are cheap.
+    pub(crate) health_router: OnceLock<axum::Router>,
 
     // Duplicate detection (per (method, path) and per handler id)
     pub(crate) registered_routes: DashMap<(Method, String), ()>,
@@ -92,24 +98,84 @@ impl Default for ApiGateway {
             router_cache: RouterCache::new(default_router),
             final_router: Mutex::new(None),
             authn_client: Mutex::new(None),
+            healthcheck_registry: OnceLock::new(),
+            health_router: OnceLock::new(),
             registered_routes: DashMap::new(),
             registered_handlers: DashMap::new(),
         }
     }
 }
 
+// Built-in health-probe paths, shared by route registration and auth policy so they can't drift.
+const HEALTH_DETAIL_PATH: &str = "/health";
+const HEALTHZ_PATH: &str = "/healthz";
+const READYZ_PATH: &str = "/readyz";
+
 impl ApiGateway {
-    fn apply_prefix_nesting(mut router: Router, prefix: &str) -> Router {
+    /// Nest `router` under `prefix`. Returns `router` unchanged when `prefix` is empty.
+    ///
+    /// Auth matching is keyed on unprefixed `OperationBuilder` paths, so the caller must
+    /// apply `router`'s middleware before this strips/adds the prefix via `nest()`.
+    fn apply_prefix(router: Router, prefix: &str) -> Router {
         if prefix.is_empty() {
-            return router;
+            router
+        } else {
+            Router::new().nest(prefix, router)
         }
+    }
 
-        let top = Router::new()
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
+    /// Standalone health-probe router (`/health`, `/healthz`, `/readyz`): all public, no
+    /// gateway middleware, no auth, and NOT part of the `OpenAPI` document. In `separate`/`both`
+    /// mode the framework binds this on the separate health listener (`health.bind_addr`) from
+    /// [`serve`](Self::serve); it is also exposed for embedders that want to serve it themselves.
+    ///
+    /// # Errors
+    /// Returns an error if [`rest_prepare`](toolkit::contracts::ApiGatewayCapability::rest_prepare)
+    /// has not run yet (the healthcheck registry is unset).
+    pub fn health_router(&self) -> Result<Router> {
+        let hc_registry = self.healthcheck_registry.get().cloned().ok_or_else(|| {
+            anyhow::anyhow!(
+                "healthcheck_registry not set; call rest_prepare before health_router()"
+            )
+        })?;
+        let config = self.get_cached_config();
+        Ok(self.build_health_router(
+            hc_registry,
+            Duration::from_millis(config.healthcheck_timeout_ms),
+        ))
+    }
 
-        router = Router::new().nest(prefix, router);
-        top.merge(router)
+    /// Build (or return the cached) standalone health router. Deps (`hc_registry`,
+    /// `healthcheck_timeout`) are fixed after `rest_prepare`, so the first build is
+    /// authoritative; later calls reuse the clone.
+    fn build_health_router(
+        &self,
+        hc_registry: Arc<toolkit::RestHealthcheckRegistry>,
+        healthcheck_timeout: Duration,
+    ) -> Router {
+        if let Some(cached) = self.health_router.get() {
+            return cached.clone();
+        }
+        let built = Self::health_routes(hc_registry, healthcheck_timeout);
+        // First build wins; a concurrent racer's `set` is a no-op (builds are sequential
+        // during lifecycle setup, so this is a belt-and-braces guard).
+        drop(self.health_router.set(built.clone()));
+        built
+    }
+
+    /// Health-probe routes as plain Axum routes (no `OperationBuilder`, so they never enter
+    /// the `OpenAPI` registry). All three are public; deps injected via `Extension`. Merged
+    /// onto the main router in `main`/`both` mode and used standalone on the separate listener.
+    fn health_routes(
+        hc_registry: Arc<toolkit::RestHealthcheckRegistry>,
+        healthcheck_timeout: Duration,
+    ) -> Router {
+        Router::new()
+            .route(HEALTH_DETAIL_PATH, get(web::health_detail))
+            .route(HEALTHZ_PATH, get(|| async { "ok" }))
+            .route(READYZ_PATH, get(web::readyz_check))
+            .layer(Extension(hc_registry))
+            .layer(Extension(web::HealthcheckTimeout(healthcheck_timeout)))
     }
 
     /// Create a new `ApiGateway` instance with the given configuration
@@ -122,6 +188,8 @@ impl ApiGateway {
             router_cache: RouterCache::new(default_router),
             final_router: Mutex::new(None),
             authn_client: Mutex::new(None),
+            healthcheck_registry: OnceLock::new(),
+            health_router: OnceLock::new(),
             registered_routes: DashMap::new(),
             registered_handlers: DashMap::new(),
         }
@@ -157,12 +225,22 @@ impl ApiGateway {
         let mut authenticated_routes = std::collections::HashSet::new();
         let mut public_routes = std::collections::HashSet::new();
 
-        // Always mark built-in health check routes as public
-        public_routes.insert((Method::GET, "/health".to_owned()));
-        public_routes.insert((Method::GET, "/healthz".to_owned()));
-
         public_routes.insert((Method::GET, "/docs".to_owned()));
         public_routes.insert((Method::GET, "/openapi.json".to_owned()));
+
+        // In main/both mode the health probes are merged onto the main router *before* the
+        // middleware stack (see `rest_finalize`), so the auth layer resolves them: mark them
+        // explicitly public so no bearer token is ever required. Auth matches on the unprefixed
+        // path, so these keys stay unprefixed even when `prefix_path` is set.
+        let config = self.get_cached_config();
+        if matches!(
+            config.health.serve,
+            HealthServeMode::Main | HealthServeMode::Both
+        ) {
+            for path in [HEALTHZ_PATH, READYZ_PATH, HEALTH_DETAIL_PATH] {
+                public_routes.insert((Method::GET, path.to_owned()));
+            }
+        }
 
         for spec in &self.openapi_registry.operation_specs {
             let spec = spec.value();
@@ -178,7 +256,6 @@ impl ApiGateway {
             }
         }
 
-        let config = self.get_cached_config();
         let requirements_count = authenticated_routes.len();
         let public_routes_count = public_routes.len();
 
@@ -256,7 +333,11 @@ impl ApiGateway {
         // 14) Propagate MatchedPath to response extensions (route_layer — innermost).
         // This copies MatchedPath from the request (populated by Axum route matching)
         // into the response so outer layer() middleware (metrics) can read it.
-        router = router.route_layer(from_fn(middleware::http_metrics::propagate_matched_path));
+        // `route_layer` panics on a routeless router — reachable when no REST provider
+        // has registered anything yet.
+        if router.has_routes() {
+            router = router.route_layer(from_fn(middleware::http_metrics::propagate_matched_path));
+        }
 
         let config = self.get_cached_config();
 
@@ -484,19 +565,14 @@ impl ApiGateway {
         }
 
         tracing::debug!("Building new router (standalone/fallback mode)");
-        // In standalone mode (no REST pipeline), register both health endpoints here.
-        // In normal operation, rest_prepare() registers these instead.
-        let mut router = Router::new()
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
-
-        // Apply all middleware layers including auth, above the router
+        // No "main" routes here — the empty router is tolerated (`has_routes` guard).
+        // Health probes are not part of this router; see `health_router`.
         let authn_client = self.authn_client.lock().clone();
-        router = self.apply_middleware_stack(router, authn_client)?;
+        let router = self.apply_middleware_stack(Router::new(), authn_client)?;
 
         let config = self.get_cached_config();
         let prefix = Self::normalize_prefix_path(&config.prefix_path)?;
-        router = Self::apply_prefix_nesting(router, &prefix);
+        let router = Self::apply_prefix(router, &prefix);
 
         // Cache the built router for future use
         self.router_cache.store(router.clone());
@@ -527,6 +603,29 @@ impl ApiGateway {
             .map_err(|e| anyhow::anyhow!("Invalid bind address '{bind_addr}': {e}"))
     }
 
+    /// Resolve the separate health listener address, if the config demands one.
+    ///
+    /// `Ok(None)` for `serve = main`. For `separate`/`both`, `health.bind_addr` must be present
+    /// and parseable — used both to fail fast at `init` and to bind at `serve`.
+    ///
+    /// # Errors
+    /// Returns an error if `serve` needs a separate listener but `health.bind_addr` is missing
+    /// or not a valid socket address.
+    fn health_bind_addr(cfg: &ApiGatewayConfig) -> anyhow::Result<Option<SocketAddr>> {
+        match cfg.health.serve {
+            HealthServeMode::Main => Ok(None),
+            HealthServeMode::Separate | HealthServeMode::Both => {
+                let raw = cfg.health.bind_addr.as_deref().ok_or_else(|| {
+                    anyhow::anyhow!(
+                        "health.serve = {:?} requires health.bind_addr to be set",
+                        cfg.health.serve
+                    )
+                })?;
+                Ok(Some(Self::parse_bind_address(raw)?))
+            }
+        }
+    }
+
     /// Get the finalized router or build a default one.
     fn get_or_build_router(self: &Arc<Self>) -> anyhow::Result<Router> {
         let stored = { self.final_router.lock().take() };
@@ -553,27 +652,53 @@ impl ApiGateway {
         let addr = Self::parse_bind_address(&cfg.bind_addr)?;
         let router = self.get_or_build_router()?;
 
-        // Bind the socket, only now consider the service "ready"
+        // Bind the main socket.
         let listener = tokio::net::TcpListener::bind(addr).await?;
         tracing::info!("HTTP server bound on {}", addr);
+
+        // Bind the separate health listener (if `serve` = separate|both) BEFORE signalling
+        // ready, so readiness reflects every listener the pod must accept traffic on.
+        let health_bound = self.bind_health_listener(&cfg).await?;
+
         ready.notify(); // Starting -> Running
 
-        // Graceful shutdown on cancel
-        let shutdown = {
-            let cancel = cancel.clone();
-            async move {
-                cancel.cancelled().await;
-                tracing::info!("HTTP server shutting down gracefully (cancellation)");
-            }
-        };
-
-        axum::serve(
+        let main_server = axum::serve(
             listener,
             router.into_make_service_with_connect_info::<SocketAddr>(),
         )
-        .with_graceful_shutdown(shutdown)
-        .await
-        .map_err(|e| anyhow::anyhow!(e))
+        .with_graceful_shutdown(Self::shutdown_signal(cancel.clone(), "HTTP server"));
+
+        // Both listeners share the runtime cancellation token, so shutdown fans out to both.
+        if let Some((health_listener, health_router)) = health_bound {
+            let health_server = axum::serve(health_listener, health_router.into_make_service())
+                .with_graceful_shutdown(Self::shutdown_signal(cancel, "health server"));
+            tokio::try_join!(
+                async { main_server.await.map_err(|e| anyhow::anyhow!(e)) },
+                async { health_server.await.map_err(|e| anyhow::anyhow!(e)) },
+            )?;
+            Ok(())
+        } else {
+            main_server.await.map_err(|e| anyhow::anyhow!(e))
+        }
+    }
+
+    /// Bind the separate health listener when `serve` = separate|both, else `None`.
+    async fn bind_health_listener(
+        &self,
+        cfg: &ApiGatewayConfig,
+    ) -> anyhow::Result<Option<(tokio::net::TcpListener, Router)>> {
+        let Some(health_addr) = Self::health_bind_addr(cfg)? else {
+            return Ok(None);
+        };
+        let health_listener = tokio::net::TcpListener::bind(health_addr).await?;
+        tracing::info!("health server bound on {}", health_addr);
+        Ok(Some((health_listener, self.health_router()?)))
+    }
+
+    /// Future that resolves when `cancel` fires, logging the graceful-shutdown of `name`.
+    async fn shutdown_signal(cancel: CancellationToken, name: &'static str) {
+        cancel.cancelled().await;
+        tracing::info!("{name} shutting down gracefully (cancellation)");
     }
 
     /// Check if `handler_id` is already registered (returns true if duplicate)
@@ -683,6 +808,10 @@ impl toolkit::Gear for ApiGateway {
         // tower-http would otherwise assert-panic during eager router
         // layering — a startup crash-loop with no pointer at the config.
         crate::cors::validate_cors_config(&cfg).map_err(|e| anyhow::anyhow!(e))?;
+        // Fail fast when health.serve needs a separate listener but health.bind_addr is
+        // missing or unparseable, rather than crashing later in serve() after other gears
+        // have started.
+        Self::health_bind_addr(&cfg)?;
         self.config.store(Arc::new(cfg.clone()));
 
         debug!(
@@ -712,16 +841,15 @@ impl toolkit::contracts::ApiGatewayCapability for ApiGateway {
         &self,
         _ctx: &toolkit::context::GearCtx,
         router: axum::Router,
+        hc_registry: Arc<toolkit::RestHealthcheckRegistry>,
     ) -> anyhow::Result<axum::Router> {
-        // Add health check endpoints:
-        // - /health: detailed JSON response with status and timestamp
-        // - /healthz: simple "ok" liveness probe (Kubernetes-style)
-        let router = router
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
+        // Store for use when health routes are added in rest_finalize. A second set
+        // means rest_prepare ran twice — a lifecycle bug; fail fast rather than mask it.
+        if self.healthcheck_registry.set(hc_registry).is_err() {
+            anyhow::bail!("healthcheck_registry already set; rest_prepare called more than once");
+        }
 
-        // You may attach global middlewares here (trace, compression, cors), but do not start server.
-        tracing::debug!("REST host prepared base router with health check endpoints");
+        tracing::debug!("REST host prepared base router");
         Ok(router)
     }
 
@@ -729,6 +857,7 @@ impl toolkit::contracts::ApiGatewayCapability for ApiGateway {
         &self,
         _ctx: &toolkit::context::GearCtx,
         mut router: axum::Router,
+        hc_registry: Arc<toolkit::RestHealthcheckRegistry>,
     ) -> anyhow::Result<axum::Router> {
         let config = self.get_cached_config();
 
@@ -736,15 +865,33 @@ impl toolkit::contracts::ApiGatewayCapability for ApiGateway {
             router = self.add_openapi_routes(router)?;
         }
 
-        // Apply middleware stack (including auth) to the final router
+        // Health probes (`main`/`both` mode): merge them onto the main router BEFORE the
+        // middleware stack, so they share the gateway's unified surface (request id, tracing,
+        // metrics, error mapping, ...). They are marked public in the route policy (see
+        // `build_route_policy_from_specs`), so the auth layer lets them through without a bearer
+        // token. Like every other route they inherit `prefix_path` via the nesting below.
+        // `separate` mode omits them here; `serve()` binds them on the dedicated health listener.
+        if matches!(
+            config.health.serve,
+            HealthServeMode::Main | HealthServeMode::Both
+        ) {
+            let health = Self::health_routes(
+                hc_registry,
+                Duration::from_millis(config.healthcheck_timeout_ms),
+            );
+            router = router.merge(health);
+        }
+
+        // Middleware on the main router before nesting (auth matching keyed on
+        // unprefixed OperationBuilder paths; layers run before nest() strips the prefix).
         tracing::debug!("Applying middleware stack to finalized router");
         let authn_client = self.authn_client.lock().clone();
         router = self.apply_middleware_stack(router, authn_client)?;
 
         let prefix = Self::normalize_prefix_path(&config.prefix_path)?;
-        router = Self::apply_prefix_nesting(router, &prefix);
+        router = Self::apply_prefix(router, &prefix);
 
-        // Keep the finalized router to be used by `serve()`
+        // Keep the finalized router to be used by `serve()`.
         *self.final_router.lock() = Some(router.clone());
 
         tracing::info!("REST host finalized router with OpenAPI endpoints and auth middleware");
--- a/gears/system/api-gateway/src/web.rs
+++ b/gears/system/api-gateway/src/web.rs
@@ -1,10 +1,19 @@
 use axum::{
+    extract::Extension,
     http::StatusCode,
-    response::{Html, Json},
+    response::{Html, IntoResponse, Json, Response},
     routing::{MethodRouter, get},
 };
 use chrono::{SecondsFormat, Utc};
-use serde_json::{Value, json};
+use serde_json::json;
+use std::sync::Arc;
+use std::time::Duration;
+use toolkit::RestHealthcheckRegistry;
+
+/// Per-check timeout (from `ApiGatewayConfig::healthcheck_timeout_ms`), injected as an
+/// `Extension`. A newtype avoids colliding with any other `Duration` extension.
+#[derive(Clone, Copy)]
+pub struct HealthcheckTimeout(pub Duration);
 
 /// Returns a 501 Not Implemented handler for operations without implementations
 #[allow(dead_code)]
@@ -20,11 +29,39 @@ pub fn placeholder_handler_501() -> MethodRouter {
     })
 }
 
-pub async fn health_check() -> Json<Value> {
-    Json(json!({
-        "status": "healthy",
-        "timestamp": Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true)
-    }))
+/// `GET /health`: runs all checks, returns JSON with per-component detail. Public (no auth) in
+/// every serving mode; the per-component detail is protected by network placement, not auth,
+/// and check messages are sanitized. 200 (Healthy/Degraded), 503 (Unhealthy).
+pub async fn health_detail(
+    Extension(registry): Extension<Arc<RestHealthcheckRegistry>>,
+    Extension(timeout): Extension<HealthcheckTimeout>,
+) -> Response {
+    let report = registry.report(timeout.0).await;
+    let status_code = status_for_report(report.is_ready());
+    let body = Json(json!({
+        "status": report.status,
+        "timestamp": Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true),
+        "components": report.components,
+    }));
+    (status_code, body).into_response()
+}
+
+/// `GET /readyz`: runs all checks, public, status code only (no detail).
+/// 200 (Healthy/Degraded), 503 (Unhealthy).
+pub async fn readyz_check(
+    Extension(registry): Extension<Arc<RestHealthcheckRegistry>>,
+    Extension(timeout): Extension<HealthcheckTimeout>,
+) -> StatusCode {
+    let report = registry.report(timeout.0).await;
+    status_for_report(report.is_ready())
+}
+
+fn status_for_report(report_ready: bool) -> StatusCode {
+    if report_ready {
+        StatusCode::OK
+    } else {
+        StatusCode::SERVICE_UNAVAILABLE
+    }
 }
 
 #[cfg(not(feature = "embed_elements"))]
--- a/gears/system/api-gateway/tests/health_endpoints.rs
+++ b/gears/system/api-gateway/tests/health_endpoints.rs
@@ -1,0 +1,504 @@
+#![allow(clippy::unwrap_used, clippy::expect_used)]
+
+//! Integration tests for health endpoints (`/health`, `/healthz`, `/readyz`).
+//!
+//! Health endpoints are unauthenticated in every serving mode. In `main`/`both` mode they are
+//! merged onto the main gateway router at stable, unprefixed paths and outside the auth layer;
+//! in `separate`/`both` mode the standalone health router (`health_router()`) is served on a
+//! dedicated listener.
+
+use async_trait::async_trait;
+use authn_resolver_sdk::{
+    AuthNResolverClient, AuthNResolverError, AuthenticationResult, ClientCredentialsRequest,
+};
+use axum::{
+    Router,
+    body::Body,
+    http::{Request, StatusCode, header},
+};
+use serde_json::json;
+use std::sync::Arc;
+use tokio_util::sync::CancellationToken;
+use toolkit::{
+    ClientHub, GearCtx, Healthcheck, HealthcheckResult, RestHealthcheckRegistry,
+    contracts::ApiGatewayCapability,
+};
+use tower::ServiceExt;
+use uuid::Uuid;
+
+// ---------- helpers ----------
+
+/// A `ConfigProvider` backed by a single JSON blob, keyed by gear name.
+struct JsonConfigProvider(serde_json::Value);
+
+impl toolkit::config::ConfigProvider for JsonConfigProvider {
+    fn get_gear_config(&self, gear: &str) -> Option<&serde_json::Value> {
+        self.0.get(gear)
+    }
+}
+
+/// Mock `AuthN` client that rejects every token. Registered for auth-enabled tests to prove
+/// health endpoints never reach the auth layer (a real request would be rejected).
+struct RejectingAuthN;
+
+#[async_trait]
+impl AuthNResolverClient for RejectingAuthN {
+    async fn authenticate(&self, _token: &str) -> Result<AuthenticationResult, AuthNResolverError> {
+        Err(AuthNResolverError::Unauthorized(
+            "health tests: auth must never be invoked for health endpoints".to_owned(),
+        ))
+    }
+
+    async fn exchange_client_credentials(
+        &self,
+        _request: &ClientCredentialsRequest,
+    ) -> Result<AuthenticationResult, AuthNResolverError> {
+        Err(AuthNResolverError::Internal("not implemented".to_owned()))
+    }
+}
+
+/// `init` + `rest_prepare` + `rest_finalize` the gateway with the given `api-gateway.config`
+/// object, an optional `AuthN` client, and a registry populated by `setup`. Returns the
+/// finalized main router, or the first error (used by fail-fast config tests).
+async fn finalize_main_router(
+    config_obj: serde_json::Value,
+    authn: Option<Arc<dyn AuthNResolverClient>>,
+    setup: impl FnOnce(&RestHealthcheckRegistry),
+) -> anyhow::Result<Router> {
+    use api_gateway::ApiGateway;
+    use toolkit::Gear;
+
+    let registry = Arc::new(RestHealthcheckRegistry::new());
+    setup(&registry);
+
+    let config = json!({ "api-gateway": { "config": config_obj } });
+    let hub = Arc::new(ClientHub::new());
+    if let Some(client) = authn {
+        hub.register::<dyn AuthNResolverClient>(client);
+    }
+    let ctx = GearCtx::new(
+        "api-gateway",
+        Uuid::new_v4(),
+        Arc::new(JsonConfigProvider(config)),
+        hub,
+        CancellationToken::new(),
+    );
+
+    let gw = ApiGateway::default();
+    gw.init(&ctx).await?;
+    let router = gw.rest_prepare(&ctx, Router::new(), registry.clone())?;
+    gw.rest_finalize(&ctx, router, registry)
+}
+
+/// Build the standalone health router (`rest_prepare` + `health_router`). This is the router
+/// the framework serves on the separate health listener in `separate`/`both` mode.
+async fn build_standalone_health_router(setup: impl FnOnce(&RestHealthcheckRegistry)) -> Router {
+    use api_gateway::ApiGateway;
+    use toolkit::Gear;
+
+    let registry = Arc::new(RestHealthcheckRegistry::new());
+    setup(&registry);
+
+    let config =
+        json!({ "api-gateway": { "config": { "bind_addr": "0.0.0.0:0", "auth_disabled": true } } });
+    let ctx = GearCtx::new(
+        "api-gateway",
+        Uuid::new_v4(),
+        Arc::new(JsonConfigProvider(config)),
+        Arc::new(ClientHub::new()),
+        CancellationToken::new(),
+    );
+
+    let gw = ApiGateway::default();
+    gw.init(&ctx).await.expect("init failed");
+    let _base = gw
+        .rest_prepare(&ctx, Router::new(), registry)
+        .expect("rest_prepare failed");
+    gw.health_router().expect("health_router failed")
+}
+
+async fn get(router: Router, path: &str) -> axum::http::Response<axum::body::Body> {
+    router
+        .oneshot(Request::builder().uri(path).body(Body::empty()).unwrap())
+        .await
+        .unwrap()
+}
+
+async fn get_with_bearer(
+    router: Router,
+    path: &str,
+    token: &str,
+) -> axum::http::Response<axum::body::Body> {
+    router
+        .oneshot(
+            Request::builder()
+                .uri(path)
+                .header(header::AUTHORIZATION, format!("Bearer {token}"))
+                .body(Body::empty())
+                .unwrap(),
+        )
+        .await
+        .unwrap()
+}
+
+async fn body_json(resp: axum::http::Response<axum::body::Body>) -> serde_json::Value {
+    let bytes = axum::body::to_bytes(resp.into_body(), usize::MAX)
+        .await
+        .unwrap();
+    serde_json::from_slice(&bytes).unwrap()
+}
+
+const HEALTH_PATHS: [&str; 3] = ["/healthz", "/readyz", "/health"];
+
+// ---------- check stubs ----------
+
+struct HealthyCheck;
+#[async_trait]
+impl Healthcheck for HealthyCheck {
+    fn name(&self) -> &'static str {
+        "always-healthy"
+    }
+}
+
+struct DegradedCheck;
+#[async_trait]
+impl Healthcheck for DegradedCheck {
+    fn name(&self) -> &'static str {
+        "always-degraded"
+    }
+    async fn check(&self) -> HealthcheckResult {
+        HealthcheckResult::degraded("cache warming up")
+    }
+}
+
+struct UnhealthyCheck;
+#[async_trait]
+impl Healthcheck for UnhealthyCheck {
+    fn name(&self) -> &'static str {
+        "always-unhealthy"
+    }
+    async fn check(&self) -> HealthcheckResult {
+        HealthcheckResult::unhealthy("database unreachable")
+    }
+}
+
+struct CodedUnhealthyCheck;
+#[async_trait]
+impl Healthcheck for CodedUnhealthyCheck {
+    fn name(&self) -> &'static str {
+        "coded-unhealthy"
+    }
+    async fn check(&self) -> HealthcheckResult {
+        HealthcheckResult::unhealthy("database unreachable").with_code("db_unreachable")
+    }
+}
+
+// ---------- default (main) mode: health on the main router, no auth ----------
+
+#[tokio::test]
+async fn default_main_serves_all_health_endpoints() {
+    let router = finalize_main_router(
+        json!({ "bind_addr": "0.0.0.0:0", "auth_disabled": true }),
+        None,
+        |reg| reg.register("my-gear", Arc::new(HealthyCheck)),
+    )
+    .await
+    .expect("finalize failed");
+
+    assert_eq!(
+        get(router.clone(), "/healthz").await.status(),
+        StatusCode::OK
+    );
+    assert_eq!(
+        get(router.clone(), "/readyz").await.status(),
+        StatusCode::OK
+    );
+
+    let resp = get(router, "/health").await;
+    assert_eq!(resp.status(), StatusCode::OK);
+    let body = body_json(resp).await;
+    assert_eq!(body["status"], "healthy");
+    assert_eq!(body["components"][0]["gear"], "my-gear");
+}
+
+#[tokio::test]
+async fn main_readyz_returns_503_when_unhealthy() {
+    let router = finalize_main_router(
+        json!({ "bind_addr": "0.0.0.0:0", "auth_disabled": true }),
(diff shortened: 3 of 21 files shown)


libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// Run all registered healthchecks concurrently and return an aggregate report.
+    ///
+    /// Each check runs in its own `tokio::task::spawn` so panics are caught.
+    /// Checks that exceed `timeout_per_check` are cancelled and reported as
+    /// [`HealthcheckStatus::Unhealthy`].
+    ///
+    /// Results are cached for [`REPORT_CACHE_TTL`] so bursty probes reuse the
+    /// most recent report instead of re-running every check on each request.
+    pub async fn report(&self, timeout_per_check: Duration) -> HealthcheckReport {
+        if let Some((ts, cached)) = self.cached_report.read().as_ref()
```
MikeFalcon77: **[RUST-CONC-001] HIGH — TOCTOU double fan-out in `report()` cache**

The read lock is acquired and released on line 242, then a write lock is acquired at the end of the method. Any N concurrent callers that all observe a stale cache within that window each independently spawn a full fan-out of all checks, running every healthcheck N times instead of once.

```rust
// Current: read-lock dropped before awaits; concurrent callers all miss cache
if let Some((ts, cached)) = self.cached_report.read().as_ref()
    && ts.elapsed() < REPORT_CACHE_TTL
{
    return cached.clone();
}
// ... all callers that passed the read-check now run all checks ...
*self.cached_report.write() = Some((Instant::now(), report.clone()));
```

**Fix:** Use a `tokio::sync::Mutex` held across the full async body so that concurrent callers queue behind the in-flight computation:

```rust
// one write-lock gates the whole body
let mut guard = self.cached_report.lock().await;
if let Some((ts, cached)) = guard.as_ref() {
    if ts.elapsed() < REPORT_CACHE_TTL {
        return cached.clone();
    }
}
let report = run_all_checks(&self.entries, timeout_per_check).await;
*guard = Some((Instant::now(), report.clone()));
report
```

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+///
+/// - `check` must not panic; panics are caught by [`RestHealthcheckRegistry`] and
+///   converted to [`HealthcheckStatus::Unhealthy`].
+/// - `check` must complete within the configured per-check timeout
+///   (default 500 ms); timeouts are converted to [`HealthcheckStatus::Unhealthy`].
+/// - `check` must never leak secrets, passwords, DSNs, or stack traces in its
+///   message; the registry sanitises messages but defence-in-depth is preferred.
+#[async_trait]
+pub trait Healthcheck: Send + Sync + 'static {
+    fn name(&self) -> &'static str {
```
MikeFalcon77: **[RUST-SEC-001] HIGH — `type_name::<Self>()` exposed in unauthenticated `/health` JSON**

The default `name()` returns the full Rust module path (e.g., `api_gateway::module::MyCheck`). This value is serialized verbatim into `HealthcheckComponentReport.check` and served on the unauthenticated `/health` endpoint, disclosing internal module structure to any caller.

```rust
// Current — emits "my_crate::module::DatabaseCheck" to external callers:
fn name(&self) -> &'static str {
    std::any::type_name::<Self>()
}
```

**Fix:** Remove the default and force implementors to supply an explicit, intentionally public name; or strip the module prefix before serializing:

```rust
// No default — implementors must supply a human-readable name:
fn name(&self) -> &'static str;
```

gears/system/api-gateway/src/gear.rs
```diff
@@ -706,17 +741,24 @@ impl toolkit::Gear for ApiGateway {
...
-        // - /healthz: simple "ok" liveness probe (Kubernetes-style)
-        let router = router
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
-
-        // You may attach global middlewares here (trace, compression, cors), but do not start server.
+        // Retrieve the healthcheck registry created in run_rest_phase before this call.
+        // Falls back to an empty registry so standalone tests still compile.
+        let hc_registry = ctx
+            .client_hub()
```
MikeFalcon77: **[RUST-ERR-001 + RUST-NO-002] HIGH — Discarded error + silent fallback masks misconfiguration in `rest_prepare`**

`unwrap_or_else(|_| ...)` discards the original `ClientHub::get` error and silently substitutes an empty registry. If `run_rest_phase` fails to register the registry (e.g., due to ordering change or a future refactor), health checks will silently be unwired with no error or log.

```rust
// Current — error thrown away, empty registry substituted silently:
let hc_registry = ctx
    .client_hub()
    .get::<toolkit::RestHealthcheckRegistry>()
    .unwrap_or_else(|_| Arc::new(toolkit::RestHealthcheckRegistry::new()));
```

**Fix:** Propagate or at minimum log the original error:

```rust
let hc_registry = ctx
    .client_hub()
    .get::<toolkit::RestHealthcheckRegistry>()
    .inspect_err(|e| tracing::warn!("healthcheck registry not found in ClientHub: {e:#}"))
    .unwrap_or_else(|_| Arc::new(toolkit::RestHealthcheckRegistry::new()));
```

Or return an error and let the runtime fail fast:

```rust
let hc_registry = ctx
    .client_hub()
    .get::<toolkit::RestHealthcheckRegistry>()
    .context("RestHealthcheckRegistry not registered in ClientHub before rest_prepare")?;
```

gears/system/api-gateway/src/gear.rs
```diff
@@ -735,10 +777,16 @@ impl toolkit::contracts::ApiGatewayCapability for ApiGateway {
...
         tracing::debug!("Applying middleware stack to finalized router");
         let authn_client = self.authn_client.lock().clone();
-        router = self.apply_middleware_stack(router, authn_client)?;
+        router = self.apply_middleware_stack(router, authn_client.clone())?;
+
+        let hc_registry = self
+            .healthcheck_registry
+            .lock()
+            .clone()
+            .unwrap_or_else(|| Arc::new(toolkit::RestHealthcheckRegistry::new()));
```
MikeFalcon77: **[RUST-NO-002] MEDIUM — Silent empty registry fallback in `rest_finalize` hides broken lifecycle**

`rest_finalize` creates a fresh empty `RestHealthcheckRegistry` when `self.healthcheck_registry` is `None`, which means if `rest_prepare` was never called the finalized router will silently serve `/health` and `/readyz` with zero checks — no error, no warning.

This field is `None` only during broken lifecycle ordering. The fallback conceals the bug instead of surfacing it.

**Fix:** Fail fast or warn:

```rust
let hc_registry = self
    .healthcheck_registry
    .lock()
    .clone()
    .unwrap_or_else(|| {
        tracing::warn!("healthcheck_registry is None in rest_finalize — rest_prepare may not have been called");
        Arc::new(toolkit::RestHealthcheckRegistry::new())
    });
```

Or return `Err` if `None` is considered a programmer error here.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+            r.iter().map(|e| (e.gear, e.check.clone())).collect()
+        };
+
+        let report = if entries.is_empty() {
+            HealthcheckReport {
+                status: HealthcheckAggregateStatus::Healthy,
+                components: vec![],
+            }
+        } else {
+            let handles: Vec<_> = entries
```
MikeFalcon77: **[RUST-PERF-001] MEDIUM — Double-spawn per check: unnecessary task overhead**

`report()` wraps each `run_one_check` call in an outer `tokio::spawn`, and `run_one_check` itself spawns an inner `tokio::spawn` for the check. Every registered check therefore allocates two tasks — the outer one adds no value since `run_one_check` doesn't panic and the inner spawn already catches check panics via `JoinError`.

```rust
// Outer spawn in report() — unnecessary:
tokio::spawn(run_one_check(gear, check, timeout_per_check))

// Inner spawn in run_one_check — this is the one that catches panics:
let mut handle = tokio::spawn(async move { check.check().await });
```

**Fix:** Remove the outer `tokio::spawn` in `report()` and drive `run_one_check` directly via `futures::future::join_all` or `FuturesUnordered`; keep only the inner spawn for panic isolation:

```rust
let futures: Vec<_> = entries
    .into_iter()
    .map(|(gear, check)| run_one_check(gear, check, timeout_per_check))
    .collect();
let components = futures::future::join_all(futures).await;
```

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+}
+
+impl RestHealthcheckRegistry {
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Register a healthcheck for the given gear.
+    pub fn register(&self, gear: &'static str, check: Arc<dyn Healthcheck>) {
```
MikeFalcon77: **[RUST-API-001] MEDIUM — `gear: &'static str` in `register()` is overly restrictive**

Requiring a `'static` string prevents callers from registering checks for dynamically-named gears (e.g., loaded from config, plugin instances, or OOP spawned gears with runtime names). Note that `HealthcheckComponentReport::gear` is already a `String`, so this restriction is artificial.

```rust
// Current — 'static only:
pub fn register(&self, gear: &'static str, check: Arc<dyn Healthcheck>) {
```

**Fix:** Accept any string:

```rust
pub fn register(&self, gear: impl Into<String>, check: Arc<dyn Healthcheck>) {
    self.entries.write().push(RegistryEntry {
        gear: gear.into(),
        check,
    });
}
```

And update `RegistryEntry` accordingly:

```rust
struct RegistryEntry {
    gear: String,   // was &'static str
    check: Arc<dyn Healthcheck>,
}
```

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+}
+
+const MAX_MESSAGE_LEN: usize = 256;
+
+// Deliberately broad, defense-in-depth blocklist for a public/unauthenticated endpoint:
+// entries like "tenant", "select ", "update ", "delete " can match benign operator
+// diagnostics (e.g. "tenant lookup slow") and collapse the message to the generic
+// "health check failed". That false-positive rate is accepted to avoid ever leaking
+// credentials, DSNs, or query fragments to an unauthenticated caller.
+static SUSPICIOUS_SUBSTRINGS: &[&str] = &[
```
MikeFalcon77, line 382: **[RUST-SEC-001] MEDIUM — Blocklist missing common secret-bearing patterns**

Several common patterns that appear in error messages from database drivers, cloud SDKs, and auth libraries are not in the blocklist and will pass through to unauthenticated callers:

- `"api_key"`, `"apikey"`, `"private_key"`, `"private"`
- `"credential"`, `"access_key"`
- `"mongodb://"`, `"redis://"`, `"amqp://"`, `"jdbc:"`
- `"aws_"`, `"client_secret"`

Consider extending the blocklist with these patterns. Alternatively, replace the blocklist approach with an **allowlist** that only passes messages matching a known-safe character set (alphanumeric, spaces, common punctuation, max length) — allowlists cannot have false negatives for novel secret formats:

```rust
fn sanitize_message(msg: &str) -> Option<String> {
    if msg.len() > MAX_MESSAGE_LEN { return None; }
    // Allow only printable ASCII, no control chars or URL-scheme chars
    if msg.bytes().all(|b| b.is_ascii_alphanumeric() || b" .,!?-_:".contains(&b)) {
        Some(msg.to_owned())
    } else {
        None  // caller substitutes generic message
    }
}
```

gears/system/api-gateway/src/web.rs
```diff
@@ -20,11 +25,36 @@ pub fn placeholder_handler_501() -> MethodRouter {
...
-pub async fn health_check() -> Json<Value> {
-    Json(json!({
-        "status": "healthy",
-        "timestamp": Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true)
-    }))
+/// Detailed readiness JSON endpoint (`GET /health`).
+///
+/// Runs all registered REST healthchecks and returns a JSON report.
+/// Returns 200 (Healthy or Degraded), 503 (Unhealthy).
+pub async fn health_detail(registry: Arc<RestHealthcheckRegistry>) -> Response {
```
MikeFalcon77: **[TOOLKIT-REST-001] MEDIUM — Closure capture instead of `Extension<T>` (non-golden-path)**

`health_detail` and `readyz_check` receive the registry via closure capture in `health_routes()` rather than via `Extension<Arc<RestHealthcheckRegistry>>`. The [ToolKit golden path](../libs/toolkit/src/lib.rs) states: *"prefer stateless handlers that receive `Extension<T>`"* — closures that capture environment are the explicitly-discouraged alternative.

```rust
// Current — closure capture (non-golden-path):
.route("/health", get(move || {
    let r = reg_h.clone();
    async move { web::health_detail(r).await }
}))
```

**Fix:** Make `health_detail` and `readyz_check` accept `Extension<Arc<RestHealthcheckRegistry>>` and attach the layer on the router:

```rust
pub(crate) async fn health_detail(
    Extension(registry): Extension<Arc<RestHealthcheckRegistry>>,
) -> Response { ... }

// In health_routes():
fn health_routes(hc_registry: Arc<RestHealthcheckRegistry>) -> Router {
    Router::new()
        .route("/health", get(health_detail))
        .route("/readyz", get(readyz_check))
        .route("/healthz", get(|| async { "ok" }))
        .layer(Extension(hc_registry))
}
```

Also change visibility to `pub(crate)` since these are only called from `gear.rs`.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    #[tokio::test]
+    async fn default_healthcheck_check_returns_healthy() {
+        let r = AlwaysHealthy.check().await;
+        assert_eq!(r.status, HealthcheckStatus::Healthy);
+    }
+
+    #[tokio::test]
+    async fn empty_registry_report_is_healthy_and_ready() {
+        let registry = RestHealthcheckRegistry::new();
```
MikeFalcon77, line 532: **[RUST-TEST-001] MEDIUM — Missing cache-TTL test and mixed-aggregate tests**

Two important behavioral properties have no test coverage:

**1. Cache TTL:** A second `report()` call within REPORT_CACHE_TTL must return the cached result without re-running checks. There is no test for this — a regression in the TOCTOU fix could silently break caching.

**2. Mixed aggregate:** One healthy + one unhealthy = Unhealthy; one healthy + one degraded = Degraded. The short-circuit in `compute_aggregate` (only Unhealthy returns early) needs a mixed-status test to catch regressions.

Please add both cases to `healthcheck.rs` unit tests.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+/// Aggregate readiness status across all registered healthchecks.
+///
+/// Kept distinct from [`HealthcheckStatus`] even though the variants currently
+/// match 1:1 (see `compute_aggregate`): per-check status and overall aggregate
+/// status are different concepts and may diverge (e.g. an aggregate-only
+/// variant for "no checks registered yet") without forcing an API change on
+/// every [`Healthcheck`] implementor.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum HealthcheckAggregateStatus {
```
MikeFalcon77: **[RUST-TYPE-001] LOW — Duplicate enum variants between `HealthcheckStatus` and `HealthcheckAggregateStatus`**

Both enums have identical variants (`Healthy`, `Degraded`, `Unhealthy`). The doc comment acknowledges this and calls out a hypothetical divergence (e.g., a `Starting` variant), but that variant is not present.

Until there is an actual divergence the two-enum design adds conversion boilerplate in `compute_aggregate` with no payoff. Either:
- Add the diverging variant now (e.g., `Starting` for the zero-checks case, currently returned as `Healthy`) to make the split immediately justified, or
- Collapse to one enum until divergence is needed

The empty-registry path currently returns `HealthcheckAggregateStatus::Healthy` — returning `Starting` (or a separate `Ok` state meaning "no checks, assumed healthy") would be semantically cleaner and would justify the two-enum split.

libs/toolkit/src/lib.rs
```diff
@@ -136,6 +136,13 @@ pub use directory::{
     ServiceInstanceInfo,
 };
 
+// REST healthcheck infrastructure
+pub mod healthcheck;
```
MikeFalcon77: **[RUST-MOD-001] LOW — Redundant dual public path for healthcheck types**

Both `pub mod healthcheck` and a flat `pub use healthcheck::{...}` are re-exported, making types reachable via two paths simultaneously:
- `toolkit::healthcheck::RestHealthcheckRegistry`
- `toolkit::RestHealthcheckRegistry`

**Fix:** Keep `pub(crate) mod healthcheck` and expose only the flat `pub use` at crate root, or expose the module without the flat re-exports. Either is fine; having both is the issue.

gears/system/api-gateway/tests/health_endpoints.rs
```diff
@@ -0,0 +1,348 @@
...
+    let resp = get(router, "/health").await;
+    assert_eq!(resp.status(), StatusCode::OK);
+    let body = body_json(resp).await;
+    assert_eq!(body["status"], "degraded");
+}
+
+// ---------- outside prefix ----------
+
+#[tokio::test]
+async fn health_endpoints_accessible_outside_prefix() {
```
MikeFalcon77: **[RUST-TEST-001] LOW — Missing negative-case assertion in prefix isolation test**

`health_endpoints_accessible_outside_prefix` asserts that `/healthz`, `/readyz`, and `/health` return 200 at root paths when the prefix is `/cf`. But it does not assert that the same endpoints are NOT served under the prefix (i.e., `/cf/healthz`, `/cf/readyz`, `/cf/health` should return 404).

Without the negative case, the test does not fully verify the prefix-isolation invariant — a bug that duplicates health routes under the prefix would pass the test.

MikeFalcon77 (changes requested):

MikeFalcon77:
.

gears/system/api-gateway/src/web.rs
```diff
@@ -20,11 +25,36 @@ pub fn placeholder_handler_501() -> MethodRouter {
...
+    let report = registry.report(HEALTHCHECK_TIMEOUT).await;
+    let status_code = status_for_report(report.is_ready());
+    let body = Json(json!({
+        "status": report.status,
+        "timestamp": Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true),
+        "components": report.components,
```
MikeFalcon77, line 44: **[HIGH] TOOLKIT-SEC-002 — `health_detail` exposes component internals to unauthenticated callers**

The `health_detail` handler serializes `report.components` — including internal check names, gear names, and diagnostic messages — into the response body of the unauthenticated public `/health` endpoint without any authorization check.

**Fix:** Either require authentication for the detail `/health` endpoint (keeping `/readyz` as the public shallow probe), or strip the `components` array from the response for unauthenticated callers before serializing.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+        } else {
+            let handles: Vec<_> = entries
+                .into_iter()
+                .map(|(gear, check)| tokio::spawn(run_one_check(gear, check, timeout_per_check)))
+                .collect();
+
+            let mut components = Vec::with_capacity(handles.len());
+            for h in handles {
+                match h.await {
+                    Ok(component) => components.push(component),
```
MikeFalcon77: **[MEDIUM] RUST-PERF-001 — Sequential collection of spawned check tasks**

The `for h in handles { h.await }` loop awaits each task one at a time, serializing the collection phase. Tasks that finish early must wait for the previous `.await` to resolve, adding latency proportional to check count.

**Fix:** Use `futures::future::join_all(handles).await` or `tokio::task::JoinSet` to collect all results concurrently.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    let (status, message) = tokio::select! {
+        result = &mut handle => match result {
+            Ok(r) => (r.status, r.message.as_deref().map(sanitize_message)),
+            Err(_join_err) => (
+                HealthcheckStatus::Unhealthy,
+                Some("health check failed".to_owned()),
+            ),
+        },
+        () = tokio::time::sleep(timeout_per_check) => {
```
MikeFalcon77, line 320: **[MEDIUM] RUST-OBS-001 — No tracing on healthcheck timeout or panic**

`run_one_check` converts check timeouts and panics to `HealthcheckStatus::Unhealthy` without emitting any tracing event. Operators see a 503 response but have no server-side log entry identifying the gear or check that failed.

**Fix:** Add a `tracing::warn!` in the timeout arm and a `tracing::error!` in the join-error arm, both including the gear name and check name as structured fields, before constructing the `HealthcheckComponentReport`.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum HealthcheckAggregateStatus {
+    Healthy,
+    Degraded,
+    Unhealthy,
+}
+
+/// Result returned by a single [`Healthcheck::check`] call.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HealthcheckResult {
```
MikeFalcon77, line 111: **[MEDIUM] RUST-DATA-001 — `HealthcheckResult` has no serialization versioning or contract documentation**

`HealthcheckResult` derives `Serialize`/`Deserialize` with public fields and is re-exported from the crate root. Any future field rename or addition is a silently-breaking published type change.

**Fix:** Add a doc comment stating these field names are part of the stable serialized contract, and consider `#[serde(deny_unknown_fields)]` for forward-compatibility discipline.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+        HealthcheckResult::healthy()
+    }
+}
+
+/// Per-component readiness report entry.
+///
+/// Represents the result of a single gear's healthcheck execution, including
+/// elapsed time and outcome status.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HealthcheckComponentReport {
```
MikeFalcon77, line 183: **[MEDIUM] RUST-DATA-001 — `HealthcheckComponentReport`/`HealthcheckReport` serialized directly to HTTP with no versioning**

Both structs have all fields `pub`, derive `Serialize`/`Deserialize`, are re-exported from the crate root, and are serialized directly by the `/health` endpoint. There is no version discriminator or format contract documentation.

**Fix:** Add doc comments declaring the JSON field names as a stable public contract. Consider an opaque `version` field or `#[serde(deny_unknown_fields)]` if forward-compatibility is required.

gears/system/api-gateway/src/gear.rs
```diff
@@ -160,6 +190,7 @@ impl ApiGateway {
         // Always mark built-in health check routes as public
         public_routes.insert((Method::GET, "/health".to_owned()));
         public_routes.insert((Method::GET, "/healthz".to_owned()));
+        public_routes.insert((Method::GET, "/readyz".to_owned()));
```
MikeFalcon77: **[MEDIUM] TOOLKIT-REST-002 — `/readyz` auth declared via hardcoded set, not `OperationBuilder.public()`**

The public auth status of `/readyz` is declared by inserting it into the `public_routes` set in `build_route_policy_from_specs` rather than at the point of route definition via `OperationBuilder.public()`. Auth declarations should be co-located with route definitions.

**Fix:** Register `/readyz` through `OperationBuilder` with `.public()` so its auth configuration is declared in one place and enforced uniformly.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+        result = &mut handle => match result {
+            Ok(r) => (r.status, r.message.as_deref().map(sanitize_message)),
+            Err(_join_err) => (
+                HealthcheckStatus::Unhealthy,
+                Some("health check failed".to_owned()),
+            ),
+        },
+        () = tokio::time::sleep(timeout_per_check) => {
+            handle.abort();
+            let _result = handle.await;
```
MikeFalcon77: **[LOW] RUST-NO-002 — `let _result = handle.await` silently discards join error after abort**

After `handle.abort()`, the result of `handle.await` is unconditionally discarded. If the join fails for a reason other than cancellation, the error is invisible.

**Fix:** Ignore `Err(e) if e.is_cancelled()` explicitly but log any other `JoinError` as a debug/warn event to surface unexpected task failures.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    #[test]
+    fn healthcheck_result_default_is_healthy() {
+        let r = HealthcheckResult::default();
+        assert_eq!(r.status, HealthcheckStatus::Healthy);
+        assert!(r.message.is_none());
+    }
+
+    #[tokio::test]
+    async fn default_healthcheck_check_returns_healthy() {
```
MikeFalcon77: **[LOW] TEST-QUALITY-1 — Constructor echo: `default_healthcheck_check_returns_healthy`**

`AlwaysHealthy` does not override `check()`, so the test only confirms that the unmodified trait default returns what it is defined to return. This tests nothing that can actually fail.

**Fix:** Replace with a test that exercises a real `check()` implementation whose logic can fail, or remove it and rely on coverage from the report-level tests.

gears/system/api-gateway/src/gear.rs
```diff
@@ -77,6 +77,8 @@ pub struct ApiGateway {
     pub(crate) final_router: Mutex<Option<axum::Router>>,
     // AuthN Resolver client (resolved during init, None when auth_disabled)
     pub(crate) authn_client: Mutex<Option<Arc<dyn AuthNResolverClient>>>,
+    // Readiness healthcheck registry — populated during REST wiring phase
+    pub(crate) healthcheck_registry: Mutex<Option<Arc<toolkit::RestHealthcheckRegistry>>>,
```
MikeFalcon77: **[LOW] RUST-NO-005 — `Mutex<Option<Arc<RestHealthcheckRegistry>>>` for a write-once value**

`healthcheck_registry` is written exactly once during `rest_prepare` and then only read. Wrapping it in a `parking_lot::Mutex<Option<...>>` imposes unnecessary lock overhead on every router build thereafter.

**Fix:** Replace with `std::sync::OnceLock<Arc<RestHealthcheckRegistry>>` so initialization is atomic and subsequent reads are zero-cost.

libs/toolkit/src/contracts.rs
```diff
@@ -60,6 +60,25 @@ pub trait RestApiCapability: Send + Sync {
...
+    /// succeeds. The returned check is registered in [`RestHealthcheckRegistry`] and run on
+    /// every `/readyz` and `/health` request.
+    ///
+    /// Return `None` (the default) to opt out of readiness probing.
+    ///
+    /// # gRPC healthcheck
+    ///
+    /// TODO: Add gRPC healthcheck hook in a follow-up PR.
+    /// This PR intentionally implements REST readiness healthchecks only.
+    fn healthcheck(
```
MikeFalcon77, line 73: **[LOW] RUST-NO-007 — New `healthcheck` method on `RestApiCapability` has no version annotation**

The new default `healthcheck` method is added to the published `RestApiCapability` trait without a `/// Added in toolkit v0.X` doc comment or corresponding CHANGELOG entry, making it impossible to track API version drift across crate releases.

**Fix:** Add a `/// Since: toolkit v0.X` doc line and a CHANGELOG entry so downstream implementors know when the method was introduced.

gears/system/api-gateway/tests/health_endpoints.rs
```diff
@@ -0,0 +1,348 @@
...
+        router: Router,
+        _openapi: &dyn OpenApiRegistry,
+    ) -> anyhow::Result<Router> {
+        Ok(router)
+    }
+    // healthcheck() is intentionally NOT overridden — tests the default
+}
+
+#[test]
+fn existing_rest_gear_compiles_without_healthcheck_impl() {
```
MikeFalcon77: **[LOW] TEST-QUALITY-1 — Constructor echo: `existing_rest_gear_compiles_without_healthcheck_impl`**

The test calls the unoverridden trait default `healthcheck()` and asserts `is_none()`, which only confirms the default return value defined in the trait. Nothing runtime-dependent is verified.

**Fix:** Remove the test or convert it to a compile-only check, since the default-returns-None contract is implicitly verified by all other tests that rely on the default.

gears/system/api-gateway/tests/health_endpoints.rs
```diff
@@ -0,0 +1,348 @@
...
+    ) -> anyhow::Result<Router> {
+        Ok(router)
+    }
+
+    fn healthcheck(&self, _ctx: &toolkit::GearCtx) -> Option<Arc<dyn Healthcheck>> {
+        Some(Arc::new(HealthyCheck))
+    }
+}
+
+#[test]
```
MikeFalcon77: **[LOW] TEST-QUALITY-1 — Constructor echo: `gear_can_return_some_healthcheck`**

`GearWithHealthcheck::healthcheck` is hardcoded to return `Some(...)`. The assertion `is_some()` merely confirms the method returns what it was hardcoded to return, testing nothing beyond what the source code directly states.

**Fix:** Replace with a test that exercises runtime behaviour dependent on the healthcheck being present, such as verifying the check is invoked and its result propagates to a registry report.

MikeFalcon77 (commented):
.

gears/system/api-gateway/src/gear.rs
```diff
@@ -484,19 +515,23 @@ impl ApiGateway {
...
         tracing::debug!("Building new router (standalone/fallback mode)");
-        // In standalone mode (no REST pipeline), register both health endpoints here.
+        // In standalone mode (no REST pipeline), register health endpoints here.
         // In normal operation, rest_prepare() registers these instead.
-        let mut router = Router::new()
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
+        let hc_registry = self
+            .healthcheck_registry
+            .lock()
```
MikeFalcon77: **`RUST-NO-004`** (HIGH)

**Issue:** parking_lot Mutex lock() is called inside async build_router. A contended parking_lot mutex blocks the OS thread and starves the Tokio worker.

**Fix:** Replace parking_lot Mutex on healthcheck_registry with tokio Mutex and await, or switch write-once fields to OnceLock.

gears/system/api-gateway/src/gear.rs
```diff
@@ -735,10 +777,16 @@ impl toolkit::contracts::ApiGatewayCapability for ApiGateway {
         // Apply middleware stack (including auth) to the final router
         tracing::debug!("Applying middleware stack to finalized router");
         let authn_client = self.authn_client.lock().clone();
-        router = self.apply_middleware_stack(router, authn_client)?;
+        router = self.apply_middleware_stack(router, authn_client.clone())?;
+
+        let hc_registry = self
+            .healthcheck_registry
```
MikeFalcon77: **`RUST-NO-004`** (HIGH)

**Issue:** parking_lot Mutex lock() is called inside rest_finalize, which runs on the Tokio executor thread. Contention blocks the thread.

**Fix:** Same as build_router - use tokio Mutex with await or restructure healthcheck_registry as OnceLock.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+struct RegistryEntry {
+    gear: &'static str,
+    check: Arc<dyn Healthcheck>,
+}
+
+/// Minimum interval between fanning out a fresh round of healthchecks.
+/// Bursty `/health` and `/readyz` probes within this window reuse the last
+/// computed report instead of re-running every registered check.
+const REPORT_CACHE_TTL: Duration = Duration::from_millis(500);
```
MikeFalcon77: **`RUST-PERF-001`** (MEDIUM)

**Issue:** REPORT_CACHE_TTL (500 ms) equals HEALTHCHECK_TIMEOUT (500 ms). Every cache miss triggers a full fan-out lasting up to 500 ms, during which concurrent callers also find the cache stale and each spawn their own full check set.

**Fix:** Set REPORT_CACHE_TTL larger than HEALTHCHECK_TIMEOUT (e.g. 2 s) and add coalescing so concurrent refreshes share one in-flight result.

MikeFalcon77 (commented):
Automated review: security, concurrency, observability, and test-coverage findings across the healthcheck feature.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    /// Run all registered healthchecks concurrently and return an aggregate report.
+    ///
+    /// Each check runs in its own `tokio::task::spawn` so panics are caught.
+    /// Checks that exceed `timeout_per_check` are cancelled and reported as
+    /// [`HealthcheckStatus::Unhealthy`].
+    ///
+    /// Results are cached for [`REPORT_CACHE_TTL`] so bursty probes reuse the
+    /// most recent report instead of re-running every check on each request.
+    pub async fn report(&self, timeout_per_check: Duration) -> HealthcheckReport {
```
MikeFalcon77, line 255: **`TOOLKIT-LIFE-001`** (MEDIUM)

**Issue:** RestHealthcheckRegistry report() spawns Tokio tasks without accepting or checking a CancellationToken. In-flight healthcheck tasks are not cancelled on shutdown.

**Fix:** Add a CancellationToken parameter and use tokio::select! on it alongside the per-check timeout.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+            let handles: Vec<_> = entries
+                .into_iter()
+                .map(|(gear, check)| tokio::spawn(run_one_check(gear, check, timeout_per_check)))
+                .collect();
+
+            let mut components = Vec::with_capacity(handles.len());
+            for h in handles {
+                match h.await {
+                    Ok(component) => components.push(component),
+                    Err(_join_err) => {
```
MikeFalcon77: **`RUST-ERR-001`** (MEDIUM)

**Issue:** The JoinError from the outer tokio::spawn in report() is discarded via _join_err, losing which gear task panicked and why.

**Fix:** Log the error with tracing::warn! including gear name before pushing the fallback unhealthy entry.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+) -> HealthcheckComponentReport {
+    let name = check.name();
+    let start = Instant::now();
+
+    let mut handle = tokio::spawn(async move { check.check().await });
+
+    let (status, message) = tokio::select! {
+        result = &mut handle => match result {
+            Ok(r) => (r.status, r.message.as_deref().map(sanitize_message)),
+            Err(_join_err) => (
```
MikeFalcon77: **`RUST-ERR-001`** (MEDIUM)

**Issue:** The JoinError from the inner tokio::spawn in run_one_check is discarded via _join_err, losing the panic payload and affected gear identity.

**Fix:** Log the join error with the gear name before returning the unhealthy status.

libs/toolkit/src/healthcheck.rs
MikeFalcon77, line 255: **`RUST-OBS-001`** (MEDIUM)

**Issue:** When a JoinHandle in report() returns Err on task panic, the fallback entry records gear as unknown, giving operators no actionable signal about which gear failed.

**Fix:** Capture the gear name before tokio::spawn and use it in the Err arm.

gears/system/api-gateway/src/gear.rs
```diff
@@ -484,19 +515,23 @@ impl ApiGateway {
         }
 
         tracing::debug!("Building new router (standalone/fallback mode)");
-        // In standalone mode (no REST pipeline), register both health endpoints here.
+        // In standalone mode (no REST pipeline), register health endpoints here.
         // In normal operation, rest_prepare() registers these instead.
-        let mut router = Router::new()
-            .route("/health", get(web::health_check))
-            .route("/healthz", get(|| async { "ok" }));
+        let hc_registry = self
```
MikeFalcon77: **`RUST-OBS-001`** (MEDIUM)

**Issue:** When healthcheck_registry is None in build_router, unwrap_or_else creates an empty registry with no log message, invisible to operators.

**Fix:** Add tracing::warn! before the fallback.

gears/system/api-gateway/src/gear.rs
```diff
@@ -92,24 +94,51 @@ impl Default for ApiGateway {
...
     }
 }
 
 impl ApiGateway {
-    fn apply_prefix_nesting(mut router: Router, prefix: &str) -> Router {
+    fn apply_prefix_nesting(
+        &self,
+        router: Router,
+        prefix: &str,
+        hc_registry: Arc<toolkit::RestHealthcheckRegistry>,
```
MikeFalcon77, line 153: **`RUST-API-001`** (MEDIUM)

**Issue:** apply_prefix_nesting now takes four parameters including Arc RestHealthcheckRegistry and Option Arc dyn AuthNResolverClient, tightly coupling prefix nesting with health-route construction and auth wiring.

**Fix:** Extract health-route creation and auth middleware into separate call-site steps; keep apply_prefix_nesting focused on router nesting only.

gears/system/api-gateway/src/web.rs
MikeFalcon77: **`RUST-MOD-001`** (MEDIUM)

**Issue:** health_detail and readyz_check are declared pub but are only ever called from closures inside gear.rs - not part of any external API surface.

**Fix:** Change to pub(crate).

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+    async fn unhealthy_component_makes_aggregate_unhealthy_and_not_ready() {
+        let registry = RestHealthcheckRegistry::new();
+        registry.register("my-gear", Arc::new(AlwaysUnhealthy));
+        let report = registry.report(Duration::from_millis(500)).await;
+        assert_eq!(report.status, HealthcheckAggregateStatus::Unhealthy);
+        assert!(!report.is_ready());
+    }
+
+    #[tokio::test]
+    async fn slow_check_times_out_and_becomes_unhealthy() {
```
MikeFalcon77, line 563: **`RUST-TEST-001`** (MEDIUM)

**Issue:** slow_check_times_out_and_becomes_unhealthy verifies aggregate status but not the component-level message field, leaving the health check timed out branch in run_one_check untested.

**Fix:** Add an assertion on report.components[0].message checking for the health check timed out string.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    #[tokio::test]
+    async fn panicking_check_becomes_unhealthy_without_panicking_caller() {
+        let registry = RestHealthcheckRegistry::new();
+        registry.register("panic-gear", Arc::new(PanickingCheck));
+        let report = registry.report(Duration::from_millis(500)).await;
+        assert_eq!(report.status, HealthcheckAggregateStatus::Unhealthy);
+    }
+
+    #[test]
```
MikeFalcon77, line 670: **`RUST-TEST-001`** (LOW)

**Issue:** No test covers compute_aggregate with a mix of Degraded and Unhealthy components to confirm Unhealthy takes priority.

**Fix:** Add a test registering both a DegradedCheck and an UnhealthyCheck and assert the aggregate is Unhealthy.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,585 @@
...
+/// # Contract
+///
+/// - `check` must not panic; panics are caught by [`RestHealthcheckRegistry`] and
+///   converted to [`HealthcheckStatus::Unhealthy`].
+/// - `check` must complete within the configured per-check timeout
+///   (default 500 ms); timeouts are converted to [`HealthcheckStatus::Unhealthy`].
+/// - `check` must never leak secrets, passwords, DSNs, or stack traces in its
+///   message; the registry sanitises messages but defence-in-depth is preferred.
+#[async_trait]
+pub trait Healthcheck: Send + Sync + 'static {
```
MikeFalcon77, line 172: **`RUST-API-001`** (LOW)

**Issue:** Both name() and check() have default implementations, so any Send+Sync+static type silently satisfies Healthcheck with a healthy default - accidental opt-in is invisible.

**Fix:** Remove the default for name() to force implementors to provide a meaningful name. Document check() default as an explicit opt-out escape hatch.

MikeFalcon77:
## Ideological Review: Handler Design & Rust Web Conventions

Beyond the automated checks, here is a design-level assessment of whether this handler pattern is idiomatic for Rust web services and convenient for gear authors.

### Ideological concerns ⚠️

#### 1. Handler functions take `Arc<RestHealthcheckRegistry>` by value — not idiomatic Axum

```rust
pub async fn health_detail(registry: Arc<RestHealthcheckRegistry>) -> Response { ... }
pub async fn readyz_check(registry: Arc<RestHealthcheckRegistry>) -> StatusCode { ... }
```

In Axum, the idiomatic way to share state is via `axum::extract::State<T>` or `Extension<T>`, not by cloning `Arc` into closures. The current code wraps these in manual closures in `health_routes()`:

```rust
get(move || {
    let r = reg_h.clone();
    async move { web::health_detail(r).await }
})
```

This works but is **not how Axum handlers are typically written**. The standard pattern:

```rust
// In gear.rs
Router::new()
    .route("/health", get(web::health_detail))
    .route("/readyz", get(web::readyz_check))
    .with_state(hc_registry)

// In web.rs
pub async fn health_detail(State(registry): State<Arc<RestHealthcheckRegistry>>) -> Response { ... }
```

**Why this matters**: `State<T>` gives you type-safe dependency injection, automatic cloning, and works with `Router::with_state()`. The current closure-wrapping approach is a workaround that adds boilerplate and makes the handlers non-reusable outside this specific wiring.

#### 2. Double `tokio::spawn` per check is unnecessary overhead

```rust
// In report():
let handles = entries.map(|(gear, check)| tokio::spawn(run_one_check(gear, check, timeout)));

// In run_one_check():
let mut handle = tokio::spawn(async move { check.check().await });
```

Each check is spawned **twice**: once in `report()` and once in `run_one_check()`. The inner spawn is needed for panic isolation + timeout via `select!`, but the outer spawn is redundant — `run_one_check` already handles panic isolation internally. The outer spawn only adds task overhead.

**Idiomatic fix**: Drop the outer spawn, use `FuturesUnordered` or `join_all` to run `run_one_check` concurrently:

```rust
let futures = entries.into_iter().map(|(gear, check)| run_one_check(gear, check, timeout));
let components = futures::future::join_all(futures).await;
```

#### 3. `HEALTHCHECK_TIMEOUT` hardcoded in `web.rs` — should be configurable

```rust
const HEALTHCHECK_TIMEOUT: Duration = Duration::from_millis(500);
```

This is a framework-level constant buried in the gateway's web module. In production Rust services, the healthcheck timeout is typically configurable via the same config struct that configures the server. Gears with slow dependencies (e.g. remote DB ping) may need a longer timeout.

**Recommendation**: Move this to `ApiGatewayConfig` or accept it as a parameter to `report()` from the caller.

#### 4. Report cache uses `RwLock` with TOCTOU — not idiomatic for async code

```rust
// Read cache
if let Some((ts, cached)) = self.cached_report.read().as_ref()
    && ts.elapsed() < REPORT_CACHE_TTL
{
    return cached.clone();
}
// ... run checks ...
// Write cache
*self.cached_report.write() = Some((Instant::now(), report.clone()));
```

Between the read and write, multiple concurrent requests can all miss the cache and fan out checks simultaneously. In Rust async web services, the idiomatic approach is `tokio::sync::Mutex` guarding the entire check-or-cache decision, or `arc_swap::ArcSwapOption` for the cached report with a single mutex for the fan-out gate.

#### 5. `Healthcheck::name()` default leaks `std::any::type_name::<Self>()`

```rust
fn name(&self) -> &'static str {
    std::any::type_name::<Self>()
}
```

This returns the full Rust type path (e.g. `my_crate::my_gear::MyHealthcheck`) which goes into the public `/health` JSON response. This is an information leak. The default should either be a simple `"healthcheck"` or require explicit implementation.

---

### Is this how it's done in Rust web services? 🦀

**Partially yes, partially no:**

| Aspect | Convention | This PR |
|--------|-----------|---------|
| State injection via `State<T>` | ✅ Axum standard | ❌ Manual closures |
| Configurable timeout | ✅ Production norm | ❌ Hardcoded |
| Cache with proper locking | ✅ `tokio::sync::Mutex` | ⚠️ TOCTOU with `parking_lot::RwLock` |

**Bottom line**: The design is **ideologically sound** — the right abstractions, the right K8s semantics, the right trait shape. The implementation has a few non-idiomatic patterns (closure wrapping instead of `State<T>`, double spawn, hardcoded timeout, TOCTOU cache) that should be addressed before merge but don't block the overall approach.

The `RestApiCapability::healthcheck()` hook is **convenient for gear authors** — implementing one method that returns `Option<Arc<dyn Healthcheck>>` is minimal friction, and the `None` default means zero breakage for existing gears.

gears/system/api-gateway/src/gear.rs
```diff
@@ -92,24 +94,51 @@ impl Default for ApiGateway {
...
+        Ok(top.merge(nested))
+    }
 
-        router = Router::new().nest(prefix, router);
-        top.merge(router)
+    fn health_routes(hc_registry: Arc<toolkit::RestHealthcheckRegistry>) -> Router {
+        let reg_h = hc_registry.clone();
+        let reg_r = hc_registry;
+        Router::new()
+            .route(
```
MikeFalcon77: **[HIGH]** `TOOLKIT-REST-001` / `TOOLKIT-REST-002` — health routes bypass `OperationBuilder` and declare auth out-of-band.

`health_routes()` registers `/health`, `/healthz`, and `/readyz` via raw `Router::new().route()` calls with bare closures. Authentication intent is then declared separately by manually inserting into `public_routes` (lines 191-193). Route definition and auth declaration are decoupled, making auth audits error-prone.

Note: `/healthz` was already registered this way before this PR. `/readyz` is new and should follow whatever pattern the team decides to standardize on. **Fix:** Register these routes via `OperationBuilder::new().public()` so auth intent is co-located with the route definition, consistent with all other ToolKit REST endpoints.

gears/system/api-gateway/src/web.rs
```diff
@@ -1,10 +1,15 @@
...
     routing::{MethodRouter, get},
 };
 use chrono::{SecondsFormat, Utc};
-use serde_json::{Value, json};
+use serde_json::json;
+use std::sync::Arc;
+use std::time::Duration;
+use toolkit::RestHealthcheckRegistry;
+
+const HEALTHCHECK_TIMEOUT: Duration = Duration::from_millis(500);
```
MikeFalcon77: **[MEDIUM]** `RUST-API-001` — `HEALTHCHECK_TIMEOUT` hardcoded at 500ms with no operator knob.

Services with slow downstream dependencies may never produce a meaningful readiness signal within 500ms, and there is no way to tune this without recompilation. Also note that `REPORT_CACHE_TTL` in `healthcheck.rs` is also 500ms — equal to the timeout — meaning a worst-case timed-out check fills the cache with an `Unhealthy` result that expires exactly when the next probe arrives, providing no buffering margin.

**Fix:** Add a `healthcheck_timeout_ms: u64` field to `ApiGatewayConfig` (default 500) and thread it through to `health_detail` and `readyz_check`. Set `REPORT_CACHE_TTL` strictly greater than the configured timeout (e.g. 2x) so bursty probes reuse the last result.

libs/toolkit/src/runtime/host_runtime.rs
```diff
@@ -439,6 +439,12 @@ impl HostRuntime {
         // use host as the registry
         let registry: &dyn crate::contracts::OpenApiRegistry = host.as_registry();
 
+        // Create the healthcheck registry and publish it to ClientHub so both
+        // the REST provider loop (below) and the API Gateway can access it.
+        let hc_registry = Arc::new(crate::healthcheck::RestHealthcheckRegistry::new());
+        self.client_hub
```
MikeFalcon77: **[HIGH]** `TOOLKIT-CLIENT-001` / `RUST-MOD-001` — `RestHealthcheckRegistry` stored in `ClientHub`, which is designed for inter-gear service clients (gRPC stubs, HTTP clients), not infrastructure registries.

Using `ClientHub` as a side-channel to pass state from `run_rest_phase` to `ApiGateway::rest_prepare` means any gear can retrieve the registry via `client_hub.get::<RestHealthcheckRegistry>()` and register checks outside the intended wiring phase, bypassing the write-once access control this PR intends to enforce.

**Fix:** Either (a) add `Arc<RestHealthcheckRegistry>` as an explicit parameter to `ApiGatewayCapability::rest_prepare` and `rest_finalize`, or (b) hold it as a field on `HostRuntime` and pass it explicitly. Do not route infrastructure state through the general-purpose service-locator.

MikeFalcon77 (changes requested):
.

gears/system/api-gateway/src/gear.rs
refur-nfn: fixed

gears/system/api-gateway/src/web.rs
refur-nfn: Clippy's redundant_pub_crate lint flags "(crate)" as redundant

refur-nfn: Clippy's redundant_pub_crate lint flags "(crate)" as redundant

libs/toolkit/src/healthcheck.rs
refur-nfn: Collapsed to a single `HealthcheckStatus`

libs/toolkit/src/healthcheck.rs
refur-nfn: add stable-contract note

libs/toolkit/src/healthcheck.rs
refur-nfn: `name()` default removed

libs/toolkit/src/healthcheck.rs
refur-nfn: Default removed

libs/toolkit/src/healthcheck.rs
refur-nfn: added stable-contract one-liners on both structs

libs/toolkit/src/contracts.rs
refur-nfn: added one-liner in the method doc

libs/toolkit/src/lib.rs
refur-nfn: Module made private `mod healthcheck;` + flat `pub use` only

gears/system/api-gateway/src/gear.rs
refur-nfn: fixed

gears/system/api-gateway/src/gear.rs
refur-nfn: added

gears/system/api-gateway/src/gear.rs
refur-nfn: Replaced by `OnceLock`

gears/system/api-gateway/src/gear.rs
refur-nfn: Registry is a required param

gears/system/api-gateway/src/gear.rs
refur-nfn: restructured

gears/system/api-gateway/src/web.rs
refur-nfn: (default 500) threaded via `HealthcheckTimeout` Extension

gears/system/api-gateway/src/web.rs
refur-nfn: `/health` no longer in `public_routes`

gears/system/api-gateway/tests/health_endpoints.rs
refur-nfn: added

gears/system/api-gateway/tests/health_endpoints.rs
refur-nfn: Rewritten as `gear_without_healthcheck_override_contributes_nothing_to_report` — exercises registry report

gears/system/api-gateway/tests/health_endpoints.rs
refur-nfn: Rewritten as `gear_with_healthcheck_override_is_invoked_via_registry` — asserts the check runs and shows in the report

libs/toolkit/src/runtime/host_runtime.rs
refur-nfn: Registry created in `run_rest_phase` and passed **explicitly** to `rest_prepare`/`rest_finalize`

libs/toolkit/src/healthcheck.rs
refur-nfn: TTL raised to 2s

libs/toolkit/src/healthcheck.rs
refur-nfn: Signature now `register(gear: impl Into<String>, …)

libs/toolkit/src/healthcheck.rs
refur-nfn: added

libs/toolkit/src/healthcheck.rs
refur-nfn: mutex is used

libs/toolkit/src/healthcheck.rs
refur-nfn: Outer spawn removed

libs/toolkit/src/healthcheck.rs
refur-nfn: join_all` Is used

libs/toolkit/src/healthcheck.rs
refur-nfn: Outer spawn removed; inner join error logged with gear+check

libs/toolkit/src/healthcheck.rs
refur-nfn: added `tracing::error!` with gear+check on panic arm

libs/toolkit/src/healthcheck.rs
refur-nfn: added  `error!` on panic, `warn!` on timeout

libs/toolkit/src/healthcheck.rs
refur-nfn: Logs any non-`is_cancelled()` `JoinError` after abort

libs/toolkit/src/healthcheck.rs
refur-nfn: Blocklist extended

libs/toolkit/src/healthcheck.rs
refur-nfn: Replaced by behavioral registry tests

libs/toolkit/src/healthcheck.rs
refur-nfn: added

libs/toolkit/src/healthcheck.rs
refur-nfn: added

libs/toolkit/src/healthcheck.rs
refur-nfn: added

libs/toolkit/src/healthcheck.rs
refur-nfn: Gear name captured before spawn and used in every arm

gears/system/api-gateway/src/gear.rs
refur-nfn: co-locating auth via .public() is a real readability win, but the mechanical fix conflicts with the root-mount + no-OpenAPI-exposure requirements. Current code compromises by de-duping paths through shared consts (HEALTH_DETAIL_PATH/HEALTHZ_PATH/READYZ_PATH) used at both the route and the public_routes site, so they can't silently drift

gears/system/api-gateway/src/gear.rs
refur-nfn: Now uses self.healthcheck_registry OnceLock, and fallback logs tracing::warn!

gears/system/api-gateway/src/gear.rs
refur-nfn: health routes (/health, /healthz, /readyz) now declare auth via OperationBuilder instead of a hardcoded set in gear.rs

gears/system/api-gateway/src/web.rs
MikeFalcon77: Actually I have a more common question : why do we need to expose all these /healthz, /ready and etc in OperationBuider? It's internal routes for internal checks. A service knows already these endpoints and service's owner will add these routes into the helm chart (or docker) to do health checks. Looks like we don't need these routes in open api.

gears/system/api-gateway/src/gear.rs
```diff
@@ -92,24 +97,83 @@ impl Default for ApiGateway {
...
-            .route("/healthz", get(|| async { "ok" }));
+    /// Standalone health-probe router (`/health`, `/healthz`, `/readyz`): all public, no
+    /// gateway middleware, no auth, and NOT part of the `OpenAPI` document. Intended to be
+    /// bound on a separate (private/management) listener by the embedding application — the
+    /// main gateway server does not serve these paths.
+    ///
+    /// # Errors
+    /// Returns an error if [`rest_prepare`](toolkit::contracts::ApiGatewayCapability::rest_prepare)
+    /// has not run yet (the healthcheck registry is unset).
+    pub fn health_router(&self) -> Result<Router> {
```
MikeFalcon77, line 135: Nothing serves this. `health_router()` has no callers outside its own test; bootstrap `run_server` and gateway `serve()` are untouched, so no listener ever binds it. Two gaps to fix:

1. No serve mode. Health is removed from the main router with no way to put it back, so single-port platforms (Railway/PaaS that probe `$PORT/health`) get 404. Add `health.serve: main | separate | both` on `ApiGatewayConfig`, default `main` (or `both`) so the common single-port case works with zero config.
2. `separate` must be served by the framework, not the embedder. Handing back a bare `Router` means every host re-implements bind + `ready.notify()` + `with_graceful_shutdown` on the same cancel token (see `serve()` in this file). `with_cancellation` only aborts the check tasks; it does not drain the HTTP server. Add `health_bind_addr` config and start the second listener in the gateway lifecycle.

MikeFalcon77 (commented):
Direction is good and the per-gear opt-in hook is easy to use. Main thing blocking this as a universal healthcheck: health is removed from the main router with no way to put it back, and the separate listener is not actually served by the framework. Inline notes below.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,642 @@
...
+//! }
+//! ```
+//!
+//! # Kubernetes probe configuration
+//!
+//! ```yaml
+//! livenessProbe:
+//!   httpGet:
+//!     path: /healthz
+//!     port: http
```
MikeFalcon77, line 53: This probe points at the main service port, but the PR removes health from the main router, so on the main port these paths are 404. Fix the snippet to use the separate health port, or (better) add a `main`/`both` mode so this stays valid.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,642 @@
...
+//!   periodSeconds: 5
+//!   timeoutSeconds: 2
+//!   failureThreshold: 3
+//! ```
+//!
+//! `/healthz` is liveness — always shallow, never runs user checks.
+//! `/readyz` is readiness — runs user REST healthchecks and removes the pod
+//! from traffic when unhealthy without triggering a restart.
+//! `/health` is the detailed diagnostic endpoint — it requires authentication
```
MikeFalcon77: There is no auth here. `health_router()` adds no middleware, so all three paths are public. Either add an optional auth layer for `/health`, or drop this line. Docs should not claim auth that does not exist.

libs/toolkit/src/healthcheck.rs
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+    status
+}
+
+const MAX_MESSAGE_LEN: usize = 256;
+
+// Deliberately broad blocklist: false positives (benign messages collapsed to the
+// generic string) are accepted to never leak secrets/DSNs on the public endpoint.
+static SUSPICIOUS_SUBSTRINGS: &[&str] = &[
```
MikeFalcon77, line 382: This blocklist is fine for a public endpoint, but too broad for a private management one: `tenant`, `select `, `delete `, `private` will turn useful operator messages ("tenant pool exhausted", "delete queue backlog") into "health check failed". Pick the threat model first, then tune. Consider a stable `code` field next to the free-text `message` so operators keep signal even when the text is redacted.

libs/toolkit/src/contracts.rs
```diff
@@ -60,29 +60,47 @@ pub trait RestApiCapability: Send + Sync {
         router: Router,
         openapi: &dyn OpenApiRegistry,
     ) -> anyhow::Result<Router>;
+
+    /// Readiness healthcheck for this gear, run on every `/readyz` and `/health`.
+    /// `None` (default) opts out. Since `cf-gears-toolkit` v0.6.13.
```
MikeFalcon77: Premature: the crate is not released with this API yet. Drop or fix the version note before merge.

libs/toolkit/src/contracts.rs
```diff
@@ -60,29 +60,47 @@ pub trait RestApiCapability: Send + Sync {
...
+    fn rest_prepare(
+        &self,
+        ctx: &crate::context::GearCtx,
+        router: Router,
+        hc_registry: std::sync::Arc<crate::healthcheck::RestHealthcheckRegistry>,
+    ) -> anyhow::Result<Router>;
 
-    /// Finalize before start: attach /openapi.json, /docs, persist the Router internally if needed.
-    /// Do NOT start the server here.
+    /// Finalize before start: attach /openapi.json, /docs, health endpoints, persist the
```
MikeFalcon77, line 98: Doc says finalize attaches health endpoints, but the code no longer mounts them on this router. Update the wording — health lives only on `health_router()`.

gears/system/api-gateway/src/gear.rs
refur-nfn: done

libs/toolkit/src/healthcheck.rs
refur-nfn: separate health port is added

libs/toolkit/src/healthcheck.rs
refur-nfn: requires authentication statement is dropped

libs/toolkit/src/healthcheck.rs
refur-nfn: code is added

libs/toolkit/src/contracts.rs
refur-nfn: removed

PR #4171, opened 2026-07-01
Add throttling middleware for rate and in-flight limit.

genericaccount-de:
Adds throttling middleware for rate and in-flight limit. Throttling is defined by common zones, which are reused by separate API endpoints by means of OperationsBuilder. 

## Summary by CodeRabbit

* **New Features**
  * Added zone-based request throttling with separate controls for steady-rate (RPS) and concurrent in-flight limits.
  * Enables per-route throttling via named zones, IP-keyed limiting, and optional “dry run” mode.
* **Bug Fixes**
  * Throttling rejections now return consistent, canonical error details with `Retry-After` and rate-limit metadata headers.
* **Documentation**
  * Updated quickstart and E2E/local configs to the zone-based throttling model; local config increases request body limit, and OpenAPI output no longer emits legacy rate-limit vendor extensions.
* **Tests**
  * Added end-to-end tests for zone-based rate limiting and in-flight rejection; updated middleware-order and other gateway tests; removed legacy rate-limit test coverage.


deny.toml
```diff
@@ -39,6 +39,8 @@ ignore = [
...
   "RUSTSEC-2026-0187", # lopdf <0.42 unbounded-recursion DoS on crafted PDFs — transitive via kreuzberg→cf-gears-file-parser→cf-gears-example-server (example app only, never parses untrusted PDFs). No usable fix: kreuzberg 4.9.x pins lopdf "^0.41" (<0.42) and the fix landed in 0.42.0; 5.x drops the bundled-pdfium feature we use. Remove once kreuzberg ships a release allowing lopdf >=0.42.
+  "RUSTSEC-2026-0195", # quick-xml <0.41 unbounded xmlns-declaration allocation (memory-exhaustion DoS) — transitive via calamine 0.35 (excel) and kreuzberg 4.9.8 directly → cf-gears-file-parser→cf-gears-example-server. No usable fix: patched only in quick-xml >=0.41.0; latest stable calamine (0.35.0) requires quick-xml "^0.39" (max 0.39.4, still <0.41) and kreuzberg 4.9.8 uses 0.40.1; only kreuzberg 5.0.0-rc exists and it drops the bundled-pdfium feature we use. Residual DoS risk is bounded by the gateway request body-size limit (defaults.body_limit_bytes), which caps the number of xmlns declarations an attacker can submit. Remove once kreuzberg ships a release allowing quick-xml >=0.41.
```
genericaccount-de, line 42: I have tried to update kreuzberg to the latest 4.9.9 build in a separate PR, but it does not help since it still uses quick-xml 0.40.1. That is why it is added to exceptions list.

gears/system/api-gateway/src/middleware/throttling.rs
MikeFalcon77, line 1: I think generic part of throttling should not be in api gateway crate

genericaccount-de: Done, moved generic parts to the api-gateway.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1040 @@
...
+            |ext| ext(req),
+        ),
+    }
+}
+
+/// Resolve the client IP: `X-Forwarded-For` (first hop) / `X-Real-IP`, then the
+/// peer address from `ConnectInfo`, else `"unknown"`.
+fn client_ip(req: &Request) -> String {
+    let headers = req.headers();
+    if let Some(ip) = headers
```
MikeFalcon77: **CRITICAL**

`client_ip` trusts the leftmost `X-Forwarded-For` (and `X-Real-IP`) value, which is fully client-controlled, as the throttling bucket key with no trusted-proxy validation.

Pre-auth IP-keyed zones are the only limit an unauthenticated caller hits; a spoofed or per-request-unique XFF both bypasses the limit entirely and lets an attacker attribute load to victim IPs.

Derive the client IP from `ConnectInfo` or from a configured trusted-proxy hop count (rightmost untrusted XFF entry), and ignore client-supplied `X-Forwarded-For`/`X-Real-IP` unless the immediate peer is a trusted proxy.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1040 @@
...
+    cfg: &RateLimitZone,
+) -> Result<Arc<RateZone>> {
+    if let Some(existing) = zones.get(name) {
+        return Ok(Arc::clone(existing));
+    }
+    let rps = NonZeroU32::new(cfg.rate_limit.rps)
+        .ok_or_else(|| anyhow!("throttling: rate_limit zone '{name}' has rps = 0"))?;
+    let burst = NonZeroU32::new(cfg.burst_limit)
+        .ok_or_else(|| anyhow!("throttling: rate_limit zone '{name}' has burst_limit = 0"))?;
+    let limiter = RateLimiter::keyed(Quota::per_second(rps).allow_burst(burst))
```
MikeFalcon77, line 446: **HIGH**

`RateLimitZone.max_keys` is validated and documented as an LRU-eviction bound, but `get_or_build_rate_zone` never applies it to the governor keyed limiter, so its internal state store grows one entry per distinct key with no eviction.

For IP-keyed pre-auth zones the key is attacker-controlled via `X-Forwarded-For`, so unbounded distinct keys exhaust process memory and take down the gateway despite the configured bound.

Enforce `max_keys` on the rate limiter (periodic `retain_recent`/`shrink` prune bounded by a lifecycle `CancellationToken`, or a bounded keyed store) so the tracked-key count cannot exceed the configured limit.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1040 @@
...
+}
+
+impl InFlightZone {
+    fn gate(&self, key: &str) -> Arc<KeyGate> {
+        if let Some(existing) = self.keys.get(key) {
+            return Arc::clone(&existing);
+        }
+        // Soft `max_keys` cap: drop gates no longer referenced by in-flight requests.
+        if self.keys.len() as u64 >= self.cfg.max_keys {
+            self.keys.retain(|_, v| Arc::strong_count(v) > 1);
```
MikeFalcon77: **HIGH**

`gate()` runs `DashMap::retain` on the request hot path once `len() >= max_keys`; it write-locks and scans every shard, and when keys are actively in-flight (`strong_count > 1`) it evicts nothing, so the map keeps growing and the full O(n) scan repeats on each subsequent new-key request.

Under a flood of distinct keys (attacker-controlled IPs pre-auth) this becomes a per-request O(n) all-shard write-locking scan that serializes gate lookups and fails to bound memory — the cap becomes a DoS amplifier.

Replace the inline retain with a bounded eviction that does not write-lock all shards per request (segmented LRU with a size cap, or a periodic background sweep gated by a `CancellationToken`), and skip the scan when it cannot free entries.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -244,15 +246,67 @@ pub struct ODataPagination<T> {
...
+///
+/// References throttling zones (defined in the API gateway configuration) by
+/// name and, for identity-keyed zones, supplies the [`IdentityKeyFn`] used to
+/// compute per-request keys. Limits themselves live in config (zones are the
+/// primary source of truth); this struct only binds an operation to zones and
+/// provides the code-side behavior that config cannot express.
+#[derive(Clone, Default)]
+pub struct ThrottlingSpec {
+    /// Name of the rate-limit zone this operation participates in.
+    pub rate_limit_zone: String,
```
MikeFalcon77: **MEDIUM**

`ThrottlingSpec` models zone references as bare `String`, with the empty string overloaded as a sentinel for "no zone in this category" (see `if thr.rate_limit_zone.is_empty()` in the gateway).

The absent-zone state is representable only by convention, so every consumer must remember the sentinel rule and an accidental empty string is a silently-valid "no throttling" — a type the compiler could have rejected.

Change both fields to `Option<String>` (or a `ZoneName` newtype) so "no zone" is expressed as `None` and callers cannot pass an ambiguous empty string.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1040 @@
...
+    if let Some(secs) = retry_after_seconds
+        && let Ok(value) = HeaderValue::from_str(&secs.to_string())
+    {
+        headers.insert(header::RETRY_AFTER, value);
+    }
+    response
+}
+
+fn log_throttled(key: &ThrottleKey, kind: &str, id: &str) {
+    tracing::debug!(
```
MikeFalcon77: **MEDIUM**

Enforced throttling rejections (real 429s) are logged at `tracing::debug!` and emit no metric, while non-enforced dry-run would-be-rejections are logged at `tracing::warn!`.

At the default info level operators get zero visibility into actual client throttling, and with no rejection counter, capacity/tuning problems on this critical path are undiagnosable in production.

Emit a rejection metric (counter keyed by zone/kind) and raise the enforced-rejection log to at least `info`, keeping the high-cardinality key value out of the message body.

gears/system/api-gateway/tests/throttling_tests.rs
```diff
@@ -0,0 +1,300 @@
...
+                    .uri("/tests/v1/slow")
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .expect("first request failed")
+    });
+
+    // Let the first request acquire the permit before firing the second.
+    tokio::time::sleep(Duration::from_millis(20)).await;
```
MikeFalcon77: **LOW**

The in-flight concurrency test coordinates the two requests with a fixed 20ms sleep, so the required ordering (first request holds the sole permit before the second arrives) is not guaranteed.

On a loaded runner the spawned first request may not acquire the permit within 20ms, causing spurious failures or masking a real regression instead of deterministically exercising the gate.

Replace the sleep with an explicit synchronization signal (a `tokio::sync::Notify` or `Barrier` the slow handler fires on entry) so the second request is issued only after the permit is confirmed held.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -1713,6 +1767,35 @@ mod tests {
...
+                in_flight_limit_zone: "ifl_identity".to_owned(),
+                identity_key_func: Some(std::sync::Arc::new(|_req: &axum::extract::Request| {
+                    "subject-123".to_owned()
+                })),
+                require_security_context: true,
+                dry_run: false,
+            });
+
+        let throttling = builder.spec.throttling.as_ref().expect("throttling set");
+        assert_eq!(throttling.rate_limit_zone, "rl_identity");
```
MikeFalcon77: **LOW**

`builder_with_throttling_sets_spec_and_extractor` asserts the zone-name fields equal the exact strings just passed to `with_throttling`, which only confirms the setter stored its input.

`with_throttling` merely assigns the spec, so echoing the same names back exercises no logic and catches no regression; only the extractor-invocation and Clone/Debug assertions carry value.

Drop the two zone-name echo assertions and keep/extend the ones that verify behavior — that the stored closure is invokable and the `Arc<dyn Fn>` field survives Clone/Debug.

MikeFalcon77:
Design-level comments: 

This is separate from the code-level findings. It concerns the throttling model itself.

1. Configuration ergonomics
The code/config split is drawn in the wrong place. Numeric limits (rps, burst, status, retry-after, max_keys) live in YAML zones — good, operators tune them without a rebuild. But the operation->zone binding, dry_run, and require_security_context all live in Rust code. dry_run is exactly what an operator wants to flip at runtime when rolling out enforcement gradually; today that requires editing the binary. This contradicts the NGINX-style model the code advertises, where the whole policy is config.

2. 3rd-party API gateway target
The model will not port to an external gateway (Kong/Envoy/AWS API GW) as-is. Enforcement lives entirely in the in-process axum middleware; ThrottlingSpec is only metadata that the local middleware executes. Hard blockers:

IdentityKeyFn = Arc<dyn Fn(&axum::extract::Request) -> String> is a Rust closure — not serializable, not callable by an external gateway. Identity keying does not translate at all.

IdentityKeyFn pulls axum::extract::Request into the toolkit contract type (operation_builder.rs:261). That leaks the runtime into the contract layer — a direct problem on feature/toolkit_contracts, where contracts are meant to be portable.

Zone definitions live in ApiGatewayConfig, not in the contract. require_security_context / dry_run are concepts of the local middleware pipeline only.

For an external target the key must be declarative (enum Key { Ip, Header(String), JwtClaim(String), … }) and the zone model must live in the contract, not the gateway gear.

3. Rust idioms (Go port)
Mostly idiomatic; two Go-isms stand out:

Empty string as a "no zone" sentinel (if thr.rate_limit_zone.is_empty()) instead of Option<String>.
axum in a public contract type (see §2).
Good/idiomatic: RateSpec newtype with custom Deserialize, KeyType/RetryAfter enums, the manual Debug, governor + tokio::Semaphore. The with_throttling(mut self) -> Self builder is consistent with the rest of OperationBuilder.

4. Group-level throttling setup
Not supported. with_throttling is per-operation and the full 5-field spec is repeated on every route; there is no router/group API and no default. Two consequences:

Numeric limits are not duplicated (defined once per zone), only the binding boilerplate is.
Sharing semantics: build_maps makes every operation on a zone share one token bucket / gate. "One zone on 10 routes" means 10 rps across all ten combined, not 10 rps each. Same-limit-but-independent per route is not expressible without separate zones.

5. Is this how Rust does it?
Building blocks are OK (governor, Semaphore, axum from_fn). Architecture - no. The per-operation-metadata → central resolver → config zones shape is more Go-framework pattern. The common Rust approach is a tower::Layer (tower_governor::GovernorLayer, GlobalConcurrencyLimitLayer) applied to a router or route group via .route_layer(), which is declarative at the composition site and gives grouping for free. Need to design config-based throttling setup with tower layers approach.

**Suggestion:**
Let's design it via toolkit ADR/desing document
It should decide:

1. The config/code boundary — specifically whether dry_run and the operation->zone binding move to config (route-policies already reserve rate_limit as a "future field").
2. A declarative key model (enum Key) that replaces the axum closure, so the contract stays runtime-agnostic and an external-gateway backend is possible.
3. Whether zone binding is expressible at the route-group level, and the intended shared-vs-per-route bucket semantics.

MikeFalcon77 (changes requested):

MikeFalcon77:
Also one important thing: throttling must support cache abstractionb from cluster primitives. It means we need implemantation of StateStorage for Governor: https://docs.rs/governor/latest/governor/state/trait.StateStore.html

parviz-93:
This PR appears abandoned (author unreachable, branch not shared), so the work has been continued in #4701: the original commit is cherry-picked with authorship preserved, and all 13 review threads from here are addressed or explicitly deferred (see the table in the new PR's description). Maintainers — please consider closing this one in favor of #4701.

PR #4184, opened 2026-07-06
fix(file-storage): P2 remediation — Tier 0 showstoppers (+ full roadmap)

ffedoroff:
## file-storage P2 remediation — Tier 0 (this PR) + full roadmap

This PR opens the remediation effort for the `file-storage` gear after the P2 merge (#4170).
It **lands Tier 0 in full** and describes the complete, tiered plan so reviewers can see where
this is going, not just what is here today. Work continues on this branch; later tiers will be
pushed as follow-up commits (or split into separate PRs if reviewers prefer).

The ordered, individually-verifiable source plan lives at
`gears/file-storage/docs/IMPLEMENTATION_PLAN_TEMP.txt` (temporary working doc, deleted once
split into tickets).

**Severity tiers:** Tier 0 = correctness/security showstoppers · Tier 1 = production-readiness
blockers · Tier 2 = correctness hardening · Tier 3 = doc reconciliation · Tier 4 = deferred/P3.

---

### ✅ Tier 0 — Correctness / security showstoppers (DONE in this PR, 11/11)

Every item ships with unit tests. Full gear suite green (101 lib + 131 integration + 12 sdk,
0 failures), `cargo fmt --check` + `cargo clippy --all-targets -D warnings` clean, sidecar builds.
36 files, +2443/−105, 5 new test files + 1 additive migration.

**Authorization & input-validation cluster** (root cause: resource-less `authorize(…, "", None)` + no validation)
- **0.7** cross-user policy/retention authz — new `actions::ADMIN_POLICY`, ownership gates (try-admin-first), per-file reauthorization for file-scope rules, fetch-then-reauthorize on delete.
- **0.9** cross-user file enumeration via `list_files` — ownership gate on `owner_id`.
- **0.10** idempotency replay bypassed authz — replay now authorized (WRITE) before any ticket is returned; key bound to `subject_id` (additive migration).
- **0.11** semantic validation — reject `max_age_days<1` (tenant-wipe), all-criteria-none, missing scope target, none-owner user policy, `*/*` mime.

**Finalize / version-integrity cluster**
- **0.1** finalize no longer trusts client size/hash — reads the blob back, verifies size + SHA-256, persists read-back-derived values.
- **0.4** double-finalize — `finalize`/`mark_available` become pending-only CAS; second call is a 409, not a silent overwrite.
- **0.8** `delete_version` with a non-matching id returned 404 instead of deleting the whole file.

**Background sweep cluster**
- **0.3** sweep-vs-complete race — session CAS runs first and gates version cleanup; `complete` rejects expired sessions; status-guarded `delete_pending_version`.
- **0.6** retention sweep now propagates a `list_versions` error (skips expiry) instead of swallowing it and orphaning blobs.

**Backend safety & multipart**
- **0.5** in-memory backend off by default (`enable_in_memory_backend`); `durable` capability; `migrate_backend` onto a non-durable target requires `ADMIN_POLICY`.
- **0.2** multipart report-part callback (token-auth route + handler + `MultipartService::report_part` + sidecar callback) so `complete_multipart_upload` no longer sees an empty parts list; documented the `local-fs multipart_native=false` caveat.

---

### 🔜 Tier 1 — Production-readiness blockers (planned next)

Cleanly implementable now, executed in dependency order:
- **1.1** local-fs fsync + atomic write (temp→fsync→rename→parent-dir-fsync).
- **1.3** `require_signing_key_seed` config guard (secure-by-default; fixes ephemeral per-replica signing keys).
- **1.4 (sweep half)** `enable_background_sweep` defaults to `true`; config templates made explicit.
- **1.6** sidecar `/healthz` + `build_router` refactor (enables in-process route tests).
- **1.2(a)** sidecar `DefaultBodyLimit` layer (lift axum's 2 MiB floor).
- **1.5** finalize-callback timeouts + bounded retries.
- **1.2(b)** real streaming `put_stream` (memory-DoS fix; incremental-adoption trait default).
- **1.11** RFC-correct sidecar range responses + stop leaking control-plane URL/error body.
- **1.10** validate declared MIME against actual bytes on the upload path.
- **1.9** GC expired `idempotency_keys` in the sweep (outboxes left untouched, lock-in test).
- **1.8** metrics/observability via the platform OTel `Meter` precedent (mini-chat) + request-id propagation.

**🛑 Blocked / needs a team decision (will NOT be silently coded around):**
- **1.7 (S3 backend)** — the largest item; requires an external-SDK security review + team choice of S3 crate (`aws-sdk-s3` vs lighter alternative) + a new heavy dependency. Tracked separately.
- **1.4 (quota half)** — no `QuotaClient` SDK crate exists in the workspace yet; blocked until it ships (cross-gear).
- **1.12 (usage accounting symmetry)** — coupled with wiring `usage_reporter` off `None` (via `usage-collector-sdk`); needs a wiring decision, done in the same change.

### 🔜 Tier 2 — Correctness hardening (13 items, planned)
Idempotency body-match on replay (2.1), unbounded `GET /versions` + full-scan `VersionRepo::get` (2.2), `migrate_backend` CAS (2.3), policy-upsert race (2.4), error-code contract drift 429/400 vs 412/422/507 (2.5), negative-size/malformed-hash validation (2.6), `delete_version` vs `bind` dangling `content_id` (2.7), orphan zero-version rows (2.8), dead superseded-version retention (2.9), silently-dropped `If-Match-Metadata` (2.10), `compute_plan` overflow from client part size (2.11), unvalidated `transfer_ownership` target (2.12), `sign_url` wrong `MultipartPart` mapping (2.13).

### 🔜 Tier 3 — Documentation reconciliation (11 items, code is authoritative)
ADR-0004 PASETO claim vs bespoke codec (3.1), DESIGN/ADR-0003 sidecar bind/finalize flow drift (3.2), multipart-coordinator doc drift (3.3), fabricated/missing endpoints (3.4), migration.sql DDL drift (3.5), README/DECOMPOSITION P2-scope (3.6), api.md status-code table (3.7), undocumented functionality (3.8), stale comments (3.9), retention-rule 404 wording (3.10), inactivity-retention-vs-doc contradiction (3.11).

### 🔜 Tier 4 — Deferred / P3 follow-ups (11 items)
EventBroker relay (4.1), owner-deletion workflow (4.2), `enabled_event_types` (4.3), signed-URL rate/conn caps (4.4), SDK trait placeholder (4.5), BLAKE3 alignment (4.6), blob↔DB reconciliation (4.7), encryption at rest (4.8), signed-URL key rotation `kid` (4.9), load/perf validation (4.10), backup/restore runbook + alerting (4.11).

---

Every commit is DCO-signed (`git commit -s`). Commit granularity: one commit per tier (per request).

## Summary by CodeRabbit

* **New Features**
  * Added a token-authenticated multipart “report part” callback endpoint.
  * Added `limit`/`offset` pagination to file version listings.
  * Added optional in-memory backend enablement and introduced S3 backend support.
* **Bug Fixes**
  * Finalization now re-verifies stored size/hash/MIME before activating versions.
  * Multipart now enforces shipped token/topology behavior and returns `403` for invalid/expired tokens.
  * REST precondition/conflict mappings updated to return `400` instead of `412`; callback failures now return `502`; download range behavior corrected.
* **Tests**
  * Expanded multipart reporting, pagination, idempotency, authorization, and pinned error-code contract coverage.


gears/file-storage/file-storage/src/domain/service/write.rs
```diff
@@ -14,9 +15,64 @@ use crate::domain::error::DomainError;
...
+
+    let mut stream = backend
+        .get_stream(backend_path)
+        .await
+        .map_err(|_| no_content_err())?;
+
+    let mut hasher = hash::Hasher::new();
+    let mut prefix: Vec<u8> = Vec::with_capacity(MIME_SNIFF_PREFIX_BYTES);
+    while let Some(chunk) = stream.next().await {
+        let chunk = chunk.map_err(|_| no_content_err())?;
```
Artifizer: **HIGH**

A mid-stream read-back error is collapsed via `map_err(|_| no_content_err())` into `validation("content", "no uploaded content found; PUT was not completed")`, discarding the original error.

This conflates a transient backend/transport failure (e.g. an S3 connection dropped partway through reading back a fully-uploaded object) with a definitive "object missing" state, and destroys the source context needed to diagnose or retry — the client is told the upload never completed when it did.

Distinguish object-not-found from a transport/read error: preserve the underlying error (map to a distinct `DomainError::backend`/internal variant for mid-stream failures) and reserve `no_content_err()` for the genuine not-found case. Same applies to the `get_stream().map_err(|_| no_content_err())` on the initial call.

ffedoroff: Good catch — fixed in 3d89000e.

Mid-stream chunk errors now map to `DomainError::backend` (preserving the original error, mirroring `put_stream`), and the initial `get_stream` failure only collapses to the not-found `validation` error when `exists()` confirms the object is genuinely absent; otherwise the underlying backend error is surfaced (5xx + tracing) instead of being masked as "never uploaded".

PR #4187, opened 2026-07-07
feat(usage-collector): BigDecimal aggregates, deterministic record ids, op-per-kind guard

capybutler:
Consolidates three related changes to the usage-collector gateway and SDK, plus one fix.

**Arbitrary-precision aggregate values.** `AggregationBucket.value` (SDK) and
`AggregationBucketDto.value` (gateway) move from fixed-precision `Option<Decimal>`
to `Option<BigDecimal>`, so aggregates no longer lose precision on large or
high-scale sums. Values round-trip over the wire as plain-decimal strings (never
scientific notation, never a bare float) via a new `bigdecimal_str_option` serde
codec.

**Deterministic, gateway-derived record id.** The usage-record id is now derived
on the gateway as a `UUIDv5` over the dedup key `(tenant_id, gts_id,
idempotency_key)` (`derive_usage_record_id`) and is no longer accepted in the
request body. The record-identity field is renamed `uuid` → `id` across the SDK,
DTOs, and OpenAPI schema. ADR-0013 documents the derivation scheme and why the
namespace must never change.

**Per-kind aggregation-op restriction.** Aggregation operations are constrained
by record kind: an `AggregationOp::is_allowed_for` matrix, an
`OP_NOT_ALLOWED_FOR_KIND` validation reason and matching SDK error, and a
`require_op_allowed_for_kind` query guard that rejects a disallowed op/kind
combination *before* the request is dispatched to a plugin.

**Fix:** correct aggregation `orderby` handling.

Design docs, feature specs, domain model, plugin SPI, SDK trait, and the OpenAPI
schema (`usage-collector-v1.yaml`) are updated to match.

## Type of Change
- [ ] Documentation update

## Testing
- [x] Unit tests pass 
- [x] Integration tests pass

## Documentation
- [ ] README updated (if applicable) — n/a
- [x] API documentation updated (OpenAPI `usage-collector-v1.yaml` + gear design/feature docs, ADR-0013)

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **New Features**
  * Usage record identifiers are now deterministic and gateway-derived (UUIDv5 from the dedup key) to strengthen idempotent retries.
  * Aggregation responses now use higher-precision numeric values for bucket results.
* **Bug Fixes**
  * Record creation no longer accepts client-supplied identity; stray IDs are rejected with `400`.
  * Pagination/sorting and cursors consistently use `(created_at, id)`.
  * Aggregation operator/kind mismatches are validated up-front (returns `400`), and idempotency conflicts report `existing_id`.
* **Documentation**
  * Updated REST/OpenAPI, SDK, and plugin/spi references from `uuid` → `id`, plus aggregation typing/value encoding.
* **Tests**
  * Updated/expanded SDK and REST handler tests for deterministic IDs and BigDecimal aggregation behavior.


gears/system/usage-collector/usage-collector/src/domain/service.rs
```diff
@@ -1180,12 +1198,22 @@ impl Service {
         // its only scan bound. Runs after authz (PDP-first posture).
         require_bounded_time_window(query)?;
 
+        let plugin = self.get_plugin().await.map_err(UsageCollectorError::from)?;
+
+        // Resolve the queried usage type before dispatch: existence (an
+        // unregistered `gts_id` lifts the plugin's `UsageTypeNotFound` to a
+        // pre-dispatch `404`) AND `kind`, so a mismatched `(op, kind)` pair is
+        // rejected as a typed `400` here and the plugin stays pure-persistence.
+        let usage_type = plugin
```
Artifizer: **MEDIUM**

Every aggregation query now issues an extra `get_usage_type` plugin round-trip before dispatch, roughly doubling backend calls on a hot metering path.

High-QPS aggregation endpoints pay a second network/DB hop per request purely to resolve `kind` for the op-per-kind guard.

Fold the kind resolution into the aggregate dispatch (return `kind` alongside the result) or cache the immutable `gts_id`→`kind` mapping; if the tradeoff is intentional, record it in the DESIGN performance section.

gears/system/usage-collector/usage-collector/src/api/rest/handlers/usage_records.rs
```diff
@@ -423,21 +435,40 @@ fn prepare_list_query(mut query: ODataQuery) -> Result<ODataQuery, CanonicalErro
...
+    // — never pairing a descending caller order with an ascending
+    // tiebreaker, which the plugin rejects as a mixed-direction keyset.
+    //
+    // Skipped when a cursor is present: the toolkit OData extractor leaves
+    // `order` empty on a cursor request (and rejects `$orderby` + `cursor`
+    // together), so the effective keyset order is reconstructed from the
+    // cursor's signed tokens in step 3 instead — and those tokens already
+    // carry the suffix minted into the cursor on the first page.
+    if query.cursor.is_none() {
+        let dir = query
```
Artifizer, line 484: **MEDIUM**

The tiebreaker direction is taken from the caller's trailing sort key, so a mixed-direction `$orderby` (e.g. `created_at asc, value desc`) produces a non-uniform keyset (`... id desc`) that the plugin rejects downstream.

The caller receives a late, non-specific plugin error instead of a typed gateway `400` naming the real cause (mixed sort directions).

Detect non-uniform directions here in `prepare_list_query` and reject up front with a typed `InvalidArgument`, or document that mixed-direction orders are unsupported and map them to a clear validation error.

gears/system/usage-collector/usage-collector/src/domain/service.rs
capybutler: Recorded the tradeoff in the DESIGN performance section (§3.11.1) rather than adding a cache.

Caching the `gts_id`→`kind` mapping would conflict with the gear's documented no-cache posture — §3.11.1 states the collector "holds no gear-owned cache … the usage-type catalog SoR is the plugin-owned `usage_type_catalog` reached per call through the Plugin SPI." Keeping the plugin as the single catalog SoR is deliberate, so I'd rather not introduce a gateway-side cache here.

The lookup itself isn't really "extra" relative to design intent: existence validation on every aggregation query is already required, and this SPI read satisfies that — the `require_op_allowed_for_kind` guard just reads `kind` off the same response at no additional cost. Net new cost per aggregate call is one catalog read, within the aggregated-query p95 budget (§3.11.2).

If aggregation QPS ever dominates plugin load, the preferred lever is folding `kind` into the aggregate SPI response (your first suggestion) rather than a gateway cache — noted as such in §3.11.1.

gears/system/usage-collector/usage-collector/src/api/rest/handlers/usage_records.rs
capybutler: Done — took the first option. `prepare_list_query` now detects a non-uniform caller order and rejects it up front with a typed `InvalidArgument` 400 naming the cause (mixing `asc`/`desc` is unsupported because the keyset requires a uniform direction), before any tiebreaker is appended. So the caller no longer gets a late, non-specific plugin error.

Added two tests: `mixed_direction_orderby_is_rejected_as_invalid_argument` and `uniform_multi_key_orderby_is_accepted_and_tiebroken` (the latter guards against over-rejecting legitimate uniform multi-key orders).

gears/system/usage-collector/usage-collector-sdk/src/id_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::id::{USAGE_RECORD_ID_NAMESPACE, derive_usage_record_id};
+use crate::models::{IdempotencyKey, UsageTypeGtsId};
+
+fn tenant() -> Uuid {
+    Uuid::parse_str("11111111-1111-1111-1111-111111111111").unwrap()
+}
+fn gts() -> UsageTypeGtsId {
+    // Must be a valid derived GTS instance id: the segment after `~` is itself
+    // a full vendor.package.namespace.type.vMAJOR[.MINOR] chain.
+    UsageTypeGtsId::new("gts.cf.core.uc.usage_record.v1~cf.mini_chat._.tokens_consumed.v1").unwrap()
```
aviator5: @capybutler could you please use `gts_id!` macro from the toolkit-gts crate for defining GTS identifiers/patterns in code? Additionally, I would suggest using gts-id crate for parsing.

capybutler: Done — adopted `gts_id!` for every in-code GTS literal, not just this line:

- Replaced the raw `UsageTypeGtsId::new("gts.…")` strings with prefix-free
  `gts_id!("cf.core.uc.usage_record.v1~…")` across the SDK tests
  (`id_tests.rs`, `models_tests.rs`) and the gateway tests (`query_tests.rs`,
  `service_tests.rs`). The `dto_tests.rs` create-request fixture now references
  the file's existing `gts_id!`-backed `SAMPLE_USAGE_TYPE_ID` const instead of a
  duplicated literal.
- Deliberately left the negative-path builders raw
  (`format!("{GTS_ID_PREFIX}…invalid…")`, `"not-a-valid-prefix"`): those assert
  `UsageTypeGtsId::new` *rejects* malformed input, and `gts_id!` would reject
  them at compile time.

On parsing: the production path already routes through the crate —
`UsageTypeGtsId::new` (and its custom `Deserialize`) parse via `GtsId::try_new`,
and `derive_usage_record_id` intentionally hashes the already-validated id's
canonical bytes rather than re-parsing. The only spot still doing string surgery
was the `uc_permission_instances()` test helper (`strip_prefix` + `starts_with`);
switched it to parse with `GtsId` and match on `get_type_id()` plus the concrete
segment's `vendor()`/`package()`/`namespace()`.

gears/system/usage-collector/docs/ADR/0013-deterministic-usage-record-id.md
```diff
@@ -0,0 +1,121 @@
...
+`status`), and the wire request omits it likewise (`deny_unknown_fields`
+rejects a stray value with `400`). `CreateUsageRecord::into_usage_record` is the
+sole point that stamps the derived `id` and the initial `status = active`. On
+the persisted `UsageRecord` the identity field is renamed `uuid → id`.
+
+### Consequences
+
+- One key ⇒ one `id`; distinct keys ⇒ distinct `id`s. Duplicate-`id` rows from
+  distinct keys are impossible by construction, so the by-id `get` /
+  `deactivate` surfaces operate on a unique identity.
```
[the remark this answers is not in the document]
capybutler: Reworded to drop the injectivity claim. The Consequences bullet now states UUIDv5 is a truncated SHA-1 digest — collision-resistant but not injective — so distinct keys are overwhelmingly likely (not guaranteed) to yield distinct ids, with a collision being cryptographically negligible (~2⁻¹²²) rather than "impossible by construction." The Confirmation section's "injectivity" is now "distinctness across the pinned key vectors" (the test asserts distinct outputs over a finite golden-vector set — that's all it can assert).

Not adding a collision-handling policy: by-id uniqueness isn't a runtime-defended hard invariant here, and a ~2⁻¹²² event on a pre-release gear doesn't justify a collision-handling path (it'd be dead code). The docs now mark the residual probability out of scope. If it ever becomes load-bearing, the plugin's storage-level uniqueness constraint on id is the enforcement point, not gateway logic.

gears/system/usage-collector/docs/features/usage-emission.md
```diff
@@ -310,7 +312,7 @@ This feature exists so that at-least-once delivery of usage records is uniformly
...
+**Input**: an inbound `POST /usage-collector/v1/records` REST request carrying the gateway-resolved `Extension<SecurityContext>`, audit-correlation headers, and a per-record attribution tuple lifted directly off each `UsageRecord` payload — caller-supplied `tenant_id` (never derived from the inbound `SecurityContext`), `ResourceRef`, optional `SubjectRef` (its mandatory `subject_id` plus the optional `subject_type` qualifier when populated), UsageType `gts_id`; OR an SDK `UsageCollectorClientV1::create_usage_record(ctx, ...)` / `create_usage_records(ctx, ...)` invocation carrying `ctx: &SecurityContext` as the first parameter and an equivalent attribution tuple lifted off each identity-free `CreateUsageRecord`.
```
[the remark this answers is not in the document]
capybutler: Fixed. The REST-boundary input in the attribution/PDP algorithm now reads UsageRecordInput (the wire create shape) instead of UsageRecord, matching the batch contract at line 198 (UsageRecordInput on the wire / CreateUsageRecord on the SDK) and the OpenAPI schema. Also corrected the same slip in the semantics-enforcement step inst-algo-semantics-read-input-v2, which read fields "from the inbound UsageRecord payload." UsageRecord is now reserved for the normalized persisted model.

gears/system/usage-collector/docs/plugin-spi.md
```diff
@@ -817,19 +827,29 @@ Taxonomy".
 
   Trace context is ambient (active `tracing::Span` / OpenTelemetry context); no explicit context parameter.
 - Structural inputs reaching the SPI:
-  - `gts_id: UsageTypeGtsId` — the typed usage-type key. The gateway has already validated usage-type existence and resolved the declared `metadata_fields`; the plugin lowers this to its `gts_id` column filter (`WHERE gts_id = $1`). The `gts_id` field on the OData filter surface is reserved and is guaranteed by the gateway to be absent from `query`.
+  - `gts_id: UsageTypeGtsId` — the typed usage-type key. The gateway has already validated usage-type existence AND resolved its `kind` via a pre-dispatch `get_usage_type`, and has rejected any `(op, kind)` pair the kind does not admit; the plugin lowers `gts_id` to its `gts_id` column filter (`WHERE gts_id = $1`). The `gts_id` field on the OData filter surface is reserved and is guaranteed by the gateway to be absent from `query`.
   - `window: TimeWindow` — validated `[from, to)` UTC range.
```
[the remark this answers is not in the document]
capybutler: Confirmed, and broader than the flagged line. The TimeWindow type and the window parameter are gone from the code entirely — the mandatory [from, to) window is now a bounded created_at ge … and created_at lt … predicate inside $filter (enforced by require_bounded_time_window; an unbounded window is MISSING_TIME_WINDOW/400), and created_at is a first-class UsageRecordFilterField. The old "filter MUST NOT carry created_at" prose was the exact inversion of current behavior.

Fixed everywhere the stale signature/prose appeared, not just here:

plugin-spi.md — Method 3 & 4 SPI signatures + input tables/bullets
sdk-trait.md — Method 3 & 4 signatures, tables, validation rules, type lists
usage-collector-v1.yaml — removed the time_range field and the TimeWindow schema from AggregationRequest; documented the mandatory created_at window in both endpoints' $filter
domain-model.md, features/usage-query.md, DECOMPOSITION.md — remaining TimeWindow type/param references

PR #4189, opened 2026-07-07
feat(cluster): standalone backend plugin and cluster gear wiring

asmith987:
Add cf-gears-standalone-cluster-plugin, the zero-infrastructure dev/test deployment for the cluster gear: a native in-process cache provider plus the SDK default leader-election, lock, and service-discovery backends. Builds against cluster-sdk with the `otel` feature so the standalone backend emits cluster metrics.

Add the cf-gears-cluster wiring crate: gear definition, profile-based configuration, provider registry, and ClientHub wiring that defaults to the standalone in-process backend plugin. Include the cluster-sdk examples (single/multi primitive, multi-profile, capability mismatch, observability, plugin-author) and the cf-studio "Nested SDK Examples" artifact autodetect rule.

## Summary by CodeRabbit

* **New Features**
  * Added config-driven cluster wiring with profile-based primitive selection (cache required; others optional) and fluent builder lifecycle.
  * Added a standalone in-process cluster backend and provider (`standalone`) for local usage/testing.
  * Added/expanded examples demonstrating single-primitive, multi-primitive, multi-profile, observability, and plugin authoring flows.

* **Bug Fixes**
  * Improved registration rollback on startup failures and strengthened shutdown/unbinding behavior (including stop/drop safety).

* **Documentation**
  * Expanded cluster and standalone plugin READMEs, plus updated traceability audit references.

### DIFF
--- a/gears/system/cluster/cluster/src/config.rs
+++ b/gears/system/cluster/cluster/src/config.rs
@@ -1,0 +1,104 @@
+//! Operator YAML schema for the cluster gear (DESIGN §3.4 / §3.11).
+//!
+//! [`ClusterConfig`] is the operator-facing contract: a map of named profiles,
+//! each binding the four coordination primitives to a backend `provider`. The
+//! `cache` binding is the required anchor; the other three may be omitted to ride
+//! the SDK default backends over that profile's cache
+//! (`cpt-cf-clst-fr-routing-omit-default`), or bound to their own provider for
+//! per-primitive routing (`cpt-cf-clst-fr-routing-per-primitive`).
+//!
+//! These types are serde-deserializable (typically via `ctx.config()` in a host
+//! gear, fed by `serde-saphyr`). They live in the wiring crate, not the SDK — the
+//! SDK coordination contract stays serde-free per `cpt-cf-clst-constraint-no-serde`.
+//!
+//! Per-provider options are **flattened** into the backend binding and parsed by
+//! the provider itself (see [`crate::provider::ClusterCacheProvider`]), so adding
+//! a backend is a new crate plus config, not a schema change here.
+
+use serde::Deserialize;
+
+/// The whole cluster section of operator YAML: a set of named profiles.
+///
+/// ```yaml
+/// cluster:
+///   profiles:
+///     default:
+///       cache: { provider: standalone }
+///       # leader_election / lock / service_discovery omitted → SDK defaults
+/// ```
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ClusterConfig {
+    /// Profile name → per-primitive backend bindings. Profile names must conform
+    /// to the cluster name rule (`[a-zA-Z0-9_-]+`); the wiring validates this at
+    /// registration time.
+    #[serde(default)]
+    pub profiles: std::collections::BTreeMap<String, ProfileConfig>,
+}
+
+/// The per-primitive backend bindings for one profile.
+///
+/// `cache` is required (it is the omit-default anchor). Each of the other three
+/// primitives may be bound to its own provider or omitted; an omitted primitive
+/// is auto-filled with the SDK default backend over this profile's cache.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ProfileConfig {
+    /// The cache backend — required. Serves as the anchor the SDK default
+    /// leader-election, lock, and service-discovery backends wrap when those
+    /// primitives are omitted.
+    pub cache: BackendBinding,
+    /// An explicit leader-election backend. Omit to use the SDK default over the
+    /// cache.
+    #[serde(default)]
+    pub leader_election: Option<BackendBinding>,
+    /// An explicit distributed-lock backend. Omit to use the SDK default over the
+    /// cache.
+    #[serde(default)]
+    pub lock: Option<BackendBinding>,
+    /// An explicit service-discovery backend. Omit to use the SDK default over the
+    /// cache.
+    #[serde(default)]
+    pub service_discovery: Option<BackendBinding>,
+}
+
+/// One primitive's binding to a backend `provider`, plus that provider's own
+/// options (flattened) and an optional credential reference.
+///
+/// The known keys are `provider` and `secret_ref`; every other key is captured
+/// into [`options`](Self::options) verbatim for the provider to parse. This keeps
+/// the schema open: a new backend defines its own option keys without changing
+/// this struct.
+#[derive(Debug, Clone, Deserialize)]
+pub struct BackendBinding {
+    /// The backend provider name, e.g. `standalone`, `postgres`, `redis`,
+    /// `k8s-lease`. Matched against the registered providers at wiring time; an
+    /// unknown provider fails startup with `ClusterError::InvalidConfig`.
+    pub provider: String,
+    /// A provisional, OPEN reference to the credential the backend uses to reach
+    /// its infrastructure (DESIGN §3 open question — credential wiring is deferred
+    /// to the OOP deployment design). Placeholder shape only; not a committed
+    /// contract. Ignored by the in-process standalone provider.
+    #[serde(default)]
+    pub secret_ref: Option<SecretRef>,
+    /// Provider-specific options captured verbatim (Design A: flattened options).
+    /// The provider deserializes the keys it understands from this map.
+    #[serde(flatten)]
+    pub options: serde_json::Map<String, serde_json::Value>,
+}
+
+/// Provisional placeholder for a backend credential reference (DESIGN §3 open
+/// question, deferred to the OOP deployment design). The concrete resolution —
+/// credstore lookup, K8s service-account fallback, rotation — is intentionally
+/// unspecified here.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SecretRef {
+    /// An opaque name the future credential layer will resolve. Treated as an
+    /// opaque string for now.
+    pub name: String,
+}
+
+#[cfg(test)]
+#[path = "config_tests.rs"]
+mod config_tests;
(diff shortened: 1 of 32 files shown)


asmith987:
@coderabbit

asmith987:
@coderabbitai review

gears/system/cluster/cluster-sdk/Cargo.toml
```diff
@@ -57,4 +57,6 @@ opentelemetry = { workspace = true, optional = true }
 # /`advance` and `#[tokio::test(start_paused = true)]` so renewal and TTL logic
 # is driven deterministically without real sleeps.
 tokio = { workspace = true, features = ["test-util"] }
+cf-gears-cluster = { path = "../cluster" }
```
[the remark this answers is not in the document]
asmith987: Fixed by moving the SDK's showcase examples (and their shared fixture module) from `cluster-sdk/examples/` into `cluster/examples/`, since that's where the concrete default backends they demonstrate already live. `cluster-sdk/Cargo.toml` no longer depends on `cf-gears-cluster` at all, so the cycle is gone entirely rather than just being a permitted dev-dependency cycle.

striped-zebra-dev (approved):

gears/system/cluster/cluster/src/config.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub provider: String,
+    /// A provisional, OPEN reference to the credential the backend uses to reach
+    /// its infrastructure (DESIGN §3 open question — credential wiring is deferred
+    /// to the OOP deployment design). Placeholder shape only; not a committed
+    /// contract. Ignored by the in-process standalone provider.
+    #[serde(default)]
+    pub secret_ref: Option<SecretRef>,
+    /// Provider-specific options captured verbatim (Design A: flattened options).
+    /// The provider deserializes the keys it understands from this map.
+    #[serde(flatten)]
```
MikeFalcon77, line 86: **MEDIUM**

The `#[serde(flatten)]` open schema for provider `options` captures every unrecognized key verbatim, and each provider reads only the keys it knows.

An operator typo in an option key (e.g. `sweep_interval_m` instead of `sweep_interval_ms`) is silently dropped with no error or log, so the intended setting never takes effect.

Deserialize `options` into a per-provider struct marked `#[serde(deny_unknown_fields)]` so unknown keys fail startup with `ClusterError::InvalidConfig`.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -0,0 +1,434 @@
...
+    ///   provider for any primitive, or if a provider rejects its options.
+    /// - Propagates [`ClusterError`] from provider construction, the SDK default
+    ///   backends (consistency guard), and backend registration (invalid name).
+    pub async fn from_config(
+        hub: Arc<ClientHub>,
+        config: &ClusterConfig,
+        providers: &ProviderRegistry,
+    ) -> Result<ClusterHandle, ClusterError> {
+        let mut builder = Self::builder(hub);
+        for (name, profile) in &config.profiles {
```
MikeFalcon77, line 116: **MEDIUM**

The cluster wiring lifecycle (per-profile backend construction, the partial-registration rollback that deregisters already-registered profiles, and `ClusterHandle::stop`) emits no tracing on this critical operational path.

When wiring fails midway and rolls back, or a profile is registered or stopped, operators have no log trail to diagnose it.

Add structured `tracing` spans/events (profile name, primitive) around profile wiring, a `warn` when rollback deregisters earlier profiles, and `info`/`debug` on registration and shutdown.

gears/system/cluster/plugins/standalone-cluster-plugin/src/cache.rs
```diff
@@ -0,0 +1,487 @@
...
+                key,
+                &CacheEvent::Changed {
+                    key: key.to_owned(),
+                },
+            );
+        }
+        outcome
+    }
+
+    async fn compare_and_delete(
```
MikeFalcon77, line 430: **MEDIUM**

`compare_and_delete` is not exercised by any test, so its value-match delete, value-mismatch no-op, and absent/expired no-op branches are entirely uncovered.

This is a documented conditional-delete safety property that could regress silently.

Add cache tests for `compare_and_delete` covering a matching value (deletes and emits `Deleted`), a mismatching value (no-op, returns false), and an expired/absent key (no-op).

gears/system/cluster/plugins/standalone-cluster-plugin/src/cache.rs
```diff
@@ -0,0 +1,487 @@
...
+                        let stored = Stored {
+                            value: new_value.to_vec(),
+                            version,
+                            expires_at: ttl.as_duration().map(|d| now + d),
+                        };
+                        let entry = stored.entry();
+                        guard.map.insert(key.to_owned(), stored);
+                        Ok(entry)
+                    } else {
+                        Err(ClusterError::CasConflict {
```
MikeFalcon77, line 407: **MEDIUM**

`compare_and_swap`'s version-match increment, version-mismatch `CasConflict`, and absent-key `CasConflict` branches have no direct test; only the post-shutdown rejection path is directly tested.

The success path is exercised only indirectly via leader election, so a regression in the CAS conflict logic would not be caught.

Add direct cache tests asserting `compare_and_swap` increments on a matching `expected_version` and returns `CasConflict` with the current entry on a version mismatch and on an absent key.

gears/system/cluster/cluster/src/wiring_tests.rs
```diff
@@ -0,0 +1,296 @@
...
+        .with_lock(lock)
+        .with_service_discovery(discovery);
+
+    let handle = ClusterWiring::builder(Arc::clone(&hub))
+        .profile(EventBroker, backends)
+        .on_stop(move || async move { plugin.stop().await })
+        .build_and_start()
+        .expect("wiring starts");
+
+    assert!(
```
MikeFalcon77: **MEDIUM**

`explicit_backends_override_defaults` builds explicit leader/lock/SD backends but its only assertion is that a lock resolves `is_ok()`, which would pass identically if the explicit backends were ignored and the SDK defaults used.

The test does not verify the override behavior it is named for, so a bug that drops explicit backends would not fail it.

Assert an observable property that distinguishes the explicitly-bound backend from the default (e.g. a marker or behavior on a custom backend).

gears/system/cluster/cluster/src/gear_tests.rs
```diff
@@ -0,0 +1,96 @@
...
+                .profile(DefaultProfile)
+                .resolve(),
+            Err(ClusterError::ProfileNotBound { .. })
+        ),
+        "stop deregisters the profile's backends"
+    );
+}
+
+#[tokio::test]
+async fn gear_with_no_config_starts_empty() {
```
MikeFalcon77, line 78: **LOW**

`gear_with_no_config_starts_empty` claims start "binds nothing" but only asserts init/start/stop return `Ok`, never that no profile is bound.

It would pass even if `start` erroneously registered a backend.

After start, resolve a profile and assert `ClusterError::ProfileNotBound` to confirm the empty config binds nothing.

gears/system/cluster/plugins/standalone-cluster-plugin/src/plugin.rs
```diff
@@ -0,0 +1,183 @@
...
+    ///
+    /// The cache `shutdown()` runs **first** so any active watch observes a
+    /// terminal `Closed(Shutdown)` (`cpt-cf-clst-fr-shutdown-revoke`, DESIGN
+    /// §3.13) before the sweeper stops and the cache is dropped. TTL bounds any
+    /// remaining in-flight cluster resources (DESIGN §3.7) — there is no
+    /// best-effort remote cleanup.
+    pub async fn stop(self) {
+        self.cache.shutdown();
+        self.shutdown.cancel();
+        let _exited = self.sweeper.await;
```
MikeFalcon77: **LOW**

The sweeper `JoinHandle` result is discarded via `let _exited = self.sweeper.await;`, so a panic in the background TTL sweeper (surfaced as a `JoinError`) is silently swallowed during shutdown.

A panicked sweeper would leave no trace, hiding a real defect.

Match on the await result and emit a `tracing::warn!`/`error!` when it is `Err(join_err)`.

gears/system/cluster/cluster/src/config.rs
asmith987: Fixed. Applied the suggested fix

gears/system/cluster/cluster/src/wiring.rs
asmith987: Fixed. Added tracing throughout profile registration, rollback and ClusterHandle::stop

gears/system/cluster/plugins/standalone-cluster-plugin/src/cache.rs
asmith987: Fixed. Added tests

gears/system/cluster/plugins/standalone-cluster-plugin/src/cache.rs
asmith987: Fixed. Added direct tests that were missing

gears/system/cluster/cluster/src/wiring_tests.rs
asmith987: Fixed. Added a custom backend that counts calls to prove the explicitly-bound instances actually receive calls.

gears/system/cluster/cluster/src/gear_tests.rs
asmith987: Fixed. Added assertion to the test

gears/system/cluster/plugins/standalone-cluster-plugin/src/plugin.rs
asmith987: Fixed. Matches on the result and logs a `tracing::error` when there is an `Err`

PR #4204, opened 2026-07-08
Port stateful credstore gear

diffora:
## Summary

Ports the stateful **credstore gateway** into this workspace. The gateway gear now owns per-secret **metadata** in its own DB, while the backend plugin becomes a **value-only** store. Adds hierarchical tenant resolution with PDP `AccessScope` enforced in SQL, optimistic concurrency (`ETag`/`If-Match`), a crash-safe write saga + background reaper, a value-fingerprint fence binding each stored value to its metadata row, OTel metrics, and a REST API under `/cf/credstore/v1/secrets`. Consumers (mini-chat, oagw) are adapted to the new runtime-provisioning contract.

## credstore

- **SDK** — `CredStorePluginClientV1` is now a pure per-tenant value store (`get`/`put`/`delete` by `tenant_id`/`key`/`owner_id`); the gateway client gains write ops (`put`/`create`/`delete`); `GetSecretResponse` gains `version`; registry-driven GTS secret typing.
- **Gateway** (`gears/credstore/credstore`) — `api/rest`, `infra/storage` (sea-orm migration `m0001`, SecureORM with PDP-in-SQL), `domain/secret` (service, write saga, reaper, fence, typing), tenant resolver with TTL/LRU ancestor-chain cache + isolation barriers, versioning, canonical error mapping.
- **Value-fingerprint fence** (`gears/credstore/docs/features/001-value-fingerprint-fence.md`, ADR 0003) — `HMAC-SHA256(fence_key, value)` binds a backend value to its metadata row; a crosswise last-writer-wins interleave fails closed (404) instead of a cross-tenant disclosure. Crypto runs through **AWS-LC (`aws-lc-rs`)**, FIPS-validated under `--features fips`.
- **static-credstore-plugin** — reduced to a value-only development backend.
- **ADRs** — stateful gateway (0001), deprovisioning saga (0002), value-fingerprint fence (0003).

## Consumers

- **mini-chat** — OAGW upstream registration now **defers and retries in the background** when a provider secret is not yet provisioned at boot (with the stateful gateway, secrets are created at runtime via the API), instead of failing `Gear::init()`. Adds unit tests for the register / defer / reuse / reconcile / paging paths.
- **oagw** — test-mock updates only, to satisfy the new SDK trait methods (no production changes).

## Config / e2e

- Registers the credstore gear across the run configs.
- New e2e suites for credstore (CRUD, sharing, types, concurrency) plus mini-chat / oagw fixtures that provision provider secrets through the gateway API at runtime.

## Correctness hardening

- Status-gated reaper: `reap_by_id` is guarded by the observed status and claims a provisioning row before touching the backend, so a slow create's `mark_active` can no longer lose its row + value to the reaper.
- Private-secret resolution pins `tenant_id == req` (own-tenant only), rather than relying on an unenforced authn invariant.
- Fence-key cache: cooldown-gated, single-flighted refresh that no longer evicts the shared key on every poisoned read.
- Saturating reaper cutoff (no panic on an absurd timeout config).

## Summary by CodeRabbit

* **New Features**
  * Added a credential store for creating, retrieving, updating, and deleting tenant-scoped secrets.
  * Added hierarchical sharing modes, inheritance metadata, secret types, expiry controls, and validation.
  * Added optimistic concurrency using generation-bound ETags and required `If-Match` headers.
  * Added crash-safe lifecycle handling and automatic cleanup for incomplete operations.
  * Added REST API documentation and configuration support across quickstarts and end-to-end environments.
* **Reliability**
  * Improved deferred provider registration and retry handling for Mini Chat.
* **Documentation**
  * Added quickstart guidance and architecture decision records for credential storage behavior.


docs/api/api.json
```diff
@@ -790,6 +790,43 @@
...
+      "CreateSecretRequestDto": {
+        "additionalProperties": false,
+        "description": "Request body for `POST /credstore/v1/secrets`.\n\n`Debug` is hand-written to redact `value` — a derived `Debug` would expose\nthe plaintext secret if this DTO is ever `{:?}`-logged by a future layer.",
+        "properties": {
+          "expires_at": {
+            "description": "Expiry instant (RFC 3339); only for expirable types.",
+            "type": [
+              "string",
+              "null"
+            ]
```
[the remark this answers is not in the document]
diffora: Addressed in code (spec is generated by `make openapi`; hand-editing `api.json` fails the `api_contracts` freshness check).

Added `#[schema(format = DateTime)]` to all three `expires_at` DTO fields (`CreateSecretRequestDto`, `UpdateSecretRequestDto`, `SecretMetadataDto`); the regenerated spec now carries `"format": "date-time"` on each.

docs/api/api.json
```diff
@@ -790,6 +790,43 @@
...
+          },
+          "reference": {
+            "description": "Secret reference key — `[a-zA-Z0-9_-]+`, max 255 characters.",
+            "type": "string"
```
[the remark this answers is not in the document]
diffora: Partially addressed in code.

Applied `#[schema(pattern = "^[A-Za-z0-9_-]+$", min_length = 1, max_length = 255)]` to `CreateSecretRequestDto.reference` (request body) — the regenerated spec now carries these constraints.

**Not applied to the `{ref}` path parameter:** toolkit`s `ParamSpec` (name/location/required/description/param_type) has no field to express `pattern`/`minLength`/`maxLength` on path params. That`s a framework limitation, out of scope for this PR. (Reminder: `api.json` is generated via `make openapi`, so edits must live in source.)

docs/api/api.json
```diff
@@ -10848,6 +10993,394 @@
...
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Problem"
+                }
+              }
+            },
+            "description": "Service Unavailable"
+          }
+        },
```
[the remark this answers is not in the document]
diffora: Partially addressed; some points rejected.

- **If-Match:** added as an optional `header` parameter on PUT and DELETE via `OperationBuilder`; the regenerated spec now documents it (with the `"<id>.<version>"` ETag shape and 409 semantics).
- **412 Precondition Failed:** rejected — the contract uses **`409 OPTIMISTIC_LOCK_FAILURE`**, not 412. GET/PUT/DELETE already register `error_409`.
- **`ETag` response header on GET:** not applied — toolkit`s `ResponseSpec` (status/content_type/description/schema_name) can`t express response headers. Framework limitation, tracked separately.

(Reminder: `api.json` is generated via `make openapi`; direct edits fail the `api_contracts` freshness check.)

gears/credstore/credstore-sdk/src/error.rs
```diff
@@ -27,19 +38,169 @@ impl CredStoreError {
...
+            self,
+            Self::ServiceUnavailable { .. } | Self::NoPluginAvailable
+        )
+    }
+
+    /// `true` if the operation may succeed on a future retry.
+    #[must_use]
+    pub fn is_retryable(&self) -> bool {
+        self.is_unavailable()
+    }
```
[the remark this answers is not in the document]
diffora: Done. `is_retryable()` now matches only `ServiceUnavailable`; `NoPluginAvailable` stays `is_unavailable()` (operator misconfiguration) but is no longer retryable. Added a test asserting `!NoPluginAvailable.is_retryable()`.

gears/credstore/credstore-sdk/src/models.rs
```diff
@@ -163,12 +171,32 @@ pub enum SharingMode {
...
+#[derive(Debug, Clone, Copy, Default)]
+pub struct WriteOptions {
+    /// Secret type reference (catalog type, full GTS id, or type UUID —
+    /// see [`SecretTypeRef`]); `None` keeps the existing secret's type on
+    /// overwrite and defaults to the generic type on create.
+    pub secret_type: Option<SecretTypeRef>,
+    /// Expiry instant; only permitted for types whose traits are
+    /// `expirable`. An overwrite replaces the stored expiry with this value
+    /// (including clearing it when `None`).
+    pub expires_at: Option<time::OffsetDateTime>,
```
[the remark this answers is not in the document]
diffora: Done — implemented the tri-state as suggested. Added `ExpiryWrite { Preserve (default), Set(t), Clear }`; `WriteOptions.expires_at` is now `ExpiryWrite`, so the convenience `put()` (default `Preserve`) no longer strips an existing expiry on a value rotation. The domain service resolves it against the stored row (`Preserve` keeps current, `Set` overrides, `Clear` drops) and validates only newly-`Set` instants. REST `PUT` keeps whole-value-replace semantics (omitted `expires_at` → `Clear`), which is the documented HTTP-PUT behaviour.

gears/credstore/credstore-sdk/README.md
```diff
@@ -6,8 +6,12 @@ SDK crate for the CredStore gear, providing public API contracts for credential
 
 This crate defines the transport-agnostic interface for the CredStore gear:
 
-- **`CredStoreClientV1`** — Async trait for consumers (get/put/delete secrets)
-- **`CredStorePluginClientV1`** — Async trait for backend storage plugin implementations
+- **`CredStoreClientV1`** — consumer-facing trait (`get`/`put`/`create`/`delete`);
+  `get` returns the value plus metadata (`owner_tenant_id`, `sharing`,
+  `is_inherited`, `version`)
```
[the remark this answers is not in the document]
diffora: Done — README now lists the full `get` metadata surface (`owner_tenant_id`, `sharing`, `is_inherited`, `id`, `version`, `secret_type`, `expires_at`).

gears/credstore/docs/PRD.md
```diff
@@ -287,83 +318,103 @@ If the tenant owns a secret with the same reference but it is **inaccessible** t
...
 
-### 5.4 P1 — Encryption & Key Management
+#### Optimistic Concurrency
 
-#### External Key Management
+- [ ] `p1` - **ID**: `cpt-cf-credstore-fr-optimistic-concurrency`
 
-- [ ] `p1` - **ID**: `cpt-cf-credstore-fr-external-key-mgmt`
+<!-- cpt-cf-id-content -->
+Each secret **MUST** carry a monotonic version. `GET` **MUST** return it as a strong `ETag`; `PUT` and `DELETE` **MUST** honour an optional `If-Match` precondition (`*` = must exist, `"<version>"` = version must match), enforced atomically with the metadata commit. A failed precondition surfaces as a conflict (409, `OPTIMISTIC_LOCK_FAILURE`); a malformed precondition is a validation error (400).
```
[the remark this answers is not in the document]
diffora: Done — updated both spots (GET/PUT/DELETE contract and UC-005) to the generation-bound `"<id>.<version>"` ETag shape.

.github/workflows/cfs.yml
```diff
@@ -65,8 +65,13 @@ jobs:
...
           if [ "${{ github.event_name }}" = "pull_request" ] && [ -n "$base_ref" ]; then
-            git fetch --no-tags --depth=1 origin "$base_ref"
-            changed_files="$(git diff --name-only "origin/$base_ref"...HEAD)"
+            # Full (not --depth=1) fetch so the three-dot diff can find a merge
+            # base: checkout uses fetch-depth: 0 (full HEAD), so a shallow base
+            # ref has no reachable common ancestor and the diff fails with
+            # "no merge base" (notably on fork PRs). Tolerate a failed diff too
+            # so this detection step never hard-fails the job.
+            git fetch --no-tags origin "$base_ref"
+            changed_files="$(git diff --name-only "origin/$base_ref"...HEAD 2>/dev/null || true)"
```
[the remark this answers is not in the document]
diffora: Done — guarded the fetch with `|| true`. The run shell is `bash -eo pipefail`, so the unguarded `git fetch` could indeed hard-fail the step despite the comment; it now tolerates a failed fetch just like the downstream `git diff`.

gears/credstore/docs/PRD.md
```diff
@@ -587,84 +654,124 @@ Secret values **MUST NOT** appear in logs, error messages, or debug output at an
...
+- **P2/Future — Metadata list endpoint**: a values-free list becomes cheap with gateway-owned metadata; must be reconciled with the anti-enumeration stance and per-type authorization.
+- **Dynamic type descriptors** (see DESIGN §9): resolve traits from the types-registry at runtime so vendors can add types without an SDK release.
 
 ## 14. Traceability
 
 - **Design**: [DESIGN.md](./DESIGN.md)
-- **ADRs**: [ADR/](./ADR/)
-- **Features**: [features/](./features/)
\ No newline at end of file
+- **ADRs**: [ADR/](./ADR/) — [ADR-0001 stateful gateway](./ADR/0001-cpt-cf-credstore-adr-stateful-gateway.md), [ADR-0002 deprovisioning saga](./ADR/0002-cpt-cf-credstore-adr-deprovisioning-saga.md)
```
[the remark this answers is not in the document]
diffora: Done — added ADR-0003 (value-fingerprint fence) to the traceability ADR list.

gears/credstore/credstore/src/infra/storage/repo_impl/reads.rs
```diff
@@ -0,0 +1,316 @@
...
+    tenant: Uuid,
+) -> Result<bool, DomainError> {
+    if scope.is_unconstrained() {
+        return Ok(true);
+    }
+    if scope.is_deny_all() {
+        return Ok(false);
+    }
+    // Fast path: direct UUID match via Eq/In filters.
+    if scope.contains_uuid(pep_properties::OWNER_TENANT_ID, tenant) {
```
Artifizer: **MAJOR**

`scope_includes_tenant` is the sole enforcement of the PDP `AccessScope` on the read/write paths (`resolve_for_get`/`find_own`/`find_for_write` all query with `allow_all`), yet it only interprets `OWNER_TENANT_ID` `Eq`/`In` (this fast path) and `InTenantSubtree` filters — any other filter in a returned constraint (`owner_id`, `resource_id`, `InGroup`) and any sibling AND-filter within the same constraint are silently ignored.

Because a `ScopeConstraint`'s filters are AND-ed, a scope such as `OWNER_TENANT_ID = T AND owner_id = X` returns `true` here on the tenant match alone, authorizing the secret for the whole tenant even though the PDP decision was narrower — a scope that is stricter than tenant granularity is not enforced.

Apply `.scope_with(scope)` to the authorization query so the full constraint is evaluated in SQL, or make `scope_includes_tenant` fail closed (return `false`) when it encounters any filter type it does not understand, instead of reducing every constraint to its tenant predicate.

gears/credstore/credstore-sdk/src/gts.rs
```diff
@@ -1,6 +1,34 @@
...
 
+/// GTS resource-type id for the credstore **secret** base type — the single
+/// source of truth for this string. Mirrored by [`SecretV1`]'s `type_id` and
+/// the `#[resource_error(...)]` marker in `infra::sdk_error_mapping` (a unit
+/// test there pins that marker to this constant). Enforcement authorizes
+/// against the secret's full *concrete* type, not this base: the impl crate's
+/// `domain::authz::secret_type_resource(gts_id)` builds the per-operation PEP
+/// `ResourceType` from the type resolved out of the types-registry, and the
+/// resolver checks a type descends from this base id.
+pub const SECRET_RESOURCE_TYPE: &str = "gts.cf.core.credstore.secret.v1~";
```
aviator5: Please use `gts_id!` macros everywhere from `toolkit-gts` crate.

gears/credstore/docs/PRD.md
```diff
@@ -74,34 +72,59 @@ REQUIREMENT LANGUAGE:
 
 ### 1.1 Purpose
 
-CredStore provides per-tenant secret storage and retrieval for the platform. It abstracts backend differences behind a unified API, enabling platform gears to store and access credentials without coupling to a specific storage technology.
+CredStore provides per-tenant secret storage and retrieval for the platform.
+The gateway owns all secret metadata (identity, sharing, ownership, lifecycle
```
aviator5: Let's don't use new-lines within the same paragraph - it's hard to read..

gears/credstore/docs/PRD.md
```diff
@@ -496,87 +569,81 @@ Secret values **MUST NOT** appear in logs, error messages, or debug output at an
...
+- Partner has `internal-admin-key` with sharing `private` (owned by PartnerAdmin); customer has no secret with this reference
 
 **Main Flow**:
-1. OAGW calls resolve for `internal-admin-key`, tenant_id `customer-123`
-2. Gateway walks up: Customer — phase 1 (private for OAGW): miss. Phase 2 (tenant/shared): miss.
-3. Partner — phase 1 (private for OAGW): miss (OAGW is not the owner). Phase 2 (tenant/shared): miss (no tenant/shared secret).
-4. Walk-up exhausted — system returns not-found error (404)
+1. OAGW resolves `internal-admin-key` for the customer
+2. The resolution query only matches private rows owned by the requesting subject; PartnerAdmin's row is invisible to OAGW
+3. No row matches → 404
```
aviator5: Looks like the current PRD (as almost all our PRDs) is overloaded with technical details. For example, I'm not sure that we need to describe here
- what HTTP codes should be returned and in what cases
- what endpoints (concrete HTTP methods + paths) should gear contains
- etc.

Drawback of such approach is duplication: all technical details are already described in `DESIGN.md`.

Perhaps not it's a good point to simplify `PRD.md` doc.

gears/credstore/credstore-sdk/src/gts.rs
```diff
@@ -10,3 +38,222 @@ use toolkit_gts::gts_type_schema;
...
+        toolkit_gts::inventory::submit! {
+            toolkit_gts::InventoryTypeSchema {
+                type_id: $gts_id,
+                schema_fn: $fn_name,
+            }
+        }
+    };
+}
+
+submit_secret_type_schema!(
```
aviator5: `gts_type_schema` macro submits type schemas to inventory automatically.

gears/credstore/credstore-sdk/src/gts.rs
```diff
@@ -10,3 +38,222 @@ use toolkit_gts::gts_type_schema;
...
+//
+// The derived schemas are submitted as raw `InventoryTypeSchema` entries
+// (not via `#[gts_type_schema(base = SecretV1, ...)]`): the macro's `base =`
+// path requires a payload-generic base struct (like `PluginV1<P>`), which the
+// unit `SecretV1` resource marker deliberately is not.
+
+/// Derived-type schema JSON for a catalog entry, in the same draft-07
+/// `allOf`/`$ref` shape the `gts_type_schema` macro generates, plus
+/// `x-gts-traits` mirroring the catalog descriptor.
+fn secret_type_schema_json(gts_id: &str) -> String {
```
aviator5: `secret_type_schema_json` looks redundant as well (please check https://github.com/constructorfabric/gears-rust/pull/4204/changes#r3587894438)

gears/credstore/credstore-sdk/src/api.rs
```diff
@@ -2,27 +2,90 @@ use async_trait::async_trait;
...
         ctx: &SecurityContext,
         key: &SecretRef,
     ) -> Result<Option<GetSecretResponse>, CredStoreError>;
+
+    /// Stores or updates a secret (upsert) with default [`WriteOptions`]:
+    /// the type is preserved on overwrite (`generic` on create) and the expiry
+    /// is preserved ([`ExpiryWrite::Preserve`](crate::ExpiryWrite::Preserve)),
+    /// so a value rotation never strips an existing expiry. Use
+    /// [`Self::put_opts`] to set or clear it explicitly.
+    async fn put(
```
aviator5, line 53: Please correct me if I'm wrong, but looks like an optimistic lock (`version` field, mentioned in the `DESIGN.md`) should appear somewhere here, right?

gears/credstore/credstore-sdk/src/types.rs
```diff
@@ -0,0 +1,412 @@
...
+use crate::error::CredStoreError;
+use crate::models::SharingMode;
+
+/// Enforceable traits of a secret type.
+///
+/// All fields are enforced by the gateway on write except
+/// `rotation_period_secs` (advisory) and `expirable`, which additionally
+/// gates reads (an expired secret resolves as not-found).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SecretTypeDescriptor {
```
aviator5, line 38: Did you consider using `#[derive(JsonSchema, serde::Serialize, GtsTraitsSchema)]` here?
Please check an example: https://github.com/GlobalTypeSystem/gts-rust/blob/main/gts-macros/tests/golden/traits_struct_literal.rs

gears/credstore/credstore-sdk/src/models.rs
```diff
@@ -163,12 +171,78 @@ pub enum SharingMode {
...
+}
+
+/// Options for typed writes ([`CredStoreClientV1::put_opts`](crate::CredStoreClientV1::put_opts) /
+/// [`create_opts`](crate::CredStoreClientV1::create_opts)).
+#[derive(Debug, Clone, Copy, Default)]
+pub struct WriteOptions {
+    /// Secret type reference (catalog type, full GTS id, or type UUID —
+    /// see [`SecretTypeRef`]); `None` keeps the existing secret's type on
+    /// overwrite and defaults to the generic type on create.
+    pub secret_type: Option<SecretTypeRef>,
```
aviator5: Why `SecretTypeRef` cannot always be `GtsId`?

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
```diff
@@ -0,0 +1,24 @@
...
+//! not a tenant-scoped resource. Queried only to check descendant membership.
+
+use sea_orm::entity::prelude::*;
+use toolkit_db::secure::Scopable;
+use uuid::Uuid;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel, Scopable)]
+#[sea_orm(table_name = "tenant_closure")]
+#[secure(unrestricted)]
+pub struct Model {
```
aviator5: Could you please clarify how the corresponding table is supposed to be created?

gears/credstore/docs/ADR/0001-cpt-cf-credstore-adr-stateful-gateway.md
```diff
@@ -0,0 +1,115 @@
+---
+status: accepted
+date: 2026-07-04
+---
+# ADR-0001: Stateful Gateway with Gateway-Owned Secret Metadata
```
aviator5: I would suggest stop using `gateway` word for the `credstore` gear since it's becoming to be a full-featured regular gear with pluggable backend storage.

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
```diff
@@ -22,64 +23,219 @@ use crate::config::ProviderEntry;
...
+enum ProviderRegistration {
+    /// Root upstream, route, and every tenant-override upstream are registered.
+    Registered,
+    /// Could not complete — the backend secret is not accessible yet (with the
+    /// stateful credstore it may be provisioned after boot) or a transient OAGW
+    /// error occurred. The caller should retry (see [`reconcile_deferred_upstreams`]).
+    Deferred,
+    /// A deterministic misconfiguration that cannot self-heal (logged at error).
+    /// The provider is left unavailable; retrying would not help.
+    Misconfigured,
```
aviator5, line 108: In the misconfiguration case I would suggest using [fast-fast approach ](https://en.wikipedia.org/wiki/Fail-fast_system) to detect an issue as soon as possible.

Actually, I'm not sure that it's ok that upstreams/routes registration depends on credstore availability. Did we have such dependency before this PR?

gears/system/oagw/oagw/src/infra/plugin/oauth2_client_cred_auth.rs
```diff
@@ -483,11 +483,43 @@ mod tests {
...
                     owner_tenant_id: CredstoreTenantId::nil(),
                     sharing: SharingMode::default(),
                     is_inherited: false,
+                    version: 1,
+                    secret_type: credstore_sdk::SecretType::generic().gts_id().to_owned(),
+                    expires_at: None,
                 }))
             }
+
+            async fn put(
```
aviator5: How about having centralized mock in the `credstore-sdk` crate (under `test-util` feature flag) and use it everywhere to avoid code duplication?

gears/credstore/docs/ADR/0003-cpt-cf-credstore-adr-value-fingerprint-fence.md
```diff
@@ -0,0 +1,144 @@
...
+  - [`SERIALIZABLE` isolation on the metadata write](#serializable-isolation-on-the-metadata-write)
+  - [Stamp the version into the backend value / key](#stamp-the-version-into-the-backend-value--key)
+  - [Value fingerprint in the gateway row (CHOSEN)](#value-fingerprint-in-the-gateway-row-chosen)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-cf-credstore-adr-value-fingerprint-fence`
+
+## Context and Problem Statement
```
aviator5, line 27: Could you please explain why we cannot always use optimistic lock to avoid data races?

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
+drop; request/response DTOs carry hand-written redacted `Debug`; the `GET`
+response sets `Cache-Control: no-store`; non-UTF-8 values are rejected with a
+typed error rather than lossily decoded.
+
+#### Canonical Error Model
+
+- [ ] `p1` - **ID**: `cpt-cf-credstore-constraint-canonical-errors`
+
+All trait-boundary and REST errors follow the platform canonical error model
+(ADR 0005): domain errors map to canonical categories with stable `reason`
```
aviator5: Suggest using full path to ADR docs, otherwise it's unclear

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
```diff
@@ -22,64 +23,219 @@ use crate::config::ProviderEntry;
...
+    ctx: &toolkit_security::SecurityContext,
+    provider_id: &str,
+    entry: &ProviderEntry,
+) -> Option<oagw_sdk::Upstream> {
+    match create_upstream(gateway, ctx, provider_id, entry).await {
+        Ok(u) => Some(u),
+        Err(CanonicalError::AlreadyExists { resource_name, .. }) => {
+            reuse_existing_upstream(gateway, ctx, provider_id, resource_name.as_deref()).await
+        }
+        Err(e) => {
```
aviator5: Correct me, but now any upstram registration fail will lead to just warning in logs + retries. In my view it's not a reliable solution..

gears/credstore/docs/DESIGN.md
striped-zebra-dev, line 1: Would be nice to see `openapi.yaml` for the service.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
+|--------|------|---------|-------------|
+| `POST` | `/credstore/v1/secrets` | `201` + `Location` | Create-only; `409` if the same sharing class already holds the reference |
+| `PUT` | `/credstore/v1/secrets/{ref}` | `204` | Upsert; honours `If-Match` |
+| `GET` | `/credstore/v1/secrets/{ref}` | `200` + `ETag`, `Cache-Control: no-store` | Hierarchical read |
+| `DELETE` | `/credstore/v1/secrets/{ref}` | `204` | Delete own secret; honours `If-Match` |
+
+**Create Secret Request (`POST`)**:
 ```json
 {
   "reference": "partner-openai-key",
```
striped-zebra-dev, line 357: Would be nice to clarify the format of the reference, I expected it to be some GTS.
On the other hand, if in some cases reference will be user-faced, requiring a GTS identifier feels too restrictive.
Something that this design should address.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
   "value": "demo-secret-value-123",
```
striped-zebra-dev, line 358: Can secret contain binary? A certificate?

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
-  "sharing": "tenant"
+  "sharing": "tenant",
+  "type": "api-key"
 }
 ```
 
-**Sharing mode values**: `"private"` (owner-only), `"tenant"` (tenant-wide, default), `"shared"` (hierarchical)
+**Update Secret Request (`PUT`)** — same body without `reference`.
+`sharing` values: `"private"`, `"tenant"` (default), `"shared"`.
```
striped-zebra-dev: In oagw we consider the following options:
```
| Parent Sharing | Child Specifies | Effective Limit                             |
|----------------|-----------------|---------------------------------------------|
| `private`      | any             | Child's limit only                          |
| `inherit`      | none            | Parent's limit                              |
| `inherit`      | own limit       | `min(parent, child)`                        |
| `enforce`      | any             | `min(parent, child)` - cannot exceed parent |

```
Reads better to me.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
+**Get Secret Response (200)**:
 ```json
 {
   "value": "demo-secret-value-456",
   "metadata": {
-    "owner_tenant_id": "partner-acme",
+    "owner_tenant_id": "22222222-2222-2222-2222-222222222222",
     "sharing": "shared",
-    "is_inherited": true
+    "is_inherited": true,
```
striped-zebra-dev, line 378: Does the client needs to know inheritance story? In what scenario?

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1074 @@ NOT IN THIS DOCUMENT (see other templates):
...
-    GW --> CS["VendorA Credstore<br/>(Go svc)"]
-    GW --> OAuth["OAuth/OIDC<br/>Provider"]
+### 4.7 Database schemas & tables
+
+The gateway owns one table, `credstore_secrets` (migration
+`m0001_initial_schema`; raw per-backend SQL to preserve `CHECK` and
+partial-index semantics; PostgreSQL and SQLite; MySQL fails fast):
+
+```sql
+CREATE TABLE credstore_secrets (
```
striped-zebra-dev, line 495:
```sql
CREATE TYPE sharing AS ENUM (
    'private',
    'tenant',
    'shared'
);
...
CREATE UNIQUE INDEX uq_credstore_nonprivate ON credstore_secrets (tenant_id, reference) WHERE sharing <> 'private';
...
```

gears/credstore/credstore/src/infra/storage/repo_impl/reads.rs
diffora: Fixed — the gate now fails closed: a constraint admits a tenant only when *every* filter in
it is an `OWNER_TENANT_ID` predicate satisfied by that tenant; any sub-tenant sibling
(`owner_id`/`resource_id`/group) or unrecognized filter makes the constraint non-admitting
(constraints still OR-ed). It stays a tenant-membership gate (row + type are resolved before
the PDP scope, and GET authorizes on the caller's tenant for inherited secrets), so
`.scope_with` can't replace it. Added tests for the sibling-`owner_id`, non-tenant-property,
and OR-alternative cases.

gears/credstore/credstore-sdk/src/gts.rs
diffora: Done — every hard-coded `"gts…"` literal now goes through `gts_id!("<suffix>")` (or
`format!("{GTS_ID_PREFIX}…")` for the two intentionally-invalid test strings). Clean under
the DE0904 `no_hardcoded_gts_prefix` dylint.

gears/credstore/docs/PRD.md
diffora: Done — reflowed the PRD to one line per paragraph (folded in with the C4 pass).

gears/credstore/docs/PRD.md
diffora: Agreed — stripped the transport/mechanism detail from the PRD (HTTP status codes, concrete
methods/paths, `ETag`/`If-Match`/`Cache-Control`/`Location`) and kept it at the WHAT/WHY
level with capability/outcome language; the concrete contract lives in DESIGN.md. Requirement
IDs, actors, and rationale unchanged. Added a PRD authoring note to keep mechanism detail in
DESIGN going forward.

gears/credstore/credstore-sdk/src/gts.rs
diffora: Done — dropped the hand-rolled `secret_type_schema_json` builder and the
`submit_secret_type_schema!` macro; the 10 derived types are now
`#[gts_type_schema(base = SecretV1, …)]` unit structs, so the macro emits and submits their
schemas automatically. Traits still come from `SECRET_TYPE_CATALOG` (single source of truth),
pinned by `every_catalog_type_has_a_registered_schema_with_matching_traits`. Tests green,
dylint clean.

gears/credstore/credstore-sdk/src/gts.rs
diffora: Done — dropped the hand-rolled `secret_type_schema_json` builder and the
`submit_secret_type_schema!` macro; the 10 derived types are now
`#[gts_type_schema(base = SecretV1, …)]` unit structs, so the macro emits and submits their
schemas automatically. Traits still come from `SECRET_TYPE_CATALOG` (single source of truth),
pinned by `every_catalog_type_has_a_registered_schema_with_matching_traits`. Tests green,
dylint clean.

gears/credstore/credstore-sdk/src/api.rs
diffora: Right, it was asymmetric — `version` was read-exposed but the write precondition existed only
on REST (`If-Match`). Added `WritePrecondition { Exists, Matches { id, version } }` to the
SDK, an optional `precondition` on `WriteOptions`, and `delete_opts(…, precondition)` (with
`delete` delegating); the local client maps them to the domain `WritePrecondition` the service
already honours. Added a client test (stale validator → `Conflict`, current → success).

gears/credstore/credstore-sdk/src/types.rs
diffora: That derive already lives on `SecretTypeTraits` — the serializable trait carrier
(`#[derive(JsonSchema, Serialize, Deserialize, GtsTraitsSchema)]`) used as the base's
`traits_schema`. The line you flagged is `SecretTypeDescriptor`, the compile-time catalog
entry (a `const`, `&'static`/`Copy`, carrying identity + traits) — deriving `GtsTraitsSchema`
there would fold the identity fields into the traits schema, which is wrong; it converts to
the carrier via `.traits() -> SecretTypeTraits`.

gears/credstore/credstore-sdk/src/models.rs
diffora: Done — removed `SecretTypeRef` and made the type reference a `GtsId` on both surfaces: SDK
`WriteOptions.secret_type: Option<GtsId>` and the REST `type` parsed as a full GTS id
(`GtsId::try_new`). The gear resolves it to the type's deterministic v5 UUID at the service
boundary (unchanged internally), so "GtsId on the wire, UUID inside". Short name and raw UUID
are no longer accepted, and the `GET` response echoes the full GTS id (symmetric). Added
`impl From<SecretType> for GtsId` for ergonomics. Docs, OpenAPI, and unit + e2e tests updated;
green.

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
diffora: credstore doesn't create `tenant_closure` — it's owned and migrated by Account Management
(AM's `m0001_initial_schema`). credstore only *reads* it, and only in the **co-located**
deployment (config `hierarchy.tenant_closure_colocated`, default `false`; DESIGN §4.4):

- co-located: the gear advertises the `TenantHierarchy` PDP capability, so grants compile to
  `InTenantSubtree` predicates resolved via a `tenant_closure` subquery.
- not co-located (default): the PDP emits a flat `In` list and credstore never touches the
  table — so a standalone credstore DB without it is supported, not a missing migration.

Expanded the entity doc-comment to spell this out.

gears/credstore/docs/ADR/0001-cpt-cf-credstore-adr-stateful-gateway.md
diffora: Agreed — dropped "gateway" across the gear (docs, ADRs, comments, READMEs, Cargo desc).
Renamed ADR-0001 to *Stateful Gear with Gear-Owned Secret Metadata* (file/ID
`…-stateful-gear`) and updated cross-refs. Left the legitimate "Gateway" uses (the OAGW actor,
the `api-gateway` config prefix). Text-only; tests/dylint green.

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
diffora: Factual: no — before this PR `oagw_provisioning` didn't touch credstore and failures were
fail-hard. This PR added both the dependency and the fail-hard→defer shift deliberately: a
provider's secret is created at runtime and may not exist at boot. For fail-fast: moved the
deterministic-misconfig check into `ProviderEntry::validate()` (runs at init, fails boot) — it
now rejects a tenant override lacking both `host` and `upstream_alias`. `Misconfigured` is now
just a defensive fallback; the "secret not yet available" path stays deferred (transient,
indistinguishable from a bad ref here).

gears/system/oagw/oagw/src/infra/plugin/oauth2_client_cred_auth.rs
diffora: Done — added `credstore_sdk::test_util::MockCredStoreClient` behind a `test-util` feature: one
configurable double (`empty` / `with_secrets` / `returning_raw_value` / `always_failing`).
Removed the five hand-rolled doubles in oagw; `test_support` now re-exports the SDK mock, so
existing call sites are unchanged. Wired the feature through. Tests + dylint green.

gears/credstore/docs/ADR/0003-cpt-cf-credstore-adr-value-fingerprint-fence.md
diffora: Optimistic locking doesn't close this race — it's a crosswise dual-write: value (backend) and
metadata (row) are separate stores with no shared transaction, so two concurrent PUTs can land
backend = Bob's value + row = Alice's label (cross-tenant disclosure). `If-Match` only guards
the metadata row's version within its own store — it can't bind the backend value to the
metadata across stores, and it's opt-in. A cross-store fence isn't available (cluster
ADR-002). Hence the read-time value fingerprint. Added a paragraph to the ADR Context.

gears/credstore/docs/DESIGN.md
diffora: Fixed — the bare cross-doc ADR references are now full relative-path links: "ADR 0005" →
`docs/arch/errors/ADR/0005-cpt-cf-adr-sdk-canonical-projection.md` (×2), and "cluster ADR-002"
→ `gears/system/cluster/docs/ADR/002-async-boundary-no-remote-in-critical-section.md`. (In-gear
ADR references were already relative `./ADR/…` links.)

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
diffora: It's structured, not a bare warn+loop: `reconcile_deferred_upstreams_with_retry` runs in the
background with exponential backoff (2s → 1min cap), is cancellation-aware, shrinks as
providers succeed, and warns once after 2min with likely causes. Unbounded by design — a
runtime-provisioned secret has no deadline, and transient-vs-permanent can't be split here
(see C12); deterministic misconfigs now fail fast at boot. If you want a stronger signal I'd
add a deferred-upstreams gauge rather than bound the retry — happy to.

gears/credstore/docs/DESIGN.md
diffora: Added a credstore-scoped OpenAPI in YAML at `gears/credstore/docs/api/openapi.yaml` (the
gear's paths + referenced components), extracted from the platform `docs/api/api.json`. Both
are generated via `make openapi` and kept fresh by the `api_contracts` check. DESIGN §4.3.1
links both.

gears/credstore/docs/DESIGN.md
diffora: `reference` is intentionally *not* a GTS id — a caller-chosen opaque label
(`^[A-Za-z0-9_-]+$`, 1–255), because it's user-facing (a GTS id would be too restrictive, as
you note). Documented the format explicitly in DESIGN §4.3.1 (was only in the OpenAPI schema
before). Uniqueness is per sharing class within a tenant.

gears/credstore/docs/DESIGN.md
diffora: Yes — the value is stored as raw bytes end-to-end, so binary secrets (e.g. a DER cert or a raw
key) are supported, governed by the per-type `utf8_only` trait (only `generic` allows
non-UTF-8 today). Caveat: REST carries `value` as a UTF-8 JSON string, so raw binary goes via
the SDK (bytes); over REST it must be text-encoded (PEM/base64). A PEM cert is UTF-8 and works
anywhere. Spelled out in DESIGN §4.3.1.

gears/credstore/docs/DESIGN.md
diffora: `is_inherited` (with `owner_tenant_id`) tells the caller whether the resolved secret is the
tenant's own or inherited from an ancestor. Scenario: deciding whether to create a
tenant-local override — e.g. an admin UI showing "using the parent's shared credential" vs
"your own", so an operator can shadow it. It also lets a caller key a cache correctly.
Read-only metadata; nothing in the write path needs it.

gears/credstore/docs/DESIGN.md
diffora: `sharing`/`status` are `SMALLINT` + `CHECK` rather than a native `CREATE TYPE … AS ENUM`
because the schema targets both PostgreSQL and SQLite from one migration set (§8), and SQLite
has no enum type — a PG enum wouldn't port. The mapping is centralized in the entity layer
(`sharing_to_i16`/`sharing_from_i16`). I'd keep it as-is, but happy to add a comment making the
encoding explicit.

gears/credstore/docs/DESIGN.md
diffora: The `private`/`inherit`/`enforce` values don't carry over to credstore — it's a different axis. Our `sharing` answers *"who can read the secret within the tenant hierarchy"*, not a parent/child limit-inheritance policy:

- `private` — owner only;
- `tenant` — tenant-wide, not inherited;
- `shared` — the tenant and its descendants, closest-wins (a child may shadow it with its own value).

There's no numeric *limit* and no parent-side policy knob here, so `inherit`/`enforce`/"Effective Limit" have nothing to map onto. (Our `shared` already behaves like your `inherit` row — the child can override; we deliberately don't have an `enforce` / non-overridable mode.)

And these aren't just doc wording — `private`/`tenant`/`shared` is the shipped contract: the `SharingMode` SDK enum, the REST `sharing` field, the `credstore_secrets.sharing` column (`SMALLINT` 1/2/3), the per-type `allow_sharing` trait, and what current consumers already send. Renaming to `private`/`inherit`/`enforce` would break the SDK, the wire contract, stored rows, and every integration — for values that mean something different anyway.

Kept the modes as-is.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1093 @@ NOT IN THIS DOCUMENT (see other templates):
...
-
-The `credentials_storage` plugin is an exception to this pattern. It is a standalone Rust microservice that implements credential merge/propagation resolution internally (own → inherited → default), along with encrypted credential storage, schema validation, field-level masking, and pluggable tenant key management via a `KeyProvider` abstraction. When this plugin is active, the Gateway delegates merge resolution to the plugin rather than performing the walk-up algorithm itself. The `KeyProvider` supports two modes: local database storage (for development/simple deployments) and external key management service integration (HashiCorp Vault, AWS KMS) for production environments requiring key–data separation. The detailed plugin architecture will be documented in `plugins/credentials-storage/DESIGN.md`.
+CredStore follows the ToolKit Gear + Plugins pattern: a **stateful gear
+gear** (`credstore`) owns all secret *metadata* (identity, sharing, ownership,
```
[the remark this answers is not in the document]
diffora: Fixed in aa684a43 — both occurrences ("stateful gear gear" in §1.1 and the `credstore` component description) now read "stateful gear".

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1093 @@ NOT IN THIS DOCUMENT (see other templates):
...
+
+| Status | Canonical category | Scenario |
+|--------|--------------------|----------|
+| 400 | `InvalidArgument` | Invalid `reference` format, malformed `If-Match`, unsupported sharing transition (private ↔ tenant/shared), non-UTF-8 stored value on `GET` |
+| 401 | `Unauthenticated` | Invalid or missing token |
+| 403 | `AccessDenied` | PDP denies the action, or the caller's scope excludes their own tenant |
+| 404 | `NotFound` | Secret not found **or** inaccessible (anti-enumeration: always 404, never 403, for per-secret access) |
+| 409 | `AlreadyExists` / `Aborted` | `POST` conflict; `If-Match` version conflict (`OPTIMISTIC_LOCK_FAILURE`) |
+| 503 | `ServiceUnavailable` | PDP evaluation failure, types-registry outage / stored secret type unresolvable (§5.4), no storage plugin registered (with stable detail, no `retry_after`), backend outage (with `Retry-After` when hinted) |
+| 500 | `Internal` | Invariant violations; diagnostic stripped from the wire |
```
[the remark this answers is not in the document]
diffora: Fixed in aa684a43 — moved "non-UTF-8 stored value on `GET`" from the 400/`InvalidArgument` row to the 500/`Internal` row, matching `GetSecretResponseDto::try_from_response`.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1093 @@ NOT IN THIS DOCUMENT (see other templates):
...
+| `ssh-key` | private, tenant | — | Deploy/automation keys. Format checks deferred. |
+| `webhook-hmac` | tenant, shared | — | Webhook signing secrets shared with descendant integrations. |
+| `connection-string` | tenant | — ; `max_size_bytes` = 4 KiB | DSNs; tenant-local by policy (contain embedded endpoints/credentials). |
+
+Adding a **built-in** type (shipped with the platform, with a short REST
+name) = adding a catalog entry + its seed registration in the SDK (one
+release). Adding a **custom** type = registering a GTS schema derived from
+the secret base type — no release; the type is addressed by its full GTS id
+(or UUID) on the API. The enforcement code changes only when a new *trait*
+is introduced.
```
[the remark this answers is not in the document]
diffora: Fixed in aa684a43 — dropped "(or UUID)": custom types are addressed by their full GTS id only, matching the transport contract and route tests.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1093 @@ NOT IN THIS DOCUMENT (see other templates):
...
+- Step 4 fails → the backend value exists but the row stays `provisioning`: never readable; a client retry overwrites it; otherwise reaped — and the reaper issues a best-effort `plugin.delete` for every row it reaps (§6.4), so the orphaned backend value is reconciled rather than leaked.
+- **Create race** (unique-index conflict on step 2): a create-only `POST` surfaces 409; a `PUT` retries `find_for_write` up to 3 times with 25 ms backoff (the winner marks active within ms) and degrades to a retry-safe 409 if the winner appears stuck.
 
-**Impact**: Wrong secret returned, data corruption, security incident
+**Overwrite path** (row of the target class exists): version pre-check
+(`If-Match`) → backend `plugin.put` **first** → `touch` (version bump +
+sharing update) gated by `version = ?`. Backend-first means a plugin failure
+leaves metadata untouched; a `touch` failure after a committed backend write
+self-heals on the next put. A row that vanished concurrently is success
+(delete won); a moved version under `If-Match` is 409.
```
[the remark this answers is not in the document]
diffora: Fixed in aa684a43 — §6.2 rewritten to describe the actual dual ordering: without `If-Match` (LWW) the write is backend-first and a concurrently vanished row surfaces as a retryable 409 (not success); with `If-Match` the version-gated `touch` claims the CAS **before** `plugin.put`, so a losing writer never reaches the backend.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,1093 @@ NOT IN THIS DOCUMENT (see other templates):
...
-**Mitigation**:
-- High-availability key service deployment
-- Readiness probe reflects KMS connectivity
-- Key caching with short TTL for read-path resilience
-- Circuit breaker for key service calls
+**Failure handling**:
 
-**Likelihood**: Low | **Impact**: Critical | **Priority**: P1
+- Step 3/4 fails → the row stays `deprovisioning`; the caller gets a retryable 503. A **retry of `DELETE` resumes the saga**: `find_own` sees the `deprovisioning` row and re-runs steps 3–4 (both idempotent). Absent a retry, the reaper completes it.
+- Crash between 2 and 4 → same recovery: reaper or client retry finishes cleanup. No state leaves a readable secret or a permanently held reference.
```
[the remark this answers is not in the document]
diffora: Fixed in aa684a43 — the failure-handling bullet now says the caller gets the mapped error (503 for an unavailable backend, 500 for an internal plugin fault) instead of guaranteeing 503; the deprovisioning/retry/reaper recovery description is unchanged.

gears/credstore/credstore-sdk/src/api.rs
aviator5: Ok, but if so, why do we need to have `put()`? If `version` is omitted, optimistic lock will not be used and races may occur, right?

gears/credstore/credstore-sdk/src/gts.rs
```diff
@@ -10,3 +38,248 @@ use toolkit_gts::gts_type_schema;
...
+    description = "CredStore secret resource — tenant-scoped, RBAC/PEP protected",
+    properties = "gts_type",
+    traits_schema = inline(SecretTypeTraits),
+    // `x-gts-traits` defaults for the base type itself: generic semantics (all
+    // sharing modes, opaque value). gts's `validate_entity_traits` requires the
+    // base of a traits-schema chain to carry values. Supplied as the typed
+    // `SecretTypeTraits` carrier (via the catalog's `generic` entry) rather than
+    // an untyped `json!`, so the base default is compile-checked and can't drift
+    // from the catalog — the idiom from gts-rust's `traits_struct_literal`.
+    traits = builtin_traits("generic"),
```
aviator5, line 83: How about marking this type as abstract? It will allow having empty traits because `builtin_traits("generic")` looks strange for the base type. Or do I miss something?

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
aviator5: So, in the co-located deployment configuration CredStore and Account Management gears will share the same DB, right?

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
aviator5: So, will `ProviderRegistration::Misconfigured` lead to the application failure on start? Please correct me, but seems like it's not, and I would suggest to change this.

gears/credstore/docs/ADR/0003-cpt-cf-credstore-adr-value-fingerprint-fence.md
```diff
@@ -0,0 +1,153 @@
...
+  - [Value fingerprint in the gear row (CHOSEN)](#value-fingerprint-in-the-gear-row-chosen)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-cf-credstore-adr-value-fingerprint-fence`
+
+## Context and Problem Statement
+
+With a stateful gear (ADR-0001) a write spans two stores with no shared
```
aviator5: Suggestions:
1) don't break sentences via "\n" in all docs
2) use full path to ADR docs everywhere

@ainetx how about having some deterministic rules in CI?

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
aviator5: In my view we should retry only on temp error, for example when Credential Store is not available yet.

gears/credstore/credstore-sdk/src/api.rs
diffora: Right — without a precondition `put` is last-writer-wins, and that's deliberate. There are really three different races behind "races may occur", and `version` is the answer to exactly one of them:

1. **Two blind writers race** (rotation/provisioning: the new value comes from outside, not derived from the stored one). The later write replaces the earlier wholesale — which is the wanted outcome. A mandatory version wouldn't change it: on 409 a blind writer can only GET the fresh version and re-PUT the same bytes, i.e. the same LWW with extra round trips; and the create path has no version to send at all. This mirrors HTTP `PUT` without `If-Match` and e.g. Vault KV's opt-in `cas`.
2. **Read-modify-write** (the caller must not overwrite a concurrent update). This is the case CAS exists for, and it's provided: `put_opts`/`delete_opts` + `WritePrecondition` (the in-process `If-Match`), failing with `Conflict` on a stale validator. The generation-bound ETag `"<id>.<version>"` also closes the ABA hole (a stale validator matching a re-created secret whose counter restarted).
3. **The dangerous one — the crosswise dual-write.** Value goes to the backend, metadata (`sharing`) to the gear DB, no shared transaction. Alice `PUT value=A, sharing=shared` and Bob `PUT value=B, sharing=tenant` can interleave so the backend keeps B while the row keeps Alice's `shared` — a descendant tenant would then read Bob's tenant-only value under a `shared` label nobody gave it. An optimistic lock does **not** close this even when supplied — it guards the metadata row's version inside one store and can't bind it to the backend value. The value-fingerprint fence (ADR-0003) closes it unconditionally: `value_fp = HMAC(fence_key, value)` is stamped in the same atomic row write as `sharing`, a read recomputes it from the value the backend returned and serves only on a match — the poisoned interleaving reads as a 404 and the next successful PUT heals it.

So the opt-in `version` targets exactly case 2; cases 1 and 3 are handled by design whether or not a precondition is sent. Made this explicit in `put`'s rustdoc (Concurrency section).

That said — if this opt-in split feels over-complicated, I'm fine flipping it: make the precondition mandatory on `put`/`delete` (every writer reads first and sends the version; `create` stays the only unversioned path, and blind writers carry a GET + retry-on-409 loop). It trades the extra round trip and retry boilerplate for a single uniform rule — say the word and I'll make the change.

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
diffora: Yes — co-located means one physical database: `tenant_closure` stays owned and migrated by Account Management, and credstore's DB role only needs `SELECT` on it. Strictly speaking, though, one shared DB isn't mandatory — the contract behind the flag is "a correct and fresh `tenant_closure` is queryable through credstore's own connection". A separate credstore database also qualifies if an external mechanism (e.g. logical replication / CDC of the AM-owned table) keeps a local copy populated and consistent: credstore never writes the table and doesn't track its freshness, so that pipeline owns integrity — closure staleness translates directly into stale subtree-authorization decisions. It's opt-in (`hierarchy.tenant_closure_colocated`, default `false`); the default deployment is a standalone credstore database without the table, where the PDP pre-expands subtree grants into a flat `In` list (degraded but correct). Spelled this out in DESIGN §4.4.

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
diffora: Done — the boot path now fails fast. `register_oagw_upstreams` returns a `ProvisioningReport { deferred, misconfigured }`, and `start()` aborts startup when `misconfigured` is non-empty (`ProvisioningReport::ensure_no_misconfigured`). "Misconfigured" now also covers deterministic OAGW rejections (see the thread below on retries): a provider whose request OAGW rejects as `invalid_argument` fails the boot instead of being silently unavailable. During the *background reconcile* a misconfigured provider is dropped from the retry set with an error log rather than crashing the running gear — fail-fast is a boot-time property.

gears/mini-chat/mini-chat/src/infra/oagw_provisioning.rs
diffora: Agreed and implemented. The blocker was that OAGW's create-upstream validation mapped "secret_ref not accessible (yet)" to `Validation` → `invalid_argument` — indistinguishable from a permanently malformed request, which is why the loop had to retry everything. OAGW now returns a dedicated `SecretRefNotAccessible` → canonical `failed_precondition` (HTTP status stays 400; only the canonical category changes), and mini-chat classifies failures: `invalid_argument` / `out_of_range` / `unimplemented` → misconfigured (error-logged, never retried, fails boot per the thread above); everything else — `failed_precondition` (secret not provisioned yet), `service_unavailable` / `internal` / timeouts (credstore or OAGW down), auth/RBAC errors (fixable at runtime) — stays in the retry set with the existing backoff. Covered by tests on both sides.

gears/credstore/credstore-sdk/src/gts.rs
diffora: Good call on abstract — done: `SecretV1` is now `gts_abstract = true` (the base is a pure trait/schema carrier; secrets are always typed by a derived id — `generic` is itself a derived type), with a guard test. On the empty traits: you're right that validation would allow it (required-trait completeness is deferred to descendants for abstract types), but the base's `x-gts-traits` aren't there for the validator — they're the inherited defaults of the chain merge (leaf wins, base fills the rest). A customer-registered type that declares only some traits inherits the generic values for the rest today; with a value-less base those holes would materialize from the trait-schema defaults (`allow_sharing: []`), silently making sparse custom types unwritable — while still registering fine. So the base keeps `traits = builtin_traits("generic")` (compile-checked against the catalog), with the comment now saying exactly that. Also made the abstractness enforced end-to-end: the secret-type resolver now rejects the base's own id as a secret type (same `UNKNOWN_SECRET_TYPE` violation as a foreign type — the types-registry SDK doesn't surface `x-gts-abstract`, so the one known abstract id is checked explicitly), with a test.

gears/credstore/docs/ADR/0003-cpt-cf-credstore-adr-value-fingerprint-fence.md
diffora: Done for the credstore docs in this PR: PRD, DESIGN, ADR-0001…0003 and QUICKSTART are reflowed to one physical line per paragraph/list item (no sentence carries a `\n`; code fences, tables and the toc blocks untouched), and every bare `ADR-000N` mention is now a markdown link to the ADR file. +1 on deterministic CI rules — happy to adopt whatever `cfs`/CI check @ainetx proposes so this stays enforced rather than reviewed by hand.

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
aviator5: Hm, but for what this `tenant_closure` is used? If for tenants' hierarchy traversal, should Tenant Resolver gear be used instead?

gears/credstore/credstore-sdk/src/api.rs
aviator5: suggest making precondition mandatory

diffora: Done — the precondition is now mandatory on `put` and `delete`: in the SDK it is a required argument, on REST a missing `If-Match` is a 400. `If-Match: *` stays as the explicit form for callers that have no version to send: rotation/provisioning (the new value does not depend on the stored one), and repairing a secret whose read fails closed with 404 after a broken write — there is no ETag to obtain for it. `PUT` no longer creates: creation is `POST` only, and create-or-replace is `POST`, then on 409 `PUT` with `If-Match: *`.

Note that the value fingerprint (ADR-0003) is still needed: the precondition guards only the metadata row, while the value and the metadata are still written to two stores without a shared transaction — a crash between the two writes, or two `If-Match: *` writers racing, still leaves one writer's value under another writer's sharing label, and only the check on read catches that.

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
diffora: Tenant Resolver is used for hierarchy traversal: when a GET walks up the ancestors to find an inherited secret, credstore asks the Tenant Resolver gear for the ancestor chain. `tenant_closure` is used for the opposite direction: when a grant says "may read secrets anywhere under tenant T", that filter has to run inside the SQL query itself (`tenant_id IN (SELECT descendant … WHERE ancestor = T)`), and a call to another gear cannot be put into a SQL filter. This is also only an opt-in optimization: by default the PDP expands the subtree into a plain list of tenant ids and credstore does not read the table at all.

nonameffh (approved):
No objections from Constructor. Our requirements for CredStorage also include:
- credential metadata listing
- a bit more complex read permissions ("metadata only" and "metada + value")

More complex read permissions for us should support the case when users can read only metadata, not values (e.g. as a Tenant Administrator I can see the list of credentials and see that some of them are owned by my tenant, some are inherited).
From my perspective it is an additive change and after merging this PR I'll open an PR to support our case

Also we have a case with another dimension of access control - only an application that defines a new secret typecan read secrets of this type (e.g. only NotificationEngine can read SMTP credentials), but this from my perspective doesn't require changing CredStorage at all and can be implemented on AuthZ level (specific PDP policies)

aviator5 (approved):

gears/credstore/credstore/src/infra/storage/entity/tenant_closure.rs
aviator5: Discussed, at a first glance by optimization reason we may avoid using closure table or executing `tenant_id IN (SELECT descendant … WHERE ancestor = T)` SQL and do prefetch select instead (please check `docs/arch/authorization/AUTHZ_USAGE_SCENARIOS.md`), but it's up to the author.

diffora: Follow-up: we went one step further and removed the co-located `tenant_closure` option altogether. Every credstore operation is a point operation in the caller's own tenant (there is no LIST, and writes never cross the tenant boundary), so the flat tenant predicates the PDP hands out are sufficient — the PDP resolves subtree grants on its own side. So the final answer to your question: hierarchy traversal uses only the Tenant Resolver gear (the upward walk for shared-secret inheritance), and credstore reads no hierarchy tables at all.

diffora:
> No objections from Constructor. Our requirements for CredStorage also include:
> 
> * credential metadata listing
> * a bit more complex read permissions ("metadata only" and "metada + value")
> 
> More complex read permissions for us should support the case when users can read only metadata, not values (e.g. as a Tenant Administrator I can see the list of credentials and see that some of them are owned by my tenant, some are inherited). From my perspective it is an additive change and after merging this PR I'll open an PR to support our case
> 
> Also we have a case with another dimension of access control - only an application that defines a new secret typecan read secrets of this type (e.g. only NotificationEngine can read SMTP credentials), but this from my perspective doesn't require changing CredStorage at all and can be implemented on AuthZ level (specific PDP policies)

1) ok please open the request
2) in this implementation you can grant the subject by secret type already

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,792 @@ NOT IN THIS DOCUMENT (see other templates):
...
+| `cpt-cf-credstore-fr-put-secret` | Write saga: insert `provisioning` metadata row → plugin `put` (value only) → mark `active`; overwrite path is backend-first, then version bump |
+| `cpt-cf-credstore-fr-get-secret` | Single SQL resolution over the ancestor chain, then one plugin `get` for the winning row |
+| `cpt-cf-credstore-fr-delete-secret` | Deprovisioning saga: mark `deprovisioning` → backend delete → row delete (§6.3) |
+| `cpt-cf-credstore-fr-tenant-scoping` | Gear derives tenant from `SecurityContext.subject_tenant_id()`; own-tenant gate + SecureORM scope clamp |
+| `cpt-cf-credstore-fr-sharing-modes` | `sharing` column in the gear metadata table; partial unique indexes let private and tenant/shared coexist under one reference |
+| `cpt-cf-credstore-fr-authz-pdp` | PDP `AccessScope` per operation (`read`/`write`/`delete` on the secret GTS resource type), enforced in SQL; fail-closed |
+| `cpt-cf-credstore-fr-optimistic-concurrency` | Monotonic `version` column; `GET` returns a strong generation-bound `ETag` (`"<id>.<version>"`, §4.10); `PUT`/`DELETE` require `If-Match` (a validator or `*`) |
```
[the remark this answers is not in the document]
diffora: Fixed in 0dc4458e — the `fr-put-secret` row now documents both orderings: version-validator writes claim the metadata CAS before `plugin.put`, `If-Match: *` is backend-first then version bump, with a pointer to §6.2.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,792 @@ NOT IN THIS DOCUMENT (see other templates):
...
 
-#### Three-Tier Sharing Model (not RBAC/ABAC)
+**Generation-bound validator.** The strong `ETag` is `"<row-id>.<version>"`. The row UUID is minted fresh per created secret, so a validator from a deleted-and-recreated secret's earlier generation never matches the new row even when the restarted version counters coincide (closing the ABA lost-update). `version` stays the per-generation optimistic-lock counter and the row-level CAS still gates on it (the id is already the UPDATE key).
 
-**Decision**: Use simple three-tier sharing modes (private/tenant/shared) instead of granular ACL.
+**Out-of-band seeding.** A row provisioned directly in the DB with `value_fp = NULL` (value placed directly in the backend) is served on trust and its fingerprint backfilled — lazily on first read, or by the reaper's bounded sweep for rows nobody reads (both via a `value_fp IS NULL` CAS that never bumps the version). A re-seed of an existing row must reset `value_fp` to NULL in the same operation. This is the only path that produces a NULL fingerprint; API writes always stamp.
```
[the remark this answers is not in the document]
diffora: Fixed in 0dc4458e — the Mechanism paragraph now states the exception inline: an out-of-band seeded row with `value_fp = NULL` is served on trust until backfilled; API-written rows always carry a fingerprint.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,792 @@ NOT IN THIS DOCUMENT (see other templates):
...
+### 6.4 Reaper
 
-#### Risk: Owner ID Migration for Existing Secrets
+The gear's lifecycle entry (`serve`) runs a cancellable loop on `reaper.tick_secs` (default 60 s; delayed missed-tick behavior). Each tick:
 
-**Impact**: Existing secrets without owner_id cannot be accessed in `private` mode
+1. **Expiry sweep**: flip expired `active` rows (`expires_at <= now`) into the deprovisioning saga (they already stopped resolving at read time; this cleans the backend value and releases the reference).
+2. **List stale non-active rows** (bounded batch) older than `provisioning_timeout_secs` / `deprovisioning_timeout_secs` (both default 300 s), via the pending partial index.
+3. **Backend reconciliation**: issue a best-effort `plugin.delete` for every stale row (closing the orphaned-value debt of §6.2); `NotFound` counts as success.
+4. **Remove rows**: `provisioning` rows unconditionally (the reference must not stay wedged); `deprovisioning` rows only after a successful backend delete — otherwise the row (and the name it holds) waits for the next tick. Counts → `provisioning_reaped` / `deprovisioning_reaped`.
```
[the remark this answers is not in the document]
diffora: This asymmetry is deliberate, and 0dc4458e documents it in §6.4 instead of changing the behavior: a `deprovisioning` row must hold the name until the backend is clean (releasing it early could let the saga's lagging delete erase a successor's value), while keeping a `provisioning` row on a failed backend delete would wedge the reference for as long as the backend stays unreachable — the reference must not stay hostage to a backend outage. The cost when step 3 fails is an orphaned backend value with no metadata row: it is never readable (resolution requires a row) and is overwritten by the next create of the same reference, mirroring the accepted no-retry orphan of §6.2.

gears/credstore/docs/DESIGN.md
```diff
@@ -66,931 +73,792 @@ NOT IN THIS DOCUMENT (see other templates):
...
 
-## 9. Traceability
+## 11. Traceability
 
 - **PRD**: [PRD.md](./PRD.md)
-- **ADRs**: [ADR/](./ADR/)
-- **Features**: [features/](./features/)
-- **Credentials Storage Plugin Design**: `plugins/credentials-storage/DESIGN.md` (planned)
+- **ADRs**: [ADR/](./ADR/) — [ADR-0001 stateful gear](./ADR/0001-cpt-cf-credstore-adr-stateful-gear.md), [ADR-0002 deprovisioning saga](./ADR/0002-cpt-cf-credstore-adr-deprovisioning-saga.md)
+- **Features**: features/ (planned)
```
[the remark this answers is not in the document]
diffora: Fixed in 0dc4458e — ADR-0003 added to the Traceability list.

gears/credstore/docs/PRD.md
```diff
@@ -246,125 +255,132 @@ The system **MUST** validate SecretRef format: alphanumeric characters, hyphens,
...
+### 5.4 P1 — Reliability & Concurrency
 
-- [ ] `p1` - **ID**: `cpt-cf-credstore-fr-authz-gateway`
+#### Crash-Safe Write Lifecycle
+
+- [ ] `p1` - **ID**: `cpt-cf-credstore-fr-write-lifecycle`
 
 <!-- cpt-cf-id-content -->
-Authorization **MUST** be enforced in the gateway layer, not in plugins. Plugins are storage adapters and MUST NOT implement authorization logic.
+A secret write that spans metadata and backend **MUST** be crash-safe: a new secret becomes readable only after its value is durably stored in the backend (`provisioning` → `active`); a failed backend write rolls the metadata back; a crash mid-write leaves a non-readable in-flight record that is swept by a periodic reaper within a configurable timeout. No failure mode may leave a readable secret without a value or permanently block the reference.
```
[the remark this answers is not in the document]
diffora: Fixed in 0dc4458e — the requirement now distinguishes the two paths: a failed backend write rolls metadata back on the create path, while a failed/half-completed overwrite leaves the secret unreadable (fail-closed) until a retried write lands both parts, matching DESIGN §6.2 and ADR-0003.

gears/credstore/docs/PRD.md
```diff
@@ -587,84 +604,124 @@ Secret values **MUST NOT** appear in logs, error messages, or debug output at an
...
-- **P2/Future**: Should secret updates support atomic compare-and-swap (CAS) validation? VendorA Credstore has a newer endpoint that validates the current secret value before allowing updates, preventing race conditions. This could be added as an optional parameter to the `put` operation without breaking existing clients.
+- **Batch retrieval**: should `get` support multiple references per call for OAGW efficiency? (Single-query resolution makes this cheap on the metadata side.)
+- **P2/Future — Human vs service access**: should human users be restricted to metadata-only for inherited shared secrets while service accounts can read values?
+- **P2/Future — Audit trails**: structured audit events (actor, tenant, outcome — never values) to a tamper-evident platform sink.
+- **P2/Future — Metadata list endpoint**: a values-free list becomes cheap with gear-owned metadata; must be reconciled with the anti-enumeration stance and per-type authorization.
+- **Dynamic type descriptors** (see DESIGN §9): resolve traits from the types-registry at runtime so vendors can add types without an SDK release.
```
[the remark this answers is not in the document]
diffora: Fixed in 0dc4458e — removed the stale open question (secret types are already registry-driven at runtime; it also pointed at a non-existent DESIGN §9).

striped-zebra-dev (approved):

PR #4220, opened 2026-07-16
feat(cluster): implement Postgres cluster plugin

### LINKED ISSUE #4393: Land the Postgres cluster plugin (merge PR #4220)
## Summary

Track landing the Postgres cluster plugin (`cf-postgres-cluster-plugin`) implementation currently in PR #4220 — a native `ClusterCacheBackend` (sqlx-backed pool) and `DistributedLockBackend` (`pg_advisory_lock`) for the cluster gear, wired into the provider registry, with the shared conformance suite extended to run against it alongside the standalone plugin.

PR #4220 is open, not yet merged. Epic #4382 previously described this plugin as "complete" citing a commit hash that doesn't correspond to a merged commit — this issue exists to track it accurately as in-review work until it actually lands.

## Scope

- [ ] Review and merge #4220
- [ ] Confirm `docs/TRACEABILITY-AUDIT.md`'s requirement-to-code mapping for Postgres-backed requirements reflects merged, not just implemented, code

## References

- PR: #4220
- Cluster epic: #4382
- Design/testing docs: `gears/system/cluster/plugins/postgres-cluster-plugin/docs/{DESIGN,TESTING}.md`

asmith987:
Closes #4393

Adds the Postgres cache and distributed-lock provider implementations for the cluster gear (sqlx-backed pool, pg_advisory_lock, LISTEN/NOTIFY watch), wires both providers into the cluster gear's provider registry, and extends the shared conformance suite so it runs against the new backend alongside the existing standalone plugin.

## Summary by CodeRabbit

* **New Features**
  * Added a PostgreSQL-backed cluster plugin for distributed caching and locking.
  * Added standalone PostgreSQL lock-provider support.
  * Added polling-based service discovery for caches without native prefix-watch support.
  * Added configurable polling, cache reaping, lock reaping, replication, and connection settings.
* **Bug Fixes**
  * Improved membership consistency during concurrent updates.
  * Improved watch, lock expiry, reconnection, and graceful shutdown behavior.
* **Documentation**
  * Added PostgreSQL plugin design, configuration, and testing documentation.

### DIFF
--- a/gears/system/cluster/cluster-conformance/src/cache.rs
+++ b/gears/system/cluster/cluster-conformance/src/cache.rs
@@ -6,38 +6,47 @@
 //! a plugin may call directly; [`run_cache_conformance`] drives the full L2 set,
 //! building a fresh backend per scenario via `make` so state never leaks.
 
+use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
 use cluster_sdk::cache::{
-    CacheConsistency, CacheEvent, CacheWatchEvent, ClusterCacheBackend, PollingPrefixWatch,
-    PutRequest, Ttl,
+    CacheConsistency, CacheEvent, ClusterCacheBackend, PollingPrefixWatch, PutRequest, Ttl,
 };
 use cluster_sdk::error::ClusterError;
 
-/// Runs every implemented L2 cache scenario against a fresh backend from `make`.
+use crate::factory::{ScenarioBackend, run_scenario};
+use crate::time::TimeControl;
+use crate::watch_poll::poll_watch;
+
+/// Runs every implemented L2 cache scenario, each against a fresh backend built
+/// by the async factory `make` and torn down afterward (see
+/// [`ScenarioBackend`]). `time` selects the clock model
+/// ([`TimeControl::Virtual`] for in-memory fixtures, [`TimeControl::Real`] for a
+/// real-I/O backend — see [`crate::time`]).
 ///
 /// Capability-gated scenarios read `consistency()`/`features()` off the backend
 /// and assert strict guarantees only where the backend claims them.
-pub async fn run_cache_conformance<F>(make: F)
+pub async fn run_cache_conformance<F, Fut>(make: F, time: TimeControl)
 where
-    F: Fn() -> Arc<dyn ClusterCacheBackend>,
+    F: Fn() -> Fut,
+    Fut: Future<Output = ScenarioBackend<dyn ClusterCacheBackend>>,
 {
-    scenario_cache_001(make()).await;
-    scenario_cache_002(make()).await;
-    scenario_cache_003(make()).await;
-    scenario_cache_004(make()).await;
-    scenario_cache_005(make()).await;
-    scenario_cache_006(make()).await;
-    scenario_cache_007(make()).await;
-    scenario_cache_008(make()).await;
-    scenario_cache_009(make()).await;
-    scenario_cache_010(make()).await;
-    scenario_cache_011(make()).await;
-    scenario_cache_012(make()).await;
-    scenario_cache_013(make()).await;
-    scenario_cache_014(make()).await;
-    scenario_cache_015(make()).await;
+    run_scenario(make(), scenario_cache_001).await;
+    run_scenario(make(), scenario_cache_002).await;
+    run_scenario(make(), scenario_cache_003).await;
+    run_scenario(make(), scenario_cache_004).await;
+    run_scenario(make(), scenario_cache_005).await;
+    run_scenario(make(), scenario_cache_006).await;
+    run_scenario(make(), scenario_cache_007).await;
+    run_scenario(make(), scenario_cache_008).await;
+    run_scenario(make(), scenario_cache_009).await;
+    run_scenario(make(), |b| scenario_cache_010(b, time)).await;
+    run_scenario(make(), |b| scenario_cache_011(b, time)).await;
+    run_scenario(make(), scenario_cache_012).await;
+    run_scenario(make(), scenario_cache_013).await;
+    run_scenario(make(), |b| scenario_cache_014(b, time)).await;
+    run_scenario(make(), |b| scenario_cache_015(b, time)).await;
 }
 
 /// SC-CACHE-001: `get` on an absent key returns `Ok(None)`, never an error.
@@ -298,8 +307,8 @@ pub async fn scenario_cache_009(backend: Arc<dyn ClusterCacheBackend>) {
 
 /// SC-CACHE-010: TTL expiry removes the entry and emits `CacheEvent::Expired` to
 /// watchers.
-pub async fn scenario_cache_010(backend: Arc<dyn ClusterCacheBackend>) {
-    tokio::time::pause();
+pub async fn scenario_cache_010(backend: Arc<dyn ClusterCacheBackend>, time: TimeControl) {
+    time.begin();
     let mut watch = backend.watch("k").await.expect("watch");
     backend
         .put(PutRequest {
@@ -309,25 +318,30 @@ pub async fn scenario_cache_010(backend: Arc<dyn ClusterCacheBackend>) {
         })
         .await
         .expect("put with ttl");
-    tokio::time::advance(Duration::from_millis(100)).await;
-    tokio::task::yield_now().await;
-    // Drain the initial Changed, then wait for the Expired emission.
-    let expired = wait_for(
-        &mut watch,
-        |event| matches!(event, CacheEvent::Expired { key } if key == "k"),
-    )
-    .await;
+    time.elapse(Duration::from_millis(100)).await;
+    // Drain the initial Changed, then wait for the Expired emission. The poll
+    // keeps reading, so under `Real` it tolerates the sweeper reclaim landing a
+    // little after the elapse (caller must set a sweep interval < the elapse).
+    let expired = poll_watch(&mut watch)
+        .until(|event| matches!(event, CacheEvent::Expired { key } if key == "k"))
+        .await;
     assert!(expired, "SC-CACHE-010: TTL expiry must emit Expired");
     assert!(
         backend.get("k").await.expect("get").is_none(),
         "SC-CACHE-010: expired entry reads as absent"
     );
-    tokio::time::resume();
+    time.end();
 }
 
 /// SC-CACHE-012: exact `watch` yields `Changed`/`Deleted` for the key,
 /// preserving per-key order.
 pub async fn scenario_cache_012(backend: Arc<dyn ClusterCacheBackend>) {
+    /// Which of the two ordered events on key `k` a poll selected.
+    enum Seen {
+        Changed,
+        Deleted,
+    }
+
     let mut watch = backend.watch("k").await.expect("watch");
     backend
         .put(PutRequest {
@@ -338,26 +352,32 @@ pub async fn scenario_cache_012(backend: Arc<dyn ClusterCacheBackend>) {
         .await
         .expect("put");
     backend.delete("k").await.expect("delete");
-    // Enforce order: Changed must precede Deleted on the same watch.
-    let mut saw_changed = false;
-    for _ in 0..64 {
-        match watch.recv().await {
-            Some(CacheWatchEvent::Event(CacheEvent::Changed { key })) if key == "k" => {
-                assert!(!saw_changed, "SC-CACHE-012: duplicate Changed event");
-                saw_changed = true;
-            }
-            Some(CacheWatchEvent::Event(CacheEvent::Deleted { key })) if key == "k" => {
-                assert!(
-                    saw_changed,
-                    "SC-CACHE-012: Deleted arrived before Changed -- order violated"
-                );
-                return; // Both events received in correct order.
-            }
-            Some(CacheWatchEvent::Closed(_)) | None => break,
-            _ => {}
+    // Select on *either* event, so the order is what the two polls assert: the
+    // first must be the Changed, the second the Deleted. A predicate that only
+    // looked for Changed would skip a too-early Deleted instead of failing.
+    let on_key = |event| match event {
+        CacheEvent::Changed { key } if key == "k" => Some(Seen::Changed),
+        CacheEvent::Deleted { key } if key == "k" => Some(Seen::Deleted),
+        _ => None,
+    };
+    match poll_watch(&mut watch)
+        .first_match(on_key)
+        .await
+        .expect_match("SC-CACHE-012: no Changed event for the key")
+    {
+        Seen::Changed => {}
+        Seen::Deleted => {
+            panic!("SC-CACHE-012: Deleted arrived before Changed -- order violated")
         }
     }
-    panic!("SC-CACHE-012: did not observe both Changed then Deleted within the event bound");
+    match poll_watch(&mut watch)
+        .first_match(on_key)
+        .await
+        .expect_match("SC-CACHE-012: no Deleted event after the Changed")
+    {
+        Seen::Deleted => {}
+        Seen::Changed => panic!("SC-CACHE-012: duplicate Changed event"),
+    }
 }
 
 /// SC-CACHE-013: `watch_prefix` yields events for matching keys when the backend
@@ -376,11 +396,9 @@ pub async fn scenario_cache_013(backend: Arc<dyn ClusterCacheBackend>) {
             .await
             .expect("put");
         assert!(
-            wait_for(
-                &mut watch,
-                |e| matches!(e, CacheEvent::Changed { key } if key == "p/a")
-            )
-            .await,
+            poll_watch(&mut watch)
+                .until(|e| matches!(e, CacheEvent::Changed { key } if key == "p/a"))
+                .await,
             "SC-CACHE-013: prefix watch yields events for matching keys"
         );
     } else {
@@ -398,8 +416,8 @@ pub async fn scenario_cache_013(backend: Arc<dyn ClusterCacheBackend>) {
 
 /// SC-CACHE-011: a `Ttl::Indefinite` entry persists well past any default TTL;
 /// it is not removed by the TTL sweeper until explicitly deleted.
-pub async fn scenario_cache_011(backend: Arc<dyn ClusterCacheBackend>) {
-    tokio::time::pause();
+pub async fn scenario_cache_011(backend: Arc<dyn ClusterCacheBackend>, time: TimeControl) {
+    time.begin();
     backend
         .put(PutRequest {
             key: "k",
@@ -408,25 +426,26 @@ pub async fn scenario_cache_011(backend: Arc<dyn ClusterCacheBackend>) {
         })
         .await
         .expect("put");
-    tokio::time::advance(Duration::from_hours(1)).await;
-    // Yield so the sweeper (if any) gets a chance to erroneously remove it.
-    tokio::task::yield_now().await;
+    // A large advance under `Virtual`; under `Real` this is capped to a short
+    // real sleep (several sweep intervals) — enough for a sweeper to run and
+    // (incorrectly) remove the entry if it mishandled the indefinite TTL.
+    time.elapse(Duration::from_hours(1)).await;
     let entry = backend.get("k").await.expect("get");
     assert!(
         entry.is_some(),
         "SC-CACHE-011: an indefinite entry must survive a large time advance"
     );
-    tokio::time::resume();
+    time.end();
 }
 
 /// SC-CACHE-014: `PollingPrefixWatch` synthesizes `Changed`/`Deleted` diffs for
 /// a backend that does not natively support prefix watches.
 /// Capability-gated on `!features().prefix_watch`.
-pub async fn scenario_cache_014(backend: Arc<dyn ClusterCacheBackend>) {
+pub async fn scenario_cache_014(backend: Arc<dyn ClusterCacheBackend>, time: TimeControl) {
     if backend.features().prefix_watch {
         return; // native prefix watch; polyfill is not the subject here
     }
-    tokio::time::pause();
+    time.begin();
     let mut watch = PollingPrefixWatch::spawn(backend.clone(), "p/", Duration::from_millis(25));
 
     // A put under the prefix must be diffed as Changed.
@@ -438,36 +457,42 @@ pub async fn scenario_cache_014(backend: Arc<dyn ClusterCacheBackend>) {
         })
         .await
         .expect("put p/a");
-    tokio::time::advance(Duration::from_millis(50)).await;
-    tokio::task::yield_now().await;
+    time.elapse(Duration::from_millis(50)).await;
     assert!(
-        wait_for(
-            &mut watch,
-            |e| matches!(e, CacheEvent::Changed { key } if key == "p/a")
-        )
-        .await,
+        poll_watch(&mut watch)
+            .until(|e| matches!(e, CacheEvent::Changed { key } if key == "p/a"))
+            .await,
         "SC-CACHE-014: put under prefix must yield a Changed event"
     );
 
     // A delete must be diffed as Deleted.
     backend.delete("p/a").await.expect("delete p/a");
-    tokio::time::advance(Duration::from_millis(50)).await;
-    tokio::task::yield_now().await;
+    time.elapse(Duration::from_millis(50)).await;
     assert!(
-        wait_for(
-            &mut watch,
-            |e| matches!(e, CacheEvent::Deleted { key } if key == "p/a")
-        )
-        .await,
+        poll_watch(&mut watch)
+            .until(|e| matches!(e, CacheEvent::Deleted { key } if key == "p/a"))
+            .await,
         "SC-CACHE-014: delete under prefix must yield a Deleted event"
     );
-    tokio::time::resume();
+    time.end();
 }
 
 /// SC-CACHE-015: watch delivery is at-most-once per mutation — a single `put`
 /// must not cause more than one `Changed` event for the same key on one watch.
-pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>) {
-    tokio::time::pause();
+pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>, time: TimeControl) {
+    // The second poll's budget is what makes the duplicate check real, and it
+    // must stay a wait rather than a drain of what is already queued: the
+    // duplicate this scenario guards against is a backend echoing the mutation
+    // locally *and* again off its notification channel, so the second copy
+    // arrives a round-trip after the first. Both budgets are
+    // backend-independent — a paused clock auto-advances to the deadline while
+    // the runtime is idle, so `Virtual` pays no wall-clock for either (see
+    // [`crate::watch_poll`]) — and both are shorter than the default because a
+    // *passing* run always spends the second one in full.
+    const FIRST_EVENT_BUDGET: Duration = Duration::from_millis(500);
+    const DUPLICATE_BUDGET: Duration = Duration::from_millis(250);
+
+    time.begin();
     let mut watch = backend.watch("k").await.expect("watch");
     backend
         .put(PutRequest {
@@ -478,48 +503,29 @@ pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>) {
         .await
         .expect("put");
 
-    let mut changed_count = 0u32;
-    // Drain all immediately queued events without advancing time — any duplicate
-    // that would arrive synchronously will be in the channel already.
-    for _ in 0..64 {
-        match tokio::time::timeout(Duration::from_millis(0), watch.recv()).await {
-            Ok(Some(CacheWatchEvent::Event(CacheEvent::Changed { key }))) if key == "k" => {
-                changed_count += 1;
-            }
-            Ok(Some(CacheWatchEvent::Closed(_)) | None) | Err(_) => break,
-            _ => {}
-        }
-    }
-    assert_eq!(
-        changed_count, 1,
-        "SC-CACHE-015: a single put must deliver exactly one Changed event (got {changed_count})"
+    // Annotated so the closure is higher-ranked over the borrow, and therefore
+    // reusable across both polls.
+    let changed = |event: &CacheEvent| matches!(event, CacheEvent::Changed { key } if key == "k");
+    assert!(
+        poll_watch(&mut watch)
+            .upto(FIRST_EVENT_BUDGET)
+            .until(changed)
+            .await,
+        "SC-CACHE-015: a single put must deliver a Changed event"
     );
-    tokio::time::resume();
+    assert!(
+        !poll_watch(&mut watch)
+            .upto(DUPLICATE_BUDGET)
+            .until(changed)
+            .await,
+        "SC-CACHE-015: a single put must deliver at most one Changed event"
+    );
+    time.end();
 }
 
 // TODO(SC-CACHE-016/017) [L4]: slow-subscriber `Lagged` and connection-loss
 //   `Reset` are fault-injection scenarios delivered with the L4 harness, not the
 //   in-process suite.
-
-/// Polls a watch for up to a bounded number of events, returning `true` once one
-/// satisfies `pred`. Bounded so a missing event fails fast rather than hanging.
-async fn wait_for<P>(watch: &mut cluster_sdk::cache::CacheWatch, pred: P) -> bool
-where
-    P: Fn(&CacheEvent) -> bool,
-{
-    for _ in 0..64 {
-        match watch.recv().await {
-            Some(CacheWatchEvent::Event(event)) => {
-                if pred(&event) {
-                    return true;
-                }
-            }
-            Some(CacheWatchEvent::Closed(_)) | None => return false,
-            Some(_) => {}
-        }
-    }
-    false
-}
 
 /// A helper plugins may use to assert a backend's self-declared consistency
 /// before handing it to the consistency-sensitive default backends.
--- a/gears/system/cluster/cluster-conformance/src/factory.rs
+++ b/gears/system/cluster/cluster-conformance/src/factory.rs
@@ -1,0 +1,98 @@
+// Created: 2026-07-15 by Constructor Tech
+//! [`ScenarioBackend`] — an async-built backend plus its teardown, the unit a
+//! `run_*_conformance` runner drives one scenario against.
+//!
+//! The runners are generic over an **async** factory `Fn() -> Future<Output =
+//! ScenarioBackend<_>>` so a backend whose construction is genuinely async
+//! (opening a pool, running migrations, starting a container) can be built
+//! fresh per scenario — an in-memory fixture just wraps a ready value. Each
+//! backend is paired with a [`Teardown`] the runner awaits once the scenario
+//! finishes, so a backend that needs an explicit async shutdown (e.g. a plugin
+//! handle whose `Drop` guard panics if `stop()` was skipped) is released
+//! deterministically between scenarios rather than leaked.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+
+/// The async teardown a runner awaits after a scenario. `Send` so the runner
+/// future stays `Send` for `multi_thread` callers; a no-op for fixtures.
+pub type Teardown = Pin<Box<dyn Future<Output = ()> + Send>>;
+
+/// A backend built for one conformance scenario, plus the teardown to run once
+/// the scenario completes.
+pub struct ScenarioBackend<T: ?Sized> {
+    /// The backend the scenario exercises.
+    pub backend: Arc<T>,
+    /// Awaited by the runner after the scenario — closes pools / stops handles /
+    /// drops containers. [`ScenarioBackend::bare`] makes this a no-op.
+    pub teardown: Teardown,
+}
+
+impl<T: ?Sized> ScenarioBackend<T> {
+    /// A backend that needs no teardown (an in-memory fixture: dropping the
+    /// `Arc` is sufficient).
+    #[must_use]
+    pub fn bare(backend: Arc<T>) -> Self {
+        Self {
+            backend,
+            teardown: Box::pin(std::future::ready(())),
+        }
+    }
+
+    /// A backend with an async `teardown` (e.g. `async move { handle.stop().await }`),
+    /// awaited by the runner once the scenario returns.
+    #[must_use]
+    pub fn with_teardown<F>(backend: Arc<T>, teardown: F) -> Self
+    where
+        F: Future<Output = ()> + Send + 'static,
+    {
+        Self {
+            backend,
+            teardown: Box::pin(teardown),
+        }
+    }
+}
+
+/// Brackets one scenario: awaits the factory to build the backend, runs `run`
+/// against it, then awaits the teardown. The single place every runner threads
+/// a `ScenarioBackend` through, so the build → run → teardown order is
+/// consistent across all suites.
+///
+/// The scenario runs behind an async panic boundary so `teardown` is awaited
+/// **even when a scenario assertion panics** (`AssertUnwindSafe` +
+/// [`FutureExt::catch_unwind`](futures::FutureExt::catch_unwind)). Without it, a
+/// panic would drop the `teardown` future un-awaited; a teardown owning a handle
+/// with a panic-on-drop guard (e.g. `PostgresClusterHandle`) would then panic
+/// during unwind and abort the whole test process, masking the original failure.
+///
+/// `teardown` is awaited behind its own panic boundary too: if it panics after
+/// a scenario panic, resuming the scenario's panic (the actual assertion
+/// failure) takes priority over the teardown's, so the original failure still
+/// surfaces rather than being replaced by a cleanup-time panic. If the scenario
+/// succeeded, a teardown panic is resumed instead.
+pub(crate) async fn run_scenario<T, Fut, S, SFut>(make_fut: Fut, run: S)
+where
+    T: ?Sized,
+    Fut: Future<Output = ScenarioBackend<T>>,
+    S: FnOnce(Arc<T>) -> SFut,
+    SFut: Future<Output = ()>,
+{
+    use futures::FutureExt;
+
+    let scenario_backend = make_fut.await;
+    let outcome = std::panic::AssertUnwindSafe(run(scenario_backend.backend))
+        .catch_unwind()
+        .await;
+    let teardown_outcome = std::panic::AssertUnwindSafe(scenario_backend.teardown)
+        .catch_unwind()
+        .await;
+    match outcome {
+        Err(panic) => std::panic::resume_unwind(panic),
+        Ok(()) => {
+            if let Err(panic) = teardown_outcome {
+                std::panic::resume_unwind(panic);
+            }
+        }
+    }
+}
--- a/gears/system/cluster/cluster/src/defaults/discovery.rs
+++ b/gears/system/cluster/cluster/src/defaults/discovery.rs
@@ -12,7 +12,9 @@ use tracing::Instrument;
 
 use crate::defaults::{SVC_KEY_PREFIX, ShutdownRevoke, identity};
 use cluster_sdk::cache::types::{PutRequest, Ttl};
-use cluster_sdk::cache::{CacheEvent, CacheWatch, CacheWatchEvent, ClusterCacheBackend};
+use cluster_sdk::cache::{
+    CacheEvent, CacheWatch, CacheWatchEvent, ClusterCacheBackend, PollingPrefixWatch,
+};
 use cluster_sdk::discovery::{
     DiscoveryFilter, InstanceState, ServiceCommandReceiver, ServiceDiscoveryBackend,
     ServiceDiscoveryFeatures, ServiceHandle, ServiceInstance, ServiceRegistration, ServiceRequest,
@@ -58,6 +60,16 @@ const DEFAULT_TTL: Duration = Duration::from_secs(30);
 )]
 const DEFAULT_RENEWAL: Duration = Duration::from_secs(DEFAULT_TTL.as_secs() / 3);
 
+/// Default poll interval for the [`PollingPrefixWatch`] fallback used when the
+/// bound cache lacks a native prefix watch (`features().prefix_watch == false`).
+///
+/// Five seconds mirrors the Postgres plugin's `sd_poll_interval` default (its
+/// DESIGN §6) — the canonical `prefix_watch: false` backend — and trades
+/// topology-change latency for a bounded `scan_prefix` + `N` `get` cost per tick
+/// (`PollingPrefixWatch`'s documented cost). Override with
+/// [`with_prefix_watch_polling`](CacheBasedServiceDiscoveryBackend::with_prefix_watch_polling).
+const DEFAULT_PREFIX_WATCH_POLL: Duration = Duration::from_secs(5);
+
 /// The in-flight command buffer for each [`ServiceHandle`].
 const COMMAND_BUFFER: usize = 4;
 
@@ -92,17 +104,23 @@ const WATCH_CAPACITY: usize = 256;
 ///
 /// # Prefix-watch dependency
 ///
-/// Discovery and watch require the cache's native `watch_prefix`. A cache that
-/// declares no prefix watch returns
-/// [`ClusterError::Unsupported`] from `watch_prefix`, so
-/// [`watch`](ServiceDiscoveryBackend::watch) surfaces that error and
-/// [`discover`](ServiceDiscoveryBackend::discover) is **best-effort**: with no
-/// maintainer to reap entries on TTL expiry, [`register`](ServiceDiscoveryBackend::register)
-/// deliberately skips its local pre-insert (an unreapable pre-insert would read
-/// as live forever), so degraded-mode `discover` returns an empty set rather
-/// than a stale view. This is a known limitation pending the
-/// `PollingPrefixWatch` polyfill (DECOMPOSITION §2.7), which is out of scope
-/// here.
+/// Discovery and watch need a prefix stream. When the bound cache has a native
+/// `watch_prefix` (`features().prefix_watch == true`) it is used directly. When
+/// the cache declares no native prefix watch (`prefix_watch == false`, e.g. the
+/// Postgres plugin's cache), the backend transparently falls back to the
+/// [`PollingPrefixWatch`] polyfill over `scan_prefix`
+/// ([`open_prefix_watch`](Self::open_prefix_watch)) — so a maintainer runs and
+/// `register`'s local pre-insert is safe (it is reaped on TTL expiry /
+/// deregistration through the polling stream) over either kind of cache. The
+/// polyfill path pays the documented poll-interval latency and `N+1`
+/// round-trips per tick; tune it with
+/// [`with_prefix_watch_polling`](Self::with_prefix_watch_polling).
+///
+/// The historical degraded, always-empty mode (skip the pre-insert; return an
+/// empty set) only remains reachable if a maintainer genuinely cannot start —
+/// a transient native-`watch_prefix` failure, or a concurrent caller still
+/// bringing the maintainer up — not merely because the cache lacks native
+/// prefix watch.
 pub struct CacheBasedServiceDiscoveryBackend {
     cache: Arc<dyn ClusterCacheBackend>,
     shared: Arc<Shared>,
@@ -124,6 +142,11 @@ pub struct CacheBasedServiceDiscoveryBackend {
     provider: &'static str,
     /// The metrics sink (default [`NoopMetrics`]).
     metrics: Arc<dyn ClusterMetrics>,
+    /// Poll interval for the [`PollingPrefixWatch`] fallback, used only when the
+    /// bound cache declares `features().prefix_watch == false`. Defaults to
+    /// [`DEFAULT_PREFIX_WATCH_POLL`]; override with
+    /// [`with_prefix_watch_polling`](Self::with_prefix_watch_polling).
+    prefix_watch_poll: Duration,
 }
 
 impl CacheBasedServiceDiscoveryBackend {
@@ -142,6 +165,39 @@ impl CacheBasedServiceDiscoveryBackend {
             tasks: Arc::new(Mutex::new(Vec::new())),
             provider: "unknown",
             metrics: Arc::new(NoopMetrics),
+            prefix_watch_poll: DEFAULT_PREFIX_WATCH_POLL,
+        }
+    }
+
+    /// Sets the poll interval for the [`PollingPrefixWatch`] fallback (used only
+    /// over a cache with no native prefix watch). Mirrors
+    /// [`with_observability`](Self::with_observability); without it the backend
+    /// uses [`DEFAULT_PREFIX_WATCH_POLL`].
+    #[must_use]
+    pub fn with_prefix_watch_polling(mut self, interval: Duration) -> Self {
+        self.prefix_watch_poll = interval;
+        self
+    }
+
+    /// Opens the prefix-watch stream backing discovery for `prefix`.
+    ///
+    /// Uses the cache's **native** `watch_prefix` when it supports prefix
+    /// watches (`features().prefix_watch == true`) — that path is unchanged.
+    /// When the bound cache declares no native prefix watch it would return
+    /// [`ClusterError::Unsupported`]; instead of propagating that (which left
+    /// discovery permanently empty over such a cache — the Postgres plugin's
+    /// gap, DESIGN §6 / DECOMPOSITION §2.7), synthesize the stream with the
+    /// [`PollingPrefixWatch`] polyfill over `scan_prefix`, so discovery actually
+    /// works — at the polyfill's documented poll-latency / `N+1`-round-trip cost.
+    async fn open_prefix_watch(&self, prefix: &str) -> Result<CacheWatch, ClusterError> {
+        if self.cache.features().prefix_watch {
+            self.cache.watch_prefix(prefix).await
+        } else {
+            Ok(PollingPrefixWatch::spawn(
+                Arc::clone(&self.cache),
+                prefix,
+                self.prefix_watch_poll,
+            ))
         }
     }
 
@@ -183,8 +239,9 @@ impl CacheBasedServiceDiscoveryBackend {
     /// view.
     ///
     /// # Errors
-    /// Propagates the cache's [`ClusterError::Unsupported`] when it has no native
-    /// prefix watch.
+    /// Propagates a native `watch_prefix` failure. A cache with no native prefix
+    /// watch does **not** error here — it uses the [`PollingPrefixWatch`]
+    /// polyfill (see [`open_prefix_watch`](Self::open_prefix_watch)).
     async fn ensure_maintainer(&self, name: &str) -> Result<MaintainerStatus, ClusterError> {
         {
             let mut registry = self.shared.lock();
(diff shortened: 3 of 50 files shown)


gears/system/cluster/cluster-conformance/src/factory.rs
```diff
@@ -0,0 +1,71 @@
...
+pub(crate) async fn run_scenario<T, Fut, S, SFut>(make_fut: Fut, run: S)
+where
+    T: ?Sized,
+    Fut: Future<Output = ScenarioBackend<T>>,
+    S: FnOnce(Arc<T>) -> SFut,
+    SFut: Future<Output = ()>,
+{
+    let scenario_backend = make_fut.await;
+    run(scenario_backend.backend).await;
+    scenario_backend.teardown.await;
```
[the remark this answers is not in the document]
asmith987: Fixed — `run_scenario` now runs the scenario behind an `AssertUnwindSafe`/`catch_unwind` boundary, awaits teardown, then resumes the panic, so teardown always runs and the original assertion failure is preserved instead of being masked by a panic-on-drop abort.

gears/system/cluster/cluster/src/defaults/discovery.rs
```diff
@@ -556,6 +629,39 @@ fn take_str(bytes: &[u8], pos: &mut usize) -> Option<String> {
...
+        let Some(instance_id) = key.strip_prefix(prefix).map(str::to_owned) else {
+            continue;
+        };
+        if let Ok(Some(entry)) = cache.get(&key).await
+            && let Some(record) = InstanceRecord::decode(&entry.value)
+        {
+            fresh.insert(instance_id.clone(), record.to_instance(instance_id));
+        }
+    }
+    shared.lock().instances.insert(name.to_owned(), fresh);
```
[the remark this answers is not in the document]
asmith987: Fixed — added a per-service generation counter (bumped by `apply` only on real membership changes) that `reconcile_view` snapshots before its sweep and re-checks before committing, so a stale sweep is discarded rather than clobbering a newer watch event.

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -304,6 +304,17 @@ fn resolve_profile_backends(
...
+            // e.g. the Postgres backend) the default SD backend drives its
+            // topology watch through the `PollingPrefixWatch` polyfill; it uses
+            // its own poll-interval default (`DEFAULT_PREFIX_WATCH_POLL`, 5s)
+            // here. The provider-specific override (e.g. the Postgres plugin's
+            // `sd_poll_interval_ms`) is not visible at this SDK-side auto-wrap
+            // site — SD is built over the type-erased `Arc<dyn
+            // ClusterCacheBackend>`, not the concrete cache's config.
+            // FOLLOW-UP(cluster): thread a per-profile SD poll interval from
+            // operator config into this auto-wrap (via `with_prefix_watch_polling`)
+            // so a non-default cadence is honoured.
```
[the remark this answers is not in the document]
asmith987: Fixed — the configured `sd_poll_interval_ms` is now threaded through `ProfileBackends::with_sd_poll_interval` into `with_prefix_watch_polling` at the auto-wrap site, so an operator-set cadence is honored instead of the fixed 5s default.

gears/system/cluster/plugins/postgres-cluster-plugin/Cargo.toml
```diff
@@ -0,0 +1,145 @@
...
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+description = "Postgres backend plugin for the cluster gear: a native ClusterCacheBackend over a sqlx::PgPool plus a native DistributedLockBackend over pg_advisory_lock, for multi-instance, no-K8s deployments"
+readme = "README.md"
+keywords = ["constructorfabric", "cf-gears"]
+categories = ["development-tools"]
+metadata.docs.rs.all-features = true
```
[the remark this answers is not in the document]
asmith987: Not applied (false positive) — no `cf-studio-path` key exists in any Cargo.toml in this workspace (including the sibling plugins); it is Claude/Studio harness config in the root CLAUDE.md, not a Cargo-manifest requirement.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+impl PostgresCache {
+    #[must_use]
+    pub fn new(pool: PgPool, schema: &str) -> Arc<Self> {
+        Arc::new(Self {
+            pool,
+            table: format!("{schema}.cluster_cache"),
+            watch_registry: WatchRegistry::new(),
+        })
```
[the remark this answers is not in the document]
asmith987: Addressed defensively — the schema is operator config (same trust boundary as the connection string, not tenant input), but it is now validated to a safe identifier charset via `validate_schema` at startup, which also keeps the migration and runtime schemas resolving to the same object.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+    match ttl {
+        Ttl::Indefinite => Ok(None),
+        Ttl::Of(duration) => {
+            let chrono_duration = chrono::Duration::from_std(duration).map_err(|err| {
+                ClusterError::InvalidConfig {
+                    reason: format!("ttl {duration:?} is out of range: {err}"),
+                }
+            })?;
+            Ok(Some(chrono::Utc::now() + chrono_duration))
+        }
```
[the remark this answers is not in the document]
asmith987: Fixed — `expires_at` is now computed in SQL as `now() + $n::interval` on the database clock rather than from `chrono::Utc::now()`, so expiry no longer depends on each instance's wall clock.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch_tests.rs
```diff
@@ -0,0 +1,132 @@
...
+    let registry = WatchRegistry::new();
+    let mut watch = registry.register("k");
+
+    registry.close_all();
+
+    assert!(matches!(
+        watch.recv().await,
+        Some(CacheWatchEvent::Closed(ClusterError::Shutdown))
+    ));
+}
```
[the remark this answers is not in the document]
asmith987: Fixed — added tests that saturate the 64-slot buffer before `dispatch(Reset)`/`close_all()` and assert the typed `Reset`/`Closed(Shutdown)` terminal event is still delivered.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
```diff
@@ -0,0 +1,421 @@
...
+                let _delivered = slot.sender.try_send(make_event());
+            }
+        }
+        self.watchers.clear();
+    }
+
+    /// Closes every active watch terminally with [`ClusterError::Shutdown`]
+    /// (DESIGN.md §10 step 2, `PG-LIFE-004`) before the LISTEN task exits.
+    pub fn close_all(&self) {
+        self.broadcast_and_clear(|| CacheWatchEvent::Closed(ClusterError::Shutdown));
```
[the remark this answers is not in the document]
asmith987: Fixed — `broadcast_and_clear` now delivers the typed terminal event via a bounded, concurrent `send` (immediate when the buffer has room, grace-bounded when momentarily full) instead of a discarded `try_send`, guaranteeing the typed terminal to every draining watcher without hanging shutdown.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -0,0 +1,891 @@
...
+/// the async mutex — so a concurrent reaper `remove` on the same key/shard can
+/// never block a shard lock held across an `.await`, which would deadlock the
+/// single worker under a current-thread runtime (`GAP-SOLUTIONS.md` §5). The
+/// mutex serializes the accessors with a plain `.await`, not a thread-blocking
+/// wait.
+struct HeldLock {
+    conn: Arc<tokio::sync::Mutex<PoolConnection<sqlx::Postgres>>>,
+    key1: i32,
+    key2: i32,
+}
```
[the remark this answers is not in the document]
asmith987: Fixed — `holder_id` is stored in `HeldLock`, threaded into the guard task, and matched in `renew`/`release` and the reaper's reclaim, so a stale guard or foreign reaper can no longer renew, release, or reclaim a newer holder's live lock.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -0,0 +1,891 @@
...
+    /// race for a given name, `remove` here returns `None` and there is simply
+    /// nothing left to reclaim.
+    pub async fn drain_held(&self) {
+        let names: Vec<String> = self.held.iter().map(|entry| entry.key().clone()).collect();
+        for name in names {
+            if let Some((_, held_lock)) = self.held.remove(&name) {
+                reaper::reclaim(&name, held_lock, &*self.metrics, self.provider).await;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed — `try_acquire` now rejects acquisitions once shutdown is observed (pre- and post-insert, self-reclaiming the connection) and `drain_held` loops until empty, so no lock can be inserted after the drain snapshot and strand a connection that blocks `pool.close()`.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -0,0 +1,891 @@
...
+        .bind(key1)
+        .bind(key2)
+        .fetch_one(&mut *conn)
+        .await
+        .map_err(map_sqlx_error)?;
+
+    if !acquired {
+        // `conn` drops here, returning to the pool.
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed — added a pool `after_release` hook running `pg_advisory_unlock_all()` (releasing any advisory lock left by a cancelled acquire) and bounded each `try_acquire` attempt by the remaining deadline, while still running the first attempt to completion.

gears/system/cluster/plugins/postgres-cluster-plugin/src/migrations/lock/0002_cluster_lock.sql
```diff
@@ -0,0 +1,23 @@
...
+-- `ttl_ms` lets the lock TTL reaper (DESIGN.md §5.2) identify expired locks.
+-- `holder_id` is a random UUID generated at acquire time; `release()` guards
+-- on it to prevent a foreign holder from releasing another's lock.
+CREATE TABLE cluster_lock (
+    name        TEXT        NOT NULL,
+    holder_id   TEXT        NOT NULL,
+    acquired_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+    ttl_ms      BIGINT      NOT NULL,
+    PRIMARY KEY (name)
+);
```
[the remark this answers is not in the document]
asmith987: Fixed — the pool now pins `search_path` to the configured schema and `CREATE SCHEMA IF NOT EXISTS` runs before the migrators, so the unqualified DDL lands in the same schema the runtime queries target (also fixed the same latent issue in the cache migration).

gears/system/cluster/plugins/postgres-cluster-plugin/src/plugin.rs
```diff
@@ -0,0 +1,303 @@
...
+        // Awaited, not fire-and-forgotten: the LISTEN connection must be
+        // confirmed live before `build_and_start` returns (DESIGN.md §3.2
+        // step 5's "no readiness gate" guarantee — see
+        // `cache::watch::spawn_listen_task`'s doc for the race this closes).
+        let listen_task = crate::cache::watch::spawn_listen_task(
+            config.connection_string.clone(),
+            cache.watch_registry(),
+            shutdown.clone(),
+        )
+        .await?;
```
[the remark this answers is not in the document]
asmith987: Fixed — on a `spawn_listen_task` failure the shared shutdown token is now cancelled and both reapers are awaited (and the pool closed) before returning the error, so they no longer detach and keep the pool alive.

asmith987:
## Review follow-up — remaining (Minor / Nitpick) items

The Critical/Major inline comments each have a threaded reply above. The Minor/Nitpick items posted in the review summary (no inline thread to reply to) are addressed as follows:

- **C5** — `delete` / `compare_and_delete` now include the live-entry predicate `AND (expires_at IS NULL OR expires_at > now())`, so an expired-but-unreaped row is treated as absent (no spurious `true`/`Deleted`).
- **C6** — Dropped the "NOTIFY overflow → empty payload" claim in `watch.rs` and DESIGN.md/TESTING.md; overflow is now described as a commit-time failure, and the empty/unrecognized-payload → `Reset` fallback is kept for bare/foreign NOTIFYs.
- **D4** — `wait_for` in the cache conformance suite now bounds the whole poll with `tokio::time::timeout`, so a missing event fails the assertion instead of hanging.
- **D5** — SC-LOCK-005 elapses 50ms (not 150ms) before renew, leaving ~150ms of the 200ms lease as margin.
- **D6** — `TimeControl::elapse` doc corrected to state `Real` sleeps `min(dur, REAL_CAP)`, not "at least `dur`".
- **D7** — `conformance.rs` overview updated to reflect the shared-container / per-schema / `drain_held` lock isolation.
- **E2** — `config.validate()` rejects zero `cache_reaper_interval_ms` / `lock_reaper_interval_ms` / `sd_poll_interval_ms` at startup (they would panic `tokio::time::interval`).
- **E3** — `pg_lock_003` now synchronizes on the waiter's registration (`__test_release_waiter_count > 0`) before releasing, so it provably exercises the NOTIFY wake path.
- **E4** — `postgres_specific` empty-payload watch test uses the standard 3s delivery timeout instead of 500ms.
- **E5** — `wait_until` bounds each in-flight `condition().await` by the overall timeout.
- **E6** — TESTING.md key-length references corrected from 8190 to the implemented 7997 (`MAX_KEY_BYTES`).

**Not applied (false positives):** the `Cargo.toml` `cf-studio-path` suggestion (no such key exists anywhere in this workspace) and the SQL-identifier hardening framed as a vulnerability (schema is operator config, same trust boundary as the connection string — still hardened defensively via `validate_schema`).

**Verification:** `cargo clippy --all-targets --features integration` clean, `make dylint` clean, and the full Postgres integration suite passes (99 tests, 0 failures) alongside all unit/conformance tests.

gears/system/cluster/cluster/src/defaults/discovery.rs
```diff
@@ -408,6 +481,28 @@ struct Registry {
...
+    fn generation(&self, name: &str) -> u64 {
+        self.generations.get(name).copied().unwrap_or(0)
+    }
+
+    /// Bumps `name`'s revision; called under the registry lock by every
+    /// maintainer mutation so an in-flight reconcile sweep can detect it raced.
+    fn bump_generation(&mut self, name: &str) {
+        let counter = self.generations.entry(name.to_owned()).or_default();
+        *counter = counter.wrapping_add(1);
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. The local pre-insert in `register` now calls `registry.bump_generation(&reg.name)` under the same lock, right after the insert. An in-flight `reconcile_view` sweep that snapshotted the generation before the pre-insert now detects the race and discards its stale rebuilt map at the commit check instead of clobbering the fresh registration. `StoreMaintainer::apply`'s generation handling is unchanged.

gears/system/cluster/cluster/src/defaults/discovery.rs
```diff
@@ -556,6 +651,53 @@ fn take_str(bytes: &[u8], pos: &mut usize) -> Option<String> {
...
+    for key in keys {
+        let Some(instance_id) = key.strip_prefix(prefix).map(str::to_owned) else {
+            continue;
+        };
+        if let Ok(Some(entry)) = cache.get(&key).await
+            && let Some(record) = InstanceRecord::decode(&entry.value)
+        {
+            fresh.insert(instance_id.clone(), record.to_instance(instance_id));
+        }
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. `reconcile_view`'s per-key loop now `match`es on `cache.get`: `Ok(Some)` includes the entry, `Ok(None)` legitimately omits a key that expired between the scan and the get, and `Err(_)` returns early **before** the commit block — preserving the existing service view rather than committing a partial one. The next reconcile/`discover` re-sweeps, so it stays self-healing.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+| `DistributedLockBackend` | Native — `pg_advisory_lock` + `cluster_lock` metadata table. Independently routable via `lock: { provider: postgres }` (§3.5), with its own pool/config — not required to be paired with the postgres cache provider | `linearizable: true` | — |
+| `ServiceDiscoveryBackend` | SDK default `CacheBasedServiceDiscoveryBackend` over Postgres cache | — | `metadata_pushdown: false` |
+
+`prefix_watch: false` means that consumers requiring `CacheCapability::PrefixWatch` cannot bind this backend without the polyfill. The service-discovery default backend uses `watch_prefix` internally and therefore falls back to `PollingPrefixWatch` on a prefix-watch-incapable cache; the wiring crate enables this fallback automatically (see §6).
+
+## 2. Domain Model
+
+### 2.1 Database Tables
+
+Three tables are owned by this plugin. All live in the schema specified by the plugin config (default: `public`). Migration is managed via `sqlx-macros` embedded migrations; the wiring crate runs them at startup before registering backends.
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. Updated to "Two tables are owned by this plugin, plus one virtual NOTIFY channel." Only the `cluster_cache` and `cluster_lock` migrations create tables; the lock-release NOTIFY channel is not a table.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/TESTING.md
```diff
@@ -0,0 +1,316 @@
...
+    // Built via the standalone lock-only path (§3.5 DESIGN.md), the same one
+    // ClusterLockProvider::build_lock uses in production — not the combined
+    // cache+lock plugin. `_cache` is ignored: this exercises the real
+    // "independently routable" shape, not a shared-pool shortcut.
+    let handle = PostgresLockPlugin::builder(test_lock_config())
+        .build_and_start()
+        .await
+        .expect("standalone lock plugin starts");
+    handle.lock()
+}
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. The TESTING.md example now uses the owning `ScenarioBackend::with_teardown(backend, async move { handle.stop().await })` pattern for all four factories (cache/lock/leader/discovery), and documents why it is required (the handle's debug `Drop` panic + background-task teardown) along with the `TimeControl::Real` argument. Note the shipped `tests/conformance.rs` was already correct — this was a stale documentation snippet, not a bug in the test code.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -0,0 +1,1042 @@
...
+async fn try_acquire(
+    pool: &PgPool,
+    table: &str,
+    held: &Arc<DashMap<String, HeldLock>>,
+    name: &str,
+    ttl: Duration,
+    ctx: &GuardContext,
+) -> Result<Option<LockGuard>, ClusterError> {
+    let (key1, key2) = lock_key(name);
+    let ttl_ms = duration_to_ttl_ms(ttl)?;
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. Added `MAX_LOCK_NAME_BYTES = 7999` and `validate_lock_name`, called at the top of `try_acquire` **before** any lock state is mutated, returning `ClusterError::InvalidName`. Both `try_lock` and `lock` route through `try_acquire`, so `release` can never reach a lock whose name will not fit the NOTIFY payload. Added boundary unit tests: accept at exactly 7999 bytes, reject at 8000, and reject a multi-byte name that is under the char count but over the byte limit.

tools/dylint_lints/lint_utils/src/lib.rs
```diff
@@ -229,6 +229,43 @@ pub fn is_in_toolkit_db_path(source_map: &SourceMap, span: Span) -> bool {
...
+    check_span_path(
+        source_map,
+        span,
+        "/gears/system/cluster/plugins/postgres-cluster-plugin/",
+    ) || check_span_path(
+        source_map,
+        span,
+        "gears/system/cluster/plugins/postgres-cluster-plugin/",
+    ) || check_span_path(source_map, span, "postgres-cluster-plugin/src/")
+}
```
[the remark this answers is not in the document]
asmith987: Fixed in d8a0856b. Replaced `str::contains` in `check_span_path` with a component-aware `path_matches` helper: a match is accepted only at the start of the path, immediately after a `/`, or via the pattern's own leading `/`. So `not-gears/system/.../postgres-cluster-plugin/` and `my-postgres-cluster-plugin/src/` no longer match. This fixes every `is_in_*_path` helper at once; the slash-bounded patterns (`/domain/`, `/gears/`, …) keep their existing behaviour. Added unit tests covering absolute / relative / simulated-dir matches, Windows separators, the slash-bounded segment patterns, and the near-miss exclusions.

On `test_comment_annotations_match_stderr`: that helper already exists in `lint_utils/src/lib.rs` and is exercised by `de0706_no_direct_sqlx`'s test suite (`tests::test_comment_annotations_match_stderr`), so no new implementation was needed — that part of the suggestion was a false positive.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -0,0 +1,1070 @@
...
+    /// interval so the guard's own timer never fires during the test) so the
+    /// correction is deterministic and never races the guard task's own
+    /// `held.get_mut`-across-`.await` (a DashMap current-thread deadlock hazard,
+    /// `GAP-SOLUTIONS.md` §5).
+    ///
+    /// Gated behind `--features integration` (PGR-M8).
+    #[cfg(feature = "integration")]
+    #[doc(hidden)]
+    pub async fn __test_reassert_synchronous_commit(&self, name: &str) {
+        let _reasserted = reassert_synchronous_commit(&self.held, name).await;
```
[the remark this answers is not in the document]
asmith987: Fixed: added a `holder_id` parameter to `reassert_synchronous_commit` and a no-op guard when the current `held` entry belongs to a different holder, so a stale guard whose lock lapsed and was re-acquired can no longer run `SET` on the successor's pinned connection (the guard task passes its own `holder_id`, and the `__test_reassert_synchronous_commit` seam resolves the current holder internally so it still exercises the matching path).

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/notify.rs
```diff
@@ -0,0 +1,338 @@
...
+/// costs a waiter its heartbeat-interval latency (`lock/mod.rs`), never
+/// correctness, so silently resuming is sufficient.
+pub fn spawn_release_listen_task(
+    connection_string: String,
+    waiters: Arc<ReleaseWaiters>,
+    cancel: CancellationToken,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(async move {
+        let Ok(mut listener) = connect_and_listen(&connection_string).await else {
+            return;
```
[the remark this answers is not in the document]
asmith987: Fixed: the initial connect now goes through the same cancellation-aware retry loop as reconnects via a new `connect_and_listen_cancellable` helper, so a startup Postgres blip falls into backoff retry instead of permanently disabling NOTIFY wakeups, and every `connect_and_listen` await (initial and reconnect) is now interruptible by `cancel` so shutdown never stalls on network I/O.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+| `compare_and_swap(key, expected_version: u64, new_value, ttl: Ttl) -> CacheEntry` | `UPDATE cluster_cache SET value = $3, version = version + 1, expires_at = $4 WHERE key = $1 AND version = $2 AND (expires_at IS NULL OR expires_at > now()) RETURNING version` — zero rows → `CasConflict` |
+| `compare_and_delete(key, expected_value) -> bool` | `DELETE FROM cluster_cache WHERE key = $1 AND value = $2` |
+| `scan_prefix(prefix) -> Vec<String>` | `SELECT key FROM cluster_cache WHERE key LIKE $1 || '%' AND (expires_at IS NULL OR expires_at > now())` — uses `LIKE` with a literal prefix; the `%` is appended by the plugin, not the caller |
```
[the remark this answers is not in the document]
asmith987: Fixed: the SQL contract table now shows the `expires_at IS NULL OR expires_at > now()` active-entry predicate on `delete`/`compare_and_delete` and documents `scan_prefix`'s Rust-side escaped pattern binding with `ESCAPE '\'` (which also removes the stray `||` that triggered the MD056 column-count warning).

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+
+This avoids busy-polling: waiters wake promptly when a holder explicitly releases. TTL-based reaper releases also trigger `NOTIFY cluster_lock_released` after unlocking.
+
+### 5.4 PgBouncer Constraint
+
+Session-level advisory locks are **incompatible with PgBouncer in transaction pooling mode** — the advisory lock is released when PgBouncer returns the connection to the pool between transactions, even if the Rust code still holds `LockGuard`. The plugin documents this prominently in config validation:
+
+- If `pgbouncer_transaction_mode: true` is set in config, `build_and_start` returns `Err(ClusterError::InvalidConfig { reason: "pg_advisory_lock requires session-mode pooling or a direct connection; transaction-mode PgBouncer incompatible with distributed locks" })`.
+- Operators using PgBouncer must either use session pooling mode for the cluster plugin's connection string, or use a different lock backend.
```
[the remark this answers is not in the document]
asmith987: Fixed: §5.4 and the §11 risk entry now explain that transaction pooling does not pin a client to one server session, so returning the connection leaves the session advisory lock attached to the pooled server session where it outlives the guard and can leak to the next client, rather than claiming the lock is released between transactions (the `build_and_start` rejection and operator guidance are unchanged).

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+
+```
+<event_type>:<key>
+```
+
+Where `<event_type>` is one of `C` (Changed), `D` (Deleted), `E` (Expired). The key must therefore be ≤ 7997 bytes (7999-byte payload limit minus the two-byte `<event_type>:` prefix); the plugin validates this constraint at write time (`cache::watch::MAX_KEY_BYTES`) and returns `ClusterError::InvalidName` for keys that would exceed it.
+
+An empty payload — a bare `NOTIFY cluster_cache_changes` (no payload) from an unrelated writer, or any value this plugin's own version never produces — is interpreted by the LISTEN task as a `Reset` signal, broadcasting `CacheWatchEvent::Reset` to all active watchers so consumers re-read their keys (ADR-003 §"NOTIFY overflow mapping"). Note this is *not* how NOTIFY queue overflow surfaces: Postgres does not emit an empty-payload notification on overflow — it aborts the committing transaction with an error ("too many notifications in the NOTIFY queue") and broadcasts nothing. Overflow recovery instead rides the LISTEN task's reconnect-then-`Reset` path (below).
```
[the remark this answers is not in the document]
asmith987: Fixed: the overflow descriptions at §2.3, §4.3, and the §11 risk now state that queue overflow aborts the producer transaction (surfacing as the failing write's `Provider` error and in PostgreSQL logs) without disconnecting the listener or incrementing `cluster_watch_resets_total`, reserving reconnect/`Reset` language for actual LISTEN connection gaps.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/notify.rs
```diff
@@ -0,0 +1,374 @@
...
+                    }
+                    Ok(None) => {
+                        // Transparent reconnect already happened inside
+                        // `try_recv`; nothing else to do.
+                        attempt = 0;
+                    }
+                    Err(_lost) => {
+                        match reconnect_with_backoff(&connection_string, &policy, &mut attempt, &cancel).await {
+                            Some(reconnected) => listener = reconnected,
+                            None => return,
```
MikeFalcon77: **MEDIUM**

On reconnect-budget exhaustion the release-listen background task returns silently, conflating a terminal connection-loss failure with graceful cancellation and emitting no warn/error log or consumer signal.

Distinguish cancellation from budget exhaustion and log at warn/error (and/or signal waiters) on exhaustion before the task exits, so the permanent loss of NOTIFY wakeups is observable.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
```diff
@@ -0,0 +1,472 @@
...
+    cancel: &CancellationToken,
+) -> Option<PgListener> {
+    while *attempt < policy.max_retries {
+        let backoff = policy.backoff_for(*attempt);
+        *attempt += 1;
+        tokio::select! {
+            () = cancel.cancelled() => return None,
+            () = tokio::time::sleep(backoff) => {}
+        }
+        if let Ok(listener) = connect_and_listen(connection_string).await {
```
MikeFalcon77: **MEDIUM**

The LISTEN task's mid-stream reconnect awaits `connect_and_listen` directly without selecting on the CancellationToken, so a shutdown fired while `PgListener::connect` reaches an unreachable host stalls `stop()` for the full connect timeout.

Wrap the reconnect connect in a cancellation-aware helper as `lock/notify.rs` does (`connect_and_listen_cancellable` selecting on `cancel.cancelled()`), returning `None` on cancellation so the task exits promptly.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
```diff
@@ -0,0 +1,472 @@
...
+                received = listener.try_recv() => match received {
+                    Ok(Some(notification)) => {
+                        attempt = 0;
+                        registry.dispatch(&parse_notification(notification.payload())).await;
+                    }
+                    Ok(None) => {
+                        // `PgListener`'s own transparent reconnect just ran and
+                        // succeeded; events during the gap may have been missed.
+                        attempt = 0;
+                        registry.dispatch(&ParsedNotification::Reset).await;
```
MikeFalcon77, line 465: **MEDIUM**

DESIGN §8 says the LISTEN task's Reset broadcasts call `ClusterMetrics::watch_reset("cache")` (backing `cluster_watch_resets_total`) and emit a `cluster.watch.reset` WARN log, but the watch task takes no `ClusterMetrics` and emits neither signal on any Reset.

Thread the `ClusterMetrics` sink into the LISTEN task/`WatchRegistry` and emit `watch_reset("cache")` plus a `cluster.watch.reset` WARN wherever a `Reset` is dispatched, or amend DESIGN §8 to state these are covered solely by the SDK combinator.

gears/system/cluster/plugins/postgres-cluster-plugin/src/plugin.rs
```diff
@@ -0,0 +1,327 @@
...
+             background tasks may leak"
+        );
+    }
+}
+
+/// Wraps a native cache in the SDK's `InstrumentedCache` decorator so its
+/// operations emit the contracted `cluster.cache.*` signals (DESIGN.md §8),
+/// mirroring `StandaloneClusterHandle`'s construction
+/// (`standalone-cluster-plugin/src/plugin.rs`).
+pub fn instrumented_cache(
```
MikeFalcon77: **LOW**

`instrumented_cache` is declared `pub` but is called only once within `plugin.rs`, is not re-exported by `lib.rs`, and exposes the internal `PostgresCache` type in its signature.

Drop `pub` (or use `pub(crate)` only if cross-module reuse is intended) so the function's visibility matches its actual use.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch_tests.rs
```diff
@@ -0,0 +1,206 @@
...
+    registry.close_all().await;
+
+    assert!(matches!(
+        watch.recv().await,
+        Some(CacheWatchEvent::Closed(ClusterError::Shutdown))
+    ));
+}
+
+#[tokio::test]
+async fn dispatch_prunes_a_watcher_whose_consumer_dropped_it() {
```
MikeFalcon77, line 127: **LOW**

The test named for pruning a dropped watcher makes no assertion and relies only on the absence of a panic, so it would pass even if the dead entry were never pruned.

Assert the registry actually removed the dropped watcher's entry after dispatch (inspect the internal subscriber map as the `notify.rs` tests do).

gears/system/cluster/plugins/postgres-cluster-plugin/src/pg_error.rs
```diff
@@ -0,0 +1,53 @@
...
+/// | `sqlx::Error::PoolTimedOut` | `Provider { Timeout }` |
+/// | `sqlx::Error::PoolClosed` | `Provider { ConnectionLost }` |
+/// | SQLSTATE `28xxx` (invalid auth) | `Provider { AuthFailure }` |
+/// | Any other `sqlx::Error` | `Provider { Other }` |
+///
+/// Takes `err` by value (not `&sqlx::Error`) so call sites can pass this
+/// directly as `.map_err(map_sqlx_error)` without a wrapping closure — the
+/// dominant call pattern across `cache/mod.rs` — rather than for its own sake.
+#[allow(clippy::needless_pass_by_value)]
+pub fn map_sqlx_error(err: sqlx::Error) -> ClusterError {
```
MikeFalcon77, line 23: I think this error mapper should live somewhere in toolkit-db. Otherwise, we'll end up duplicating these errors and the pooling parameters.

gears/system/cluster/plugins/postgres-cluster-plugin/src/pg_setup.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// A [`PgPoolOptions`] with the `after_connect`/`before_acquire`/`after_release`
+/// hooks wired (DESIGN.md §3.4) — callers still need to chain
+/// `.max_connections(...)`, `.acquire_timeout(...)`, and `.connect(...)`.
+///
+/// `after_connect` pins every connection's `search_path` to `schema` so the
+/// migrations' unqualified `CREATE TABLE`s land in the same schema the runtime
+/// queries target with their `{schema}.cluster_*` qualifiers (PGR-L4). `schema`
+/// must already have passed [`validate_schema`] — it is interpolated unquoted.
+pub fn base_pool_options(schema: &str) -> PgPoolOptions {
```
MikeFalcon77, line 22: The same thing about pgx pooling parameters. We already have some pooling utils in toolokit-db

gears/system/cluster/cluster/src/wiring.rs
```diff
@@ -125,6 +142,23 @@ impl ClusterWiring {
...
+            // the omit-default auto-wrap (PGR-D3). The interval lives in the
+            // cache provider's own options (e.g. the Postgres plugin's
+            // `sd_poll_interval_ms`); read it generically here so a non-default
+            // cadence reaches `with_prefix_watch_polling` instead of being
+            // silently forced to the 5s default. A zero/absent value keeps the
+            // default.
+            if let Some(interval_ms) = profile
+                .cache
+                .options
+                .get("sd_poll_interval_ms")
```
MikeFalcon77: **LOW**

The service-discovery poll cadence is read from the cache provider options via the hard-coded literal `"sd_poll_interval_ms"`, duplicating `PostgresClusterConfig`'s field name across a crate boundary with no shared constant.

If that config field is ever renamed, this read silently falls back to the default with no compile error, so an operator-configured cadence would be ignored.

Expose the key as a `pub const` in `cluster-sdk` (or a documented provider-options convention) and reference it from both the config field and this read so the coupling is single-sourced.

gears/system/cluster/plugins/postgres-cluster-plugin/tests/postgres_specific.rs
```diff
@@ -0,0 +1,713 @@
...
+/// private implementation detail (not re-exported from this crate's public
+/// API), so this test cannot *force* a synthetic collision from outside;
+/// instead it exercises the property the DESIGN doc claims — two
+/// unrelated, ordinary lock names never contend with each other — which is
+/// what "the collision is mechanically unreachable in practice" actually
+/// predicts. A real collision test would require reaching into
+/// `postgres_cluster_plugin::lock::lock_key` directly, which is exactly
+/// the private surface DESIGN.md §5.1 says is an implementation detail.
+#[tokio::test]
+async fn pg_spec_003_lock_hash_collision_is_documented_not_forced() {
```
MikeFalcon77: **LOW**

`pg_spec_003` acquires two differently-named locks and releases them but asserts nothing about the advisory-lock key-collision property it claims to cover.

Two arbitrary distinct names not contending would pass even if the 64-bit `(key1,key2)` hash split were broken, so the test cannot fail when the property under test regresses, and adds no coverage beyond the `lock_tests.rs` unit tests.

Either drop it in favor of the existing unit tests, or expose `lock_key` via a test seam and assert the actual `(key1,key2)` mapping / a forced-collision contention path.

gears/system/cluster/plugins/postgres-cluster-plugin/src/pg_error.rs
MikeFalcon77: I mean generic classification (which SQLSTATE/variant means what) is the reusable part and may live in toolkit-db.  But for now there's no duplication today and the plugin deliberately doesn't depend on toolkit-db.
So this descision is up to you.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/notify.rs
asmith987: Fixed.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
asmith987: Fixed. Mid stream reconnect now uses a cancellation aware connect helper that prevents shutdown from stalling on connect attempts.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
asmith987: Fixed. `ClusterMetrics` is now threaded into each LISTEN task so every Reset emits the proper metric + warning.

gears/system/cluster/plugins/postgres-cluster-plugin/src/plugin.rs
asmith987: Fixed. Dropped pub from `instrumented_cache`

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch_tests.rs
asmith987: Fixed. Added assertion that the registry actually prunes the dropped watcher's entry.

gears/system/cluster/plugins/postgres-cluster-plugin/src/pg_error.rs
asmith987: Since this plugin purposely doesn't depend on toolkit-db, I think if another plugin ends up needing a similar SQLSTATE mapping, we can revisit it there at that point.

gears/system/cluster/cluster/src/wiring.rs
asmith987: Fixed. Added a shared constant so there is only a single source instead of duplicating the string literal.

gears/system/cluster/plugins/postgres-cluster-plugin/tests/postgres_specific.rs
asmith987: Fixed. Removed the test in favor for the `lock_tests.rs` unit tests since those tests actually exercise the collision property against `lock_key` directly.

gears/system/cluster/cluster-conformance/src/cache.rs
```diff
@@ -478,11 +487,19 @@ pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>) {
         .await
         .expect("put");
 
+    // Per-`recv` budget: `Virtual` fixtures deliver synchronously into the
+    // channel during `put().await`, so a zero-timeout drain sees every event
+    // already queued. A `Real` backend delivers the `Changed` asynchronously
+    // (e.g. a Postgres NOTIFY round-trip), so wait a real grace window for the
+    // first event — and keep draining that same window afterward, so a
+    // spurious duplicate (the bug this guards against) is still caught.
+    let recv_budget = match time {
```
striped-zebra-dev: This feels odd for two reasons:
1. The test now depends on the internal behavior of TimeControl.
2. A zero timeout is treated as correct for Virtual, while Real gets 500 ms, even though both paths ultimately verify the same asynchronous result.

Could we instead use one backend-independent strategy—for example, wait up to 500 ms for the first event, then drain immediately available events? That would preserve the duplicate check without encoding backend-specific timing assumptions into the test.

gears/system/cluster/cluster-conformance/src/cache.rs
```diff
@@ -494,31 +511,46 @@ pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>) {
...
     P: Fn(&CacheEvent) -> bool,
 {
-    for _ in 0..64 {
-        match watch.recv().await {
-            Some(CacheWatchEvent::Event(event)) => {
-                if pred(&event) {
-                    return true;
+    const WAIT_BUDGET: Duration = Duration::from_secs(3);
+    let poll = async {
+        for _ in 0..64 {
```
striped-zebra-dev: Suggestion: rather than just naming `64`, fold both bounds into a small builder so the numbers stay named and are overridable per call site if a future scenario needs a different budget:

```rust
wait_for(&mut watch, pred).await                                   // defaults: 3s / 64
wait_for(&mut watch, pred).upto(Duration::from_millis(500)).await  // override
```

```rust
const DEFAULT_WAIT_BUDGET: Duration = Duration::from_secs(3);
const DEFAULT_MAX_SKIPPED_EVENTS: usize = 64;

/// Builder for `wait_for`. Polls a watch for up to a bounded number of
/// events, resolving `true` once one satisfies `pred`.
///
/// Bounded two ways so a notification/polling regression fails the assertion
/// instead of hanging the whole suite (PGR-D4): `max_skipped` limits how many
/// wrong events it will skip, and `budget` wraps the entire poll in a timeout
/// so an event that never arrives (each individual `recv().await` is
/// otherwise unbounded) resolves to `false` rather than blocking forever.
/// Under `TimeControl::Virtual` a paused clock auto-advances to the timeout
/// when no event is pending, so it still fails fast; the defaults are
/// generous enough for a real backend's NOTIFY / polling round-trip - call
/// `upto` to override when a scenario needs a tighter or looser bound.
struct WaitFor<'a, P> {
    watch: &'a mut cluster_sdk::cache::CacheWatch,
    pred: P,
    budget: Duration,
    max_skipped: usize,
}

fn wait_for<P>(watch: &mut cluster_sdk::cache::CacheWatch, pred: P) -> WaitFor<'_, P>
where
    P: Fn(&CacheEvent) -> bool,
{
    WaitFor {
        watch,
        pred,
        budget: DEFAULT_WAIT_BUDGET,
        max_skipped: DEFAULT_MAX_SKIPPED_EVENTS,
    }
}

impl<'a, P> WaitFor<'a, P>
where
    P: Fn(&CacheEvent) -> bool,
{
    /// Overrides the default overall wait budget.
    fn upto(mut self, budget: Duration) -> Self {
        self.budget = budget;
        self
    }
}

impl<'a, P> std::future::IntoFuture for WaitFor<'a, P>
where
    P: Fn(&CacheEvent) -> bool + 'a,
{
    type Output = bool;
    type IntoFuture = std::pin::Pin<Box<dyn Future<Output = bool> + 'a>>;

    fn into_future(self) -> Self::IntoFuture {
        let WaitFor { watch, pred, budget, max_skipped } = self;
        Box::pin(async move {
            let poll = async {
                for _ in 0..max_skipped {
                    match watch.recv().await {
                        Some(CacheWatchEvent::Event(event)) => {
                            if pred(&event) {
                                return true;
                            }
                        }
                        Some(CacheWatchEvent::Closed(_)) | None => return false,
                        Some(_) => {}
                    }
                }
                false
            };
            tokio::time::timeout(budget, poll).await.unwrap_or(false)
        })
    }
}
```

`IntoFuture` makes the builder awaitable directly, so every existing `wait_for(...).await` call site is unaffected.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+
+### 2.1 Database Tables
+
+Two tables are owned by this plugin, plus one virtual NOTIFY channel. All live in the schema specified by the plugin config (default: `public`). Migration is managed via `sqlx-macros` embedded migrations; the wiring crate runs them at startup before registering backends.
+
+#### `cluster_cache`
+
+```sql
+CREATE TABLE cluster_cache (
+    key        TEXT        NOT NULL,
```
striped-zebra-dev, line 76: `key` is `PRIMARY KEY (key)` here, so it's the btree that carries whatever ends up in this column — that's a real constraint, but not quite "TEXT is bad in an index." Postgres stores `TEXT` and `VARCHAR(n)` identically on disk; `VARCHAR(n)` is just `TEXT` plus a length check. The actual risk with an unbounded value in a btree key is the index-tuple size limit (roughly 1/3 of a page, ~2712 bytes by default) — insert fails outright once a key exceeds that, TOASTing doesn't save you here the way it would for a non-indexed column.

Given that, I don't think we can move off unbounded `TEXT` (or add a tight `VARCHAR(255)`) without checking what can actually land in this column first — and it's not capped the way it looks:

Scoping (`ClusterCacheV1::scoped(prefix)`, `cluster-sdk/src/cache/facade.rs`) is a pure client-side string transform: it wraps the backend in a `ScopedCacheBackend` (`cluster-sdk/src/cache/scoped.rs`) that does `format!("{prefix}{key}")` on every write, with no re-validation and no length check on the result. Scoping composes with unbounded nesting depth — `cache.scoped("a")?.scoped("b")?.scoped("c")?...` — so by the time a key reaches this plugin, it's `<prefix1>/<prefix2>/.../<key>` with no fixed upper bound.

What the SDK actually validates (`cluster-sdk/src/scope.rs`) is only:
- each individual scope prefix passed to `.scoped(...)` — max 255 bytes, `[a-zA-Z0-9_-]` segments
- the raw leaf `key` argument to `get`/`put`/etc — max 255 bytes, same charset

Neither bounds the composed string this plugin actually stores. So `VARCHAR(255)` would be a correctness bug (rejects legitimate multiply-scoped keys), and plain unbounded `TEXT` leaves this table one deeply-nested-scope away from a hard btree insert failure instead of a clean `ClusterError`.

Given the SDK doesn't cap composed length today, I'd add a `CHECK (octet_length(key) <= N)` here (a few KB, well under the btree limit) so an oversized key fails as a normal backend error rather than a Postgres insert panic — and flag on the SDK side that scope composition has no depth/length cap, which seems like the real gap.

gears/system/cluster/plugins/postgres-cluster-plugin/src/migrations/lock/0002_cluster_lock.sql
```diff
@@ -0,0 +1,25 @@
...
+-- plugin's already-applied `0001_cluster_cache.sql` version recorded there.
+--
+-- `ttl_ms` lets the lock TTL reaper (DESIGN.md §5.2) identify expired locks.
+-- `holder_id` is a random UUID generated at acquire time and used as an
+-- end-to-end ownership fence (PGR-L1): `renew()`, `release()`, and the reaper
+-- all match on it, so a stale guard whose lock lapsed and was re-acquired by a
+-- newer holder cannot renew, release, or reclaim the successor's live lock.
+CREATE TABLE cluster_lock (
+    name        TEXT        NOT NULL,
+    holder_id   TEXT        NOT NULL,
```
striped-zebra-dev: `holder_id` is only ever a `Uuid::new_v4().to_string()` (`lock/mod.rs:408`) — nothing else writes to this column. Since every value is a valid UUID by construction, this should be a native `UUID` column, not `TEXT`: fixed 16-byte storage instead of a ~36-byte string, cheaper equality comparison (used in every `renew`/`release`/reaper query as `WHERE name = $1 AND holder_id = $2`), and the column type itself documents and enforces the contract instead of leaving it as an implicit convention.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/reaper.rs
```diff
@@ -0,0 +1,284 @@
...
+/// of whether this process could act on the advisory lock for each one).
+async fn sweep(
+    pool: &PgPool,
+    table: &str,
+    held: &Arc<DashMap<String, HeldLock>>,
+    metrics: &dyn ClusterMetrics,
+    provider: &'static str,
+) -> Result<usize, ClusterError> {
+    let expired: Vec<(String, String)> = sqlx::query_as(&format!(
+        "DELETE FROM {table} WHERE acquired_at + ttl_ms * interval '1ms' < now() \
```
striped-zebra-dev: Same root issue as the `TEXT` PK comment elsewhere on this table: `acquired_at`/`ttl_ms` isn't just unindexable, it also means the reaper can only find out "is anything expired" by polling on a fixed interval and re-deriving the deadline for every row, every tick.

What if we bucket expiry into a configurable granularity instead of storing raw `acquired_at`+`ttl_ms`? E.g. with a 1h granularity, a lock acquired at 12:33 with any TTL landing before 13:00 gets `expire_slice = 13:00`; one acquired at 12:55 with a TTL past 13:00 gets `expire_slice = 14:00`. Store `expire_slice` directly (computed once at insert/renew), index it like `cluster_cache.expires_at` already is, and the reaper's sweep becomes a plain `WHERE expire_slice <= now()`. Bonus: the reaper can compute `SELECT min(expire_slice) FROM cluster_lock` and sleep until that timestamp instead of polling blindly, so idle periods don't cost anything.

Trade-off is coarser precision — an expired lock can live up to one granularity window past its true TTL — so granularity needs to be chosen relative to how long a stale lock is tolerable, not just sweep cost. Worth discussing vs. just adding an expression index on the current columns.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,695 @@
...
+| Cache-watch LISTEN connection (1 dedicated, combined plugin only) | Receives all `NOTIFY cluster_cache_changes` events; never used for queries | A dedicated `sqlx::PgListener`, outside the pool |
+| Lock release-wake LISTEN connection (1 dedicated) | Receives all `NOTIFY cluster_lock_released` events, feeding the in-process `ReleaseWaiters` registry that wakes blocked `lock()` callers (§5.3) | A dedicated `sqlx::PgListener`, outside the pool |
+| Lock connections (session-affine) | `pg_advisory_lock` is session-scoped; each acquired lock holds its advisory slot for the connection duration | Borrowed from the write pool but pinned to one connection per lock |
+
+Advisory locks are session-scoped in Postgres, which means the connection that acquired a lock must be the same connection that releases it. The plugin maps each `LockGuard` to a specific pool connection that it pins (borrows exclusively) for the duration of the lock. The pool must therefore be sized to accommodate the maximum number of simultaneously held locks; the plugin warns at startup if `pool_max_size < expected_concurrent_locks` (a config-level advisory, not a hard failure).
```
striped-zebra-dev, line 277: Exploring an alternative to "one pinned connection per held lock": `pg_advisory_lock` is session-scoped, but a single session can hold *multiple distinct* advisory locks at once — nothing requires a 1:1 mapping between locks and connections. Given today's default `pool_max_size: 5`, 5 held locks already consumes the entire write pool (see my connection-count question above), and this would fix that directly.

I worked through a concrete sketch (single dedicated connection per instance behind an actor, a generation counter instead of the pinned-connection-per-lock model) plus a separate correctness finding in the TTL reaper's cross-instance sweep race that's worth tracking regardless of which connection model we land on: https://gist.github.com/striped-zebra-dev/54c23819523b575c5edec63f2d7fb34d

Happy to discuss either piece here if useful.

asmith987: Okay, this makes sense and ultimately the better approach.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/reaper.rs
```diff
@@ -0,0 +1,98 @@
...
+use opentelemetry::metrics::Histogram;
+use sqlx::PgPool;
+use tokio_util::sync::CancellationToken;
+
+use crate::cache::watch::{self, NotifyEvent};
+use crate::pg_error::map_sqlx_error;
+
+/// Runs one sweep: deletes all rows past their `expires_at` and issues one
+/// `NOTIFY` per deleted key in the same transaction (DESIGN.md §4.2).
+async fn sweep(pool: &PgPool, table: &str) -> Result<usize, ClusterError> {
```
striped-zebra-dev: Two structural issues with bundling the delete and every `NOTIFY` into one transaction here, on top of the missing `LIMIT`:

**Lock-hold time scales with backlog size.** The module doc says the reaper "uses one connection from the write pool per sweep, releasing it immediately after" — but with this loop, "immediately after" now means after every expired key's `NOTIFY` round-trip has run sequentially on that same connection. With a large backlog that's one of the write pool's connections tied up for however long N sequential round-trips take. Worse, `DELETE ... RETURNING` leaves those rows' locks held until commit — standard Postgres MVCC means a concurrent `INSERT ... ON CONFLICT (key) DO UPDATE` targeting the *same key* blocks until this transaction ends. So a `put`/`put_if_absent` racing a sweep on a key that's mid-batch doesn't wait for a quick row lock, it waits for the rest of the `NOTIFY` loop to finish.

**All-or-nothing means a backlog can never make partial progress.** Any single failure anywhere in the loop rolls back the *entire* batch, not just the failing row — e.g. 99,999 legitimately-expired keys get rolled back because key 100,000's `NOTIFY` hit a transient error. Nothing is lost (they're all still expired, retried next tick), but the sweep makes zero forward progress on that tick. If expiry rate ever outpaces one sweep's completion time, every subsequent tick re-attempts the same, now-larger, all-or-nothing batch — there's no way for this reaper to eat into a backlog incrementally.

(I also went looking for a related failure mode — an oversized key blowing the 7999-byte `NOTIFY` payload limit mid-transaction — and that one's already closed: `validate_key_len` (`cache/watch.rs`) caps every key at 7997 bytes and runs on all three insert-capable paths, so nothing that could ever land in this table can hit that limit. Good defensive work there.)

Suggest chunking the sweep — `DELETE ... LIMIT n` (or a `SELECT ... LIMIT n FOR UPDATE SKIP LOCKED` + delete) in a loop, each chunk its own short transaction. Bounds lock-hold time and `NOTIFY`-burst size per chunk regardless of total backlog, and a stuck/failing row only blocks its own chunk instead of the whole sweep.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/reaper.rs
striped-zebra-dev: Same missing `LIMIT` as the cache reaper's sweep — this `DELETE ... RETURNING` has no bound either, so a large backlog means one statement holding locks on however many rows match, for as long as that takes. It doesn't have the cache reaper's *compounding* problem though (no per-row work bundled into the same transaction — `reclaim()` runs after, independently, per name, off its own pinned connection), so it's a narrower version of the same fix: worth a `LIMIT` + loop here too, just without the all-or-nothing amplification.

striped-zebra-dev (approved):

gears/system/cluster/cluster-conformance/src/cache.rs
asmith987: I went ahead and implemented a single strategy that is independent of backend. There are two polls: a 500ms wait for the `Changed` event,then a second wait for 250ms to verify that backend delivers at most once. The second one has a wait rather than an immediate drain of what's already queued since a duplicate arrives asynchronously, so an immediate drain would cause the test to pass regardless.

gears/system/cluster/cluster-conformance/src/cache.rs
asmith987: I adopted the approach, but created a more generic builder so that cache, discovery, and leader suites shared a single strategy. This can be seen in `cluster-conformance/src/watch_poll`

gears/system/cluster/plugins/postgres-cluster-plugin/src/migrations/lock/0002_cluster_lock.sql
asmith987: Made `holder_id` a UUID.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
asmith987: Fixed. Added a check to the cluster_cache table and tracking the gap as part of the SDK. The SDK update will be done as a follow up PR

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/reaper.rs
asmith987: I agree with the issue. This was fixed by storing the exact deadline to mirror how `cluster_cache` table is handling this. Additionally, I added the sleep-until-due that you mentioned.

gears/system/cluster/plugins/postgres-cluster-plugin/docs/DESIGN.md
asmith987: I will refactor the lock implementation for the postgres plugin to make use of a single connection as a follow up PR. It needs a deeper review and it would be cleaner to do so as a follow up

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/reaper.rs
asmith987: Fixed. Added the limit and now the sweep deletes in bounded batches + loops until a batch comes back short.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/reaper.rs
asmith987: Done. Chunked in 512-row transactions, so lock-hold time and NOTIFY bursts are bounded per chunk. Additionally a failing chunk no longer rolls back the whole backlog.

gears/system/cluster/plugins/postgres-cluster-plugin/src/cache/watch.rs
```diff
@@ -0,0 +1,555 @@
...
+use dashmap::DashMap;
+use rand::RngExt as _;
+use sqlx::postgres::PgListener;
+use tokio_util::sync::CancellationToken;
+
+use crate::limits::MAX_INDEXED_KEY_BYTES;
+use crate::pg_error::map_sqlx_error;
+
+/// The Postgres NOTIFY channel this plugin's cache uses (DESIGN.md §4.3).
+pub const CHANNEL: &str = "cluster_cache_changes";
```
striped-zebra-dev, line 40: If we to choose a single DB for all gears (one of the options that we discussed) then a noisy neighbour will overload all the other gears.

PR #4224, opened 2026-07-17
Support authz code flow for oauth2 in OAGW.

genericaccount-de:
Support authz code flow for oauth2 in OAGW.

## Summary by CodeRabbit

* **New Features**
  * Added interactive OAuth 2.0 authorization-code enrollment endpoints (begin, callback, connection status, revocation) with per-user token handling.
  * Introduced an OAuth 2.0 auth-code built-in authentication option for protected upstream services.
  * Extended credstore capabilities with secret write/delete (private, tenant, and shared scoping).
* **Bug Fixes**
  * Preserved meaningful trailing slashes in normalized proxy paths.
  * Improved secret deletion scoping to respect tenant/global fallback and private isolation.
* **Documentation**
  * Updated OpenAPI specs and OAuth-related security policy details.
* **Chores**
  * Added support for HTTP 501 “Unimplemented” canonical errors.


gears/credstore/credstore-sdk/src/api.rs
```diff
@@ -25,4 +25,29 @@ pub trait CredStoreClientV1: Send + Sync {
...
+    /// Stores a secret, creating it or overwriting an existing value.
+    ///
+    /// The owning tenant and (for [`SharingMode::Private`]) the owner are
+    /// derived from `ctx`. Backends that only support a static, read-only
+    /// catalog return [`CredStoreError::Internal`].
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` for infrastructure failures or unsupported backends.
+    async fn put(
```
MikeFalcon77, line 41: **HIGH**

New put/delete added as required (no default) methods to the published consumer SDK trait CredStoreClientV1.

Any out-of-repo implementor of this V1 SDK trait fails to compile after upgrade — silent contract breakage.

Give put/delete default impls returning CredStoreError::Internal (as the docs already describe for read-only backends), or bump to a V2 trait.

gears/credstore/credstore-sdk/src/plugin_api.rs
```diff
@@ -16,4 +16,28 @@ pub trait CredStorePluginClientV1: Send + Sync {
...
+    /// Stores a secret in the backend, creating or overwriting it.
+    ///
+    /// The owning tenant and (for [`SharingMode::Private`]) the owner are
+    /// derived from `ctx`. Read-only backends return
+    /// [`CredStoreError::Internal`].
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` for backend failures or unsupported (read-only) backends.
+    async fn put(
```
MikeFalcon77, line 32: **HIGH**

New required put/delete added to the published CredStorePluginClientV1 plugin trait with no default impl.

Every external credstore plugin backend breaks on upgrade — accidental breaking API expansion on a V1 boundary.

Add default method bodies returning an unsupported/Internal error so read-only plugins stay source-compatible, or introduce a versioned trait.

libs/toolkit-auth/src/oauth2/authcode.rs
```diff
@@ -0,0 +1,530 @@
...
+        .next()
+        .ok_or_else(|| {
+            TokenError::InvalidResponse(
+                "protected-resource metadata advertised no authorization_servers".to_owned(),
+            )
+        })?;
+    let issuer = Url::parse(&issuer_str).map_err(|e| {
+        TokenError::InvalidResponse(format!("invalid authorization server URL: {e}"))
+    })?;
+    discover_authorization_server(client, &issuer).await
```
MikeFalcon77, line 239: **MEDIUM**

discover_from_resource follows the remote-supplied authorization_servers[0] and POSTs the auth code, PKCE verifier and refresh token to the discovered endpoints without validating the returned issuer matches the requested one (RFC 8414 §3.3) or enforcing https.

Malicious protected-resource metadata can redirect the token exchange to attacker-controlled or internal endpoints (OAuth mix-up / SSRF), exfiltrating the code and tokens.

Validate AuthorizationServerMetadata.issuer equals the requested issuer and that all discovered endpoints share that origin and use https before use.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+        let registration_endpoint = meta.registration_endpoint.as_ref().ok_or_else(|| {
+            DomainError::validation(
+                "authorization server does not support dynamic client registration",
+            )
+        })?;
+        let registered = authcode::register_client(
+            &http,
+            registration_endpoint,
+            &client_name,
+            std::slice::from_ref(&redirect_uri),
```
MikeFalcon77, line 229: **MEDIUM**

The caller-supplied redirect_uri from BeginOAuthRequest is passed straight into dynamic client registration and the authorize URL with no validation that it is an absolute https URL or an allowlisted callback.

Unvalidated redirect_uri is the canonical authorization-code interception vector; the authorization server delivers the code to whatever destination the request names.

Validate redirect_uri as an absolute https URL against an explicit allowlist (or OAGW's own callback) before registration and authorize-URL construction.

gears/system/oagw/oagw/src/infra/plugin/oauth2_auth_code_auth.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Default safety margin: refresh once the token is within this window of
+/// expiry so it does not lapse mid-request.
+const DEFAULT_REFRESH_MARGIN: Duration = Duration::from_secs(60);
+
+/// Per-user token record persisted in credstore (JSON) by the OAGW management
+/// API and consumed by this plugin.
+///
+/// Self-contained so the plugin needs only the credstore `token_ref` in its
+/// binding config; the authorization-server coordinates travel with the record.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
MikeFalcon77: **MEDIUM**

OAuthTokenRecord derives Debug while holding plaintext access_token, refresh_token and client_secret as String.

A single debug log of this record leaks live bearer and refresh tokens, contradicting the module's own claim that secret material is never logged.

Wrap the token fields in SecretString or implement a redacting Debug so tokens and client secrets cannot appear in formatted output.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+/// used to discover the authorization server.
+const CONFIG_RESOURCE: &str = "resource";
+/// Config key naming the credstore `SecretRef` under which the per-user token
+/// record is stored (the `oauth2_auth_code` plugin reads the same key).
+const CONFIG_TOKEN_REF: &str = "token_ref";
+
+/// Short-lived state persisted between `begin` and `complete`, keyed by the
+/// CSRF `state`. Holds everything needed to complete the code exchange without
+/// re-discovering or re-registering.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
MikeFalcon77: **MEDIUM**

PendingAuthorization derives Debug while holding plaintext client_secret and code_verifier as String.

Any accidental debug-formatting leaks OAuth secret material, defeating the SecretString redaction discipline used elsewhere.

Store client_secret and code_verifier as SecretString, or drop the Debug derive / implement a redacting Debug.

gears/system/oagw/oagw/src/infra/plugin/oauth2_auth_code_auth.rs
```diff
@@ -0,0 +1,242 @@
...
+pub struct OAuthTokenRecord {
```
MikeFalcon77: **MEDIUM**

OAuthTokenRecord is a durably-persisted (credstore JSON) contract written by the enrollment service and read independently by the plugin, but carries no schema/version field.

Any rename or type change to a required field silently fails to deserialize previously stored records at request time (surfaced only as 'corrupt token record'), and the two readers/writers can diverge.

Add an explicit version discriminator to the serialized record and handle unknown versions explicitly on read.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+use toolkit_auth::oauth2::types::SecretString;
+use toolkit_security::SecurityContext;
+use url::Url;
+use uuid::Uuid;
+
+use crate::domain::error::DomainError;
+use crate::domain::services::{
+    ControlPlaneService, OAuthBeginOutcome, OAuthConnectionStatus, OAuthEnrollmentService,
+};
+use crate::infra::plugin::oauth2_auth_code_auth::OAuthTokenRecord;
```
MikeFalcon77: **MEDIUM**

infra/oauth/enrollment.rs imports the shared serialization type OAuthTokenRecord from the sibling plugin module infra/plugin/oauth2_auth_code_auth, so the persistence contract written here is owned by an unrelated plugin module.

The token record is the shared contract between the enrollment writer and the plugin reader; burying it in one plugin couples the two infra submodules and obscures ownership, inviting drift.

Move OAuthTokenRecord (and now_unix) into a shared module (e.g. infra/oauth) that both the plugin and the enrollment service depend on.

gears/system/oagw/oagw-sdk/src/api.rs
```diff
@@ -154,4 +166,50 @@ pub trait ServiceGatewayClientV1: Send + Sync {
...
+    // -- Interactive OAuth authorization management --
+    //
+    // Default implementations return an `unimplemented` canonical error so
+    // that transport shims and test doubles need not implement them; the
+    // in-process gateway facade overrides all four.
+
+    /// Begin an interactive OAuth authorization-code flow for an upstream on
+    /// behalf of the calling user (discovery + dynamic client registration +
+    /// PKCE), returning the browser authorization URL and CSRF state.
+    async fn begin_oauth_authorization(
```
MikeFalcon77: **MEDIUM**

The four new OAuth methods on ServiceGatewayClientV1 default-implement to CanonicalError::internal(...), masking a missing implementation as a runtime 500 and misusing the 'internal' category for an unimplemented feature.

An implementor that forgets to override gets a silent runtime 500 (misclassified as internal, not unimplemented/501) instead of a build failure; the type system no longer enforces trait completeness.

Make the methods required (no default), or if defaults are kept return an unimplemented/not-implemented canonical category rather than internal.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+                &token_ref,
+                SecretValue::new(json),
+                SharingMode::Private,
+            )
+            .await
+            .map_err(|e| DomainError::internal(format!("credstore put error: {e}")))?;
+
+        // Best-effort cleanup of the pending record; failure here does not
+        // invalidate the successful enrollment.
+        let _ = self.credstore.delete(ctx, &state_ref).await;
```
MikeFalcon77: **MEDIUM**

Pending-record cleanup `let _ = self.credstore.delete(ctx, &state_ref).await;` swallows the delete error with no log.

A failed cleanup silently leaves a pending record containing secret material (code_verifier, client_secret) in credstore with zero operational visibility, so leaks accumulate invisibly.

Log the delete error at warn (with state_ref context) before continuing instead of discarding the Result.

gears/system/oagw/oagw/src/infra/plugin/oauth2_auth_code_auth.rs
```diff
@@ -0,0 +1,242 @@
...
+                    scope: tokens.scope.clone().or_else(|| record.scope.clone()),
+                };
+                self.persist(ctx, &secret_ref, &refreshed).await?;
+                inject_bearer(ctx, &refreshed.access_token);
+                Ok(())
+            }
+            Err(PluginError::AuthorizationRequired(_)) => {
+                // Refresh rejected or unavailable: drop the stale record and
+                // require the user to re-authorize.
+                let _ = self
```
MikeFalcon77: **MEDIUM**

The stale-record eviction after a rejected refresh uses `let _ = self.credstore.delete(...).await;`, discarding the error silently.

A failed delete leaves an invalid token record that re-fails on every subsequent request with no signal, hiding a persistent credstore problem.

Capture and log the delete failure at warn before returning AuthorizationRequired.

gears/system/oagw/oagw/src/infra/plugin/oauth2_auth_code_auth.rs
```diff
@@ -0,0 +1,242 @@
...
+            .as_ref()
+            .map(|s| SecretString::new(s.clone()));
+        let refresh_ss = SecretString::new(refresh.clone());
+        let scopes: Vec<String> = record
+            .scope
+            .as_deref()
+            .map(|s| s.split_whitespace().map(String::from).collect())
+            .unwrap_or_default();
+
+        let http = toolkit_http::HttpClientBuilder::with_config(self.http_config.clone())
```
MikeFalcon77: **MEDIUM**

A fresh HttpClient is built on every token refresh via HttpClientBuilder inside refresh(), discarding the connection pool and re-running TLS/root-cert setup each call.

authenticate() runs on the proxy request path; rebuilding the client per refresh loses keep-alive reuse to the token endpoint and repeats expensive TLS/cert init, adding request latency.

Build the HttpClient once (OnceCell/lazily-initialized field or at wiring time) and reuse the pooled client across refresh calls.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+    ) -> Self {
+        Self {
+            cp,
+            credstore,
+            http_config,
+        }
+    }
+
+    fn http(&self) -> Result<toolkit_http::HttpClient, DomainError> {
+        toolkit_http::HttpClientBuilder::with_config(self.http_config.clone())
```
MikeFalcon77: **MEDIUM**

OAuthEnrollmentServiceImpl::http() constructs a new HttpClient per operation, so each begin/complete flow rebuilds the client and pool for its multiple sequential requests.

Rebuilding reqwest-backed clients repeats TLS/root-cert setup and prevents connection reuse across the discovery/registration/exchange round-trips, adding avoidable latency.

Construct the HttpClient once (OnceCell or at wiring time) and reuse it across operations.

libs/toolkit-auth/src/oauth2/authcode.rs
```diff
@@ -0,0 +1,530 @@
...
+            .field("verifier", &"[REDACTED]")
+            .field("challenge", &self.challenge)
+            .finish()
+    }
+}
+
+impl Pkce {
+    /// Generate a fresh PKCE pair from 32 bytes of randomness.
+    #[must_use]
+    pub fn generate() -> Self {
```
MikeFalcon77, line 376: **MEDIUM**

Pkce::generate has no unit test verifying the S256 contract (challenge == base64url(SHA256(verifier))) or that the verifier is URL-safe/high-entropy.

PKCE is the core defense of the authorization-code flow; a broken challenge derivation would ship silently since nothing asserts the verifier/challenge relationship.

Add a test that recomputes URL_SAFE_NO_PAD(SHA256(verifier)) and asserts it equals challenge, and asserts method()=="S256".

libs/toolkit-auth/src/oauth2/authcode.rs
```diff
@@ -0,0 +1,530 @@
...
+
+/// Build the browser authorization-request URL (RFC 6749 §4.1.1 + PKCE).
+///
+/// Only scopes advertised by the server should be passed in `scopes`.
+///
+/// # Errors
+///
+/// Returns [`TokenError::ConfigError`] if `authorization_endpoint` is not a
+/// valid URL.
+pub fn build_authorize_url(
```
MikeFalcon77, line 513: **MEDIUM**

build_authorize_url has no unit test asserting the emitted query params (response_type=code, state, code_challenge, code_challenge_method=S256, scope join) or its ConfigError path.

This function builds the security-critical browser authorization request; wrong/missing state or code_challenge params would not be caught since no test inspects the produced URL.

Add tests parsing the returned Url and asserting each expected query pair (including code_challenge_method=S256 and space-joined scopes), plus an Err case for a malformed endpoint.

gears/system/oagw/oagw/src/infra/oauth/enrollment_tests.rs
```diff
@@ -0,0 +1,426 @@
...
+    let svc = service_with(
+        Some(upstream_with_auth(&resource, TOKEN_REF)),
+        credstore.clone(),
+    );
+
+    let outcome = svc
+        .begin(
+            &test_ctx(),
+            Uuid::nil(),
+            vec!["read".to_owned()],
```
MikeFalcon77: **MEDIUM**

The begin tests only pass an already-supported scope or an empty list, so the scope-intersection filter in begin (drops scopes not in scopes_supported) is never exercised and the authorize-URL scope param is never asserted.

Scope intersection is a security control (prevents requesting unadvertised/over-broad scopes); with only supported/empty inputs the filter could be removed and every test would still pass.

Add a begin test requesting a mix of supported and unsupported scopes (e.g. ["read","admin"]) and assert authorization_url contains scope=read and not admin.

gears/system/oagw/oagw/src/infra/oauth/enrollment_tests.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+#[tokio::test]
+async fn begin_then_complete_persists_token_record() {
+    let server = MockServer::start();
+    mount_discovery(&server);
+    server.mock(|when, then| {
+        when.method(POST)
+            .path("/token")
+            .body_includes("grant_type=authorization_code")
```
MikeFalcon77, line 398: **MEDIUM**

The complete-flow token mock matches only grant_type and code, so the PKCE code_verifier that begin generated is never asserted to be sent on the exchange.

PKCE binding is only effective if the verifier is actually transmitted on code exchange; the test would still pass if code_verifier were dropped from the request.

Add .body_includes("code_verifier=") (and ideally redirect_uri) to the /token mock matcher so the exchange is bound to the PKCE verifier.

gears/system/oagw/oagw/src/infra/plugin/oauth2_auth_code_auth_tests.rs
```diff
@@ -0,0 +1,140 @@
...
+            .header("content-type", "application/json")
+            .body(r#"{"access_token":"at-new","refresh_token":"rt-new","expires_in":3600,"token_type":"Bearer"}"#);
+    });
+
+    let record = record_json(now_unix() - 10, Some("rt-old"), &token_ep);
+    let plugin = plugin_with(vec![(TOKEN_REF.to_owned(), record)]);
+    let mut ctx = ctx_with_token_ref();
+
+    plugin.authenticate(&mut ctx).await.unwrap();
+    assert_eq!(ctx.headers.get("authorization").unwrap(), "Bearer at-new");
```
MikeFalcon77, line 239: **MEDIUM**

expired_with_refresh_injects_new_bearer uses MockCredStoreClient whose put/delete are no-ops, so it only asserts the injected header and never verifies the rotated record (refresh_token=rt-new) is persisted back.

Refresh-token rotation persistence is security-critical (the old refresh token must be replaced); with a no-op store the persist() call could be deleted and this test would still pass.

Use a stateful credstore mock and assert the stored record contains access_token=at-new and refresh_token=rt-new after authenticate.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,361 @@
...
+    client_id: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    client_secret: Option<String>,
+    code_verifier: String,
+    redirect_uri: String,
+    #[serde(default)]
+    scopes: Vec<String>,
+}
+
+fn now_unix() -> i64 {
```
MikeFalcon77: **LOW**

now_unix() is duplicated verbatim in both infra/oauth/enrollment.rs and infra/plugin/oauth2_auth_code_auth.rs.

The same time-conversion logic (including i64::try_from saturation) is implemented twice, so a change must be made in two places.

Extract now_unix() into a single shared helper and import it from both call sites.

gears/system/oagw/oagw-sdk/src/api.rs
striped-zebra-dev: That 100% shall not be a part of the api.

MikeFalcon77: doc reference: https://github.com/constructorfabric/gears-rust/pull/4225

MikeFalcon77:
Don't add these to the public SDK, and rework the token/state storage.

1. The 4 new methods on ServiceGatewayClientV1 are not needed.

The REST handlers already call the internal OAuthEnrollmentService directly (state.oauth.begin(...)), not the SDK trait. So begin_oauth_authorization and the other three methods on ServiceGatewayClientV1 are a second, duplicate entry point into the same internal service — they serve no real caller.

They also can't work cross-gear: the default impl returns unimplemented, so any HTTP transport shim returns an error. They only work for an in-process caller — which is the same process as OAGW and would just use the internal service or REST anyway.

And the flow itself is a browser redirect + user consent. That is an HTTP/REST interaction (frontend/BFF), not an in-process Rust call between gears. An in-process caller has nothing to do with authorization_url.

=> Remove the 4 trait methods and oagw-sdk/src/oauth.rs. Keep enrollment on REST + the internal OAuthEnrollmentService. The cross-gear contract stays unchanged. Re-auth is signaled through the existing CanonicalError / FailureReason::AuthorizationRequired.

2. POST /oauth/complete sends the raw authorization code through the consumer.

In a standard OAuth client, the callback (GET /oauth/callback?code&state) is the completion. There is no separate complete step, and the code never leaves OAGW. The current design routes the code back through the caller, which is weaker and non-standard.

=> Replace POST /oauth/complete with a standard callback endpoint.

3. Storage layering is wrong.

OAuthTokenRecord (the shared storage contract) is defined inside the data-plane plugin, and the enrollment service imports it from there. This is an inverted dependency. The record type doesn't belong to the plugin.
The storage location is a token_ref config string, parsed independently in two places (plugin parse_config and enrollment resolve_binding, both doing strip_prefix("cred://")). If they drift, the plugin reads a different ref than enrollment wrote. This is misconfigurable by design.
Transient PKCE/CSRF PendingAuthorization is stored in CredStore. CredStore is durable secret storage with no TTL. Putting single-use, 5-minute state there is a misuse: if complete never happens, the record is orphaned (the manual expires_at_unix + best-effort delete does not fix this).
=> Introduce one UserTokenStore seam that owns the record type and derives the key from (subject, upstream_id) — used by both the plugin and enrollment, so there is no string convention to keep in sync. Keep the pending PKCE/CSRF state in a separate in-memory TTL store, not in CredStore.

The plugin and enrollment should stay decoupled (they communicate only through the token store — that part is fine), but through one shared seam, not a record type owned by the plugin plus a duplicated config key.

The draft in #4225 already describes exactly this: no SDK growth, standard callback, dedicated UserTokenStore, ephemeral pending store. I'd align this PR with that.

gears/system/oagw/oagw/src/api/rest/routes/oauth.rs
```diff
@@ -0,0 +1,91 @@
...
+        .handler(handlers::oauth::connection_status)
+        .json_response_with_schema::<dto::OAuthConnectionStatusResponse>(
+            openapi,
+            http::StatusCode::OK,
+            "OAuth connection status",
+        )
+        .standard_errors(openapi)
+        .register(router, openapi);
+
+    if writable {
```
mattgarmon: We can ignore `writable` for these endpoints and include them anyway.
This flag is used to manage CRUD for upstreams/routes when using GTS provisioned entities.

mattgarmon:
Hi @genericaccount-de,

As you know, @striped-zebra-dev has a draft design for this plugin here: #4225 
Ideally, we would resolve the open design questions and merge it first, although @striped-zebra-dev is away on PTO this week.

My take:
- We are mostly aligned now, aside from a few divergences that affect scenarios
- There are some unresolved questions which may change how this implementation is done - some of these can come as follow-up PRs

Critical divergences (should be fixed in this PR):
- Re-auth contract. Map PluginError::AuthorizationRequired to a new AUTHORIZATION_REQUIRED reason (distinct from AUTH_PLUGIN_FAILED/AUTH_PLUGIN_INTERNAL), project it to FailureReason::AuthorizationRequired { authorization_url }, and emit a WWW-Authenticate challenge at the proxy boundary. It currently collapses into PLUGIN_FAILED (proxy/service.rs), which fails scenarios 9.13 and 9.15.
- Callback completes the browser flow. Make code/error optional in OAuthCallbackQuery, handle ?error=... redirects, and return a browser-completing response (HTML or 302) instead of 204. Fixes scenario 9.11.

Other divergences (should be aligned and/or fixed in this PR):
- `return_to` - currently missing entirely; consumer-supplied + allowlisted, callback redirects to it.
- `redirect_uri` - currently caller-supplied with no allowlist; must be deployment config/allowlist, never from the caller. (Security-sensitive)
- `resource` - currently mandatory; must be optional with discovery fallback from the upstream's own 401 WWW-Authenticate.
- `scope` source - currently taken from the request body; should come from the upstream auth.config.

Other points (should also be fixed in this PR):
- FIPS/crypto. The FIPS verification failure is the newly-direct sha2 dependency in toolkit-auth - the cargo-deny deny-fips.toml gate flags it (the DE0708 Dylint allow-list is a separate gate, which is why Dylint Tests is also still red and needs its fixture updated). Compute the PKCE SHA-256 via the FIPS-validated provider already in the graph rather than pulling RustCrypto sha2. That removes the direct dep and clears both FIPS verification and Dylint Tests.
- Single-instance constraint. I am okay keeping the in-memory pending store for the moment (see follow-ups), please document the single-instance constraint.

Follow-ups:
- Stateless-state variant (#4225 open question #3). I believe we should move toward this. It bundles a pre-registered client (config) + a self-contained encrypted state, which lets us drop the in-memory pending store and removes the single-instance constraint. This is also where the client-model work belongs: the current DCR-only implementation is fine for now (all near-term upstreams support DCR), and pre-registered is only needed once we go stateless or need to enroll a non-DCR authorization server. We have no multi-instance deployments today, so the in-memory store is fine for single-instance/dev, and this can be a follow-up PR once the design is ratified in #4225.
- Durable token persistence. The in-memory/static CredStore is fine for dev, and this works in Acronis Cloud by pointing the credstore plugin at the actual Cred Store service. The adapter (acronis-credstore-plugin in oagw-ng) currently implements only the read path, so durable put/delete needs to be added there for tokens to persist in production - can be done as a separate PR in that repo.
- (Forward-looking) Client ID Metadata Documents. Not in #4225, but the latest MCP spec (2025-11-25) deprecates DCR in favor of CIMD (an HTTPS URL as client_id). Worth tracking as a later enhancement, not required now.
- (Design, pending #4225) Callback topology. Decide whether OAGW's callback is ever publicly exposed vs relayed via a BFF. If relayed, the BFF should bind the flow to the initiating user's session and verify it matches the subject_id - this mitigates cross-user OAuth session fixation, which PKCE alone does not prevent.

We will also want to create missing ADRs in #4225, so the follow-ups have a ratified design to build against.

gears/system/oagw/oagw/src/infra/proxy/service.rs
```diff
@@ -1255,6 +1263,19 @@ async fn execute_auth_plugin(
                     instance: instance_uri.to_string(),
                 }
             }
+            crate::domain::plugin::PluginError::AuthorizationRequired(_) => {
```
mattgarmon: (NON-BLOCKING)

The `AUTHORIZATION_REQUIRED` typed reason is emitted, but the resource hint is dropped here and the 401 carries no `WWW-Authenticate: Bearer` challenge, so the re-auth signal #4225 describes isn't fully here. 

This is open question 6 (#4225). Likely fine to defer if the current typed reason is enough for the consumer - flagging so it's a conscious call.

The correct fix, for when the challenge shape is pinned: 
- Add a `resource: Option<String>` field to `AuthenticationFailed` and stop discarding it here (bind the value instead of _, threading through the resource from `PluginError::AuthorizationRequired`). 
- In the canonical mapping attach it via `.with_resource(...)` so it lands as `resource_name`. Then emit the `WWW-Authenticate` header naming the resource plus the OAGW authorize endpoint (POST /oagw/v1/upstreams/{id}/oauth/authorize) - on both the axum error_response path and the Pingora error path, since they build responses separately. 
- Note the data plane only knows the resource, not a fully-formed `authorization_url` (that's minted by begin), so the challenge should reference the authorize endpoint, not a ready-to-follow URL. Optionally project it into the SDK as `FailureReason::AuthorizationRequired { resource_name }`, keeping the flat `as_wire`/`from_wire` round-trip intact (return the bare wire constant, reconstruct with an empty payload) so the round-trip tests still pass.

libs/toolkit-auth/Cargo.toml
```diff
@@ -41,6 +41,9 @@ url = { workspace = true }
 # OAuth2 token lifecycle
 rand = { workspace = true }
 base64 = { workspace = true }
+# FIPS-validated SHA-256 for the PKCE S256 code-challenge (same crypto provider
+# the platform installs at bootstrap); avoids a direct RustCrypto `sha2` dep.
+aws-lc-rs = { workspace = true }
```
mattgarmon: (BLOCKING)

This adds a direct `aws-lc-rs` dependency and computes the S256 challenge via `aws_lc_rs::digest::SHA256`, with a comment claiming it's "same crypto backend the platform installs at bootstrap." 

That's only true in non-FIPS builds, see https://github.com/constructorfabric/gears-rust/blob/main/libs/toolkit/src/bootstrap/crypto.rs#L113

Please route the PKCE hash through the already-installed `rustls::crypto::CryptoProvider`, which exposes SHA-256 via the `crypto::hash::Hash` trait (every backend implements it). 

Suggested shared helper in `toolkit::bootstrap::crypto`:
```
use rustls::SupportedCipherSuite;
use rustls::crypto::CryptoProvider;
use rustls::crypto::hash::HashAlgorithm;
 
/// Compute a SHA-256 digest through the process-wide rustls [`CryptoProvider`]
/// installed by [`init_crypto_provider`].
///
/// FIPS-safe way for gears to hash arbitrary bytes: the digest is produced by
/// the same validated module backing TLS (aws-lc-fips on Linux, Apple
/// corecrypto on macOS, Windows CNG, or aws-lc-rs in non-FIPS builds) rather
/// than a directly-linked hash crate. Do **not** depend on
/// `sha2`/`aws-lc-rs`/`ring` directly for hashing.
///
/// # Panics
///
/// Panics if no crypto provider is installed (call [`init_crypto_provider`]
/// first) or if it exposes no SHA-256 cipher suite — both are startup
/// invariants; failing closed beats hashing outside the validated module.
#[must_use]
pub fn sha256(data: &[u8]) -> [u8; 32] {
    let provider = CryptoProvider::get_default()
        .expect("no CryptoProvider installed — call toolkit::bootstrap::init_crypto_provider() first");
 
    let hash = provider
        .cipher_suites
        .iter()
        .find_map(|cs| {
            let common = match cs {
                SupportedCipherSuite::Tls13(cs) => &cs.common,
                SupportedCipherSuite::Tls12(cs) => &cs.common,
            };
            (common.hash_provider.algorithm() == HashAlgorithm::SHA256)
                .then_some(common.hash_provider)
        })
        .expect("installed CryptoProvider has no SHA-256 cipher suite");
 
    let output = hash.hash(data);
    let mut buf = [0u8; 32];
    buf.copy_from_slice(output.as_ref());
    buf
}
```

gears/system/oagw/oagw/src/infra/oauth/pending_store.rs
```diff
@@ -0,0 +1,114 @@
+//! In-memory, TTL-bounded store for transient OAuth authorization state.
```
mattgarmon, line 1: (NON-BLOCKING)

`PendingAuthorizationStore` is in-process only, but `begin` and `callback` are separate requests that can land on different replicas (especially if there is a BFF relaying the callback). 

This is open question 1 (#4225) and a real runtime constraint - multi-instance without a shared store breaks enrollment (callback can't find the pending state). 

Please add a comment stating this is single-instance-only (or requires sticky/state-routed callbacks), referencing the "shared/distributed pending store" (or stateless-state variant (open question 3)) future consideration.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,350 @@
...
+/// Config key on the upstream auth binding naming the protected-resource URL
+/// used to discover the authorization server. Optional — when absent (or
+/// blank), discovery falls back to probing the upstream's own endpoint for an
+/// RFC 9728 `WWW-Authenticate` challenge.
+const CONFIG_RESOURCE: &str = "resource";
+
+/// Config key on the upstream auth binding carrying the space/comma-separated
+/// OAuth scopes to request. Scopes are sourced here (never from the caller) so
+/// grants are governed by the upstream's stored configuration.
+const CONFIG_SCOPES: &str = "scopes";
```
mattgarmon: (NON-BLOCKING)

Design in #4225 uses `scope` here. It's fine either way, but we should be consistent.

In general, we will need merge the design with additional ADRs for follow-up work.

gears/system/oagw/oagw/src/infra/oauth/enrollment.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        // `return_to` is consumer-supplied but must be an exact member of the
+        // deployment allowlist before the browser is ever redirected to it.
+        if !self.return_to_allowlist.iter().any(|a| a == &return_to) {
+            return Err(DomainError::validation(
+                "return_to is not in the configured allowlist",
+            ));
+        }
+
+        let upstream = self.cp.get_upstream(ctx, upstream_id).await?;
```
mattgarmon, line 156: (NON-BLOCKING)

Let's add a validation error if the upstream's auth plugin is is not `oauth2_auth_code`

diffora:
Looks this PR will be conflicting with this one https://github.com/constructorfabric/gears-rust/pull/4204/

mattgarmon:
Let's add corresponding design documents based on #4225

Artifizer:
agreed with @striped-zebra-dev to close this PR

PR #4226, opened 2026-07-18
Bug-fix batch: api-gateway CORS expose-headers, AM listing / IdP errors / conversions / child counts, resource-group PEP type

diffora:
A batch of independent bug fixes, one commit per fix, spanning api-gateway, account-management (+SDK), resource-group, and toolkit-db.

## CORS-expose the ETag response header (api-gateway)
Browsers may only read non-safelisted response headers when the server sends `Access-Control-Expose-Headers`; the gateway sent none, so every ETag-guarded (`If-Match`) write was impossible for a cross-origin browser client while curl/Postman hid the problem. Adds `cors.exposed_headers` (default `["ETag"]`, so deployments need no config change), config-load validation for wildcard+credentials combinations (previously a tower-http assert panic at startup — a crash-loop with no pointer at the offending config), and warnings for unparseable CORS list entries instead of a silent `filter_map` drop.

## `$orderby=status` on the tenant children listing (account-management + toolkit-db + resource-group)
Sorting the tenant children listing by status returned 400 `unsupported $orderby field`. Root cause: the cursor codec was bound to the wire `String` kind while the cursor value carries the storage `SMALLINT` ordinal, so the token could not round-trip. Adds `ODataFieldMapping::cursor_kind` (default = the wire kind, so no existing mapper changes behavior) and uses it on both sides of the cursor codec; `encode_cursor_value` also accepts `SmallInt`/`Int` under `FieldKind::I64` so mappers need not hand-widen narrow columns. `status` becomes orderable by its lifecycle ordinal (`active` < `suspended` < `deleted`). resource-group's three type fields carried the same latent wire-String/storage-SmallInt mismatch (orderable, un-overridden — `$orderby=type` broke with `InvalidCursor` on any page overflow) and now declare `cursor_kind = I64`. OpenAPI updated accordingly.

## Typed IdP uniqueness / password-policy rejections (account-management SDK + gear)
`POST /tenants/{id}/users` with a duplicate username surfaced as a redacted generic 400 (`request`/`VALIDATION`), so clients could not attribute the failure to a field; duplicate email and password-policy rejections collapsed into the same catch-all. The `#[non_exhaustive]` `IdpUserOperationFailure` gains `DuplicateUser { field: Username | Email | UsernameOrEmail }` and `PasswordPolicy`; `Rejected` stays as the fallback, so providers that never emit the new variants keep today's behavior unchanged. AM maps `DuplicateUser` to 409 `already_exists` on the user resource (the stable colliding-field token as `resource_name`, never the caller-supplied value; the curated detail derives from the same typed field) and `PasswordPolicy` to 400 with a `password`/`PASSWORD_POLICY` field violation. `UsernameOrEmail` covers Keycloak's combined `ModelDuplicateException` constant — on KC 26 the only 409 shape `createUser` produces directly. Raw provider text stays digest-only in logs, same redaction posture as the existing arms. The IdP-plugin half that actually emits the classified variants lives in the consuming application; until it lands, every path still produces `Rejected` and nothing changes on the wire.

## Soft-delete cancels the pending conversion in-TX (account-management)
`DELETE /tenants/{id}` tombstoned the tenant but left its pending mode-conversion row dangling. `schedule_deletion`'s SERIALIZABLE TX now terminal-resolves it (`cancelled_by` = the deleting actor); the idempotent already-Deleted path runs the same fenced cancel, so pre-fix orphans are healed by simply retrying the DELETE. The audit event is emitted after commit (no double-count on serialization retry, no phantom event on rollback).

## Platform-scope the user-groups cascade system actor (account-management)
The retention hard-delete reaper's resource-group cascade ran scoped to the tenant being deleted; the resolver's subtree grant rooted at a Deleted tenant materializes to zero rows and fails closed, silently deferring the whole hard-delete backlog forever. The cascade context is now platform-scoped (nil home tenant resolves to Global at the resolver). The resolver-side half lives in the consuming application.

## Count self_managed direct children in `child_count` (account-management)
The public `child_count` read-shape field undercounted a parent whose direct children include a `self_managed` one — a parent with two active children where one is self_managed reported `child_count = 1` while `GET /tenants/{id}/children` listed both. Root cause: `count_children_grouped` clamped each child row by the caller's barrier-respecting scope, and a self_managed tenant's inbound closure edges carry `barrier = 1` even from its own parent, so the self_managed direct child dropped out of the count while the listing (direct-child carve-out) still showed it. The count now gates on the PARENT's Respect-reachability and then counts all its direct children; the `parent_id` predicate pins the count to depth 1, so nothing below a barrier leaks. Regression tests run under a real Respect-barriers scope (prior coverage ran under `allow_all`, barriers off, and could not catch this).

## Restore rg-namespace group resource type for PEP evaluation (resource-group)
The gts-rust v0.11.0 migration routed the group resource-type literal through the crate-name-derived SDK constant, silently renaming the PEP-evaluated type from `gts.cf.core.rg.group.v1~` to `gts.cf.core.resource_group.group.v1~`. Deployed role grants are written against the documented `gts.cf.core.rg.*` family, so the rename 403'd all group CRUD for tenant members and failed AM's tenant-delete ownership probe closed (surfaced as 503) — verified on a live deployment: a grant on `gts.cf.core.rg.*` is denied while `gts.cf.core.resource_group.*` is allowed, same caller and flow. Pins `GROUP_RESOURCE_TYPE` back to `cf.core.rg.group.v1~`, consistent with the untouched siblings (`rg.group_membership.v1~`, `rg.type.v1~`), with a comment so a future mechanical migration does not rename it again.

> Note for maintainers: if crate-derived GTS namespaces are genuinely the intended direction, that should be a separate coordinated migration — the whole type family at once, plus seed data and a data migration of existing grants — not a mechanical rewrite of a single literal (which is exactly how this regression happened).

## Verification
Branch is rebased onto current `main`. `cargo test` / `cargo clippy --all-targets` / `cargo fmt --check` are green across the six touched crates (toolkit-db, api-gateway, account-management, account-management-sdk, resource-group, resource-group-sdk); `cargo dylint --all` (incl. the GTS-layer lints DE0901/DE0904) green on the resource-group crates. The AM, gateway, and resource-group fixes were additionally exercised end-to-end against a live deployment (REST-level checks for the CORS header, `$orderby=status` incl. a cursor walk across ordinal groups, the duplicate/password-policy classification, the delete-cancels-conversion flow, and the self_managed `child_count` agreement between the single read and the listing).

## Summary by CodeRabbit

* **New Features**
  * Added clearer account-management errors for duplicate users and password-policy rejections, with sensitive provider details redacted.
  * Tenant deletion now automatically cancels pending conversion requests and records the responsible actor.
  * Tenant child listings support ordering and cursor pagination by lifecycle status.
  * CORS now exposes ETag by default and supports configurable exposed headers.

* **Bug Fixes**
  * Improved scoped child counts to include eligible direct children without exposing deeper subtrees.
  * Corrected resource-type identifiers and cursor handling for resource-group listings.

* **Documentation**
  * Clarified filtering and ordering behavior for categorical fields.

### DIFF
--- a/gears/system/account-management/account-management-sdk/src/field.rs
+++ b/gears/system/account-management/account-management-sdk/src/field.rs
@@ -45,6 +45,12 @@ pub const ROOT_TENANT_CANNOT_CHANGE_STATUS: &str = "ROOT_TENANT_CANNOT_CHANGE_ST
 /// `provisioning_metadata.realm_name`) — see [`PROVISIONING_METADATA_FIELD`].
 pub const IDP_INVALID_INPUT: &str = "IDP_INVALID_INPUT";
 
+/// The `IdP` rejected the supplied password against its configured
+/// password policy. Carried on the [`PASSWORD_FIELD`]
+/// field-violation so clients can attribute the failure to the
+/// password input; the raw policy text stays provider-side.
+pub const PASSWORD_POLICY: &str = "PASSWORD_POLICY";
+
 // ---------------------------------------------------------------------------
 // `field_violations[].field` attribution keys.
 //
@@ -68,6 +74,10 @@ pub const METADATA_FIELD: &str = "metadata";
 /// ([`ROOT_TENANT_CANNOT_DELETE`] / [`ROOT_TENANT_CANNOT_CONVERT`] /
 /// [`ROOT_TENANT_CANNOT_CHANGE_STATUS`]).
 pub const TENANT_ID_FIELD: &str = "tenant_id";
+
+/// `password` field for `IdP` password-policy rejects (carries
+/// [`PASSWORD_POLICY`]).
+pub const PASSWORD_FIELD: &str = "password";
 
 /// Shared fallback field for [`IDP_INVALID_INPUT`] when the `IdP`
 /// plugin cannot localise the violation to a specific sub-key — the
@@ -100,6 +110,8 @@ pub enum ValidationReason {
     RootTenantCannotChangeStatus,
     /// See [`IDP_INVALID_INPUT`].
     IdpInvalidInput,
+    /// See [`PASSWORD_POLICY`].
+    PasswordPolicy,
     /// Unmodeled / future reason — preserves the raw wire string.
     Unknown(String),
 }
@@ -117,6 +129,7 @@ impl ValidationReason {
             ROOT_TENANT_CANNOT_CONVERT => Self::RootTenantCannotConvert,
             ROOT_TENANT_CANNOT_CHANGE_STATUS => Self::RootTenantCannotChangeStatus,
             IDP_INVALID_INPUT => Self::IdpInvalidInput,
+            PASSWORD_POLICY => Self::PasswordPolicy,
             other => Self::Unknown(other.to_owned()),
         }
     }
@@ -132,6 +145,7 @@ impl ValidationReason {
             Self::RootTenantCannotConvert => ROOT_TENANT_CANNOT_CONVERT,
             Self::RootTenantCannotChangeStatus => ROOT_TENANT_CANNOT_CHANGE_STATUS,
             Self::IdpInvalidInput => IDP_INVALID_INPUT,
+            Self::PasswordPolicy => PASSWORD_POLICY,
             Self::Unknown(s) => s.as_str(),
         }
     }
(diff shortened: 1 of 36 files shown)


diffora:
All three review findings addressed and folded into their originating commits (history stays one-commit-per-fix):

- `ValidationReason` gains a `PasswordPolicy` variant wired into `from_wire`/`as_wire` and the round-trip test — the new reason no longer decodes as `Unknown` (folded into the IdP-classification commit).
- `account-management-v1.yaml` `child_count` description now documents the parent-gated semantics: all direct children (incl. self_managed) counted when the parent is Respect-reachable, `0` for a tenant seen only via the carve-out (folded into the child_count commit).
- `count_children_grouped` is now a single statement: the Respect-scoped parent gate rides as an `IN`-subquery inside the grouped count, so the authorization check and the count observe one database snapshot — no gate-then-count race (folded into the child_count commit).

Tests, fmt, clippy green on the touched crates; branch force-pushed.

Artifizer (approved):

PR #4230, opened 2026-07-20
CI: Add per-job path filtering to skip unaffected CI jobs

maurolacy:
Add a lightweight `changes` job using dorny/paths-filter that detects which areas of the repo changed.  Downstream jobs now conditionally run based on these outputs:

  - `clippy`/`test`/`integration`/`coverage`/`shear`: Rust source or CI changes.
  - `test-fips`: FIPS-related files, deps, or CI changes.
  - `security`: Dependency files or CI changes.
  - `dylint`: dylint sources, rust sources, or CI changes.

`workflow_dispatch` always runs all jobs (no filtering).

This avoids re-running the full pipeline when only a subset of the codebase is touched (e.g. skipping FIPS, dylint, and security checks when only gear source code changes).

Other smaller changes / improvements:

  - Caching for dylint/shear — rust-cache + sccache (minus sccache for dylint).
  - Shear advisory-only — `continue-on-error: true`.
  - Free disk space — added to all heavy compilation jobs.
  - Dynamic cache keys — `hashFiles('rust-toolchain.toml')` instead of hardcoded version.
  - Centralised nightly version — single definition for all nightly jobs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI efficiency by running only the checks that correspond to the files changed.
  * Documentation-only updates can now skip unnecessary validation workflows.
  * Enhanced runner stability with improved disk-space handling and more consistent build/test caching.

* **Chores**
  * Standardized nightly toolchain selection across coverage and analysis-related checks.
  * Improved cache key determinism for stable and FIPS lanes.
  * Updated compiler-caching behavior for analysis runs to avoid toolchain conflicts; caching is enabled conditionally where safe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,8 @@ env:
   SCCACHE_CACHE_SIZE: "2G"
   # Usually better determinism and smaller cache churn in CI.
   CARGO_INCREMENTAL: "0"
+  # Pinned nightly toolchain used by coverage, dylint, and shear jobs.
+  RUSTUP_TOOLCHAIN_NIGHTLY: nightly-2026-04-16
 
 jobs:
   # ── Change detection ──────────────────────────────────────────────────
@@ -450,7 +452,7 @@ jobs:
       contents: read
       id-token: write
     env:
-      RUSTUP_TOOLCHAIN: nightly-2026-04-16
+      RUSTUP_TOOLCHAIN: ${{ env.RUSTUP_TOOLCHAIN_NIGHTLY }}
     steps:
       - name: Free disk space
         uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -27,17 +31,77 @@ env:
...
+
+  # ── Jobs ──────────────────────────────────────────────────────────────
+
   # Fast clippy: ubuntu only, 2-pass strategy (~5-6 min vs 30-40 min before).
   # Linting output is platform-independent; platform-specific paths (fips) are
   # covered by the test-fips job. Full feature matrix runs nightly (clippy-nightly.yml).
   clippy:
     name: Clippy
+    needs: changes
+    if: needs.changes.outputs.rust == 'true' || needs.changes.outputs.ci == 'true' || github.event_name == 'workflow_dispatch'
```
MikeFalcon77: If the `changes` job ever fails (paths-filter error, GitHub API hiccup), every downstream job with `needs: changes` is skipped — because this `if:` does not use `always()`/`!cancelled()`, so a failed `needs` short-circuits the job before the condition is evaluated. That means even a manual `workflow_dispatch` run would skip everything, since the `github.event_name == 'workflow_dispatch'` branch never gets a chance to run. The `changes` job becomes a single point of failure for the whole pipeline.

Consider guarding against a failed dependency, e.g. `if: always() && (needs.changes.result == 'skipped' || ...)`, or fall back to running the job when `changes` did not succeed. At minimum, please trigger a manual `workflow_dispatch` on this branch to confirm the `changes` job does not fail on non-PR events (paths-filter behavior without an explicit `base` is not obvious there).

This applies to all gated jobs, not just this one.

MikeFalcon77 (changes requested):

.github/workflows/ci.yml
maurolacy: Good point. Changed it to run all jobs if the `changes` job fails for whatever reason.

MikeFalcon77 (approved):

PR #4233, opened 2026-07-20
docs(settings-service): add Settings Service PRD

entropyshift:
Adds the Product Requirements Document for the **Settings Service** gear — the
platform's typed configuration gear providing a single surface to manage
settings (categories, typed keys, defaults), apply staged value changes safely,
and resolve per-tenant overrides through the tenant scope hierarchy.
Scope of this PR is **documentation only**: a new `gears/system/settings-service/docs/PRD.md`.
No code, contracts, or gear wiring are introduced.

Key content:
- **Problem & goals**: one configuration hub (not per-gear surfaces), safe
  change management, multi-tenancy by design, type-safe configuration, and
  compliance-readiness, with GA success metrics.
- **Actors**: platform/tenant admins, compliance reviewer, and system actors
  (contributing modules, tenant/authn/authz/license resolvers, types registry),
  each with fail-closed availability expectations.
- **Functional requirements**: settings & category model, GTS typed values +
  traits & validation, standard/advanced mode & search, defaults & revert,
  staged-change → credential-verified apply (values only; declaration edits are
  immediate), multi-tenant cascading overrides & scope enforcement,
  security/secrets/audit, and module-contributed settings.
- **Scope Class** (global / cascading / local) deriving override & cascade
  behaviour deterministically, secure-by-default.
- **NFRs**: read-path cache p95 < 50ms sustaining 1,000 reads/s over 5,000
  settings / 100,000 tenants / depth 10; fail-safe staged apply; scope isolation.
- Data-classification (public/PII/secret), risk entries (tenant-resolver
  outage, concurrent apply, descendant invalidation, audit volume), and clear
  out-of-scope boundaries (API schemas/data models deferred to DESIGN,
  per-user preferences owned by `simple-user-settings`, boot-critical config
  owned by deployment tooling).

## Type of Change
- [x] Documentation update

### DIFF
--- a/gears/settings-service/docs/PRD.md
+++ b/gears/settings-service/docs/PRD.md
@@ -1,0 +1,892 @@
+# PRD — Settings Service
+
+
+<!-- toc -->
+
+- [1. Overview](#1-overview)
+  - [1.1 Purpose](#11-purpose)
+  - [1.2 Background / Problem Statement](#12-background--problem-statement)
+  - [1.3 Goals (Business Outcomes)](#13-goals-business-outcomes)
+  - [1.4 Glossary](#14-glossary)
+- [2. Actors](#2-actors)
+  - [2.1 Human Actors](#21-human-actors)
+  - [2.2 System Actors](#22-system-actors)
+- [3. Operational Concept & Environment](#3-operational-concept--environment)
+- [4. Scope](#4-scope)
+  - [4.1 In Scope](#41-in-scope)
+  - [4.2 Out of Scope](#42-out-of-scope)
+- [5. Functional Requirements](#5-functional-requirements)
+  - [5.1 Settings & Category Model](#51-settings--category-model)
+  - [5.2 Typed Values & Validation](#52-typed-values--validation)
+  - [5.3 Standard/Advanced Mode & Discoverability](#53-standardadvanced-mode--discoverability)
+  - [5.4 Defaults & Revert](#54-defaults--revert)
+  - [5.5 Staged Change → Apply](#55-staged-change--apply)
+  - [5.6 Multi-Tenant Overrides & Cascading Inheritance](#56-multi-tenant-overrides--cascading-inheritance)
+  - [5.7 Security, Secrets & Audit](#57-security-secrets--audit)
+  - [5.8 Module-Contributed Settings](#58-module-contributed-settings)
+  - [5.9 Additional Scope Items](#59-additional-scope-items)
+- [6. Non-Functional Requirements](#6-non-functional-requirements)
+  - [6.1 Gear-Specific NFRs](#61-gear-specific-nfrs)
+  - [6.2 NFR Exclusions](#62-nfr-exclusions)
+- [7. Public Library Interfaces](#7-public-library-interfaces)
+  - [7.1 Public API Surface](#71-public-api-surface)
+  - [7.2 External Integration Contracts](#72-external-integration-contracts)
+- [8. Use Cases](#8-use-cases)
+- [9. Acceptance Criteria](#9-acceptance-criteria)
+- [10. Dependencies](#10-dependencies)
+  - [10.1 Launch Prerequisites (blocking)](#101-launch-prerequisites-blocking)
+- [11. Assumptions](#11-assumptions)
+- [12. Risks](#12-risks)
+- [13. Open Questions](#13-open-questions)
+- [14. Traceability](#14-traceability)
+
+<!-- /toc -->
+
+## 1. Overview
+
+### 1.1 Purpose
+
+Settings Service is the platform's typed configuration gear. It provides a single surface for platform and tenant administrators to manage settings (categories, typed keys, defaults), apply staged changes safely, and resolve per-tenant overrides through the tenant scope hierarchy — including feature flags and other platform/tenant customization consumed by other gears.
+
+### 1.2 Background / Problem Statement
+
+The platform has no configuration service yet: each gear owns its own configuration through config files, environment variables, or ad-hoc endpoints. Without a central service, every new gear adds another configuration surface with its own change model, no central audit trail, no staged-preview-then-apply safety net, and no per-tenant override behavior. Administrators cannot discover all configuration from one place or preview the blast radius of a change before it goes live, and tenants have no cascading-override model for configuring their own scope. Comparable platforms that grew this way (see survey below) show how costly it is to retrofit a unified configuration model later.
+
+**Key problems solved**:
+
+- **Configuration discoverability**: administrators cannot find or reason about all platform configuration from one place.
+- **Unsafe changes**: without a staged-then-apply model, configuration changes take effect immediately and can destabilize a live platform.
+- **Multi-tenant configuration**: configuration models in comparable platforms (see survey below) are single-tenant, retrofitted with folders rather than purpose-built cascading inheritance; this platform needs cascading inheritance from the start.
+- **Type safety**: today, configuration values are stored without validation against a declared type — this PRD makes typed validation mandatory ([§5.2](#52-typed-values--validation)).
+- **Granular access**: there is no per-setting access governance — who may see or change which setting is all-or-nothing per surface. This PRD delivers role/visibility/licence gating and per-setting `tenant-visible`/`tenant-overridable` flags; finer per-integration visibility is a recorded follow-up.
+- **Compliance**: mutations to platform configuration are not consistently audited or gated behind credential verification.
+
+> **Note**: Boundary between "a setting" and "a managed entity" (e.g. quotas/limits) — a value is a *setting* when it configures platform/service behaviour and has no independent lifecycle. Entities with their own lifecycle, quota/policy semantics, and API remain owned by their respective gears; settings can reference them via entity-reference traits but MUST NOT duplicate them.
+
+**Neutral survey of comparable configuration/administration systems** (informal internal product research — observations from public product behaviour, not independently sourced or vendor-verified; patterns worth adopting, pitfalls worth avoiding): VMware vCloud Director offers per-tenant override as a first-class concept but a heavy, modal-driven UX with no global search or pre-apply preview. OpenNebula cleanly splits bootstrap/system config from runtime-adjustable settings, but keeps file-based and UI-based config as two sources of truth. Nutanix Prism aggregates configuration across clusters with good basic/advanced disclosure, but still requires CLI for some configuration. VMware vSphere makes inheritance visible via folder/group cascading, but retrofits a single-tenant model onto folders. Cloud org models (AWS Organizations, SCPs) offer effective-policy views analogous to effective-value-with-trace, but scatter configuration across many services.
+
+**Differentiators** (cross-cutting principles applied throughout this PRD):
+
+1. **Discoverability** — global cross-field search with match context; never force users to learn the category tree.
+2. **Inheritance transparency** — always show effective source with a drill-into walk.
+3. **Staged apply with preview** — show what changes, which services it touches, and which descendants it affects, before committing.
+4. **Multi-tenancy by design** — model the tenant scope hierarchy natively; do not retrofit a single-tenant abstraction.
+5. **One hub, not many** — all configuration in one place; embedded feature affordances read/write the same service.
+6. **Inform, don't block** — surface consequences (cascading impact) as non-blocking warnings; reserve hard blocks for the irreversible.
+7. **Type-safe & extensible** — GTS types + traits grow the configuration surface without code changes.
+
+### 1.3 Goals (Business Outcomes)
+
+- **One configuration hub, not many**: prevent configuration from fragmenting into per-gear surfaces as the platform grows. All platform configuration is discoverable, searchable, and governed in one place, reducing administrator error and onboarding time.
+- **Safe change management**: a staged change → explicit apply model (for **value** changes) with a pre-apply preview and credential step-up prevents accidental, unreviewed, or partially-applied configuration changes to a live platform. Defining a setting is a separate authoring action: descriptive-metadata edits apply immediately, resolution-affecting fields (Schema Default, type, Scope Class) are immutable (changed only via a replacement declaration or new major type version), and retire/reactivate require credential step-up.
+- **Multi-tenant by design**: tenants configure their own scope; values cascade down the tenant scope hierarchy, with full transparency about where an effective value came from. This is a differentiator versus single-tenant configuration models retrofitted onto folders.
+- **Type-safe configuration**: every setting value is validated against a GTS type before it is stored or applied, catching invalid configuration at write time rather than in production.
+- **Compliance-ready**: every mutation is audited (secrets masked), every apply is credential-verified, and visibility/override permissions are governed per setting.
+
+**Success Metrics** (measured at GA unless noted):
+
+
+| Goal                            | Measurable Success Criterion                                                                                                                | Target                                                                                                                                               |
+| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
+| One configuration hub, not many | Gears exposing their in-scope platform configuration through the Settings Service, vs. through gear-specific configuration surfaces         | 100% of in-scope platform configuration registered in the Settings Service at GA; zero standalone per-gear configuration surfaces introduced post-GA |
+| Safe change management          | Value mutations that go through the staged-change → credential-verified Apply path, vs. any direct-write path that bypasses staging (declaration authoring is immediate by design and outside this metric) | 100% of value mutations staged-then-applied; 0 direct-write paths for values at GA                                                                                    |
+| Multi-tenant by design          | Cascading settings resolvable via the inheritance walk with an exposed effective source                                                     | 100% of settings declared `cascading` support cascading inheritance with source trace at GA                                                          |
+| Type-safe configuration         | Setting values rejected at write time when they fail their declared GTS type, vs. accepted and only failing later                           | 100% of writes validated against the declared GTS type before persistence; 0 untyped writes at GA                                                    |
+| Compliance-ready                | Mutating operations that produce a complete audit record (actor, target, pre/post, timestamp, outcome, request id)                          | 100% of mutating operations audited; 0 audit-trail gaps found in the first post-GA compliance review                                                 |
+
+
+### 1.4 Glossary
+
+
+| Term                                    | Definition                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Setting (Key)**                       | A single configurable item with a unique key within its category, a typed value, an independent default, and metadata (mode, description, permissions, Scope Class, last-change timestamp).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+| **Category**                            | A named, flat (single-level) grouping of settings — categories do not nest. Removable only when empty.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
+| **Effective Value**                     | The value in force for a setting at a given scope. For a `cascading` setting it is the override resolved via the inheritance walk if present, otherwise the Schema Default; `global` and `local` resolve differently — see the Scope Class resolution table ([§5.6](#56-multi-tenant-overrides--cascading-inheritance)).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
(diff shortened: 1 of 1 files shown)


gears/settings-service/docs/PRD.md
```diff
@@ -0,0 +1,826 @@
...
+
+**Key problems solved**:
+
+- **Configuration discoverability**: administrators cannot find or reason about all platform configuration from one place.
+- **Unsafe changes**: without a staged-then-apply model, configuration changes take effect immediately and can destabilize a live platform.
+- **Multi-tenant configuration**: configuration models in comparable platforms (see survey below) are single-tenant, retrofitted with folders rather than purpose-built cascading inheritance; this platform needs cascading inheritance from the start.
+- **Type safety**: configuration values are not validated against a declared type before being stored or applied.
+- **Compliance**: mutations to platform configuration are not consistently audited or gated behind credential verification.
```
Artifizer, line 62: I'd also add key problems here and later describe it in more details

- **Granular access**: different actors may need to have different access permissions, e.g. a setting defined by integration A not necessary needs to be visible by integration B, or a tenant setting X not necessary to be visible by user with role Y 

- **Runtime data model adjustment**: integrations of platform extensions may need to register new data types associated with tenants or maybe with other objects in runtime. For example integration A needs to store some data per tenant basis

- **Settings Pattern Unification**: different platform objects such as tenants, users, hosts, VMs, subscriptions, etc. may need to have different runtime-level settings that are similar in it's nature and semantics - schema control, versioning, access management, defaults/inheritance rules, lock rule, audit or notification event, setting value validation (e.g. min/max), etc. It makes no much sense to repeat such a complex logic in different services responsible for the domain model elements and it would make sense to support a general purpose multi-domain setting service

gears/settings-service/docs/PRD.md
```diff
@@ -0,0 +1,826 @@
...
+| **Defaults & Revert**                          | `p2`         | Independent Schema Defaults; revert at tenant scope (to nearest ancestor) and platform scope (to Schema Default), with fallback preview.                                                                                                                 |
+| **Domain Affinity Filtering**                  | `p3`         | Optional per-setting domain affinity; hub filters categories by the admin's current domain; cross-domain hidden by default with an "All domains" platform-admin view.                                                                                    |
+| **Internal Activation Endpoints**              | `p3`         | Restricted, token-only inter-service operations (integrity-verified apply, tenant cache invalidation).                                                                                                                                                   |
```
Artifizer: Please consider to add:

## 1. Multi-domain subject support, not tenant-only scope (p1)

The model should not be restricted to tenant settings only. It would be more aligned with the reference requirements to model each setting against a generic subject:

- tenant_id
- subject_type
- subject_id

Where subject_type can be a GTS type such as Tenant, User, Storage, Application, etc., and subject_id is the UUID of the concrete object the setting belongs to.

## 2. Boolean schema and AND-composition across hierarchy not addressed (p1)

The reference requirements (R-SS4, R-TH4) define two related capabilities that do not appear to be carried forward in this PRD:

Boolean schema — an alternative to a full JSON/GTS schema that allows a setting to be declared as a boolean-typed policy flag with specific behavioral rules.

AND-resolution across the tenant hierarchy — when a boolean-typed setting is cascading, the effective value for a given tenant is the logical AND of the tenant's own value and all ancestor values in the hierarchy. For example:

```
T1.partner_ga = true
T2.partner_ga = false  (T2 child of T1)
T3.partner_ga = true   (T3 child of T2)
 
effective(T2) = AND(true, false)  = false
effective(T3) = AND(true, false, true) = false
```

The PRD's Scope Class model and cascading inheritance walk resolve the nearest ancestor override as the effective value, which is not equivalent to AND-composition.

## 3. GTS based setting type access control (p1)

The idea is straightforward - grant access based on GTS type name with respect of wildcards
Please make it mentioned more explicitly

## 4. Setting value lock / immutability (p2)

Some settings are effectively write-once or irreversible once enabled.
For example, if storage immutability is set to true, that value may need to become locked and no longer editable through normal admin flows.

## 5. Bulk retrieval implicitly assumed but not a stated requirement (p2)

It seems like it's not a part of functional requirements / scenarios

## 6. Self-service tenant inheritance rules

Sometimes the `self-service` partner tenants (those that are not managed by parent tenants) may need to inherit settings from their parent tenant, even when normal tenant hierarchy barriers would otherwise block that.

Example:

- Parent partner defines branding/settings.
- Child self-service tenant should still be able to receive those inherited settings.
- This behavior should be configurable, with a default behavior clearly defined.

The current PRD describes general tenant hierarchy inheritance, but I don’t see a clear rule for the barriers and this self-service tenant case.

## 7. MFA (sudo-session) requirement for setting modification (p3?)

The idea is that some of the settings would need to have short-living MFA (sudo) session for any change, so that just log in into the system would not be enough

Artifizer (changes requested):

vzsergeyg:
**Proposal to unblock this PR (and #4280): merge-first with a phase register.**

The open review asks on both PRs are, with two exceptions, requests to widen the charter — not objections to what is written. Absorbing them into this document would grow it past mergeability while the asks themselves deserve real design attention. The inverse — merging with a blanket "out of scope" — would discard genuinely good requirements. We propose the middle path already used on #4280 for the residual review findings: **defer with record**.

**Phase 1 — amend this PR with wording-level edits, then merge.** Eight concrete edits, none structural:

1. **E1 (consistency):** the scope table still lists *Internal Activation Endpoints* (p3), but the companion DESIGN-activation in #4280 explicitly removed internal REST endpoints and the service-token surface (§5.2). Replace the row with *Cross-Replica Cache Coherence*, which is what the design actually contains.
2. **E2 (editorial):** §1.2 — reword "configuration values are not validated…" so it cannot be misread as a spec statement: "Today, values are stored without validation against a declared type — this PRD makes typed validation mandatory."
3. **E3 (FR gap):** bulk effective-value retrieval is implemented in the design (`resolve_bulk`, `get_effective_bulk`, browse endpoint) but never stated as a requirement. Add one FR line.
4. **E4 (explicitness):** §5.7 — one sentence stating that access policies can target settings **by GTS type id, including wildcard patterns**, enforced through the platform authorization layer.
5. **E5 (forward-compat guard, subjects):** under the Versatility NFR: *"v1 value identity is (declaration, scope) with scope = platform | tenant. The model MUST NOT preclude extending value identity with additional subject dimensions (e.g. GTS-typed domain objects) in later phases."* One matching sentence in DESIGN §4.7 (the id scheme and partial unique indexes tolerate an added subject column). This makes the multi-domain-subject ask retrofit-safe without designing it now.
6. **E6 (forward-compat guard, resolution):** same section: *"Effective-value resolution is a function of Scope Class; additional resolution strategies (e.g. boolean AND-composition across the hierarchy) MAY be added as new classes/traits without changing existing settings' semantics."*
7. **E7 (problem statement):** §1.2 — add a granular-access bullet scoped to what v1 actually delivers (role/visibility/licence gating, tenant-visible/overridable), with per-integration visibility recorded as a follow-up.
8. **E8 (phasing note):** short §4 note: v1 = this document; recorded follow-ups live as issues labeled `settings-service` (list below).

We are happy to contribute these edits as a small PR against this branch, or the author applies them — whichever is easier.

**Phase 2 — "governance & semantics" bundle**, one issue each (we volunteer to file them, crediting the requesting reviewer): value locks & conditional mutability (incl. lock/unlock API, write-once/irreversible values) · per-type step-up policy with session TTL · boolean AND-composition as an additional resolution strategy · self-service inheritance barriers (co-designed with `tenant-resolver`, where barrier semantics live) · per-edition/per-tenant default sets (already named as a candidate follow-up in Out-of-Scope) · per-type event fan-out refinement with affected-set cost analysis · public (auth-less) read behind a threat review · activation-delivery timing NFR.

**Phase 3 — subject-scoped settings** as its own PRD increment: value identity `(declaration, tenant, subject_type, subject_id)` with GTS-typed subjects, per-subject-type scope classes, lifecycle cleanup on subject removal (generalizing the existing `tenant_deleted` subscription), authz scoping. This is the structural generalization several comments converge on; guard E5 keeps it non-breaking. It should be designed with the teams that hold the concrete use cases.

**Two asks we propose to decline, with reasons** (detailed in the threads): re-framing the service as a general "serverless DB" (that charter belongs to `simple-resource-registry`; a settings gear that becomes a second general-purpose store dissolves the settings-vs-managed-entities boundary this PRD deliberately draws), and dropping namespaced keys (the key is a GTS *instance* id, not a type id — see thread reply).

**Merge order:** this PR (amended) first; #4280 rebases its PRD copy on the result. #4280 itself needs no content change for the open review asks — its author's defer-with-record triage stands.

Per-thread dispositions follow on each comment so no ask is left unanswered. If the reviewers and author agree with this path, we can have the edit batch and the issue register up within a day.

---

**UPDATE (2026-08-13) — revised position after a review round on our side.** Three items above change; everything else stands as proposed.

**1. Subject-type support moves from "Phase 3 + guard" into this PRD (supersedes E5's guard-only posture).** Proposed in-line change for the author:

> *Value identity:* "A setting value is identified by `(declaration, scope)` where scope is `platform`, `tenant`, or `(tenant, subject_type, subject_id)`. `subject_type` MUST be a GTS-registered type and `subject_id` a GTS instance identifier. Scope Class is definable per subject type. Deletion of a subject triggers cleanup of its values (generalizing the existing `tenant_deleted` subscription)."

The *model, API shape, and uniqueness indexes* land subject-aware in v1; the *implementation* MAY phase (tenant scope first, subject scopes following) — so p1 delivery is not enlarged, but nothing ships that would need migration. This replaces edit E5's "MUST NOT preclude" with actual model support, and it means the "multi-domain subjects" ask is accepted into the document rather than deferred.

**2. GTS coherence made normative (one sentence).** Proposed addition wherever types are introduced: "Subject types, setting value types, and event payload schemas resolve through the platform type system (GTS): keys remain GTS instance ids, event payloads carry a registered `dataschema`, and schema evolution follows GTS major/minor compatibility rules." This is affirming the document's existing direction — stating it once, normatively, keeps later additions (incl. subject types) on the same rails.

**3. Hierarchy barriers: base semantics into this PRD, as a *downstream-only default seam* (supersedes "Phase 2 co-design" for the base mechanism).** Proposed in-line FR:

> "Hierarchy barriers block upward and lateral setting visibility. A barriered (isolated/unmanaged) tenant MAY receive **default, non-sensitive** settings from its parent chain through an explicit barrier seam: the tenant resolver exposes the barrier relationship, and only declarations explicitly marked barrier-traversable resolve downward — as read-only defaults, never values carrying the `secret` trait or a sensitivity classification. Nothing traverses upward or laterally; the traversal set is enumerable and auditable."

Advanced barrier governance (locks, conditional mutability, per-type overrides) stays in the Phase-2 bundle as before.

We will propose these as concrete in-line PRD changes (the edit-batch PR offered above) asap.

gears/settings-service/docs/PRD.md
vzsergeyg: Agreed on the first, and proposing homes for the other two:

- **Granular access** — accepted for this PR (edit E7 in the plan comment above): a §1.2 problem bullet scoped to what v1 delivers (role/visibility/licence gating, tenant-visible/overridable flags), with per-integration visibility recorded as a Phase-2 follow-up. Access decisions already flow through the platform authorization layer; E4 makes GTS-pattern targeting explicit.
- **Runtime data-model adjustment** (integrations registering new per-tenant data types at runtime) — as stated, this is generic typed per-tenant storage, which is `simple-resource-registry`'s charter; duplicating it here would create two general-purpose stores with one boundary between them undefined. The settings-shaped part of the need — typed, validated, inheritable, audited *configuration* attached to an integration-owned subject — is exactly the Phase-3 increment (subject-scoped settings) in the plan comment.
- **Settings pattern unification** (tenants/users/hosts/VMs/subscriptions on one machinery) — we read this as the north star, and propose it as the explicitly recorded Phase-3 PRD increment rather than a v1 widening. Guard E5 (value-identity extensibility) goes into this PR now so nothing merged today obstructs it. The per-user slice already has its own recorded home (`gears/simple-user-settings` in Out-of-Scope) — the same carve-out pattern.

---

**UPDATE (2026-08-13):** revised on our side — the settings-pattern-unification direction is now accepted *into this PRD* at the model level: value identity extends to `(declaration, tenant, subject_type, subject_id)` with GTS-typed subjects (implementation may phase; model/API/indexes land subject-aware). Details and paste-able wording in the updated plan comment on this PR. The per-user carve-out (`gears/simple-user-settings`) is unchanged.

gears/settings-service/docs/PRD.md
vzsergeyg: Point-by-point, against the plan comment above:

1. **Multi-domain subjects (p1)** → **Phase 3, guarded now.** Attaching values to `tenant_id + subject_type + subject_id` changes value identity, the resolution walk, authz scoping, lifecycle and cache keying at once — a PRD-sized increment, not an amendment. What we do now: edit E5 adds a normative guard that v1's model must not preclude additional subject dimensions, and DESIGN §4.7's id scheme + partial unique indexes already tolerate an added subject column. So the generalization lands later without breaking anything merged today.
2. **Boolean AND-composition (p1)** → **Phase 2, guarded now (E6).** Agreed it is not expressible today — nearest-ancestor-wins cannot represent "every ancestor must consent". It is additive: a new resolution strategy selected via Scope Class/trait alongside the existing one. The design questions worth doing properly: interaction with `tenant-overridable`, revert semantics, and effective-source tracing ("which ancestor turned this off"). One ask back: the referenced requirements (R-SS4 / R-TH4) don't resolve to any document in this repository — could you restate them as self-contained text in the Phase-2 issue, so the requirement can be evaluated on its own terms by everyone reviewing it?
3. **GTS-based access control (p1)** → **accepted now (E4).** The mechanism exists (policy decisions through the platform authorization layer); the PRD gains an explicit sentence that policies can target setting types by GTS id with wildcard patterns.
4. **Value lock / immutability (p2)** → **Phase 2.** Declaration-level immutability already exists (behavior-affecting fields immutable; retire/reactivate step-up-gated); *value-level* write-once/irreversible is genuinely new and pairs with the lock/unlock API ask on #4280 — one "locks & conditional mutability" issue covers both.
5. **Bulk retrieval (p2)** → **already implemented, requirement line added (E3).** `resolve_bulk` / `get_effective_bulk` (per-key outcomes, never all-or-nothing) and the browse endpoint exist in DESIGN §4.5–4.6; the PRD just never stated the FR.
6. **Self-service inheritance barriers** → **Phase 2, cross-gear.** Barrier/self-service semantics are tenant-model concepts; if the settings gear invents them locally they will disagree with `tenant-resolver` later. The issue gets filed as a co-design item across both gears.
7. **MFA / sudo-session (p3)** → **Phase 2.** The mechanism is already in the design: step-up is IdP re-authentication behind a swappable `StepUpVerifier` (DESIGN §4.2), required today for apply and behavior-affecting declaration actions. The delta is a per-type trait plus a session TTL — same bundle as the step-up-configurability ask on #4280.

Net: nothing here is dismissed; two items land in this PR as text, five become registered increments with your name on the request, and the two structural ones get forward-compat guards merged now.

---

**UPDATE (2026-08-13) — two items revised after a review round on our side:**

**Item 1 (multi-domain subjects):** upgraded from "Phase 3, guarded now" to **model-level acceptance in this PRD** — value identity extends to `(declaration, tenant, subject_type, subject_id)` with GTS-typed subjects and per-subject-type Scope Class; implementation may phase, the model does not. Paste-able wording is in the updated plan comment.

**Item 6 (self-service inheritance barriers):** the *base* barrier semantics move into this PRD as a **downstream-only default seam** — a barriered tenant may receive default, non-sensitive (never `secret`-trait) settings from its parent chain via an explicit, auditable seam backed by the tenant resolver; nothing traverses upward or laterally. Advanced barrier governance remains Phase 2 as proposed.

Items 2–5 and 7 stand exactly as previously dispositioned.

vzsergeyg:
@Artifizer — heads-up: following a review round on our side we've **updated our comments above** (and the two thread replies): subject-type support is now accepted into the PRD at the model level, GTS coherence is stated normatively, and hierarchy barriers gain a downstream-only default seam for non-sensitive settings (tenant-resolver-backed). Everything else stands as previously dispositioned. We'll propose the corresponding PRD changes inline asap.

vzsergeyg:
The edit batch promised above is up: https://github.com/entropyshift/gears-rust/pull/1 — one commit against this PR's branch, carrying E2/E3/E4/E7/E8 plus the three revised items from the 2026-08-13 update (subject-scoped value identity at model level, the normative GTS-coherence sentence, and the downstream-only barrier default seam). E1 was already resolved by the activation/replica-coherence commit; E5's guard is superseded by the subject-type model. Merging it lands everything inside this PR; every edit is independent — take what's useful.

gears/settings-service/docs/PRD.md
Artifizer: @vzsergeyg , I can't see PRD update, only comments

shaneemberton:
@coderabbitai pause

Artifizer (approved):

Artifizer (approved):

gears/settings-service/docs/PRD.md
```diff
@@ -0,0 +1,892 @@
...
+
+> **Note**: Settings do not carry per-setting version history or one-action rollback to a previous applied value in v1. Operational safety before activation is provided by the staged-change → credential-verified Apply model; after apply, the supported corrections are revert-to-Schema-Default (platform scope) and revert-to-ancestor (tenant scope), with the prior value recoverable from the audit trail. Per-setting version history + rollback is a candidate follow-up.
+
+- **Rationale**: A clear, independent default and predictable revert-to-ancestor/default behavior let administrators undo a bad override without guessing where the value will land.
+- **Actors**: `cpt-cf-settings-service-actor-platform-admin`, `cpt-cf-settings-service-actor-tenant-admin`
+
+### 5.5 Staged Change → Apply
```
nonameffh, line 346: [MAJOR] This section blocks non-interactive value writes by integrations. 

Settings are not determined solely by administrators viewing the hub. For example, they can be also defined by tenant provisioning workflows - which apply values for each tenant during the creation phase (time zone, language locale, default branding settings) even before any administrator logs in; as well as by configurable integrations - partner-level integrations used to configure settings for each tenant.

Normalising every credential (human and service) into a single security context is necessary - and worth stating explicitly in the PRD, since authn-resolver today returns only "a validated identity" - but it is not sufficient while the actor model prohibits the operation and Apply requires human re-verification.

Suggestion: add a value-writing machine actor (the step-up contract defines a verification method for non-interactive principals) or/and support marking a declaration as not-staged; authorized writes take effect immediately

gears/settings-service/docs/PRD.md
```diff
@@ -0,0 +1,892 @@
...
...der the appropriate input and pre-validate; structured (object/array) values **MUST** be supported, not only scalar values. A setting whose type carries the `secret` trait **MUST** be stored encrypted at rest and **MUST** be masked in every **administrative** read, search, or audit response; its plaintext **MUST NOT** be retrievable through any administrative or human-facing path (single get, bulk get, search, list-by-category, audit). Plaintext secret resolution **MUST** be available only through one explicit, authenticated **machine-only runtime path** — the Settings Read SDK / internal in-process reader ([§7.1](#71-public-api-surface)) — and only for a consuming service authorized to that specific setting; every such plaintext resolution **MUST** be recorded as a secret-use audit event ([§5.7](#57-security-secrets--audit)) and its value **MUST NOT** be cached in plaintext ([§6.1](#61-gear-specific-nfrs)). Subject types, setting value types, and event payload schemas **MUST** resolve through the platform type system (GTS): setting keys remain GTS instance identifiers, event payloads carry a registered `dataschema`, and schema evolution follows GTS major/minor compatibility rules.
```
nonameffh, line 297: We have credstore gear, from my perspective this functionality should stay there, otherwise we implement twice encryption-at-rest, rotation and secret-use audit

gears/settings-service/docs/PRD.md
```diff
@@ -0,0 +1,892 @@
...
+| **Scope**                               | The point in the tenant hierarchy a setting value applies to: platform (root) or a specific tenant.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
+| **Subject Scope**                       | An optional third scope dimension: a value attached to `(tenant, subject_type, subject_id)` where the subject type is GTS-registered (e.g. an integration or host). Model-level in v1 (identity, API shape, indexes); implementation may phase. See `cpt-cf-settings-service-fr-subject-scoped-values` ([§5.6](#56-multi-tenant-overrides--cascading-inheritance)). |
```
nonameffh, line 118: [ MAJOR] This closes access-level dimensions to very specific model that doesn't fully satisfy Constructor needs. We need to express a multi-dimensional, AuthZ-driven access model - read and write decided by the PDP over the principal's attributes and the setting's attributes. Two examples that are ordinary requirements for us and are not expressible today:

- a principal of type `user` holding a given `role` may read every setting visible in its scope;
- an application may read only the settings labelled for it: application `foo` may read settings labelled `bar`, and is denied to read settings labelled `baz`.

I would suggest enhancing the declaration model to support open, vendor-defined metadata: the declaration carries a label map authored by an operator (not developer), forwarded to the PDP as resource attributes

nonameffh (approved):
I have three comments; please review them. 
Approved on the condition that there are no fundamental objections to the comments and consiquent resolution of these comments in subsequent PRs.

PR #4237, opened 2026-07-21
feat(toolkit-db): implement DB-native advisory locks with safe ownership

modcrafter77:
Replace the file-only advisory lock path with PostgreSQL/MySQL session locks and a cancellation-safe file-marker fallback, including unambiguous v2 lock keys and shared SQLite DSN/path identity.

## Summary by CodeRabbit

* **New Features**
  * Added advisory lock support for SQLite, PostgreSQL, and MySQL.
  * Added configurable keepalive settings and database-scoped locking.
  * Added retry policies with exponential backoff and configuration validation.
  * Added explicit lock release with clear error reporting.

* **Bug Fixes**
  * Improved cleanup after dropped locks, connection loss, shutdown, and contention.
  * Improved SQLite path handling for memory databases, relative paths, and symlinks.

* **Tests**
  * Expanded coverage for native locks, keepalive behavior, retries, configuration, and path handling.

### DIFF
--- a/libs/toolkit-db/src/advisory_locks.rs
+++ b/libs/toolkit-db/src/advisory_locks.rs
@@ -1,18 +1,146 @@
-//! Advisory locks implementation with namespacing and retry policies.
+//! Advisory locks with namespacing and retry policies.
 //!
-//! Cross-database advisory locking with proper namespacing and configurable
-//! retry/backoff.
+//! Backends:
+//! - **`PostgreSQL`** (`pg`): session-level `pg_try_advisory_lock` on a pinned pool connection
+//! - **`MySQL`** (`mysql`): session-level `GET_LOCK(name, 0)` on a pinned pool connection
+//! - **`SQLite` / fallback**: file marker locks under the OS cache dir
 //!
-//! ## Security policy
-//! This crate forbids plain SQL outside migration infrastructure. Therefore, locks
-//! are implemented **purely as file-based locks** (no DB-native advisory locks).
+//! ## Public API semantics
 //!
-//! Notes:
-//! - Prefer calling `guard.release().await` for deterministic unlock;
-//!   `Drop` provides best-effort cleanup only (may be skipped on runtime shutdown).
-//! - File-based locks use `create_new(true)` semantics and keep the file open,
-//!   then remove it on release. Consider using `fs2::FileExt::try_lock_exclusive()`
-//!   if you want kernel-level advisory locks across processes.
+//! - [`LockManager::lock`] — exactly one non-blocking acquisition attempt.
+//!   Contention → [`DbLockError::AlreadyHeld`]. On PG/MySQL, SQL/protocol errors map to
+//!   `DbLockError::Database`.
+//! - [`LockManager::try_lock`] — owns retry/backoff; returns `Ok(None)` when limits are exhausted.
+//!   `max_wait` bounds retry scheduling and time between **completed** acquisition attempts. It
+//!   does **not** currently impose a cancellation-safe timeout on pool acquisition or an
+//!   in-flight database query.
+//! - Native backends never use blocking `pg_advisory_lock` or `GET_LOCK` with a positive timeout.
+//!
+//! ## Single-session model (native backends)
+//!
+//! Each PG/MySQL [`LockManager`] owns a **dedicated pool sized `max=1`/`min=1`** with idle and
+//! lifetime reaping disabled. One database session therefore holds **all** advisory keys for the
+//! process. A held lock does **not** pin the connection: `pg_try_advisory_lock` /
+//! `GET_LOCK(_, 0)` return immediately and the session merely remembers the key, so the single
+//! connection is free to service the next lock/unlock op. Guards carry only a cheap
+//! `Arc<..Session>` handle plus the key — never a `PoolConnection`.
+//!
+//! The lock session is established **lazily**, on the first `.lock()` / `.try_lock()` call (see
+//! `LockSource`). A `DbHandle` that never takes an advisory lock opens no
+//! extra connection and spawns no keepalive task — important where many gears point at the same
+//! connection-limited database. The cost is that connection failures for the lock session surface
+//! from the first lock attempt (as `DbLockError::Database`) rather than at `connect()` time.
+//!
+//! ### Reconnect accounting (generation epoch)
+//!
+//! If the single connection dies, sqlx re-establishes a new backend and **all** advisory locks
+//! held by the old session are gone server-side. To avoid acting on stale "I hold key X"
+//! beliefs, the pool's `after_connect` callback bumps a shared [`AtomicU64`] generation on every
+//! newly established physical connection (initial + each reconnect). Each [`DbLockGuard`] records
+//! the generation observed right after its acquire; on release, if the current generation no
+//! longer matches, the lock was already released by the disconnect and the unlock SQL is skipped.
+//!
+//! ### Keepalive + unlock draining
+//!
+//! A background maintenance task keeps the session warm with a periodic `SELECT 1` (defeating
+//! proxy/PG idle-killers and bounding dead-session detection latency) and drains a pending-unlock
+//! queue. [`DbLockGuard`] `Drop` performs a non-blocking, runtime-free send of `(key, generation)`
+//! into that queue instead of running unlock SQL inline; the task performs the generation-checked
+//! `pg_advisory_unlock` / `RELEASE_LOCK`. This makes `Drop` panic-free even after the runtime has
+//! shut down and removes the previous `close_on_drop` / `mem::forget` connection-fate machinery.
+//! The task stops when the last [`LockManager`] / `DbHandle` clone drops the shared `Arc<..Session>`.
+//!
+//! Prefer awaiting [`DbLockGuard::release`] on the normal path for deterministic unlock; `Drop`
+//! is best-effort via the queue.
+//!
+//! ### In-process exclusivity (held-key registry)
+//!
+//! Session advisory locks are **re-entrant**: `pg_try_advisory_lock` / `GET_LOCK` grant a key the
+//! same session already holds and just bump the server's hold count. Since one session now serves
+//! the whole process, the database can no longer express in-process exclusivity on its own, so each
+//! session keeps a registry of claimed keys (see `HeldKeys`) that is consulted before the SQL runs.
+//!
+//! A claim is released only where the server is **known** to no longer hold the key: a confirmed
+//! unlock, a lost session, or a generation bump. When an unlock fails on a live session the key
+//! stays claimed, because PG still holds it and re-acquiring would raise the hold count to two —
+//! a leak no release could balance. Claims are recorded per generation, so such a claim is not
+//! permanent: the next reconnect drops every server-side lock and frees the key.
+//!
+//! One consequence for native backends: after `Drop`, re-acquiring the same key in the same process
+//! fails until the maintenance task has drained that unlock. Await [`DbLockGuard::release`] when the
+//! key is re-acquired immediately.
+//!
+//! **Connection pooling caveat:** session advisory locks require a real session. A
+//! transaction-pooling proxy (e.g. `PgBouncer` in `transaction` mode) breaks them entirely — use
+//! session pooling or a direct connection for the lock pool.
+//!
+//! ### File-marker backend limits
+//!
+//! The SQLite/fallback backend uses an exclusive create of a marker file (not `fs2` kernel
+//! locks). Process termination or cancellation during filesystem acquisition may leave a stale
+//! marker — the file backend does not provide kernel-owned lock cleanup.
+//!
+//! After `open(...).await` returns successfully, ownership is transferred to the guard without
+//! another await point. Cancellation while the filesystem open itself is in flight may still
+//! leave a marker, depending on Tokio blocking-filesystem cancellation behavior.
+//!
+//! Implicit [`DbLockGuard`] Drop removes the marker **synchronously** (file cleanup does not
+//! depend on a spawned Tokio task). Explicit [`DbLockGuard::release`] for the file backend is
+//! likewise synchronous after taking ownership.
+//!
+//! #### Recovering a stale marker
+//!
+//! There is **no automatic reclaim**: a marker left by `SIGKILL`, a power loss, or cancellation
+//! mid-`open` keeps the key un-acquirable until it is deleted. This is deliberate — the two
+//! automatic policies on offer are both unsound here. A TTL cannot be chosen safely because the
+//! library does not know how long a legitimate holder may hold a key, and PID liveness checks
+//! misfire once PIDs are reused; either one can reclaim a *live* lock, which is a worse failure
+//! than a stuck one. The file backend is only ever selected for `SQLite`, so the participants are
+//! processes on one machine sharing one database file — a human (or the service's own start-up
+//! script) can see the whole picture that a TTL cannot.
+//!
+//! The operational fallback is to remove the marker while no holder is running:
+//!
+//! ```text
+//! <cache_dir>/cf-gears/locks/{database_scope:016x}/{xxh3_64(canonical_lock_input):016x}.lock
+//! ```
+//!
+//! `<cache_dir>` is `dirs::cache_dir()` (falling back to the temp dir), e.g.
+//! `%LOCALAPPDATA%` on Windows and `~/.cache` on Linux. Deleting the whole
+//! `cf-gears/locks/{database_scope:016x}` directory clears every key for one database and is the
+//! usual recovery step; deleting `cf-gears/locks` clears all of them. Doing this while a holder
+//! is alive drops that holder's mutual exclusion without telling it, so scope the deletion to a
+//! window where the participating processes are down.
+//!
+//! If crash recovery ever needs to be automatic, the sound mechanism is kernel-owned locks
+//! (`flock` / `LockFileEx`), which the OS releases on process death — not a TTL.
+//!
+//! ## Semver note
+//!
+//! This module introduces a **breaking** public API change versus prior `0.8.x` releases:
+//! [`DbLockGuard::release`] now returns [`Result`], and `DbLockError` (including
+//! `Database(sqlx::Error)` when `pg`/`mysql` features are enabled) is publicly re-exported.
+//! Additionally, the public `lock_keepalive: Option<Duration>` field was added to both
+//! [`crate::ConnectOpts`] and [`crate::config::DbConnConfig`]; neither type is `#[non_exhaustive]`,
+//! so this is source-breaking for callers that build them with an exhaustive struct literal
+//! (config deserialization stays compatible via `#[serde(default)]`). Bump the crate major/minor
+//! appropriately when publishing (intended: `0.9.0`). Local path/patch validation against consumers
+//! pinned to `0.8.4` may keep the package version at `0.8.4` until a coordinated Gears release.
+//!
+//! ## Stable lock namespace
+//!
+//! ```text
+//! canonical_lock_input =
+//!   "cf-gears-toolkit-db:v2:{database_scope:016x}:g{gear_utf8_len}:{gear}:k{key_utf8_len}:{key}"
+//! ```
+//!
+//! Length prefixes are UTF-8 byte lengths so `gear`/`key` values that contain `:` cannot collide
+//! (e.g. `("a:b","c")` ≠ `("a","b:c")`). The `v2` prefix replaces the ambiguous `v1` colon-joined
+//! encoding.
+//!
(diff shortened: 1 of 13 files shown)


libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -60,102 +167,1035 @@ impl Default for LockConfig {
...
+        | sqlx::Error::Tls(_)
+        | sqlx::Error::PoolClosed
+        | sqlx::Error::PoolTimedOut => true,
+        // PostgreSQL: connection_exception class `08*`; admin shutdown / crash `57P01/02/03`.
+        sqlx::Error::Database(db) => db.code().as_deref().is_some_and(|code| {
+            code.starts_with("08") || matches!(code, "57P01" | "57P02" | "57P03")
+        }),
+        _ => false,
+    }
+}
```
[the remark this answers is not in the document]
MikeFalcon77: Fixed in `d3437f8b`.

`PoolTimedOut` is out of the connection-lost arm; `Io`, `Tls`, `PoolClosed` and the `08*` / `57P01..03` SQLSTATE checks are unchanged. `is_connection_lost_classification` now asserts `PoolTimedOut` is **not** connection-lost, and the reason is recorded above the predicate so it does not get re-added: with `max_connections(1)` a timeout fires routinely while the session is alive and still holding the key, so treating it as released would skip the unlock and report success.

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -269,55 +1296,127 @@ impl LockManager {
...
+            Err(error) => {
+                tracing::warn!(
+                    backend = "mysql",
+                    key_fingerprint = %fingerprint,
+                    %error,
+                    "GET_LOCK failed"
+                );
+                Err(DbLockError::Database(error))
+            }
+        }
```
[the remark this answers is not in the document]
MikeFalcon77: Fixed in `d3437f8b`. This one was real and reproducible — `postgres_single_session_model` was failing on `assert!(contended.is_none())` against a real Postgres before the change.

Added a per-session held-key registry: `HeldKeys<K>`, a `Mutex<HashSet<K>>` on `PgSession` (keyed by `i64` lock key) and `MySqlSession` (keyed by lock name). `try_lock_postgres` / `try_lock_mysql` claim the key before running the SQL and unclaim on every path that returns no guard, so a second acquire of a held key returns `None` — which `lock()` maps to `AlreadyHeld` and `try_lock()` folds into `Ok(None)`, same as the file backend. `HashSet::insert` under the mutex *is* the claim and there is no `await` between it and the SQL, so two tasks racing the same key cannot both reach the database. Distinct keys stay independent.

One deliberate deviation from the suggestion: the claim is dropped when the guard relinquishes ownership (`release_pg` / `release_mysql` and the `Drop` enqueue path), not after the queued unlock completes. Waiting for the unlock would deadlock the key — on the `Drop` path the unlock is skipped entirely when the session has reconnected (generation mismatch), so the key would stay claimed forever. Unclaiming early is safe because PG and MySQL 5.7.5+ count advisory locks: every acquire has exactly one matching unlock, so every interleaving ends at zero. It also means the server-side count never dips to zero mid-handoff, so another process cannot steal the key in the gap, while in-process exclusivity is held by the registry.

Validated: `postgres_single_session_model` and `mysql_single_session_model` both pass against real containers, plus unit coverage in `held_keys_claim_is_exclusive_until_unclaimed` and `held_keys_unclaims_string_keys_by_borrow`.

libs/toolkit-db/src/lib.rs
```diff
@@ -306,6 +316,7 @@ pub(crate) struct DbHandle {
     engine: DbEngine,
     dsn: String,
     sea: DatabaseConnection,
+    locks: advisory_locks::LockManager,
```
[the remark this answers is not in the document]
MikeFalcon77: Same defect as the thread on `advisory_locks.rs` (shared session makes native locks re-entrant) — fixed by the same change in `d3437f8b`.

`HeldKeys` on `PgSession` / `MySqlSession` is consulted before `pg_try_advisory_lock` / `GET_LOCK`; already-held keys are rejected and surface as `AlreadyHeld` from `lock()`, distinct keys stay independent, and failed acquisitions are unclaimed so nothing leaks into the registry. Details and the one deviation on unclaim timing are in the other thread.

libs/toolkit-db/src/options.rs
```diff
@@ -85,13 +95,17 @@ impl DbConnectOptions {
 
                 let sqlx_pool = pool_opts.connect_with(opts.clone()).await?;
 
+                let filename = opts.get_filename().display().to_string();
+                let dsn = format!("sqlite://{filename}");
+                let database_scope = crate::advisory_locks::database_scope_from_dsn(&dsn);
+                let locks = crate::advisory_locks::LockManager::file(database_scope);
```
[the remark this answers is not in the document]
MikeFalcon77: Not changing this — following it would introduce the split it is meant to prevent.

Query parameters are stripped in `parse_sqlite_path_from_dsn` before `.filename()`, so they never reach `connect_with(opts)`: `sqlite:app.db?mode=rwc` and `sqlite:app.db` open the same file. If the scope encoded those parameters, the two would hash to different lock markers and two processes on one database file would stop excluding each other.

For file-backed databases the two entry points already agree — `stable_file_path_identity` canonicalizes the path whichever way the DSN was written. The only remaining difference is private in-memory databases, and there separate scopes are correct: those are genuinely separate databases, and collapsing their identities would make unrelated in-memory databases contend for one lock namespace.

Added `sqlite_scope_ignores_dsn_query_parameters` to pin the contract this rests on. If parameters ever start reaching `connect` (real `mode=memory&cache=shared` support), that test fails and the scope will have to encode them — which is the point at which this comment becomes valid.

libs/toolkit-db/src/lib.rs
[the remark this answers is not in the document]
modcrafter77: Agreed. Fixed.

A claim is now released only where the server is known to no longer hold the key: confirmed unlock, lost session, or generation bump. A failed unlock on a live session keeps the key claimed. For the Drop path the claim moved owner - enqueue_native_unlock no longer unclaims, drain_pg_unlock / drain_mysql_unlock do it after the unlock is confirmed; enqueue still unclaims only where no drain will run (generation mismatch, or queue closed during teardown).

Two deviations worth flagging:

Claims are scoped to the session generation, otherwise a retained claim would never expire and a failed unlock would block the key in-process forever. On reconnect the old connection's locks are gone server-side, so an older-generation claim is stale and gets taken over. Test: held_keys_claim_from_a_dead_generation_is_stale.
RELEASE_LOCK → 0/NULL and pg_advisory_unlock → false still unclaim. Those mean this session is not the holder, so there is no stale hold of ours to protect; retaining would only block our own process. The invariant is "unclaim iff the server is known not to hold it for us".
Documented consequence: after Drop, re-acquiring the same key in-process fails until the queued unlock drains — await release() if you re-acquire immediately.

MikeFalcon77 (approved):

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -68,102 +219,1181 @@ impl Default for LockConfig {
...
+async fn release_pg(
+    session: &PgSession,
+    lock_key: i64,
+    generation: u64,
+    key_fingerprint: &str,
+) -> Result<(), DbLockError> {
+    // The in-process claim is given up only where PG is known to no longer hold the key. If the
+    // unlock cannot be confirmed on a live session the key stays claimed — see [`HeldKeys`].
+    if session.generation.load(Ordering::Acquire) != generation {
+        session.held_keys.unclaim(&lock_key);
```
Artifizer: **[MAJOR]**

Stale-generation cleanup removes a key from `HeldKeys` without verifying which generation currently owns its registry entry. After a reconnect, `HeldKeys::claim` permits a new guard to replace the old generation's claim. If the old guard is subsequently released or dropped, its unconditional `unclaim` (in `release_pg`/`release_mysql`'s "session reconnected, lock already gone" branch, `drain_pg_unlock`/`drain_mysql_unlock`'s generation-mismatch branches, and `enqueue_native_unlock`'s generation-mismatch branch) removes the new guard's claim regardless of which generation actually holds it. A third acquisition on the shared session can then succeed re-entrantly while the second (newer-generation) guard is still alive, producing multiple live guards for the same key and breaking mutual exclusion. The same race exists in queued unlock draining for both PostgreSQL and MySQL.

**Suggested fix:** Make removal generation-aware, e.g. `unclaim_if_generation(key, generation)`, and remove the entry only when its stored generation equals the caller's generation. Use that operation throughout guard release, drop enqueue, and pending-unlock draining. Add a test that claims a key in generation 1, replaces it in generation 2, attempts cleanup for generation 1, and verifies generation 2 remains claimed.

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -277,55 +1495,155 @@ impl LockManager {
...
+        // by another process.
+        if !session.held_keys.claim(lock_key, generation) {
+            return Ok(None);
+        }
+
+        let result: Result<bool, sqlx::Error> =
+            sqlx::query_scalar("SELECT pg_try_advisory_lock($1)")
+                .bind(lock_key)
+                .fetch_one(&mut *conn)
+                .await;
```
Artifizer: **[MAJOR]**

`try_lock` is documented as cancellation-safe, but native acquisition creates the in-process claim (`held_keys.claim(...)`) before awaiting the database query, and creates the cleanup-capable `DbLockGuard` only after that await completes. If the future is cancelled at the query await, the claim is never removed. Worse, the server may already have acquired the session lock even though no guard is returned, leaving the key stranded until the session reconnects. The MySQL path has the same ownership gap around `GET_LOCK`.

**Suggested fix:** Introduce an acquisition cleanup token immediately after claiming the key. Its `Drop` should queue a generation-checked unlock, and it should be disarmed only after ownership is transferred into the returned `DbLockGuard`. This covers cancellation both before and after the server processes the query. Add cancellation tests for both native backends.

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -68,102 +219,1181 @@ impl Default for LockConfig {
...
     }
 
-    /// Deterministically release the lock (preferred).
-    pub async fn release(mut self) {
+    /// Deterministically release the lock (preferred path).
+    ///
+    /// # Errors
+    /// Returns [`DbLockError`] if unlock fails or the lock was not held.
+    pub async fn release(mut self) -> Result<(), DbLockError> {
         if let Some(inner) = self.inner.take() {
```
Artifizer: **[MAJOR]**

`DbLockGuard::release` removes `inner` from the guard before awaiting native unlock work. If the release future is cancelled while acquiring the session connection or executing the unlock query, the owned `GuardInner` is simply dropped. `GuardInner` has no cleanup implementation, and the now-empty `DbLockGuard` cannot enqueue cleanup, so the server lock and held-key claim can remain indefinitely. This is especially easy to trigger when callers wrap deterministic release in a shutdown timeout or `tokio::select!`.

**Suggested fix:** Keep a drop-enabled cleanup token armed across every await in `release_inner`, enqueueing the native unlock if the explicit release future is cancelled. Disarm it only after a confirmed unlock or confirmed session loss. Alternatively, give native guard ownership a dedicated RAII type whose `Drop` always enqueues cleanup, regardless of whether it resides in `DbLockGuard` or an in-flight release future.

libs/toolkit-db/src/lib.rs
```diff
@@ -341,16 +352,28 @@ impl DbHandle {
...
+        let lock_keepalive = opts
+            .lock_keepalive
+            .unwrap_or(advisory_locks::DEFAULT_LOCK_KEEPALIVE);
         match engine {
             #[cfg(feature = "pg")]
             DbEngine::Postgres => {
                 let o = PgPoolOptions::new().apply(&opts);
                 let pool = o.connect(dsn).await?;
+                let database_scope = advisory_locks::database_scope_from_dsn(dsn);
+                let locks = advisory_locks::LockManager::postgres_session_dsn(
```
Artifizer: **[MAJOR]**

`DbHandle::connect` (and identically `DbConnectOptions::connect` in `libs/toolkit-db/src/options.rs:114-143`) now unconditionally opens a **second, dedicated** PG/MySQL connection and spawns a perpetual keepalive/maintenance task (`spawn_pg_maintenance` / `spawn_mysql_maintenance`, default ping every `DEFAULT_LOCK_KEEPALIVE = 5s`) for every single Postgres/MySQL `DbHandle`, whether or not that handle's owner ever calls `.lock()` / `.try_lock()`. This is a real behavioral regression versus the code this PR replaces: the prior implementation used purely file-based locking for *all* backends, so no extra DB connection was ever opened. Now every gear wired through `DbManager` (`libs/toolkit-db/src/manager.rs:140` → `build_db_handle` → `DbConnectOptions::connect`) permanently reserves one extra connection from the target database for the lifetime of the process, purely on the chance that advisory locking might be used later. In a deployment with many gears pointed at a connection-limited managed Postgres/MySQL instance, this multiplies the connection footprint by "+1 per gear" and adds constant background chatter (a `SELECT 1` every 5s per gear by default) for a feature most gears will never invoke. There is also no way to opt out: `DbConnConfig`/`ConnectOpts` only expose `lock_keepalive` to tune the ping interval, not to skip creating the lock session entirely.

**Suggested fix:** Make the dedicated lock session lazy: only establish the PG/MySQL lock connection and spawn its maintenance task the first time `.lock()` / `.try_lock()` is actually called on a `DbHandle` (e.g. via `OnceCell`/`tokio::sync::OnceCell` around `LockManager`), rather than eagerly in `connect()`. If eager creation is intentional for latency reasons, at minimum add a config knob (e.g. `lock_native: bool` defaulting to `false`, or reuse of the existing pool instead of a dedicated one) so operators on connection-constrained databases aren't forced to pay for a feature they don't use, and call this cost out explicitly in the module's semver/behavior notes.

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -277,55 +1495,155 @@ impl LockManager {
...
+        let mut conn = session.pool.acquire().await?;
+        // Read AFTER acquire so a reconnect during acquire is reflected in the recorded generation.
+        let generation = session.generation.load(Ordering::Acquire);
+        let lock_key = stable_lock_key(canonical);
+        let fingerprint = key_fingerprint(canonical);
+
+        // `pg_try_advisory_lock` is re-entrant on the session every guard shares, so in-process
+        // exclusivity is decided here, before the SQL. Contended → `None`, exactly like a key held
+        // by another process.
+        if !session.held_keys.claim(lock_key, generation) {
```
Artifizer: **[CRITICAL]`

`try_lock_postgres` and `try_lock_mysql` register a `HeldKeys` claim (`session.held_keys.claim(...)` at line 1541 / 1594) *before* awaiting the actual `pg_try_advisory_lock` / `GET_LOCK` query (the `fetch_one(...).await` at lines 1548-1549 / 1601-1602), and only release that claim from the code that runs *after* the query resolves (`Ok(false)`/`Err` arms, or via the `DbLockGuard` that owns it on `Ok(true)`). If the enclosing future is dropped while suspended at that `.await` — which happens under perfectly ordinary usage such as `tokio::time::timeout(...)` or the `tokio::select!` cooperative-shutdown pattern the crate's own `try_lock` doc comment (lines 1391-1396) recommends callers use — execution never reaches the `match result { ... }` block. The claim taken at line 1541/1594 is never released, and no `DbLockGuard` was ever constructed to release it via `Drop`/`release()` either. The key becomes permanently unacquirable in this process (for this session generation) even though no lock is actually held by any live guard — a self-inflicted, undetectable lockout that only clears on the next reconnect (a session generation bump), which may never happen for a healthy connection. Because the doc comment at lines 1384-1396 explicitly asserts "This future is cancellation-safe. Dropping it mid-retry will not leak a lock," this is also a correctness contract violation, not just a theoretical race — it is meaningfully different from the generation-unaware `unclaim` issue already flagged above (which concerns overwriting a *newer* claim, not orphaning a claim with no owner at all).

**Suggested fix:** Introduce a small RAII claim guard that owns the `HeldKeys` entry from the moment `claim()` succeeds and unclaims on `Drop` unless explicitly disarmed:

```rust
struct ClaimGuard<'a, K: Eq + std::hash::Hash + Clone> {
    held_keys: &'a HeldKeys<K>,
    key: Option<K>,
}
impl<K: Eq + std::hash::Hash + Clone> Drop for ClaimGuard<'_, K> {
    fn drop(&mut self) {
        if let Some(k) = self.key.take() {
            self.held_keys.unclaim(&k);
        }
    }
}
```

Acquire the claim through this guard immediately after `claim()` returns `true`, run the query, and only call `guard.key.take()` (disarm, without unclaiming) once a `DbLockGuard` is actually being constructed to take over ownership. Because Rust drops all live locals — including this guard — when the containing future is cancelled at any `.await` point, this makes claim release unconditional on cancellation, matching the crate's documented guarantee. Add a test that starts `try_lock_postgres`/`try_lock_mysql`, cancels it via `tokio::time::timeout` while the query is in flight (e.g. against a slow/blocked connection), and asserts the key is claimable again afterward without waiting for a reconnect.

libs/toolkit-db/src/advisory_locks.rs
MikeFalcon77: Fixed.  Added `HeldKeys::unclaim_if_generation(key, generation)` and switched every native cleanup path to it: guard-drop enqueue, queued-unlock draining, and the generation-mismatch branches of release for both PG and MySQL. It removes the entry only when the stored generation still equals the caller's, so a stale generation can no longer evict a newer generation's live claim. Covered by the new `unclaim_if_generation_preserves_a_newer_claim` test (claim gen1 -> replace gen2 -> stale cleanup gen1 -> gen2 still held).

libs/toolkit-db/src/advisory_locks.rs
MikeFalcon77: Fixed. The claim taken before the `pg_try_advisory_lock` / `GET_LOCK` await is now owned by a `ClaimGuard` whose Drop calls `unclaim_if_generation`. If the future is cancelled at that await, the guard drops and the claim is released; it is only `disarm()`ed once a `DbLockGuard` takes ownership on success. This matches the documented cancellation-safety contract. SQL-level cancellation tests need a live DB (those sit in the ignored integration set); the guard mechanism itself is covered by `claim_guard_retracts_on_drop_unless_disarmed` and `claim_guard_drop_does_not_evict_a_newer_generation`.

libs/toolkit-db/src/advisory_locks.rs
MikeFalcon77: Fixed, same root cause as the CRITICAL below. Both `try_lock_postgres` and `try_lock_mysql` now wrap the pre-await claim in a `ClaimGuard` that retracts on Drop unless a `DbLockGuard` takes over, so cancelling at the query await no longer strands the key.

libs/toolkit-db/src/advisory_locks.rs
MikeFalcon77: Fixed. `release()` now keeps `inner` inside a `ReleaseCleanup` token held across the unlock awaits. On normal completion it disarms the token; if the release future is cancelled mid-await, the token's Drop runs the same runtime-free `enqueue_native_unlock` that `DbLockGuard::drop` uses, so the server lock and the claim are no longer dropped without cleanup.

libs/toolkit-db/src/lib.rs
MikeFalcon77: Fixed. The PG/MySQL lock session and its keepalive task are now established lazily on the first `.lock()`/`.try_lock()` via a `tokio::sync::OnceCell` behind an `Arc` shared across `LockManager` clones (`PgLockSource`/`MySqlLockSource`). A handle that never locks opens no extra connection and runs no keepalive ping. The constructors became sync and were renamed `postgres_lazy(_dsn)` / `mysql_lazy(_dsn)`. Trade-off: lock-session connect errors now surface from the first lock attempt instead of `connect()`, which is called out in the module docs. Went with lazy rather than a config flag, since after lazy init an opt-out flag adds little.

libs/toolkit-db/src/advisory_locks.rs
```diff
@@ -277,55 +1692,156 @@ impl LockManager {
...
+
+    /// Single non-blocking PG acquire on the shared session. A held lock does not pin the
+    /// connection — it returns to the pool immediately while the session remembers the key.
+    #[cfg(feature = "pg")]
+    async fn try_lock_postgres(
+        session: &Arc<PgSession>,
+        display_key: &str,
+        canonical: &str,
+    ) -> Result<Option<DbLockGuard>, DbLockError> {
+        let mut conn = session.pool.acquire().await?;
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

PR #4249, opened 2026-07-23
fix(resource-group): accept caller-supplied id on REST create-group

ffedoroff:
## What

Expose an optional caller-supplied `id` on the REST create-group endpoint, matching the SDK contract.

The SDK `CreateGroupRequest` already carries an optional `id`, but the REST `CreateGroupDto` dropped it and the `From` conversion hard-coded `id: None`. This PR adds `id` to the DTO and passes it through.

## Why

Two problems with the previous divergence:

1. **Silent data loss over REST.** Any SDK call with `id: Some(..)` routed over a REST transport would have the field silently discarded, because the DTO had no `id`. Keeping SDK ⊇ DTO avoids this.
2. **Blocks the platform-sync use case.** For tenant-typed groups `tenant_id == group.id` (`group_service.rs`: `effective_tenant_id = if is_tenant_type { group_id } else { tenant_id }`). To align an RG tenant scope with an existing platform tenant identifier, the caller must be able to supply the id at create time — impossible over REST until now.

## Changes

- Add `CreateGroupDto.id: Option<Uuid>` (`skip_serializing_if = "Option::is_none"`).
- Pass `dto.id` through in `From<CreateGroupDto> for CreateGroupRequest` (was hard-coded `None`).
- Tests: id passthrough through the conversion, omitted-id → `None`, deserialize-without-id → `None`.

Omitting `id` preserves the previous behaviour: the server generates a UUID.

## Notes

- The hand-written design doc `docs/openapi.yaml` already documented this optional `id`; the generated `docs/api/api.json` was the stale artifact and is regenerated in this PR (`make openapi`) — so SDK, DTO and the published spec now agree.
- Security: for tenant-typed groups a caller-supplied `id` sets the tenant scope, so exposure should stay gated to privileged/system callers (platform sync), not arbitrary end-users.

## Test plan

- [x] `cargo check -p cf-gears-resource-group --tests`
- [x] `cargo test -p cf-gears-resource-group --lib dto` — 9 passed
- [x] regenerate `docs/api/api.json` (`make openapi`) — clean 8-line diff, only the new `id` property

## Summary by CodeRabbit

* **New Features**
  * Group creation requests can now include an optional caller-provided ID.
  * If the ID field is omitted, the server will generate a UUID automatically.
* **Documentation**
  * Updated the API schema to clarify the optional ID behavior for group creation.
* **Tests**
  * Added/extended DTO tests to verify ID deserialization and correct passthrough when converting the request.

### DIFF
--- a/docs/api/api.json
+++ b/docs/api/api.json
@@ -681,6 +681,14 @@
       "CreateGroupDto": {
         "description": "REST DTO for creating a new resource group.",
         "properties": {
+          "id": {
+            "description": "Optional caller-supplied ID. If omitted, the server generates a UUID.\n\nFor tenant-typed groups the ID becomes the group's tenant scope\n(`tenant_id == group.id`), so a platform sync can align the RG tenant\nscope with an existing tenant identifier by supplying it here.",
+            "format": "uuid",
+            "type": [
+              "string",
+              "null"
+            ]
+          },
           "metadata": {
             "description": "Type-specific metadata."
           },
--- a/docs/api/api.json
+++ b/docs/api/api.json
@@ -682,7 +682,7 @@
         "description": "REST DTO for creating a new resource group.",
         "properties": {
           "id": {
-            "description": "Optional caller-supplied ID. If omitted, the server generates a UUID.\n\nFor tenant-typed groups the ID becomes the group's tenant scope\n(`tenant_id == group.id`), so a platform sync can align the RG tenant\nscope with an existing tenant identifier by supplying it here.",
+            "description": "Optional caller-supplied ID. If omitted, the server generates a UUID.",
             "format": "uuid",
             "type": [
               "string",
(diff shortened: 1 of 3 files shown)


docs/api/api.json
```diff
@@ -681,6 +681,14 @@
       "CreateGroupDto": {
         "description": "REST DTO for creating a new resource group.",
         "properties": {
+          "id": {
+            "description": "Optional caller-supplied ID. If omitted, the server generates a UUID.\n\nFor tenant-typed groups the ID becomes the group's tenant scope\n(`tenant_id == group.id`), so a platform sync can align the RG tenant\nscope with an existing tenant identifier by supplying it here.",
```
nonameffh: I think this line is not neccessary here: 
> For tenant-typed groups the ID becomes the group's tenant scope\n(`tenant_id == group.id`), so a platform sync can align the RG tenant\nscope with an existing tenant identifier by supplying it here.

nonameffh (approved):

docs/api/api.json
ffedoroff: fixed

aviator5 (approved):

PR #4262, opened 2026-07-24
feat(sql cte): Proposal for safe CTE implementation is added

MikeFalcon77:
## Summary by CodeRabbit

* **Documentation**
  * Added an Architectural Decision Record outlining how to support Common Table Expressions in the Secure ORM while preserving tenant isolation.
  * Defined safe CTE construction rules, execution behavior, and semantic constraints to keep query scope consistent.
  * Documented allowed/forbidden CTE usage levels, including restrictions on raw `WITH` usage and notes on future controlled escape mechanisms.
  * Clarified that `WITH RECURSIVE` is not supported for hierarchy traversal; recommended alternative modeling instead.

### DIFF
--- a/docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
+++ b/docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
@@ -1,0 +1,381 @@
+---
+status: proposed
+date: 2026-07-24
+decision-makers: Constructor Fabric steering committee
+---
+
+# ADR-0001: Safe CTE Support in the Secure ORM
+
+**ID**: `cpt-cf-adr-secure-cte-policy`
+
+## Context and Problem Statement
+
+The Secure ORM (`libs/toolkit-db/src/secure/`) enforces multi-tenant isolation
+through a single load-bearing invariant: **every access to a table passes through a
+scope condition**. This is guaranteed mechanically, not by convention:
+
+- A typestate transition `Unscoped → Scoped` — a query physically cannot be executed
+  until `.scope_with()` is called
+  ([select.rs:151-188](../../../../libs/toolkit-db/src/secure/select.rs#L151-L188)).
+  The `Scoped` marker carries an `Arc<AccessScope>` so related-entity queries inherit
+  the same scope
+  ([select.rs:22-25](../../../../libs/toolkit-db/src/secure/select.rs#L22-L25)).
+- `build_scope_condition::<E>()` attaches a `WHERE` for the concrete entity `E` via
+  `E::resolve_property()`
+  ([cond.rs:54-83](../../../../libs/toolkit-db/src/secure/cond.rs#L54-L83)).
+
+A Common Table Expression (`WITH x AS (SELECT ... FROM sensitive_table) ...`) breaks
+this invariant. The body of a CTE is an **independent `SELECT` over arbitrary tables**
+to which the outer query's scope `WHERE` does **not** apply. If a gear could build a
+CTE freely, the scope filter would land on the outer entity while the tables read
+inside `WITH` stay unfiltered — a direct tenant-isolation hole.
+
+The naive workaround — "expose `into_inner()` and assemble the CTE by hand on
+sea_query"
+([select.rs:415-418](../../../../libs/toolkit-db/src/secure/select.rs#L415-L418)) —
+also violates the platform guardrails. The rule **"No plain SQL in
+handlers/services/repos. Raw SQL is allowed only in migration infrastructure"** is
+explicit and repeated in
+[11_database_patterns.md:9](../../../toolkit_unified_system/11_database_patterns.md#L9)
+(and again at lines 161 and 167). Raw SQL escapes both the typestate guarantee and
+this rule.
+
+There is real, tracked demand for capabilities that today would reach for a CTE:
+
+- `account-management/.../tr_plugin/queries.rs:26-39` builds a parent map in memory and
+  walks it pre-order on the client, explicitly deferring the SQL recursive CTE "as a
+  follow-up once `toolkit-db` exposes a safe raw-SQL hook".
+- `account-management/.../repo_impl/conversion.rs:1221` recomputes barriers per row and
+  notes "Replace with self-join on `tenant_closure` once the recursive-CTE work lands".
+
+**Scope of this ADR.** This policy governs **standard user gears** — the
+handlers / services / repos that build queries against their own domain tables. It
+does **not** govern `toolkit-db` itself. The one existing CTE in the crate is the
+outbox writer
+([outbox/store.rs:264-273](../../../../libs/toolkit-db/src/outbox/store.rs#L264-L273)):
+that is raw dialect-specific SQL living **inside the system library that implements the
+Secure ORM**, where sea_query and hand-written SQL are the implementation substrate.
+That is correct and out of scope here — it is not user-gear code and is not a precedent
+for user-gear code.
+
+Today there is **no CTE support in the Secure ORM's user-facing API** and no policy
+governing it for user gears. This ADR settles the policy and the API shape **before**
+ad-hoc solutions accrete in gears.
+
+## Decision Drivers
+
+- **Preserve the typestate invariant** — a CTE must be as impossible to construct
+  unscoped as an `Unscoped` select is impossible to execute.
+- **No raw SQL in user-gear code** — honor
+  [11_database_patterns.md:9](../../../toolkit_unified_system/11_database_patterns.md#L9);
+  in gears everything goes through the Secure ORM / sea_query builder (the `toolkit-db`
+  system library that implements it is a separate layer).
+- **Reuse what exists** — `build_scope_condition` already emits nested subqueries via
+  `sea_query::Query::select()`
+  ([cond.rs:104-185](../../../../libs/toolkit-db/src/secure/cond.rs#L104-L185)); the
+  "scope as a sea_query expression" mechanic is proven.
+- **Do not regress hierarchy handling** — hierarchy traversal already works via
+  materialized closure tables (`tenant_closure`, `resource_group_closure`); do not
+  reintroduce recursion the isolation model cannot cover.
+- **Reviewability** — any escape from the safe path must be visible at the API surface
+  so a reviewer sees that isolation was considered.
+
+## Considered Options
+
+- **Option A** — `SecureCte`, constructible only from `SecureSelect<E, Scoped>`; scope
+  is embedded inside each CTE body.
+- **Option B** — Controlled escape-hatch: a CTE over a non-`Scopable` source with a
+  mandatory `.scope_via_exists::<J>()` join predicate to a scoped entity.
+- **Option C** — Raw `WITH` from a string in gear code.
+- **Option D** — `WITH RECURSIVE` for hierarchy traversal.
+
+## Decision Outcome
+
+Chosen option: **Option A**, because it is the only option that keeps the
+tenant-isolation invariant a **compile-time** guarantee while using the sea_query
+builder exclusively. The principle: **do not apply scope outside the CTE — embed scope
+inside the body of every CTE.** Then any table a CTE touches is already filtered.
+
+### API shape
+
+```rust
+// 1. A CTE body can be built ONLY from an already-scoped query.
+//    scope is already baked into the body via build_scope_condition.
+//    The name is `&'static str` (see "CTE name"); injection safety comes from
+//    sea_query identifier quoting, not from the lifetime.
+impl<E: ScopableEntity + EntityTrait> SecureSelect<E, Scoped> {
+    /// Turn a scoped query into a named CTE. Retains the query's `Arc<AccessScope>`
+    /// so `with_ctes` can enforce that every CTE shares the outer query's scope.
+    pub fn into_cte(self, name: &'static str) -> SecureCte { /* ... */ }
+}
+
+// 2. The outer query accepts CTEs and returns a DISTINCT type — NOT `Self`.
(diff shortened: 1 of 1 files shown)


docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,285 @@
...
+// 1. A CTE body can be built ONLY from an already-scoped query.
+//    scope is already baked into the body via build_scope_condition.
+impl<E: EntityTrait> SecureSelect<E, Scoped> {
+    /// Turn a scoped query into a named CTE.
+    pub fn into_cte(self, name: &str) -> SecureCte { /* ... */ }
+}
+
+// 2. The outer query — also SecureSelect<Root, Scoped> — accepts CTEs.
+impl<E: EntityTrait> SecureSelect<E, Scoped> {
+    pub fn with_ctes(self, ctes: impl IntoIterator<Item = SecureCte>) -> Self { /* ... */ }
```
Artifizer: **HIGH**

`with_ctes` returns `Self`, i.e. a `SecureSelect<E, Scoped>` that is still backed internally by `sea_orm::Select<E>` (see `select.rs:62-65`). But the "Feasibility constraint" section right below (lines 138-141) says the opposite: this value must be executed through a *separate* path (`find_by_statement`/`FromQueryResult`), not through `Select<E>::all()`, because `sea_orm::Select<E>` has no `.with()` method.

Why it matters: if `with_ctes` really returns `SecureSelect<E, Scoped>`, calling `.all()`/`.one()`/`.count()` on it runs `self.inner.all(db)` on the plain `Select<E>`, which never saw the `WithClause`. The CTE silently disappears from the executed query instead of raising a compile error — the worst kind of bug for a security-sensitive feature. This is the one thing in the ADR that will trip up whoever implements Option A first.

Suggested fix: give `with_ctes` its own return type, e.g. `SecureCteSelect<E, Scoped>`, wrapping a `sea_query::SelectStatement` + `WithClause` instead of `sea_orm::Select<E>`, with its own `.all()/.one()` going through `find_by_statement`. Update this code sketch to show that type instead of `Self` so the API shape and the feasibility constraint actually agree before implementation starts.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,285 @@
...
+- Requires a CTE-aware execution path that bypasses `Select<E>` (via
+  `find_by_statement`/`FromQueryResult`), diverging from the existing `.all()` path.
+- The CTE body is scoped, but the outer query cannot post-filter the CTE's contents —
+  the scope must be correct at body-build time.
+
+**Risks:**
+
+- The correlation between a CTE and the outer query (e.g. joining the CTE to the outer
+  entity on the right tenant key) is **not** compiler-verified; misuse can still under-
+  or over-select rows. Mitigate with tests and reviewer guidance.
```
Artifizer: **MEDIUM**

Worth spelling out explicitly here: because every `SecureCte` bakes its own scope into its own body (via `build_scope_condition`), a wrong join predicate between the CTE and the outer query can produce wrong *results* (missing or duplicate rows) — but it cannot leak another tenant's data, since the CTE body never contained that data to begin with. That's a materially different (and better) guarantee than a typical "unscoped join = isolation hole" bug, and it's the main reason Option A is safer than it might look at first glance.

Why it matters: as written, "misuse can still under- or over-select rows" reads like a security risk, when it's really a correctness risk. A reviewer implementing this needs to know which one it is to judge whether "tests and reviewer guidance" is an adequate mitigation.

Suggested fix: add one sentence confirming the isolation boundary is per-CTE-body and independent of outer-join correctness. If there's any scenario where that's *not* true (e.g. a future change lets CTEs from different scopes be combined), call that out as a separate, real risk instead.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,285 @@
...
+inside the body of every CTE.** Then any table a CTE touches is already filtered.
+
+### API shape
+
+```rust
+// 1. A CTE body can be built ONLY from an already-scoped query.
+//    scope is already baked into the body via build_scope_condition.
+impl<E: EntityTrait> SecureSelect<E, Scoped> {
+    /// Turn a scoped query into a named CTE.
+    pub fn into_cte(self, name: &str) -> SecureCte { /* ... */ }
```
Artifizer: **MEDIUM**

`into_cte(self, name: &str)` takes an unvalidated `&str`, but the Risks section further down (lines 180-181) says the CTE name "must be guarded (accept `&'static str` or validate)". The code sketch and its own stated risk mitigation don't match, so an implementer copying this signature won't get the guard.

Why it matters: this is exactly the kind of detail that gets lost between "proposed" and "implemented" if it isn't nailed down here.

Suggested fix: pick one now and reflect it in this snippet — either change the signature to `into_cte(self, name: &'static str)` to rule out runtime-built names, or validate the name inside `into_cte` (alphanumeric + underscore, length limit) and return a `Result`. Either works; just make the sketch match the decision.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
Artifizer: **MEDIUM**

`with_ctes` accepts `impl IntoIterator<Item = SecureCte>` with nothing constraining all the `SecureCte`s (or the outer query) to share the same `AccessScope`. Each individual CTE body is still correctly scoped on its own, but nothing here says whether combining CTEs built from two different scopes (e.g. two different tenants) into one query is intentional, disallowed, or just not considered yet.

Why it matters: this ADR's stated purpose is to settle the policy before ad-hoc solutions appear in gears — this is exactly the kind of edge case that will otherwise get decided ad hoc later.

Suggested fix: add one sentence either way ("intentionally unconstrained" or "must match"), and if it should be constrained, consider a debug-only assertion in `with_ctes` comparing `scope_arc()` across the outer query and all supplied CTEs.

MikeFalcon77: Fixed. `with_ctes` now returns a distinct `SecureCteSelect<E, Scoped>` that wraps a `sea_query::SelectStatement` + `WithClause` and executes via `find_by_statement`/`FromQueryResult`. It never goes through `Select<E>::all()`, so the `WITH` can no longer be dropped at execution. Updated the API sketch and the invariant list.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
MikeFalcon77: Fixed. Signature is now `into_cte(self, name: &'static str)`. That rules out runtime-built and injection-shaped names at compile time, no `Result` needed. Added a "CTE name" note and updated the Risks bullet to match.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
MikeFalcon77: Good point, reworded. The Risks bullet now says this is a correctness risk (wrong rows), not an isolation one: each CTE embeds its own scope in its body, so a bad join predicate cannot leak another tenant's data. Added a sentence that the isolation boundary is per-CTE-body and independent of outer-join correctness.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
MikeFalcon77: Added a "Same-scope constraint" section: the outer query and all CTEs must share the same `AccessScope`. Enforce it structurally (seed each CTE from the outer query's `Arc<AccessScope>` so a mismatch is unrepresentable), or as a fallback a hard runtime check returning `Err` on mismatch in all build profiles (not `debug_assert`, which is compiled out in release). A dylint lint cannot verify runtime scope equality, so it is not used for this.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,354 @@
...
+
+```rust
+// 1. A CTE body can be built ONLY from an already-scoped query.
+//    scope is already baked into the body via build_scope_condition.
+//    The name is `&'static str` (see "CTE name" below) so runtime-built /
+//    injection-shaped names are impossible at compile time.
+impl<E: EntityTrait> SecureSelect<E, Scoped> {
+    /// Turn a scoped query into a named CTE. Retains the query's `Arc<AccessScope>`
+    /// so `with_ctes` can enforce that every CTE shares the outer query's scope.
+    pub fn into_cte(self, name: &'static str) -> SecureCte { /* ... */ }
```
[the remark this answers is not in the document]
MikeFalcon77: Reworded. `&'static str` is no longer presented as injection validation — it's just an ergonomic nudge toward literals (and it isn't even strictly compile-time: a `&'static str` can be produced at runtime by leaking). Injection safety doesn't depend on it: the name is rendered by sea_query via `Alias`/`Iden` as a quoted, escaped identifier (embedded quotes doubled), so a hostile name is an inert identifier, never executable SQL. Added a hostile-name test to Confirmation. Fixed in a0940294.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,354 @@
...
+        self,
+        ctes: impl IntoIterator<Item = SecureCte>,
+    ) -> SecureCteSelect<E, Scoped> { /* ... */ }
+}
+
+// 3. The CTE-aware result type: its own execution path, separate from `Select<E>`.
+impl<E: EntityTrait> SecureCteSelect<E, Scoped> {
+    pub async fn all(self, runner: &impl DBRunner) -> Result<Vec<E::Model>, ScopeError> { /* find_by_statement */ }
+    pub async fn one(self, runner: &impl DBRunner) -> Result<Option<E::Model>, ScopeError> { /* find_by_statement */ }
+}
```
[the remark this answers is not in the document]
MikeFalcon77: Good catch — consumption was undefined. Added a "Referencing a CTE" section: `with_ctes` only attaches definitions; the outer query references each CTE by its `&'static str` name via FROM/JOIN, and the tying predicate is the caller's responsibility (the correctness risk already listed, not an isolation one). The exact typed reference helper is left out of the ADR as an implementation detail. Fixed in a0940294.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,354 @@
...
+  ([cond.rs:191+](../../../../libs/toolkit-db/src/secure/cond.rs#L191)) that assert the
+  scope condition is present in the body of **every** CTE (assert against the built
+  SQL, per backend).
+- Codify the prohibition on raw and recursive CTEs in gear code in
+  [11_database_patterns.md](../../../toolkit_unified_system/11_database_patterns.md),
+  alongside the existing "No plain SQL" rule.
+- Back the Level-C prohibition with a custom `dylint` lint in CI (flag raw `WITH`
+  strings and `into_inner()`/`into_query()` reaching a raw-SQL sink in gear crates), so
+  the guardrail is machine-checked rather than left to review. The same-scope invariant
+  is *not* covered by the lint and must be enforced structurally / at runtime instead.
```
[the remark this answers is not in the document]
MikeFalcon77: Added. Confirmation now requires an execution-path test on `SecureCteSelect::all`/`one` that captures the emitted statement and asserts it contains the `WITH`, so a regression that falls back to plain `Select<E>` and drops the CTE is caught — separate from the scope-predicate-in-body test. Fixed in a0940294.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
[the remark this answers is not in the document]
MikeFalcon77: Right that the bound wasn't visible at the CTE methods. The invariant was already enforced upstream: `SecureSelect<E, Scoped>` is only reachable via `scope_with`/`scope_with_arc`, which are bounded `E: ScopableEntity + EntityTrait` (select.rs:151-155), so a non-Scopable entity can't reach Scoped in the first place. Made it explicit anyway — `into_cte`/`with_ctes` now carry `ScopableEntity + EntityTrait` in the sketch, plus an invariant note that it's inherited from the transition. Fixed in a840f0de.

docs/arch/secure-orm/ADR/0001-secure-cte-policy.md
```diff
@@ -0,0 +1,370 @@
...
+
+### CTE name
+
+`into_cte` takes `&'static str`, not `&str` — an ergonomic nudge toward string literals,
+not a security control (a `&'static str` can still be produced at runtime by leaking).
+Injection safety does not depend on it: the name goes through sea_query's `Alias`/`Iden`,
+which renders it as a quoted, escaped identifier (embedded quotes are doubled), so a
+hostile name becomes an inert identifier, never executable SQL — confirm with a
+hostile-name test (see Confirmation). A gear that needs a truly dynamic name is out of
+scope for Level A (escape-hatch, Level B); validate against an allowlist there.
```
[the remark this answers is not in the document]
MikeFalcon77: Both done. (1) The CTE name section now pins the rendering to sea_query 0.32.7: the quote char is `"` for Postgres/SQLite and a backtick for MySQL, and an embedded quote char is escaped by doubling it (types.rs:31-39). (2) Added the missing hostile-name test to Confirmation — pass a name like `foo"; DROP TABLE x --` and assert the built statement renders it as a single quoted, escaped identifier, not executable SQL, per backend. Fixed in 730c4b88.

PR #4284, opened 2026-07-29
fix(resource-group): map duplicate group id to 409

parviz-93:
## Description

`create_group` lets the caller supply the group `id`. The group primary key is
global across tenants, so that id may already be taken. Until now every insert
error — including a duplicate-key violation — became `DomainError::database`,
which reaches the client as a generic **500**.

A conflict the caller caused is not a server fault. The contract already said so:
the route declares `.error_409(openapi)` and `openapi.yaml` lists `201, 400, 409`
for `createGroup`. The implementation simply never produced the 409.

This change makes the implementation match the contract it already publishes.
No schema change, no route change, no spec change — the generated OpenAPI is
byte-identical.

**What changed**

- New typed variant `DomainError::GroupAlreadyExists { id }`, shaped like its
  neighbour `GroupNotFound`.
- `GroupRepository::insert` now classifies the failure with the shared
  `ScopeError::is_unique_violation` helper instead of matching the driver's
  message text inline. The helper reads SQLSTATE first (23505 on PostgreSQL,
  1555/2067 on SQLite, 1062 and friends on MySQL) and only falls back to text
  when a proxy has stripped the code.
- `MembershipRepository::insert` switched to the same helper, so this decision is
  made in one place rather than two. Its inline matching was case-sensitive and
  had no branch for MySQL's "Duplicate entry".
- The new variant maps to `already_exists` (409), placed with the other
  duplicate-on-create variants in `api/rest/error.rs`.

**Why the message cannot be wrong**

`id` is the only unique constraint on `resource_group`, and the composite primary
key is the only one on `resource_group_membership`. Both tables come from a single
migration; every other index is non-unique and the remaining constraints are
foreign keys. So a unique violation on either insert can only be the collision the
error names.

**Version bump — maintainer call**

`DomainError` is `pub` and not `#[non_exhaustive]`, so adding a variant is
breaking by the letter of the versioning guide: a downstream exhaustive `match`
would stop compiling.

The exposure is nominal. `docs/toolkit_unified_system/02_gear_layout_and_sdk_pattern.md`
marks `domain/` as "internal business logic" and states consumers need "only one
dependency" — the `-sdk` crate. Both reverse dependencies on crates.io,
`cf-gears-account-management` and `cf-gears-rg-tr-plugin`, follow that and depend
on `cf-gears-resource-group-sdk`; neither references `DomainError`. Reaching this
type at all means bypassing the SDK contract.

Flagging it rather than deciding it — pre-1.0 a strict reading gives `0.3.0`.

**Overlap with #4269**

That PR rewrites this same `insert` to drop the post-insert re-read. The two
changes touch adjacent lines, so whichever lands second gets a three-line
conflict. Happy to rebase on top of it.

## Type of Change
- [ ] Documentation update

Ticked "Bug fix" for the wire behaviour, which is a pure fix. See the version-bump
note above for the Rust surface — that box is a maintainer decision.

## Testing
- [x] Unit tests pass
- [x] Integration tests pass
- [ ] Manual testing completed — n/a, no live instance was exercised

`make check` passes end to end: 9305/9305 tests across 332 binaries, plus fmt,
cfs-validate, clippy, lychee, cargo-deny, dylint, dylint-test and gts-docs.

Four tests added, all on the group path. The one that matters is
`group_create_duplicate_id_is_already_exists` in `tests/group_service_test.rs`.
It creates a group with an explicit id in one tenant, then tries the same id from
a **different** tenant, and asserts the typed variant — so it pins the id as
globally unique rather than unique per tenant, which is the premise this whole fix
rests on. (Thanks to the review bot for catching that the first version used one
tenant for both creates and therefore proved the weaker property.)

I checked that it actually bites: disabling the classification makes it fail,
restoring it makes it pass. The same check on `membership_add_duplicate`
(`tests/membership_service_test.rs`) confirms the membership change is covered by
that existing test, so no new one was needed there.

The other three tests cover the variant and its `Display`, the SDK projection, and
the wire `Problem` status.

## Documentation
- [ ] Code is documented with rustdoc comments — n/a, no rustdoc on production
      code; the variant documents itself via `#[error]`, as all variants here do
- [ ] README updated (if applicable) — n/a
- [ ] API documentation updated (if applicable) — n/a, 409 was already declared
      on the route and in `openapi.yaml`

## Checklist
- [x] Self-review completed
- [x] No linting errors (`cargo clippy`) — and `cargo dylint`

## Related Issues

None.

## Follow-up

`DomainError` here is not `#[non_exhaustive]`, so every future error variant
raises the same version-bump question this PR does. Worth noting this looks like
a repo-wide default rather than an oversight in this gear: of the 17 gears with a
`domain/error.rs`, only `account-management` and `credstore` carry the attribute.
That seems better settled as a repo-wide decision than inside a bug fix — happy to
raise it separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate group IDs now return a clear “already exists” error instead of a generic database error.
  * Conflict responses now include the affected resource ID and use the appropriate HTTP 409 status.
  * Improved detection and handling of duplicate membership records.

* **Tests**
  * Added coverage for duplicate group creation, error details, and API response mappings.

### DIFF
--- a/gears/system/resource-group/resource-group/src/api/rest/error.rs
+++ b/gears/system/resource-group/resource-group/src/api/rest/error.rs
@@ -128,6 +128,11 @@ impl From<DomainError> for CanonicalError {
             DomainError::DuplicateMembership { key, message } => {
                 RgError::already_exists(message).with_resource(key).create()
             }
+            DomainError::GroupAlreadyExists { id } => {
+                RgError::already_exists(format!("Resource group with id '{id}' already exists"))
+                    .with_resource(id.to_string())
+                    .create()
+            }
             DomainError::TenantRootAlreadyExists {
                 existing_root_id,
                 detail,
--- a/gears/system/resource-group/resource-group/src/domain/error.rs
+++ b/gears/system/resource-group/resource-group/src/domain/error.rs
@@ -28,6 +28,9 @@ pub enum DomainError {
     #[error("Group not found: {id}")]
     GroupNotFound { id: uuid::Uuid },
 
+    #[error("Group already exists: {id}")]
+    GroupAlreadyExists { id: uuid::Uuid },
+
     #[error("Membership not found: {key}")]
     MembershipNotFound { key: String },
 
@@ -140,6 +143,11 @@ impl DomainError {
     #[must_use]
     pub fn group_not_found(id: uuid::Uuid) -> Self {
         Self::GroupNotFound { id }
+    }
+
+    #[must_use]
+    pub fn group_already_exists(id: uuid::Uuid) -> Self {
+        Self::GroupAlreadyExists { id }
     }
 
     pub fn membership_not_found(key: impl Into<String>) -> Self {
--- a/gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
+++ b/gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
@@ -649,9 +649,16 @@ impl GroupRepositoryTrait for GroupRepository {
             ..Default::default()
         };
 
+        // The group PK is global, so a caller-supplied `id` may already be taken.
         toolkit_db::secure::secure_insert::<ResourceGroupEntity>(model, &scope, db)
             .await
-            .map_err(|e| DomainError::database(e.to_string()))?;
+            .map_err(|e| {
+                if e.is_unique_violation() {
+                    DomainError::group_already_exists(id)
+                } else {
+                    DomainError::database(e.to_string())
+                }
+            })?;
 
         self.find_model_by_id(db, id)
             .await?
--- a/gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
+++ b/gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
@@ -114,8 +114,7 @@ impl MembershipRepositoryTrait for MembershipRepository {
         toolkit_db::secure::secure_insert::<MembershipEntity>(model, &scope, db)
             .await
             .map_err(|e| {
-                let msg = e.to_string();
-                if msg.contains("duplicate key") || msg.contains("UNIQUE constraint") {
+                if e.is_unique_violation() {
                     DomainError::duplicate_membership(
                         format!("({group_id}, type_id={gts_type_id}, {resource_id})"),
                         format!(
@@ -123,7 +122,7 @@ impl MembershipRepositoryTrait for MembershipRepository {
                         ),
                     )
                 } else {
-                    DomainError::database(msg)
+                    DomainError::database(e.to_string())
                 }
             })?;
 
--- a/gears/system/resource-group/resource-group/tests/api_rest_test.rs
+++ b/gears/system/resource-group/resource-group/tests/api_rest_test.rs
@@ -9,7 +9,7 @@
 mod common;
 
 use std::sync::Arc;
-use toolkit_gts::gts_id;
+use toolkit_gts::{gts_id, gts_uri};
 
 use async_trait::async_trait;
 use axum::Router;
@@ -45,6 +45,8 @@ macro_rules! rg_type_id {
         format!("{}{}", gts_id!("cf.core.rg.type.v1~"), format_args!($($arg)*))
     };
 }
+
+const ALREADY_EXISTS_TYPE: &str = gts_uri!("cf.core.errors.err.v1~cf.core.err.already_exists.v1~");
 
 // ── Noop OpenAPI Registry for tests ─────────────────────────────────────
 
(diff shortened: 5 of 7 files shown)


ffedoroff:
Nice change — turning a PK collision into a typed conflict is right, and replacing the hand-rolled `duplicate key` / `UNIQUE constraint` substring match in `membership_repo` with the shared `is_unique_violation` classifier is the better half of it: that detector only ever covered Postgres and SQLite wording and would have silently missed MySQL.

The tiers you picked are the right ones per `docs/toolkit_unified_system/12_unit_testing.md`. The three `domain_unit_test.rs` cases are **unit / pure logic** — no DB, no async (`12:22`, `12:48`). `group_create_duplicate_id_is_already_exists` is **unit / service-level against SQLite `:memory:`**, which `12:26-28` names as the preferred tool at the repository↔DB boundary. No objection to either.

Four things I'd add, in rough order of value. Tier is named for each so it's clear where they belong:

**1. 409 at the REST layer — unit tier, REST-level `Router::oneshot`.** Same suite as the others, in `tests/api_rest_test.rs`, which already has the harness. `12:54` and `12:369` place response-shape checks at this level; it is *not* E2E despite living in `tests/`. Today the only thing pinning 409 is `domain_to_problem_group_already_exists_is_409`, which exercises the mapping in isolation and would still pass if the handler path stopped producing it. Two `POST /groups` with the same `id`, asserting 409 **and** the `problem+json` body — type, status, `resource_name`.

**2. The same-tenant duplicate — unit tier, service-level with SQLite.** Extends what you already wrote, same file and style. The current test uses two tenants, which is the interesting edge; a caller reusing an id inside their own tenant is the common one and reaches the insert through a different preflight. Both are worth pinning.

**3. A secondary-artifact assertion — unit tier, inside the service test you already have.** No new test, just more assertions on the existing one. `12:251-334` asks for more than the error variant: after the rejected second create, confirm with a **direct entity query** that the first group is still present and unmodified (name, tenant). A failed insert must not have touched it. `12:95` specifically calls for the direct query rather than a service read here, since a scoped read could mask the state.

**4. Coverage for the `membership_repo` swap — unit tier, service or repository level with SQLite.** The behaviour is claimed unchanged and I believe it is, but nothing in the diff demonstrates it. If an existing duplicate-membership test already covers that path, naming it in the description is enough — no new test needed.

**Deliberately not asking for an E2E test.** By `13_e2e_testing.md` that tier means a running server against real PostgreSQL, and its goal is one call per API method plus integration seams — `POST /groups` is already covered there. A status-code change on an already-covered route doesn't earn a new E2E case; items 1–4 sit closer to the defect and run in seconds.

**One follow-up, explicitly out of scope for this PR — PG integration tier, behind `--features integration`.** `is_unique_violation` is two-stage: `sea_orm::SqlErr::UniqueConstraintViolation` from the parsed SQLSTATE first, string matching only as a fallback (`libs/toolkit-db/src/secure/error.rs:46-70`). SQLite reaches only the fallback, so the branch that runs in production is unexercised by any of the above. resource-group has no Postgres tier on `main` — no `integration` feature, no `pg_*` tests, no `make test-rg-pg` — so standing it up here would dwarf a 92-line fix. The template is account-management: `tests/conversion_integration_pg.rs` gated by `[features] integration`, plus a Makefile target modelled on `test-users-info-pg`. Happy to open that as its own issue.

For the description rather than the code: which of the new tests fail on `main` without the fix. I expect all four do, but `12:40-41` makes that the bar and it's cheap to record.

And treat the list above as a floor rather than a specification — if you see cases I've missed, or think a different shape of test pins the behaviour better, add them. You're closer to this change than I am.

parviz-93:
All four addressed — items 1–3 in the follow-up commit, item 4 needed no code.

**Which tests fail without the fix.** Reverting only the `is_unique_violation` discriminator in `group_repo.rs`, leaving everything else in place:

| test | without the fix |
|---|---|
| `create_group_duplicate_id_returns_409` | fails — `left: 500, right: 409` |
| `group_create_duplicate_id_is_already_exists` | fails |
| `group_create_duplicate_id_same_tenant_is_already_exists` | fails |
| `domain_error_group_already_exists_format` | **passes** |
| `domain_to_sdk_group_already_exists` | **passes** |
| `domain_to_problem_group_already_exists_is_409` | **passes** |

So your expectation that all of them fail does not hold, and item 1 was pointing at a real hole. The three unit cases build the `DomainError` by hand, so they pin the mapping and nothing that reaches it. On `main` proper they do not compile, since the variant does not exist yet — but that is weaker than catching the regression.

**Item 4 — coverage for the `membership_repo` swap.** `membership_add_duplicate` (TC-MBR-03, `tests/membership_service_test.rs:120`) calls `add_membership` twice and reaches `MembershipRepository::insert`; there is no preflight ahead of it, `membership_service.rs:214` defers duplicate detection to the repo. Its assertion accepts any of `Conflict`, `ConflictActiveReferences` or `DuplicateMembership`, so it pins the path but not the variant. Left as is rather than widen this PR.

**Postgres tier for the SQLSTATE branch.** Agreed it is out of scope here, and happy for you to open it as its own issue. Worth noting we have a direct interest: that branch is the one that runs in production, so it decides whether this fix works there at all.

Took the list as given rather than adding to it, to keep the diff easy to read.

Local `make check` is green — 9307 tests, 368 skipped.

ffedoroff (approved):
approved

nonameffh (approved):

PR #4333, opened 2026-07-29
feat(bss-rate-provider): add FX rate-provider gear

bit4flip:
## Description

Adds the BSS **FX rate-provider adapter gear** — the missing implementation of the
ledger-owned `RateProviderV1` seam. Until now the ledger shipped only the consuming
side (`RateProviderV1`, `RateSyncJob`, the local rate store, and the fail-safe), so
every cross-currency post blocked with `FX_RATE_UNAVAILABLE`. This closes that gap.

Four new crates, all at `version = "0.1.0"`:

- **`bss-rate-provider`** (core gear) — discovers source plugins through the
  types-registry, filters them by `source_vendor`, orders them by each plugin's
  `priority`, and re-registers the result as a single composite `RateProviderV1`
  plugin. Ordered fallback: the first source returning a whole document wins;
  documents are never merged per pair.
- **`bss-rate-provider-sdk`** — the source-plugin GTS spec plus shared source
  utilities: exact-decimal conversion (`rust_decimal`, banker's rounding, never
  `f64`), ISO-4217 currency-shape gating, error mapping, fetch/health metrics, and
  an HTTPS-only outbound client builder that rejects a plain-http `base_url` at gear
  `init()` rather than at first fetch.
- **`bss-rate-provider-ecb-plugin`** — ECB daily reference rates (XML feed).
- **`bss-rate-provider-http-json-plugin`** — generic JSON feed driven by a
  dotted-path `mapping`, with `none` / `bearer` / `header-key` auth.

The gear is stateless and fetch-only: no persistence, no REST surface, no accounting.
Sources emit only their natively published pairs — cross-base derivation stays the
ledger's triangulation concern.

**Ledger side.** `resolve_rate_provider` performs lazy discovery on each sync tick and
degrades to the fail-safe `UnconfiguredRateProviderV1` (`provider_id() == "none"`, so
the sync job stays silent and raises no alarm) when the registry is unavailable or no
plugin matches.

**Provenance is carried per rate.** Each source stamps its own id onto every
`ProviderRate`, so the serving upstream travels with the data and the ledger records it
per row. An earlier design read provenance back through a later `provider_id()` call,
which was only correct while exactly one caller alternated `fetch_latest` →
`provider_id`; the composite is registered process-wide in `ClientHub`, so a second
concurrent consumer made the ledger stamp the wrong source onto financial records.
`provider_id()` is now the composite's own constant identity.

Configuration example:

```yaml
bss-rate-provider:
  config:
    vendor: cf.bss          # advertised to the ledger's plugin selection
    priority: 100
    source_vendor: cf.bss   # selects which source plugins to compose
    id: bss-rate-provider

bss-rate-provider-ecb-plugin:
  config:
    id: ecb
    vendor: cf.bss
    priority: 100           # lower is tried first
    base_url: https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml
    timeout_ms: 5000
```

Adding a further plain REST feed is configuration-only — no code change, no redeploy of
consuming modules.

62 files changed, +6386 / −34.

## Type of Change
- [x] Documentation update

All rate-provider crates are new at `0.1.0`, so no existing public surface changes.
`bss-ledger-sdk` gains an explicit `version = "0.1.0"` (it was previously implicit
`0.0.0`); its API is unchanged, so this is a packaging change, not a semver event.

## Testing
- [x] Unit tests pass
- [x] Integration tests pass

`cargo test -p bss-rate-provider -p bss-rate-provider-sdk -p bss-rate-provider-ecb-plugin
-p bss-rate-provider-http-json-plugin -p bss-ledger` → **932 passed, 0 failed, 365
ignored** (the ignored set is the ledger's Postgres testcontainer suite, which needs a
running Docker daemon). Re-verified after rebasing onto current `main`.

New test coverage added with the feature:

- `composite_tests.rs` — ordered fallback and last-error propagation for both
  `fetch_latest` and `health`, short-circuit on the first healthy source, and the
  empty-composite degradation path
- `http_client_tests.rs` — https-only gate (case-insensitive scheme, host required)
  and the guarantee that a rejection error never echoes a possibly secret-bearing URL
- `conversion_tests.rs`, `currency_tests.rs`, `error_tests.rs`, `fetch_tests.rs`,
  `metrics_tests.rs` — SDK units
- `source_tests.rs` + `tests/*_integration.rs` for both plugins
- `tests/discovery.rs` — core-gear discovery over a real `ClientHub`
- `module_tests.rs` — the ledger's lazy resolution and its fail-safe fallback

## Documentation
- [ ] API documentation updated (if applicable)

Spec-driven artifacts per CONTRIBUTING §2.2 are included:
`gears/bss/rate-provider/docs/PRD.md` (472 lines) and `docs/DESIGN.md` (1005 lines).
No REST surface is exposed, so there is no API documentation to update.

## Checklist
- [x] Self-review completed
- [x] `make check-packaging-metadata` passes (`OK: 74 publishable crates checked`)
- [x] Branch is rebased on current `main`
- [x] Commits are signed off (DCO)

## Summary by CodeRabbit

* **New Features**
  * Added configurable FX rate-provider support with ECB and generic HTTP/JSON sources.
  * Added ordered fallback, lazy per-sync discovery, vendor matching, and priority selection.
  * Added per-rate source provenance and configurable provider vendor, defaulting to `cf.bss`.
  * Added secure HTTPS feeds, authentication options, publication-date validation, and precise rate conversion.
  * Added FX sync heartbeat, duration metrics, and improved request correlation in failures.
* **Documentation**
  * Added rate-provider design documentation and local configuration examples.
* **Bug Fixes**
  * Enforced a minimum 60-second FX synchronization interval.
* **Tests**
  * Added coverage for discovery, fallback, parsing, validation, authentication, errors, and metrics.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,10 @@ members = [
     "gears/llm-gateway/llm-gateway-sdk",
     "gears/bss/ledger/ledger",
     "gears/bss/ledger/ledger-sdk",
+    "gears/bss/rate-provider/rate-provider-sdk",
+    "gears/bss/rate-provider/rate-provider",
+    "gears/bss/rate-provider/plugins/ecb-plugin",
+    "gears/bss/rate-provider/plugins/http-json-plugin",
     "gears/bss/libs/coord",
     "gears/system/event-broker/event-broker-sdk",
     "gears/system/event-broker/event-broker",
@@ -311,7 +315,8 @@ oagw-sdk = { package = "cf-gears-oagw-sdk", version = "0.5.5", path = "gears/sys
 usage-collector-sdk = { package = "cf-gears-usage-collector-sdk", version = "0.3.2", path = "gears/system/usage-collector/usage-collector-sdk" }
 
 # BSS gears (ported from vhp-core)
-bss-ledger-sdk = { path = "gears/bss/ledger/ledger-sdk" }
+bss-ledger-sdk = { path = "gears/bss/ledger/ledger-sdk", version = "0.1.0" }
+bss-rate-provider-sdk = { path = "gears/bss/rate-provider/rate-provider-sdk", version = "0.1.0" }
 coord = { path = "gears/bss/libs/coord" }
 # FIPS-validated SHA-256 for the ledger idempotency payload fingerprint.
 aws-lc-rs = "1.17"
--- a/apps/cf-gears-example-server/Cargo.toml
+++ b/apps/cf-gears-example-server/Cargo.toml
@@ -30,6 +30,14 @@ otel = ["toolkit/otel"]
 account-management = ["dep:account_management"]
 static-idp = ["dep:static-idp-plugin"]
 file-storage = ["dep:file_storage"]
+# The FX rate-provider subsystem: the core gear + its source plugins. Independent
+# of bss-ledger — the ledger discovers the composite lazily at runtime via the
+# types-registry, so neither feature requires the other.
+bss-rate-provider = [
+    "dep:bss_rate_provider",
+    "dep:bss_rate_provider_ecb_plugin",
+    "dep:bss_rate_provider_http_json_plugin",
+]
 bss-ledger = ["dep:bss_ledger"]
 usage-collector = ["dep:usage_collector"]
 timescaledb-usage-collector = ["dep:timescaledb_usage_collector_plugin"]
@@ -87,6 +95,9 @@ static-idp-plugin = { package = "cf-gears-static-idp-plugin", path = "../../gear
 
 # Optional BSS Billing Ledger gear
 bss_ledger = { package = "bss-ledger", path = "../../gears/bss/ledger/ledger", optional = true }
+bss_rate_provider = { package = "bss-rate-provider", path = "../../gears/bss/rate-provider/rate-provider", optional = true }
+bss_rate_provider_ecb_plugin = { package = "bss-rate-provider-ecb-plugin", path = "../../gears/bss/rate-provider/plugins/ecb-plugin", optional = true }
+bss_rate_provider_http_json_plugin = { package = "bss-rate-provider-http-json-plugin", path = "../../gears/bss/rate-provider/plugins/http-json-plugin", optional = true }
 
 # Optional Usage Collector gear + its TimescaleDB storage plugin
 usage_collector = { package = "cf-gears-usage-collector", path = "../../gears/system/usage-collector/usage-collector", optional = true }
--- a/gears/bss/ledger/ledger/src/config.rs
+++ b/gears/bss/ledger/ledger/src/config.rs
@@ -140,6 +140,13 @@ pub enum ConfigError {
         max: u64,
         reason: &'static str,
     },
+    /// A field fell below its allowed minimum (for bounds stricter than "> 0").
+    #[error("config: {field} must be >= {min} ({reason})")]
+    BelowMin {
+        field: &'static str,
+        min: u64,
+        reason: &'static str,
+    },
 }
 
 impl JobsConfig {
@@ -302,6 +309,10 @@ pub struct FxConfig {
     /// Deterministic provider fallback order (by `provider_id`); empty until an
     /// adapter is configured.
     pub provider_order: Vec<String>,
+    /// Vendor of the FX rate-provider plugin the `RateSyncJob` discovers each tick
+    /// from the types-registry (`RateProviderPluginSpecV1` instances). Must match
+    /// the external `bss-rate-provider` core gear's advertised `vendor`.
+    pub provider_vendor: String,
 }
 
 impl Default for FxConfig {
@@ -313,6 +324,7 @@ impl Default for FxConfig {
             rate_sync_tick_secs: 3_600,
             revaluation_run_tick_secs: 86_400,
             provider_order: Vec::new(),
+            provider_vendor: "cf.bss".to_owned(),
         }
     }
 }
@@ -323,6 +335,15 @@ impl Default for FxConfig {
 /// window overflows chrono's millisecond representation) — bound it at `validate`.
 const MAX_STALE_G10_HOURS: u64 = 168;
 
+/// Lower bound on `rate_sync_tick_secs`. The `FxNoFetchAttempted` alert reads a
+/// `3 × rate_sync_tick_secs` window of the adapter's completion metrics, and one
+/// source attempt can stay in flight for up to the adapter's 30 s per-source
+/// timeout ceiling — with a sub-minute tick a single slow attempt could span the
+/// whole alert window and read as "no attempt occurred". A 60 s floor keeps that
+/// window at ≥ 180 s, and nothing needs sub-minute FX syncs against feeds that
+/// publish daily.
+const MIN_RATE_SYNC_TICK_SECS: u64 = 60;
+
 impl FxConfig {
     /// Validate the FX tunables.
     ///
@@ -330,10 +351,11 @@ impl FxConfig {
     /// Returns `Err` if `stale_default_max_days > 7` (F3: a threshold above the
     /// 7-day bound must be rejected, never silently clamped), if `stale_g10_hours`
     /// exceeds [`MAX_STALE_G10_HOURS`] (a larger window overflows
-    /// `chrono::Duration::hours` and panics in `rate_source::is_stale`), or if any
-    /// of `stale_g10_hours` / `rate_sync_tick_secs` / `revaluation_run_tick_secs`
-    /// is `0` (a zero staleness window admits any rate; a zero tick panics
-    /// `tokio::time::interval`).
+    /// `chrono::Duration::hours` and panics in `rate_source::is_stale`), if
+    /// `rate_sync_tick_secs` is below [`MIN_RATE_SYNC_TICK_SECS`] (the
+    /// `FxNoFetchAttempted` alert-window bound — see the constant's docs), or if
+    /// `stale_g10_hours` / `revaluation_run_tick_secs` is `0` (a zero staleness
+    /// window admits any rate; a zero tick panics `tokio::time::interval`).
     pub fn validate(&self) -> Result<(), ConfigError> {
         if self.stale_default_max_days > 7 {
             return Err(ConfigError::AboveMax {
@@ -354,9 +376,11 @@ impl FxConfig {
                 reason: "7-day bound",
             });
         }
-        if self.rate_sync_tick_secs == 0 {
-            return Err(ConfigError::MustBePositive {
+        if self.rate_sync_tick_secs < MIN_RATE_SYNC_TICK_SECS {
+            return Err(ConfigError::BelowMin {
                 field: "fx.rate_sync_tick_secs",
+                min: MIN_RATE_SYNC_TICK_SECS,
+                reason: "FxNoFetchAttempted alert-window bound",
             });
         }
         if self.revaluation_run_tick_secs == 0 {
--- a/gears/bss/ledger/ledger/src/module.rs
+++ b/gears/bss/ledger/ledger/src/module.rs
@@ -140,11 +140,15 @@ pub(crate) struct LedgerRuntime {
     pub recognition_run_tick: Duration,
     /// Chain-verifier tick cadence.
     pub verify_tick: Duration,
-    /// FX rate-provider plugin (Slice 5) resolved from `ClientHub` — the
-    /// fail-safe `UnconfiguredRateProviderV1` default when no adapter-gear is
-    /// registered. The `RateSyncJob` pulls `fetch_latest` from it each
-    /// `fx_rate_sync_tick` into the local rate store.
-    pub rate_provider: Arc<dyn RateProviderV1>,
+    /// Process `ClientHub` handle used to LAZILY discover the FX rate-provider
+    /// plugin each rate-sync tick (the composite the external `bss-rate-provider`
+    /// core gear registers as a `RateProviderPluginSpecV1` instance). Resolving
+    /// per tick — rather than once at init — means a late-registered adapter is
+    /// picked up on the next tick and there is no init-ordering dependency; absent
+    /// an adapter the resolver yields the fail-safe `UnconfiguredRateProviderV1`.
+    pub rate_provider_hub: Arc<toolkit::client_hub::ClientHub>,
+    /// Vendor selecting which rate-provider plugin instance the ticker resolves.
+    pub rate_provider_vendor: String,
     /// FX rate store (the `RateSyncJob` upsert target / the lock-time
     /// `RateSource` read target).
     pub fx_repo: crate::infra::storage::repo::FxRepo,
@@ -169,6 +173,11 @@ pub(crate) struct LedgerRuntime {
 // broker-free no-op/metrics-only publisher). Declaring it would make the gear
 // unschedulable (the orchestrator can't satisfy a dep no gear provides). Re-add
 // it here once the broker gear exists and `events_enabled` is wired to it.
+//
+// NOTE: the FX rate-provider adapter is NOT a `deps` edge — the `RateSyncJob`
+// discovers it lazily from the types-registry each tick (see `resolve_rate_provider`),
+// so a late-registered adapter self-heals and the ledger stays decoupled from
+// (and schedulable without) the external `bss-rate-provider` gear.
 #[toolkit::gear(name = "bss-ledger", capabilities = [db, rest, stateful], deps = [types_registry, authz_resolver, account_management], lifecycle(entry = "serve", stop_timeout = "30s"))]
 pub struct BssLedgerGear {
     /// Typed runtime built inside [`Gear::init`] and consumed by
@@ -182,6 +191,88 @@ impl Default for BssLedgerGear {
     fn default() -> Self {
         Self {
             runtime: ArcSwapOption::from(None),
+        }
+    }
+}
+
+/// Discover the FX rate-provider plugin from the types-registry (the composite
+/// the external `bss-rate-provider` core gear registers), selecting the instance
+/// by `vendor` + `priority`. Called fresh each rate-sync tick, so a
+/// late-registered adapter is picked up on the next tick — no init-ordering
+/// dependency. Falls back to the fail-safe
+/// [`UnconfiguredRateProviderV1`](bss_ledger_sdk::UnconfiguredRateProviderV1)
+/// whenever the registry is unavailable or no plugin is registered, so a missing
+/// adapter is a benign deployment state (the sync job logs at debug, no alarm),
+/// never a books defect.
+async fn resolve_rate_provider(
+    hub: &toolkit::client_hub::ClientHub,
+    vendor: &str,
+) -> Arc<dyn RateProviderV1> {
+    let fallback = || Arc::new(UnconfiguredRateProviderV1) as Arc<dyn RateProviderV1>;
+    // Every branch below logs before falling back. Silently returning the
+    // fail-safe made a types-registry outage or a malformed plugin instance
+    // indistinguishable from the benign "no adapter configured" state, so FX
+    // rates would go stale with nothing to diagnose from. Levels split by
+    // meaning: `warn` for something broken, `debug` for the expected
+    // not-configured case (the sync job already reports that one).
+    let registry = match hub.get::<dyn types_registry_sdk::TypesRegistryClient>() {
+        Ok(registry) => registry,
+        Err(e) => {
+            tracing::warn!(
+                vendor,
+                error = %e,
+                "bss-ledger: types-registry client unavailable; using the fail-safe FX \
+                 rate provider (rates will go stale)"
+            );
+            return fallback();
+        }
+    };
+    let type_id = bss_ledger_sdk::RateProviderPluginSpecV1::gts_type_id();
+    let instances = match registry
+        .list_instances(
+            types_registry_sdk::InstanceQuery::new().with_pattern(format!("{type_id}*")),
+        )
+        .await
+    {
+        Ok(instances) => instances,
+        Err(e) => {
+            tracing::warn!(
+                vendor,
+                error = %e,
+                "bss-ledger: listing FX rate-provider plugin instances failed; using the \
+                 fail-safe FX rate provider (rates will go stale)"
+            );
+            return fallback();
+        }
+    };
+    match toolkit::plugins::choose_plugin_instance::<bss_ledger_sdk::RateProviderPluginSpecV1>(
+        vendor,
+        instances.iter().map(|e| (e.id.as_ref(), &e.object)),
+    ) {
+        Ok(gts_id) => hub
+            .try_get_scoped::<dyn RateProviderV1>(&toolkit::client_hub::ClientScope::gts_id(
+                &gts_id,
+            ))
+            .unwrap_or_else(|| {
+                // Partial registration: the plugin published its instance but
+                // never registered (or failed before registering) its scoped
+                // client. Not the benign unconfigured state — warn.
+                tracing::warn!(
+                    vendor,
+                    instance_id = %gts_id,
+                    "bss-ledger: FX rate-provider plugin instance has no scoped client \
+                     registered; using the fail-safe FX rate provider"
+                );
+                fallback()
+            }),
+        Err(e) => {
+            tracing::debug!(
+                vendor,
+                error = %e,
+                "bss-ledger: no FX rate-provider plugin instance matched; using the \
+                 fail-safe default"
+            );
+            fallback()
         }
     }
 }
@@ -247,15 +338,41 @@ impl BssLedgerGear {
                     biased;
                     () = token.cancelled() => break,
                     _ = iv.tick() => {
+                        // Heartbeat FIRST, before discovery: this must count the tick
+                        // itself. A stalled ticker attempts no fetch, so it produces no
+                        // fetch error and never raises FX_SNAPSHOT_MISSING — the rate
+                        // store just ages until a post blocks on staleness. Incrementing
+                        // after resolve/fetch would go flat on a discovery failure too,
+                        // making a dead ticker indistinguishable from an empty registry.
+                        rt.metrics.fx_rate_sync_ticked();
+                        // Timed from here, alongside the heartbeat, so the sample covers
+                        // the whole pass — discovery included. The composite tries its
+                        // sources sequentially, so the pass, not any single source's
+                        // fetch, is what has to fit inside `fx_rate_sync_tick`; nothing
+                        // else measures it. The loop awaits the pass inline, so a pass
+                        // that outgrows the interval cannot overlap the next one — it
+                        // delays it, and this histogram is where that shows up first.
+                        let started = std::time::Instant::now();
+                        // Discover the rate-provider plugin fresh each tick (lazy,
+                        // self-healing) — the fail-safe default when none is registered.
+                        let provider = resolve_rate_provider(
+                            &rt.rate_provider_hub,
+                            &rt.rate_provider_vendor,
+                        )
+                        .await;
                         let job = crate::infra::jobs::rate_sync::RateSyncJob::new(
                             rt.db.clone(),
-                            Arc::clone(&rt.rate_provider),
+                            provider,
                             rt.fx_repo.clone(),
                             Arc::clone(&rt.publisher),
                         );
                         if let Err(e) = job.run().await {
                             tracing::error!(error = %e, "bss-ledger: rate-sync job tick failed");
                         }
+                        // Recorded on the failure path too: a pass that fails slowly is
+                        // exactly the one worth seeing, and dropping those samples would
+                        // bias the p95 the alert reads.
+                        rt.metrics.fx_rate_sync_duration(started.elapsed().as_secs_f64());
                     }
                 }
             }
@@ -946,18 +1063,17 @@ impl Gear for BssLedgerGear {
         let rt_enforcer = Arc::clone(&enforcer);
         let rt_metrics = Arc::clone(&metrics);
 
-        // FX rate-provider plugin (Slice 5): resolved from `ClientHub` like the
-        // other cross-gear clients, but with a fail-safe DEFAULT — the external
-        // ECB/bank adapter-gear is out of Slice 5 scope, so absent the adapter the
-        // gear falls back to `UnconfiguredRateProviderV1` (every fetch errors → the
-        // local rate store stays empty → FX-needing posts block at lock time,
-        // never a silent wrong rate). Unlike authz, a missing rate adapter is NOT
-        // fatal to init. The `RateSyncJob` (serve loop) pulls into `fx_repo`, whose
-        // db clone is captured here before `db` moves into the client below.
-        let rate_provider: Arc<dyn RateProviderV1> = ctx
-            .client_hub()
-            .get::<dyn RateProviderV1>()
-            .unwrap_or_else(|_| Arc::new(UnconfiguredRateProviderV1));
+        // FX rate-provider plugin (Slice 5): discovered LAZILY from the
+        // types-registry each rate-sync tick (see `resolve_rate_provider`), not
+        // resolved once here — so a late-registered adapter is picked up next tick
+        // and the ledger needs no `deps` edge on it. Absent an adapter the resolver
+        // yields the fail-safe `UnconfiguredRateProviderV1` (every fetch errors →
+        // the local store stays empty → FX-needing posts block at lock time, never
+        // a silent wrong rate); a missing adapter is NOT fatal to init. The hub
+        // handle + selection vendor are captured for the serve-loop ticker. The
+        // `fx_repo` db clone is captured before `db` moves into the client below.
+        let rate_provider_hub = ctx.client_hub();
+        let rate_provider_vendor = cfg.fx.provider_vendor.clone();
         let fx_repo = crate::infra::storage::repo::FxRepo::new(db.clone());
 
         // Slice 7 Phase 3: the three launch-blocking control feeds (issued-invoice
@@ -1420,7 +1536,8 @@ impl Gear for BssLedgerGear {
             aged_alarm_tick: cfg.jobs.aged_alarm_interval(),
             recognition_run_tick: cfg.recognition.recognition_run_interval(),
             verify_tick: cfg.jobs.verify_interval(),
-            rate_provider,
+            rate_provider_hub,
+            rate_provider_vendor,
             fx_repo,
             fx_rate_sync_tick: cfg.fx.rate_sync_interval(),
             fx_config: cfg.fx.clone(),
@@ -1560,3 +1677,7 @@ impl RestApiCapability for BssLedgerGear {
         }
     }
 }
+
+#[cfg(test)]
+#[path = "module_tests.rs"]
+mod tests;
--- a/gears/bss/rate-provider/docs/DESIGN.md
+++ b/gears/bss/rate-provider/docs/DESIGN.md
@@ -1,0 +1,1331 @@
+<!-- CONFLUENCE_TITLE: [BSS]: FX Rate Provider (Adapter Gear) — Technical Design -->
+<!-- Related: ../../ledger/docs/DESIGN.md, ../../ledger/docs/design/06-fx-multicurrency.md, ../../ledger/docs/PRD.md | Owners: @vstudzinskyi (BSS Billing Platform team) -->
+
+# Technical Design — FX Rate Provider (Adapter Gear)
+
+<!-- toc -->
+
+- [1. Architecture Overview](#1-architecture-overview)
+  - [1.1 Architectural Vision](#11-architectural-vision)
+  - [1.2 Architecture Drivers](#12-architecture-drivers)
+  - [1.3 Architecture Layers](#13-architecture-layers)
+- [2. Principles & Constraints](#2-principles--constraints)
+  - [2.1 Design Principles](#21-design-principles)
+  - [2.2 Constraints](#22-constraints)
+- [3. Technical Architecture](#3-technical-architecture)
+  - [3.1 Domain Model](#31-domain-model)
+  - [3.2 Component Model](#32-component-model)
+  - [3.3 API Contracts](#33-api-contracts)
+  - [3.4 Internal Dependencies](#34-internal-dependencies)
+  - [3.5 External Dependencies](#35-external-dependencies)
+  - [3.6 Interactions & Sequences](#36-interactions--sequences)
+  - [3.7 Database schemas & tables](#37-database-schemas--tables)
+  - [3.8 Deployment Topology](#38-deployment-topology)
+- [4. Additional context](#4-additional-context)
+  - [Security & AuthZ](#security--authz)
+  - [Feature metrics](#feature-metrics)
+  - [Testing architecture](#testing-architecture)
+  - [Decision register](#decision-register)
+  - [Companion ledger change (hard dependency, from O-3)](#companion-ledger-change-hard-dependency-from-o-3)
+- [5. Traceability](#5-traceability)
+
+<!-- /toc -->
+
+- [ ] `p1` - **ID**: `cpt-cf-bss-rate-provider-design-main`
+
+> **Canonical design entry point.** This document is the FX rate-provider gear's technical
+> design and the anchor for spec traceability. The gear is small enough that the design is
+> **self-contained** — there is no slice set; component, contract, and sequence detail is
+> normative here.
+>
+> **Status**: DRAFT — decisions recorded (O-1 & O-3 decided; all other defaults accepted
+> 2026-07-08). Ready for implementation planning. The O-3 companion `bss-ledger` change
+> (§4 "Companion ledger change") is a linked hard dependency.
+>
+> **Implementation revision (2026-07-23) — `*-plugin` pattern.** The gear was
+> built on the platform's plugin pattern (types-registry `PluginV1` instances +
+> `ClientHub` scoped registration + host-side discovery), which revises three
+> decisions below. This note is authoritative where it conflicts with the older
+> prose:
+>
+> - **Two crates → four.** Instead of one composite gear with a config `sources[]`
+>   list, each source is its own **plugin gear** — `bss-rate-provider-ecb-plugin`
+>   and `bss-rate-provider-http-json-plugin` — and a **core gear**
+>   (`bss-rate-provider`) discovers and composes them. Shared source utilities
+>   (conversion, error mapping, fetch metrics) and the source-plugin GTS spec live
+>   in `bss-rate-provider-sdk`.
+> - **O-1 revised — scoped plugin registration, two layers.** *Level 1:* each
+>   source plugin registers a `PluginV1<RateProviderSourcePluginSpecV1>` instance
+>   in the types-registry and a scoped `RateProviderV1` client (`priority` = the
+>   fallback order). The core gear lists those instances, orders by `priority`,
+>   resolves each via `get_scoped`, and composes them (first whole successful
+>   document; provenance stamped per rate — see §3.2). *Level 2:* the core gear
+>   registers its composite as a `PluginV1<RateProviderPluginSpecV1>` (spec in
+>   `bss-ledger-sdk`) + scoped `RateProviderV1`; the ledger discovers **that**.
+> - **O-7 revised — no `deps` edge.** The ledger discovers the composite **lazily
+>   on every `RateSyncJob` tick** (types-registry `list_instances` → vendor/priority
+>   select → `get_scoped`), falling back to `UnconfiguredRateProviderV1`. A
+>   late-registered adapter self-heals on the next tick, so the startup-ordering
+>   concern is gone and the ledger stays decoupled (schedulable without the adapter).
+> - **O-12 revised — per-plugin config, not `sources[]`.** Source assembly is now
+>   plugin discovery ordered by each plugin's `priority`; the `fx.provider_order` /
+>   `sources[]` cross-gear order check is replaced by matching `vendor` between the
+>   core gear, its source plugins, and the ledger's `fx.provider_vendor`. Unknown
+>   `kind` no longer exists (each source is a distinct plugin crate). Config
+>   validation is per plugin (http-json requires `mapping` + https `base_url`).
+>
+> The domain model (§3.1), the fetch-only/direct-pairs/deterministic-conversion
+> principles (§2.1), the metrics (§4), and the O-3 triangulation boundary are
+> unchanged.
+
+## 1. Architecture Overview
+
+### 1.1 Architectural Vision
+
+The **FX rate-provider gear** is a **stateless adapter**: it implements the ledger's
+`RateProviderV1` contract (`bss-ledger-sdk`, GTS `gts.cf.bss.ledger.rate-provider.v1`) and
+registers an `Arc<dyn RateProviderV1>` into the platform `ClientHub`, so the ledger's
+background `RateSyncJob` resolves a live adapter instead of the fail-safe
+`UnconfiguredRateProviderV1` default. The registered instance is a
+**`DiscoveringRateProvider`** that lazily discovers source plugins and builds a
+**`CompositeRateProvider`** to try its ordered sources in the PRD-ratified provider
+order (2026-06-10) and returns the **first whole document** that succeeds (one source per
+document, never a merge; O-1 — a rule about provenance, not about coverage, see §2.1). The
+fallback **mechanism** ships in v1, but the **v1
+configuration is ECB-only** (O-10): the bank / PSP feed is a *future* source, added later
+by deploying/configuring one more source-plugin gear (its own crate, its own `vendor` +
+`priority`) — no change to this gear's code. It performs no persistence, no HTTP surface,
+and no accounting logic.
+
+**This gear only fetches rates.** The ledger declares the FX rate provider out of its own
+scope ("The FX rate provider itself (integration, feeds) — external; the ledger consumes
+rates and snapshots them",
+[`../../ledger/docs/design/06-fx-multicurrency.md`](../../ledger/docs/design/06-fx-multicurrency.md))
+and already ships the consuming seam: the `RateProviderV1` SDK trait, the `RateSyncJob`
+that pulls it, the `ledger_fx_rate` local store, the lock-time `RateSource`, staleness /
+provider-precedence resolution, and the immutable `rate_snapshot`. Everything ledger-owned
+stays there and is NOT restated here:
+
+- Functional-currency **translation** and the dual-column balance.
+- **Triangulation** through EUR (X→EUR→Y) — ledger-owned (O-3); requires the companion
+  ledger change (§4 "Companion ledger change").
+- **Staleness** rules (G10 > 24 h; others ≤ 7 d) and `stale` marking.
+- **Provider precedence / fallback-order** resolution over the local store.
+- **`rate_snapshot`** freezing, `ledger_fx_rate` upsert, per-tenant fan-out.
+- Realized / unrealized FX, revaluation runs.
+- The `RateSyncJob` tick cadence and its `FX_SNAPSHOT_MISSING` alarm (ledger-side).
+
+Also out of scope: **pricing-side FX / rate-lock governance** (Catalog module) and
+**provider commercial contracts / credential procurement** (ops).
+
+### 1.2 Architecture Drivers
+
+Requirements from the ledger [`PRD.md`](../../ledger/docs/PRD.md) and the FX slice design
+that significantly shape this gear.
+
+#### Functional Drivers
+
+| Requirement | Design Response |
+|-------------|-----------------|
+| `cpt-cf-bss-ledger-fr-multi-currency-fx` | The ledger needs a live rate feed to translate transaction currency into functional currency; this gear supplies it through the fixed `RateProviderV1` seam — implement `provider_id()`, `fetch_latest()`, `health()` (§3.3). |
+| `cpt-cf-bss-ledger-fr-fx-rate-source-failure` | Provider outage must never produce a silent wrong rate. The composite returns the last `RateProviderError` when **all** sources fail; the ledger job then alarms and FX posts block (`FX_RATE_UNAVAILABLE`) — fail-safe by absence (§2.2). |
+| Rate-source fallback ([ledger FX design](../../ledger/docs/design/06-fx-multicurrency.md)) | The ledger resolves precedence over its **local store**; cross-source fallback at fetch time is this gear's `CompositeRateProvider` — ordered sources, first whole successful document, true-source provenance (§3.2). |
+| Provider onboarding without code change | Plugin-discovery source assembly: the core gear discovers registered source-plugin instances by `vendor`, ordered by `priority`; a plain REST feed is onboarded by configuring `bss-rate-provider-http-json-plugin` alone (no code), a new provider *family* costs one new plugin crate (§3.2). |
+
+#### NFR Allocation
+
+| NFR theme | Allocated to | Design Response |
+|-----------|--------------|-----------------|
+| Post-path isolation (hard) | Consumption model | `fetch_latest` is called only by the background `RateSyncJob`, never on the posting path; a provider outage fails the job (ledger alarms), never a post. |
+| Feed freshness | Fetch path + ledger tick | A successful fetch SHOULD complete within one `rate_sync_tick` (ledger default 1 h) so G10 pairs never cross the 24 h staleness window under normal operation. |
+| Fetch latency | Sources + HTTP client | `fx_provider_fetch_duration_seconds{provider}` p95 ≤ 2 s **per source** (draft; confirm against ECB response times). One bounded attempt per source per call, no unbounded retry. The composite's worst case is the **sum of the configured per-source `timeout_ms`** (every source down). No shared total deadline is imposed in v1 — see the budget rule below for why that is safe and where it stops being safe. |
+| Availability | Ledger fail-safe | Best-effort; the ledger's fail-safe (block, not guess) absorbs adapter downtime. |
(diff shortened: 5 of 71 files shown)


gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+### 1.2 Background / Problem Statement
+
+The ledger's multi-currency posting (ledger PRD § Money, Rounding & Foreign Exchange)
+requires a live rate feed to translate transaction currency into functional currency, but
+the ledger deliberately declares the provider integration out of its own scope — it ships
+only the consuming seam (`RateProviderV1`, `RateSyncJob`, the local rate store, and the
+fail-safe that blocks FX posts when no rate is available). Until an adapter implements
+that seam, every FX post blocks (`FX_RATE_UNAVAILABLE`). This gear closes that gap.
+
+### 1.3 Goals (Business Outcomes)
```
ainetx, line 83: The PRD states that deploying the adapter with one configured source enables cross-currency posting, while §12 and DESIGN’s companion dependency clarify that non-EUR-functional tenants still require ledger-side triangulation.

Please either constrain the acceptance criterion to EUR-functional/direct-pair scenarios or make the companion ledger change an explicit readiness condition for non-EUR tenants.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+## 12. Risks
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| ECB feed outage with no fallback configured (v1 is ECB-only) | FX posts block until feed recovers | Fail-safe by absence + ledger alarm; fallback mechanism is config-ready (deploy/configure another source-plugin gear) |
+| EUR-based pairs only until ledger triangulation/inversion lands | Non-EUR-functional tenants cannot post FX | Sequencing tracked as a hard companion dependency (DESIGN §4) |
+| Provider format drift breaks parsing | Fetch fails, feed goes stale | Whole-document parse failure surfaces as an error → fallback / alarm, never partial data |
+
+## 13. Open Questions
+
```
ainetx, line 579: Banker’s rounding is specified as mandatory p1 behavior with deterministic acceptance criteria, but Finance/audit sign-off remains an open question.

Please clarify whether that sign-off is a release gate. If it is intentionally deferred, mark half-to-even as the accepted provisional default and identify the event that could still revise it.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+only the consuming seam (`RateProviderV1`, `RateSyncJob`, the local rate store, and the
+fail-safe that blocks FX posts when no rate is available). Until an adapter implements
+that seam, every FX post blocks (`FX_RATE_UNAVAILABLE`). This gear closes that gap.
+
+### 1.3 Goals (Business Outcomes)
+
+- **Unblock multi-currency billing**: with the adapter deployed and one source configured,
+  cross-currency invoice posting works without manual rate seeding.
+- **Auditable rates**: every synced rate is traceable to a named provider and its original
+  publication timestamp — no fabricated or silently stale values.
```
ainetx, line 91: “Cheap provider onboarding” is a stated business outcome, but the acceptance criteria cover configuration failures rather than the successful onboarding path.

Please add an acceptance scenario showing that a new plain REST feed can be configured through the existing http-json plugin, placed in fallback order, and picked up without changing or redeploying consuming modules.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+  "a key that would never be sent" are **config-load (deserialization) errors** rather than
+  runtime checks in `init()`. Invalid states are unrepresentable instead of validated.
+- **As implemented — narrower than the original wording:** there is no single "source
+  list" or "source kind" anymore (see DESIGN.md's Implementation-revision note and §3.2),
+  so **an unknown source kind and an empty source list are no longer things to validate at
+  all** — a `RateProviderV1` source is a plugin crate, not a config value, so there's no
+  "unknown kind" to reject. Zero matching source plugins for the configured vendor is now a
+  **first-fetch runtime error**, not a startup failure (§3.2's "Empty result is a runtime
+  error" — deliberately, so plugin registration order never has to be enforced). A
+  `vendor` mismatch between the core gear and a source plugin is not rejected anywhere
```
ainetx, line 311: A core/source `vendor` mismatch is silently filtered with no log line, even though operator needs include clear configuration failures and fetch observability.

Please add a diagnostic signal for this case—at minimum a warning or metric identifying that source instances were discovered but excluded by vendor mismatch. Otherwise a simple typo appears only as a generic “no matching sources” first-fetch failure.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+**Alternative Flows**:
+- **All sources fail**: the gear returns the last error; the ledger alarms and FX posts
+  block (fail-safe by absence).
+
+## 9. Acceptance Criteria
+
+- [ ] With the adapter deployed and ECB configured, one ledger sync tick populates the
+  ledger's rate store and a cross-currency invoice post locks a rate (no
+  `FX_RATE_UNAVAILABLE`).
+- [ ] With all sources down, FX posts block, the feed-freshness alarm fires, and non-FX
```
ainetx: Severity: MAJOR

This acceptance criterion implies that an upstream outage immediately blocks FX posts. Elsewhere, the PRD and DESIGN place the provider off the posting path and assign rate persistence and staleness handling to the ledger.

Please state the actual blocking condition—for example, all sources are unavailable *and* the ledger has no still-valid stored rate, or its available rate has crossed the staleness threshold. Otherwise the criterion can produce incorrect tests and outage expectations.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+All metrics exposed as Prometheus scrape targets. (Provider **fallback** selection is
+measured ledger-side as `ledger_fx_provider_fallback_total{provider}`, emitted at lock time
+by `RateSource`; the adapter measures the fetch itself.)
+
+| Vector | Metric | Description | Target Threshold |
+|--------|--------|-------------|------------------|
+| **Efficiency** | `fx_provider_fetch_duration_seconds{provider}` | Outbound fetch+parse latency | p95 ≤ 2 s |
+| **Performance** | `fx_provider_rates_returned{provider}` | Pairs returned per successful fetch | — |
+| **Reliability** | `fx_provider_fetch_errors_total{provider,kind}` | Fetch failures by `RateProviderError` kind | — |
+| **Reliability** | `fx_provider_last_success_timestamp{provider}` | Unix time of last successful fetch (feed-freshness gauge) | — |
```
ainetx: Severity: MAJOR

The design covers fetch failures and exposes `fx_provider_last_success_timestamp`, but it does not define an independent watchdog or alert for `RateSyncJob` no longer running. In that failure mode no fetch error is produced, so the documented provider-failure alarm may never fire.

Please define the missed-tick or last-success threshold, alert owner, and acceptance test for scheduler/job liveness.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+
+- Each source-plugin gear's `init()` registers a `PluginV1<RateProviderSourcePluginSpecV1>`
+  instance (carrying its own `vendor` + `priority` in the instance JSON) and a scoped
+  `RateProviderV1` client keyed by that instance's GTS id
+  (`bss_rate_provider_sdk::registration::register_rate_provider_plugin`, shared by every
+  plugin gear including the core gear itself).
+- The core gear's `DiscoveringRateProvider` (built at `init()`, discovery deferred to the
+  first `fetch_latest`) lists all `RateProviderSourcePluginSpecV1` instances, **keeps only
+  the ones whose `vendor` matches its own configured `source_vendor`**, sorts the survivors
+  by `priority` (lower = tried first; a tie is broken by registry list order and logged as
```
ainetx: Severity: MAJOR

The PRD requires sources to be tried in configured order, but equal priorities are resolved by registry list order and only produce a warning. Unless registry order is a stable contract, provider selection—and therefore the applied rate and provenance—can change without any configuration change.

Please reject duplicate priorities or define a deterministic secondary ordering key.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+
+#### NFR Allocation
+
+| NFR theme | Allocated to | Design Response |
+|-----------|--------------|-----------------|
+| Post-path isolation (hard) | Consumption model | `fetch_latest` is called only by the background `RateSyncJob`, never on the posting path; a provider outage fails the job (ledger alarms), never a post. |
+| Feed freshness | Fetch path + ledger tick | A successful fetch SHOULD complete within one `rate_sync_tick` (ledger default 1 h) so G10 pairs never cross the 24 h staleness window under normal operation. |
+| Fetch latency | Sources + HTTP client | `fx_provider_fetch_duration_seconds{provider}` p95 ≤ 2 s **per source** (draft; confirm against ECB response times). One bounded attempt per source per call, no unbounded retry. The composite's worst case is the **sum of the configured per-source `timeout_ms`** (every source down); this is acceptable because the fetch runs only inside the background `RateSyncJob`, never on the posting path — no shared total deadline is imposed in v1. |
```
ainetx: Severity: MAJOR

The design bounds each source attempt but explicitly imposes no shared composite deadline. Because both source count and `timeout_ms` are configuration-driven, the total fetch duration can exceed `rate_sync_tick` and undermine the single non-concurrent ticker assumption.

Please define a composite timeout budget, cap the configured sum of source timeouts, or document how overlapping/missed ticks are prevented when the sum exceeds the scheduling interval.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+|-----------|-------|--------|--------------|
+| `fetch_latest` | `ctx`, `pairs`, `request_id` | `Vec<ProviderRate>` | GET `base_url` (with `auth`) → parse JSON → apply `mapping` → convert each to `ProviderRate` stamped with this source's `id`. `pairs = &[]` ⇒ whole document; a non-empty `pairs` filters **case-insensitively** (matching the ECB source). An entry that fails mapping is skipped (logged with its quote + reason, plus an aggregate count), never fabricated; a document where **zero** entries map ⇒ `RateProviderError::Internal` (behavioral rules below). |
+| `provider_id` | — | `&str` | The configured `id`. |
+| `health` | `ctx`, `request_id` | `()` | **Not overridden** — uses the trait's default, i.e. a full `fetch_latest(&[])`. There is no cheap HEAD/minimal-GET probe for this source (unlike ECB); a health check re-parses the whole document. |
+
+**Behavioral rules:**
+
+- **Base-currency shape.** Many free feeds are single-base. Config states the base; a
+  requested pair whose base ≠ the feed base is **omitted**, never synthesized here (O-3).
+  If filtering leaves nothing at all, that is an `Internal` error, not `Ok([])`, so the
```
ainetx: Severity: MAJOR

Fallback is triggered by any `Err`; there is no retriable-error classification. However, empty requested-pair filtering has contradictory semantics: ECB returns `Ok([])` and suppresses fallback, while HTTP-JSON returns `Internal` and advances to the next source.

Please define one provider-independent contract for an unsupported requested pair and align both source plugins and their tests.

Also, why does v1 intentionally perform no same-source retry before fallback? Please document the rationale and trade-off—latency budget, source diversity, background rescheduling, or another reason—and state whether any error classes may become retryable later.

gears/bss/rate-provider/plugins/ecb-plugin/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "bss-rate-provider-ecb-plugin"
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+description = "BSS FX rate-provider ECB source plugin: fetches ECB daily reference rates and registers as a RateProviderV1 plugin instance"
```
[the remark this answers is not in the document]
bit4flip: Not applicable — this looks like a false positive.

`cf-studio-path` is not a Cargo manifest field. It comes from the repo's
`CLAUDE.md`, where it configures the Constructor Studio tooling, not the
package metadata.

No manifest in this repository sets it:

    $ grep -rl "cf-studio-path" --include=Cargo.toml .
    (0 files)

Adding it under `[package]` would put an unknown key in the manifest, which
Cargo warns about, and would make this one plugin inconsistent with all 74
other crates. Skipping

gears/bss/rate-provider/rate-provider-sdk/src/error.rs
```diff
@@ -0,0 +1,61 @@
...
+        // `base_url` may have an operator-embedded secret spliced into it (no
+        // source here supports query-string API-key auth), so never echo the
+        // URL itself into an error the sync job logs/alarms on — only the
+        // structured `kind` + diagnostic `reason`.
+        HttpError::InvalidUri { kind, reason, .. } => {
+            RateProviderError::Internal(format!("invalid request URL ({kind:?}): {reason}"))
+        }
+        // `HttpError` is `#[non_exhaustive]`; local faults + any future variant
+        // fall through here as an internal fault (never a silent success).
+        other => RateProviderError::Internal(other.to_string()),
```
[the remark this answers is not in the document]
bit4flip: **Not an issue — already handled.**

### `InvalidUri` — the URL is already dropped

```rust
HttpError::InvalidUri { kind, reason, .. } =>
    RateProviderError::Internal(format!("invalid request URL ({kind:?}): {reason}")),
```

The `..` discards `url`; only the structured `kind` and the diagnostic `reason` survive.
Guarded by `invalid_uri_maps_to_internal_and_never_echoes_the_raw_url`.

### The other arms have no URL to leak

| Arm | Wraps | URL in `Display`? |
|---|---|---|
| `Transport` | `hyper::Error`, `hyper_util::client::legacy::Error` | no |
| `Tls` | rustls / TLS errors | no |
| `InvalidScheme` | — | no, scheme only |

Printing the request URL is `reqwest` behaviour, and `toolkit-http` does not use `reqwest`
on this path.

### Conclusion

No `base_url` — or a credential spliced into it — can reach a log line through these arms.
Replacing them with fixed strings would drop useful transport diagnostics without removing
a real risk.

The one open point is the `#[non_exhaustive]` wildcard: a future variant could carry a URL.
That deserves a code comment, not a change today.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+  the original (pre-plugin-pattern) design, which required a startup-time empty-list check
+  — there is no equivalent of "unknown `kind`" or "empty `sources[]`" to validate anymore,
+  since there is no `kind` and no list.
+- **Discovery re-runs on every tick, not just the first.** Neither a successful nor a
+  failed discovery is cached across calls — every `fetch_latest`/`health` re-lists
+  instances and rebuilds the composite from scratch. A source plugin that registers,
+  deregisters, or changes its `priority` between ticks takes effect on the very next tick,
+  never requiring a gear restart (§3.8 "Deployment Topology"). A *failed* discovery (zero
+  matches) leaves the previous successfully-discovered composite in place rather than
+  clearing it, so `provider_id()` keeps reporting the last real upstream through a
```
ainetx: Severity: MAJOR

This paragraph says the previous composite is retained so `provider_id()` keeps reporting the last real upstream. That conflicts with the component contract below, where the composite `provider_id()` is the core gear’s constant id and true-source provenance is carried per rate.

Please remove or correct the stale last-upstream statement. Leaving both semantics documented risks recreating the call-order/concurrency provenance bug that O-7a resolved.

gears/bss/rate-provider/docs/PRD.md
bit4flip: duplicates https://github.com/constructorfabric/gears-rust/pull/4333#discussion_r3678185976

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+  that specific control is not implemented; a misconfigured `base_url` pointed at an internal
+  host is only caught by network reachability at fetch time, not rejected up front. (An
+  earlier draft of this document claimed such a check existed "unless explicitly allow-listed
+  for dev" — it does not, and there is no allow-list config either.)
+- **Redirect safety (this part IS implemented, inherited, not gear-specific code):** the
+  shared `toolkit-http` client's *default* `RedirectConfig` (`same_origin_only: true`,
+  `strip_sensitive_headers: true`, `allow_https_downgrade: false`) is what every source
+  plugin gets by not overriding it — cross-host redirects are refused unless allow-listed,
+  `Authorization`/`Cookie` are stripped on any redirect that does cross an origin, and an
+  HTTPS→HTTP downgrade is blocked. Config is operator-supplied (not tenant input), so this
```
ainetx, line 1070: Severity: MAJOR

The design notes that `SecretString` protects credentials only after deserialization, while raw effective-config dumps may still expose them. However, it does not establish a normative mitigation or assign ownership for it.

Could we require raw configuration output to redact or prohibit credential fields and identify the component responsible for enforcing this before production readiness? Otherwise, a routine diagnostic operation could silently disclose provider credentials.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,472 @@
...
+
+- **Rationale**: The ledger requirement `cpt-cf-bss-ledger-fr-multi-currency-fx`
+  (ledger PRD § Multi-currency & FX) needs a live feed; this gear is that feed.
+- **Actors**: `cpt-cf-bss-rate-provider-actor-ledger-rate-sync`
+
+#### Provider publication time
+
+- [ ] `p1` - **ID**: `cpt-cf-bss-rate-provider-fr-provider-publication-time`
+
+Every returned rate **MUST** carry the provider's original publication timestamp (UTC) —
```
ainetx, line 199: Severity: MAJOR

The contract requires the original provider publication timestamp and correctly rejects missing or malformed values, but it does not define how timestamps materially ahead of the current time are handled.

A future-dated upstream document could bypass the ledger’s age-based staleness protection and appear fresh for an extended period. Could we define the maximum accepted positive clock skew, assign validation ownership to either the Rate Provider or Ledger, and require fail-safe rejection plus an acceptance test for values beyond that bound?

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+  duplicate it is). Dropping every entry surfaces the documented empty-table `Internal`
+  error rather than an empty `Ok`.
+- **A currency row is bound to its own `<Cube time=…>` block.** Rows under a second,
+  non-retained date — or under a `time` that failed to parse, or with no enclosing date at
+  all — are omitted rather than folded into the retained date's table under the wrong
+  `as_of`. An unparseable `time` is logged with its raw value, so the eventual
+  "missing publication date" error isn't the only clue. A `Cube` carrying only one of
+  `currency`/`rate` (a truncated feed row) is likewise logged and ignored — no anomaly in
+  this parser is dropped silently.
+- **An unconvertible rate is skipped per entry, not fatal for the document** — matching the
```
ainetx, line 743: Severity: MAJOR

The architecture describes source selection as “first whole document” and “all-or-nothing per source,” but the parsing rules allow individual invalid rates to be skipped while the remaining document is treated as successful.

This can prevent fallback even when the primary source returns only partial coverage, leaving affected pairs stale in the ledger without presenting the tick as a source failure. Could we clarify which semantic is normative and define the required coverage for a successful document—for example, requiring all requested pairs or an explicit minimum expected set before accepting the source?

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+Primary source fails (`Err` for the whole fetch) → the composite tries the next source in
+priority order → the first `Ok(document)` is returned **whole**, with every rate in it
+already stamped `provider = <that source's id>` → the ledger stores that id per row, so the
+synced rows record the true upstream regardless of how calls from other consumers interleave.
+
+#### All sources fail → ledger alarm
+
+- [ ] `p1` - **ID**: `cpt-cf-bss-rate-provider-seq-all-sources-fail`
+
+Every source returns `Err` → the composite returns the last `RateProviderError` → the
```
ainetx, line 989: Severity: MAJOR

When every source fails, the composite returns only the last source’s error. This makes the caller-visible failure depend on source ordering and can hide the primary or most actionable cause of the outage.

Could we either return a structured aggregate containing every attempted source and error, or define a normative observability contract that preserves all attempts under the same request ID and exposes that correlation in the ledger alarm? Per-source counters alone may not be sufficient to reconstruct a specific failed tick.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+  of the stored EUR→X rate is part of the ledger's triangulation (O-3; §4 "Companion
+  ledger change").
+- **Currency codes are gated on the ISO-4217 shape.** The `currency` XML attribute arrives
+  verbatim from the remote feed and becomes a tenant-store column value, so a code that
+  isn't three ASCII letters is dropped and logged; survivors are uppercased, and dedup keys
+  off the normalized code (so a feed publishing both `USD` and `usd` is caught as the
+  duplicate it is). Dropping every entry surfaces the documented empty-table `Internal`
+  error rather than an empty `Ok`.
+- **A currency row is bound to its own `<Cube time=…>` block.** Rows under a second,
+  non-retained date — or under a `time` that failed to parse, or with no enclosing date at
```
ainetx: Severity: MAJOR

The ECB parser accepts a document containing multiple distinct publication dates, keeps whichever date appears first, and only logs the ambiguity.

Since the feed contract is a single daily rate document, accepting an order-dependent date can associate rates with the wrong `as_of`, affecting ledger staleness decisions and audit provenance. Could we treat multiple distinct publication dates as a source-level parse failure and trigger fallback, while still allowing duplicate occurrences of the same date if needed?

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1005 @@
...
+**No `format` field exists.** v1 ships **XML-only** (`parse_ecb_xml`) — there is no `format`
+config knob and no SDMX parser in the implementation; O-2's "SDMX optional" / "Frankfurter
+allowed for dev" were never built. Add a `format` field only if a non-XML feed is actually
+needed.
+
+| Operation | Input | Output | Key Behavior |
+|-----------|-------|--------|--------------|
+| `provider_id` | — | `&str` | Returns the configured stable id. |
+| `fetch_latest` | `ctx`, `pairs: &[CurrencyPair]`, `request_id` | `Vec<ProviderRate>` | GET latest feed → parse → convert, stamping each rate with this source's `id`. `pairs = &[]` ⇒ return the **whole** published table. Requested pairs the source cannot serve are **omitted** (not an error). Map transport failures to `Unreachable` / `UpstreamStatus`. |
+| `health` | `ctx`, `request_id` | `()` | Overrides the trait's default (which would delegate to a full `fetch_latest(&[])`) with a cheap `HEAD` probe against the same feed URL — never re-parses the published table just to check reachability. |
```
ainetx: Severity: MAJOR

The ECB implementation reports `health()` from a `HEAD` response without fetching or parsing the rate document. It can therefore remain healthy while every real sync fails because the body is malformed, empty, or otherwise unusable. Other source plugins may inherit a full-fetch health check, so the same interface also provides different guarantees by vendor.

Could we define `health()` explicitly as reachability-only and require a separate readiness/fetch-freshness signal with an alert threshold, or make readiness validate that a usable document can actually be fetched and parsed? Otherwise the platform may show a green provider during a sustained rate-feed outage.

gears/bss/rate-provider/docs/PRD.md
bit4flip: Named in all four places: **a change to the ledger's rounding requirement.**
Nothing else revises it. If that platform default ever moves, this adapter follows it — the dependency runs one way, and a decision to deviate cannot start here.

Docs only — no behaviour change.

gears/bss/rate-provider/docs/PRD.md
bit4flip: Docs only, no behaviour change — the capability itself was already implemented and tested. `tests/discovery.rs` covers per-tick rediscovery, `priority` ordering, and self-heal after a failed tick. What was missing was the criterion that ties those tests to the business goal they satisfy.

Cargo.toml
```diff
@@ -110,6 +110,10 @@ members = [
     "gears/llm-gateway/llm-gateway-sdk",
     "gears/bss/ledger/ledger",
     "gears/bss/ledger/ledger-sdk",
+    "gears/bss/rate-provider/rate-provider-sdk",
+    "gears/bss/rate-provider/rate-provider",
+    "gears/bss/rate-provider/plugins/ecb-plugin",
+    "gears/bss/rate-provider/plugins/http-json-plugin",
```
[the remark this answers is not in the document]
bit4flip: fixed

gears/bss/rate-provider/rate-provider-sdk/src/http_client.rs
```diff
@@ -0,0 +1,50 @@
...
+        uri.scheme_str() == Some("https"),
+        "base_url must use the https scheme"
+    );
+    anyhow::ensure!(uri.authority().is_some(), "base_url must include a host");
+    Ok(HttpClient::builder()
+        .deny_insecure_http()
+        .timeout(Duration::from_millis(timeout_ms))
+        .with_otel()
+        .build()?)
+}
```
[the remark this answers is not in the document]
bit4flip: fixed

apps/cf-gears-example-server/Cargo.toml
```diff
@@ -30,6 +30,14 @@ otel = ["toolkit/otel"]
...
 static-idp = ["dep:static-idp-plugin"]
 file-storage = ["dep:file_storage"]
+# The FX rate-provider subsystem: the core gear + its source plugins. Independent
+# of bss-ledger — the ledger discovers the composite lazily at runtime via the
+# types-registry, so neither feature requires the other.
+bss-rate-provider = [
+    "dep:bss_rate_provider",
+    "dep:bss_rate_provider_ecb_plugin",
+    "dep:bss_rate_provider_http_json_plugin",
+]
```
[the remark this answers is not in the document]
bit4flip: **Not applicable — `cf-studio-path` is not a Cargo manifest key.**

It's a Constructor Studio *project marker*. It is declared once at the repo root, in
`AGENTS.md:3` / `CLAUDE.md:3` (inside the generated `<!-- @cf:root-agents -->` block),
and is then used as a path variable: `.cf-studio/config/AGENTS.md` documents
`{cf-studio-path}/config` as the "root path for Constructor Studio Adapter navigation",
and `.cf-studio/whatsnew.toml` calls it a "project marker ... when resolving
project-local Studio installs". It says where the Studio config directory lives — it is
not a property of a crate.

The finding's own evidence points the other way: all 95 tracked `Cargo.toml` files lack
the key, because none should have it. Adding it to one manifest would make that manifest
the sole outlier, not bring it into line.

It would also regress the build. Cargo does not recognise the key — on the pinned
1.97.0 toolchain a manifest containing it emits:

    warning: unused manifest key: cf-studio-path

on every `cargo check` / `build` / `clippy` (verified on a scratch crate; the warning
disappears when the key is removed).

No change made.

Cargo.toml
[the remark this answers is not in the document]
bit4flip: **Not applicable — `cf-studio-path` is not a Cargo manifest key, and no such guideline exists.**

It is a Constructor Studio *project marker*, declared once at the repo root in
`AGENTS.md:3` / `CLAUDE.md:3` (inside the generated `<!-- @cf:root-agents -->` block).
Everywhere else it appears only as a path variable — `{cf-studio-path}/config/rules/...`,
`{cf-studio-path}/.core/requirements/...`. I searched `.cf-studio/config/rules/`, `docs/`,
`.github/`, and `.cf-studio/config/pr-review.toml`: nothing ties `cf-studio-path` to
`Cargo.toml`, so there is no `**/Cargo.toml` guideline to satisfy. 0 of the 95 tracked
`Cargo.toml` files contain the key.

The cited insertion points would break the build (verified on the pinned 1.97.0 toolchain):

- `Cargo.toml#L113-L116` is *inside* the `[workspace] members = [ ... ]` array. A
  `key = value` there is not valid TOML:
  `error: unexpected = in array, expected value, ]` — the root manifest stops parsing
  and the entire workspace fails to build, not just one crate.
- `ledger-sdk/Cargo.toml#L3` is inside `[package]`, so the key lands as
  `package.cf-studio-path` → `warning: unused manifest key: package.cf-studio-path`.
- `rate-provider-sdk/Cargo.toml` starts with `[package]` → same warning.

The TOML block at the top of the root `CLAUDE.md` is input for the `cfs` tool, not a
template for Cargo manifests. No change made.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,559 @@
...
+  - [5.2 Sources & Fallback](#52-sources--fallback)
+- [6. Non-Functional Requirements](#6-non-functional-requirements)
+  - [6.1 Gear-Specific NFRs](#61-gear-specific-nfrs)
+  - [6.2 NFR Exclusions](#62-nfr-exclusions)
+- [7. Public Library Interfaces](#7-public-library-interfaces)
+  - [7.1 Public API Surface](#71-public-api-surface)
+  - [7.2 External Integration Contracts](#72-external-integration-contracts)
+- [8. Use Cases](#8-use-cases)
+  - [Fallback serve with provenance](#fallback-serve-with-provenance)
+- [9. Acceptance Criteria](#9-acceptance-criteria)
```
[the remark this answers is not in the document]
bit4flip: fixed

gears/bss/rate-provider/docs/PRD.md
bit4flip: Fixed — the case is no longer silent, though I went with a slightly different signal than a warning; 

Two things now happen. A vendor-filtered instance is logged, carrying the instance id, the vendor it advertised and the vendor that was expected — so the exclusion is visible and attributable, not dropped on the floor. And the failure message is no longer generic: instead of "no matching sources" it names the expected vendor, how many source-plugin instances were actually in the registry, which vendors those instances advertised, and the hint that this gear's `source_vendor` has to equal each source plugin's `vendor`. A typo now reads as a one-line diagnosis rather than something you go hunting through YAML for.

The one place I did not follow the suggestion literally is the level: that log is `debug`, not `warn`. Discovery re-runs on every sync tick, and a registry can legitimately hold source plugins belonging to another composite, so a warning per foreign instance per tick would fire continuously in a correctly configured deployment — the noise would train people to ignore it. The signal that something is genuinely wrong is the failure message above, which only appears when nothing matched, and which now carries the mismatch detail. I left the reasoning in a comment at the log site so the choice isn't re-litigated blind later.

gears/bss/rate-provider/docs/PRD.md
bit4flip: Yes, the criterion contradicted the rest of the docs, and it now states the actual blocking condition:

With all sources down, the sync tick fails, the feed-freshness alarm fires, and non-FX posting is unaffected. FX posting is **not** immediately blocked: this gear is off the posting path, and the ledger locks rates from its own store. A post blocks with `FX_RATE_UNAVAILABLE` only once that store has no rate for the pair, or the rate it has has crossed the ledger's staleness threshold (`stale_g10_hours` / `stale_default_max_days`). So the outage window a deployment tolerates is set by the ledger's staleness rule, not by this gear.

The threshold parameters are named explicitly so a test written against this criterion turns the right knob. Docs only - no behavior change.

gears/bss/rate-provider/docs/PRD.md
bit4flip: Fixed — took both halves of your suggestion.

The acceptance criterion is now constrained instead of open-ended: it claims a locked rate only for an invoice post on a directly published pair for an EUR-functional tenant. And posting for a non-EUR functional currency is now called out explicitly as out of scope until the ledger triangulation/inversion companion change lands — "it is not a criterion this gear can satisfy alone" — so that change is stated as a readiness condition for non-EUR tenants rather than left as an implication.

I also made the same boundary read consistently across the document: the goal in §1.3 is scoped to EUR-functional tenants, triangulation is listed as out of scope in §4.2, and the companion change is tracked as a hard p1 dependency in §10 next to the §12 risk you pointed at.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Sources are now ordered by `priority` first, then alphabetically by the plugin's GTS instance id. Registry order no longer decides anything, so the serving source and its provenance are a pure function of configuration.

Duplicate priorities are still accepted (a deployment shouldn't fail over two plugins both left at `100`), but the warning now says the tiebreak is the instance id and that distinct priorities are how you choose deliberately — the tiebreak is deterministic, not meaningful.

Covered by a test that registers the tied sources in both registry orders and asserts the same one serves.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Fixed - added a ledger-side heartbeat so liveness no longer depends on fetch errors.

`ledger_fx_rate_sync_ticks_total` increments once per `RateSyncJob` tick, before the provider is resolved. Threshold: no increase over 3 × `fx.rate_sync_tick_secs` (`FxRateSyncStalled`), plus a separate absent-series alert for the fresh-deploy case (`FxRateSyncNeverRan`). Owner for both is platform on-call. Acceptance test is specified in the ledger design: stop the ticker and assert the heartbeat stops while no fetch error is recorded and no `fx-snapshot-missing` alarm fires.

Because the tick is counted before discovery, the heartbeat also separates two states error-based alerting could not: flat means the scheduler is dead, rising with a flat fetch count means discovery yields no sources. Your point about `fx_provider_last_success_timestamp` is documented explicitly - it is a feed-freshness gauge and keeps its own alert with a threshold set by publication cadence.

Definitions live in the ledger design (the job lives there); the rate-provider design now references them. The alert rules themselves remain deployment-side - the repo holds the metric and the spec.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Unsupported requested pair has one provider-independent answer: `PairUnavailable`, returned by both plugins, case-insensitively, as an `Err` so the composite advances. `Ok([])` is unreachable — ECB rejects a zero-entry document at parse time and http-json errors on zero mappable rates, so no source can pass an empty table off as a complete one. The `Internal` you describe was the earlier http-json behaviour and was replaced for exactly the reason you give, plus one more: its `internal` error-kind label made a routine unsupported pair indistinguishable from a real defect on the fetch-error metric. Both plugins' tests assert the shared contract.

The no-retry rationale is documented in DESIGN under "Why no same-source retry before falling back", and it names the three trade-offs you list: the chain already retries against a *different* upstream (source diversity beats attempt count for feed outages, rate limiting and bad publisher deploys); the next tick re-runs the whole chain within the sync interval, far inside the 24 h G10 staleness window, so a transient failure costs a refresh rather than a posting outage; and one attempt per source keeps the worst case at the sum of the per-source timeouts instead of multiplying it by the retry count.

Future retryability is stated there too: `UpstreamStatus` with `429`/`503` and a `Retry-After` inside the tick budget is the narrow case worth adding; `Unreachable` should keep falling through, since that is precisely when another source is the faster answer; `PairUnavailable`, `InvalidPair` and `Internal` are never retriable.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Fixed - removed the stale sentence.

The code matches that first statement: nothing is retained between calls, and `provider_id()` returns the gear's own constant configured id, so it never reported an upstream — last or otherwise. The paragraph now states what actually happens on zero matches: the tick fails, the ledger alarms, and the next tick re-runs discovery from scratch. Per-rate provenance stays the only source-identity mechanism, per O-7a.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Addressed with two changes.

**Per-source ceiling.** `build_source_http_client` now rejects a `timeout_ms` above 30 s at the plugin's `init()`. Sources are fetched sequentially, so one misconfigured source could otherwise stall a whole pass. The *sum* stays unchecked on purpose: the timeouts live in each source plugin's config, the interval in the ledger's, so no single `init()` sees both halves of the rule.

**The pass is now measured.** `ledger_fx_rate_sync_duration_seconds` records the wall time of a full `RateSyncJob` pass — discovery plus every sequential source attempt — on success and failure alike. This is the series that can be compared against `fx.rate_sync_tick_secs`; the adapter's per-source histogram cannot, since it times one HTTP round-trip. `FxRateSyncOverrunning` warns when its p95 passes half the interval.

**Overlap was never possible.** `spawn_rate_sync_ticker` awaits the job inline with `MissedTickBehavior::Delay`, so a second pass cannot start while one runs and missed ticks do not queue. An overrun delays the next refresh; it never produces concurrent fetches. Between the tick heartbeat (job alive), the new duration histogram (job keeping up), and `FxNoFetchAttempted` (fetches stopped entirely), all three states are now distinguishable.

gears/bss/rate-provider/docs/PRD.md
bit4flip: **Bound: 26 hours ahead of our own clock.** Derived, not picked: 14 h for the widest civil timezone offset (UTC+14), since a date-only feed like ECB is anchored at midnight UTC and can legitimately read ahead of us, plus 12 h of allowance for a badly-synchronised host, so a local NTP failure degrades into wrong-but-served rather than a feed-wide outage.

**Owned by the source plugin, checked at parse time before any rate is built.** Deliberately not the ledger: once a rate is a stored row, the ledger cannot tell "the feed published a bad timestamp" from "we have not synced in a while" — only the plugin sees the value arrive off the wire on this tick.

**Fail-safe.** A rejection fails the whole document, so the composite falls through to the next source and nothing is stored. It counts as a fetch error in the metrics, and the message names both the offending timestamp and the bound it broke.

**Timestamps in the past are deliberately not rejected here.** Age is the ledger's staleness policy; duplicating it in the adapter would put one rule in two layers. This guard closes only the direction an age check cannot see — a future date has a negative age, so it would otherwise read as permanently fresh.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Both halves of this are now in place — the second option you offer (a normative observability contract) was the design's choice, and the one gap in it is fixed.

**Structured aggregate: deliberately not taken.** Widening `RateProviderError` into a per-source list would push that list through a ledger-owned SDK type that maps to a single alarm anyway. The decision and its reasoning are recorded in the design's all-sources-fail sequence.

**The contract instead.** Every failed source attempt is logged at warn before the next source is tried, carrying the attempted source's id, the tick's request_id, and the error. The request_id is created by the ledger's `RateSyncJob` and passed unchanged through the composite to each source, so all attempts of one tick share it, and filtering the adapter's logs by that one id reconstructs the full sequence in priority order — including the sources whose errors the returned value discarded.

**What was genuinely missing, and is now fixed.** The alarm did not carry that id, so it was an entry point to nothing: its provider field is the composite's id rather than the failing source, and its error is the last source's. `FX_SNAPSHOT_MISSING` now carries the tick's request_id in its detail, as does the ledger's failure log line, with a pointer to what filtering by it yields. The chain is alarm -> request_id -> every attempt of that tick. The design's contract now requires this of the ledger rather than leaving the alarm side implicit, and a test pins the alarm wording.

You are right that per-source counters are not sufficient — `fx_provider_fetch_errors_total{provider,kind}` counts failures over time but cannot reconstruct a specific tick. That is exactly the split: counters for trends, the correlated log lines for one incident, and the alarm as the way in.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: Current behaviour of the ECB parser:

- A second **distinct** publication date fails the whole document with a source-level error naming both dates, so the composite falls through to the next source and nothing is stored.
- A repeat of the **same** date is accepted — only a distinct date is a conflict.
- An unparseable date means that `Cube` is ignored with a warn; if it was the feed's only date, the parse then fails as "missing publication date" rather than vanishing.
- Currency rows are bound to the date block they sit in: a row after that block closes, or under a `time` that failed to parse, is omitted rather than stamped with an `as_of` that is not its own.


Covered by tests: a second distinct date fails, a repeated identical date does not, rows after the date block are omitted, rows under an undated inner block are omitted, and an unparseable date surfaces as missing.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: **How it currently works:**

The composite tries sources in `priority` order and takes the **first one that returns `Ok`** — that document whole, never merged with another source's. Fallback is triggered only by an `Err`.

Inside a source, the parser drops individual bad entries and returns the rest as `Ok`: a malformed currency code, a duplicate after normalization, a truncated row (a `Cube` carrying only one of `currency`/`rate`), a row with no enclosing `<Cube time=…>`, a non-numeric rate value. Every skip is logged at `warn` with its reason, plus an aggregate `skipped=N`.

`Err` — and therefore fallback — happens only on: a transport failure, a non-2xx status, a whole-document parse failure, two different publication dates in the ECB feed, `Internal` when **every** entry failed conversion, and `PairUnavailable` when a non-empty requested-pairs list matched nothing the source publishes.

There is no coverage check anywhere in the code — neither a pair count nor a minimum expected set. So 49 pairs instead of 50 is an `Ok`, the composite stops at that source, and the ledger records 49.

On the real call path there are no requested pairs at all: `RateSyncJob` calls `fetch_latest(&[])` — "return the whole published table".

A pair that wasn't refreshed stays in the store with its old `as_of`, crosses the ledger's staleness window, and yields `FX_RATE_UNAVAILABLE` — the post is blocked, never priced at a wrong rate.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: `health()` no longer has a default implementation.

Reachable is defined as **the endpoint answered**: any HTTP response, `4xx` and `5xx` included, since a `503` proves the request arrived. Only a transport failure — DNS, connect, TLS, timeout — is an error. 

The previous ECB implementation did not actually match its own doc here: it required a `2xx`, which answers "is the feed serving correctly?" rather than "did anything get through". That rule is also what makes a HEAD probe viable for an arbitrary configured feed, since a JSON API commonly answers `405` to HEAD while serving GET.

Freshness stays a separate signal, as before: `fx_provider_fetch_errors_total{provider}`, and `FxFeedStale` over `fx_provider_last_success_timestamp{provider}`, thresholded against the feed's publication cadence rather than the sync interval. Neither can be satisfied by a HEAD. 

Tests pin the rule from both sides on each source: every HTTP status is reachable, a dead port is not, and a GET-only server covers the 405 case directly.

gears/bss/rate-provider/docs/DESIGN.md
bit4flip: The normative rule and the owner added in DESIGN §2.2: raw configuration output must redact credential-shaped fields, it must be enforced inside the dump (a gear cannot do it — the dump runs before typed config exists), and the owner is the platform / `toolkit` maintainers. Tracked as a risk in PRD §12; not exploitable in v1, since ECB is a public feed with no credential.

Your comment did surface a real gap on our side: the `${VAR}` path the design mandates was not actually wired. The plugin loaded config via `ctx.config()` and the config type had no `ExpandVars`, so `api_key: "${BANK_X_KEY}"` would be sent as that literal string — pushing operators to hardcode the key, i.e. straight into the leak path. Fixed: `ctx.config_expanded()` plus a hand-written `ExpandVars` for the `Auth` enum (the derive covers only named-field structs); an unresolvable variable fails `init()`.

gears/bss/ledger/ledger/src/module.rs
```diff
@@ -186,6 +195,88 @@ impl Default for BssLedgerGear {
...
+            tracing::debug!(
+                vendor,
+                error = %e,
+                "bss-ledger: no FX rate-provider plugin instance matched; using the \
+                 fail-safe default"
+            );
+            fallback()
+        }
+    }
+}
```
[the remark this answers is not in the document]
bit4flip: **Rejection:**

Both factual premises here don't hold.

The only production implementation of `TypesRegistryClient` is the in-process `TypesRegistryLocalClient` (registered via `ClientHub` in `types-registry/src/gear.rs`); its `list_instances` is a synchronous in-memory `service.list()` — no network, no await points, nothing that can hang. A `tokio::time::timeout` around an in-memory lookup guards against nothing and adds a dead failure path.

"Before the metric bucket starts" is not what the code does: `started = Instant::now()` (`module.rs:355`) precedes `resolve_rate_provider` (`:358`), so the duration histogram covers discovery — that placement is commented at the call site. And the tick heartbeat increments before discovery (`:347`), so even a hypothetically stuck discovery goes flat on `ledger_fx_rate_sync_ticks_total` and `FxRateSyncStalled` fires; the loop awaits the pass inline with `MissedTickBehavior::Delay`, so no overlap either way.

gears/bss/rate-provider/plugins/http-json-plugin/src/infra/source.rs
```diff
@@ -0,0 +1,322 @@
...
+                // for real defects. Still an `Err`, so the composite moves on to
+                // a source that may publish the pair.
+                if rates.is_empty()
+                    && let Some(first) = pairs.first()
+                {
+                    return Err(RateProviderError::PairUnavailable {
+                        base: first.base.clone(),
+                        quote: first.quote.clone(),
+                    });
+                }
```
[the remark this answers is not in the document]
bit4flip: **Rejection:**

This proposes reversing a settled contract rather than fixing a defect. DESIGN §2.1 "Direct pairs only" is explicit: a requested pair the source cannot serve is **omitted**; `PairUnavailable` fires only when a non-empty `pairs` list matches nothing. Both plugins implement and test exactly that, and coverage is deliberately not a success criterion (O-13) — settled in the partial-coverage review thread on this PR.

The failure mode is already fail-safe without merging: a pair that wasn't refreshed keeps its old `as_of`, ages past the ledger's staleness threshold, and blocks the post with `FX_RATE_UNAVAILABLE` — never priced at a wrong rate. Failing the whole document over one unserved pair would do the opposite of what the chain is for: discard every good rate the source did publish and stall the refresh for all other currencies (the exact stall O-13 avoids), while "advancing" to a source that most likely doesn't publish the exotic pair either.

It is also off the real call path: the ledger's `RateSyncJob` always calls `fetch_latest(&[])` — whole table, no requested pairs at all.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,579 @@
...
+- [ ] Ledger rate rows record the source that actually served each rate, and stay correct
+  even when a second consumer fetches from the same adapter concurrently.
+- [ ] A currency code the feed publishes that is not an ISO-4217-shaped identifier is
+  dropped and logged, never written into a tenant's rate store.
+- [ ] A feed serving a document dated more than 26 hours ahead of the gear's clock fails
+  that source's fetch with an error naming the offending timestamp; nothing from it is
+  stored, and the composite falls through to the next source. A document dated up to a full
+  day ahead — the legitimate case for a date-only feed anchored at 00:00 UTC — still serves.
+  Both bounds are asserted, because a test for the far-future case alone would pass with the
+  window set arbitrarily narrow.
```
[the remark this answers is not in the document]
bit4flip: **Rejection:**

The requested cases already exist in `publication_time_tests.rs`, on a fixed controlled clock (`now()` is a constant timestamp): `a_publication_time_exactly_at_the_bound_is_accepted` (`now + 26 h`, the bound is inclusive) and `one_second_past_the_bound_is_rejected` (`now + 26 h + 1 s`), alongside the 24 h-ahead accept and the far-future reject. A 24- or 25-hour implementation fails the exactly-at-the-bound test; a looser-than-26 h one fails the one-second-past test — the bound is pinned from both sides. The PRD criterion is a summary of that suite, not its full enumeration.

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1317 @@
...
+
+| State | `fetch_duration_seconds_count` | `fetch_errors_total` |
+|---|---|---|
+| Healthy | rising | flat |
+| Upstream outage | flat | **rising** |
+| **Nothing is attempting a fetch** | flat | flat |
+
+| Alert | Condition | Owner | Rationale |
+|-------|-----------|-------|-----------|
+| `FxNoFetchAttempted` | `rate(fx_provider_fetch_duration_seconds_count[3 × fx.rate_sync_tick_secs]) == 0 and rate(fx_provider_fetch_errors_total[3 × fx.rate_sync_tick_secs]) == 0` | Platform on-call | "No attempt completed at all", which is what separates a dead feed from a slow one. Read **together with** the ledger's tick heartbeat: with the heartbeat rising, this narrows to "the job is alive but discovery yields no sources". |
```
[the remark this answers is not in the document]
bit4flip: **Rejection:**

The in-flight window can't span the alert window. The alert reads a `3 × fx.rate_sync_tick_secs` range (hours at the default), while every attempt is bounded by the per-source HTTP timeout, capped at `MAX_TIMEOUT_MS = 30 s` at `init()` — an attempt lands in one of the two series within 30 seconds, so "in progress leaves both flat" cannot hold for the window the alert evaluates.

The second half of the suggestion is already the documented contract: the alert's definition says to read it **together with** the ledger's tick heartbeat (`ledger_fx_rate_sync_ticks_total`, incremented before the provider is resolved — before any I/O). That heartbeat *is* the pre-I/O attempt-independent signal; a separate attempt counter inside `fetch_and_parse` would duplicate it one step downstream.

gears/bss/rate-provider/plugins/ecb-plugin/src/infra/source.rs
```diff
@@ -0,0 +1,454 @@
...
+                );
+            }
+            Event::End(e) if e.local_name().as_ref() == b"Cube" => {
+                if retained_depth == Some(depth) {
+                    retained_depth = None;
+                }
+                // `saturating_sub`: a stray `</Cube>` in a malformed document
+                // must not underflow the counter.
+                depth = depth.saturating_sub(1);
+            }
```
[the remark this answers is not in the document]
bit4flip: The scenario is real, but it is an over-omission in a malformed document, not a correctness defect — and the omission is the parser's deliberate bias.

The invariant this code exists to hold — settled in the publication-dates review thread — is that no row is ever stamped with an `as_of` that is not its own. That invariant survives this scenario: the affected rows are dropped, each with a `warn` naming the currency plus an aggregate `skipped` count, and if everything drops the parse fails (`Internal` / missing publication date) so the composite falls through to the next source. Nothing is stored wrong; the worst case is fewer pairs from a corrupted document, visible in logs, with the ledger's staleness rule blocking any post the missing pairs would have priced.

The triggering shape also cannot come from the feed this plugin parses. The daily `eurofxref` document carries exactly one `time` Cube; a nested Cube with a *valid but different* date already fails the whole document (the multiple-publication-dates check), so the only trigger is an *unparseable* `time` attribute sitting inside the retained block *before* valid sibling rows — a malformation that has never been observed and that the real feed's structure has no place for.

A per-depth date stack would make the parser accept more rows from a document that is already demonstrably corrupted. On a money path the current bias — omit and log rather than salvage — is the one we want, so the scalar stays.

gears/bss/rate-provider/rate-provider-sdk/src/fetch_tests.rs
```diff
@@ -0,0 +1,334 @@
...
+    // No response ever arrived, so there is no status to count and no success
+    // instrument may move — a timed-out feed must look failed, never fresh.
+    assert!(
+        !events.iter().any(|e| matches!(
+            e,
+            Event::UpstreamStatus { .. }
+                | Event::Fetch { .. }
+                | Event::RatesReturned { .. }
+                | Event::LastSuccess { .. }
+        )),
```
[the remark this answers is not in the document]
bit4flip: Deliberate, not an omission. `fx_provider_fetch_duration_seconds` measures served documents, not attempts — its `_count` is what the alerting reads, and the documented outage signature (duration flat, errors rising) only holds if failures never touch it. Recording `elapsed()` on the error path can't be done without bumping `_count`, which destroys that separation.

Nothing is lost about the slow attempts either: a timed-out attempt's duration is known in advance — it is `timeout_ms` (capped at 30 s at `init()`) — and the attempt is counted in `fx_provider_fetch_errors_total{kind="unreachable"}`. Slow *failures* are measured where slowness matters: `ledger_fx_rate_sync_duration_seconds` records the full sync pass on success and failure alike.

The tests pin this contract from both sides on purpose: a timeout moves only the error instrument; a success records no error.

gears/bss/rate-provider/docs/PRD.md
```diff
@@ -0,0 +1,579 @@
...
+
+- [ ] `p1` - **ID**: `cpt-cf-bss-rate-provider-interface-rate-provider-v1`
+
+- **Type**: Rust trait implementation (`bss-ledger-sdk` `RateProviderV1`,
+  GTS `gts.cf.bss.ledger.rate-provider.v1`), registered in the platform `ClientHub`.
+- **Stability**: stable
+- **Description**: The gear's only consumable surface — latest-rates fetch, health probe,
+  and serving-provider identity. The health probe is **reachability only**: it reports the
+  endpoint answered, which any HTTP status proves, and says nothing about whether a usable
+  document could be fetched. Feed freshness is a separate signal and MUST NOT be inferred
```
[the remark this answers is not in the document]
bit4flip: Fixed. The RateProviderV1 description no longer lists "serving-provider identity" as a surface: it now states that provider identity arrives as per-rate provenance (`ProviderRate.provider`, the serving source's id) on every returned rate, and that `provider_id()` is only this gear's constant configured id — it never says who served a fetch. The health-probe and latest-rates descriptions are unchanged, per O-7a and the provenance thread on this PR.

gears/bss/ledger/ledger/src/config.rs
```diff
@@ -323,17 +335,27 @@ impl Default for FxConfig {
...
     /// exceeds [`MAX_STALE_G10_HOURS`] (a larger window overflows
-    /// `chrono::Duration::hours` and panics in `rate_source::is_stale`), or if any
-    /// of `stale_g10_hours` / `rate_sync_tick_secs` / `revaluation_run_tick_secs`
-    /// is `0` (a zero staleness window admits any rate; a zero tick panics
-    /// `tokio::time::interval`).
+    /// `chrono::Duration::hours` and panics in `rate_source::is_stale`), if
+    /// `rate_sync_tick_secs` is below [`MIN_RATE_SYNC_TICK_SECS`] (the
+    /// `FxNoFetchAttempted` alert-window bound — see the constant's docs), or if
+    /// `stale_g10_hours` / `revaluation_run_tick_secs` is `0` (a zero staleness
+    /// window admits any rate; a zero tick panics `tokio::time::interval`).
```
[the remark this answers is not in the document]
bit4flip: The premise doesn't hold: sequential fallback does not extend the flat-counters interval. `fetch_and_parse` runs per source, and every failed attempt lands in `fx_provider_fetch_errors_total` on all failure branches the moment it completes — so in "slow source, then fallback to the next", the error counter moves every ≤ 30 s. The interval during which both counters are flat while an attempt is live is bounded by ONE attempt (the 30 s per-source ceiling), not by the chain; a longer chain puts *more* completions inside the window, never fewer. The 180 s window covers the single-attempt bound six times over.

The complete fallback duration is already accounted for where it belongs: the alert's `for:` hold is specified as "at least one full sync pass — the 30 s per-source ceiling × the configured source count", which scales with exactly the quantity you're asking the floor to cover.

A completion metric at the composite `fetch_latest` boundary would reverse the settled instrument-ownership rule (§4): the core gear deliberately emits no instruments of its own — fetch series belong to the source plugins, and the pass-level duration already exists ledger-side (`ledger_fx_rate_sync_duration_seconds`, recorded on success and failure alike).

gears/bss/rate-provider/docs/DESIGN.md
```diff
@@ -0,0 +1,1317 @@
...
+This is one contract for both source plugins. `PairUnavailable` rather than `Internal`
+because it is a routine "not served here", not a fault in this gear — the distinction is
+visible on `fx_provider_fetch_errors_total{kind}`, where these must not be counted as
+`internal` and hide real defects. An `Err` rather than `Ok([])` because an empty document
+reads as a complete answer and would stop the fallback chain on the first source, when the
+next one may publish exactly that pair.
+
+Whole-table calls (empty `pairs`, which is what the ledger's sync job makes) are unaffected:
+with no requested pair there is nothing to report unavailable, so an empty feed stays
+`Ok([])`.
```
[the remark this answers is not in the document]
bit4flip: Fixed — the three stale spots claiming `Ok([])` now state the implemented contract: a non-empty requested-`pairs` list matching nothing is `PairUnavailable` in both plugins; a document with zero usable entries is `Internal`; and `Ok([])` is unreachable even on the whole-table path (ECB rejects a zero-entry table at parse, http-json errors when nothing maps), so the composite's fallback is never suppressed.

ainetx (approved):

diffora (approved):

PR #4339, opened 2026-07-30
(DRAFT)fix(account-management): open the tenant-metadata envelope payload so typed derived schemas can register

### LINKED ISSUE #4338: GTS/AM: typed derived tenant-metadata schemas cannot be registered — OP#12 treats the base envelope as a closed empty object
Summary

AM's extensible tenant metadata (AM PRD §5.7) is designed around GTS-registered derived schemas with validated payloads ("branding, contacts" as the PRD's examples). In practice, a derived metadata schema that declares any typed property cannot be registered: both documented/plausible authoring conventions fail, so the only registrable derived schema is a free-form type: object — which defeats the "GTS-validated payload" feature. Confirmed as "looks like a bug" by Diffora on Discord (2026-07-29).

Environment
gears-rust main @ 71218154 (2026-07-29), also reproduced on the previous release
assembly: cf-gears-example-server-style build (20 gears), SQLite and Postgres profiles
entity seeded via types-registry.config.entities
Reproduction

Add one entity to types-registry.config.entities:

yaml
- "$id": "gts://gts.cf.core.am.tenant_metadata.v1~cf.studio.workspace.settings.v1~"
  "$schema": "http://json-schema.org/draft-07/schema#"
  type: "object"
  properties:
    automation_level:
      type: "string"
      enum: ["manual", "recommendations", "autonomous"]
  x-gts-traits:
    inheritance_policy: "override_only"

Start the server → switch_to_ready fails, gear init aborts:

ERROR types_registry::gear: GTS validation error
  gts_id=gts.cf.core.am.tenant_metadata.v1~cf.studio.workspace.settings.v1~
  Schema '...workspace.settings.v1~' is not compatible with base
  'gts.cf.core.am.tenant_metadata.v1~': property 'automation_level': derived schema
  adds new property but base 'gts.cf.core.am.tenant_metadata.v1~' has
  additionalProperties: false
Error: post-init failed for gear 'types-registry'
Caused by: Failed to switch to ready mode: Ready commit failed with 1 errors

The base envelope (gears/system/account-management/docs/schemas/tenant_metadata.v1.schema.json) declares no properties and no top-level additionalProperties; the OP#12 narrowing check treats it as a closed empty object, so no derived schema may add a single payload field.

Attempt 2 — the convention referenced in AM's metadata_schema_registry.rs comments ("derived metadata schemas carry $schema: gts://gts.cf.core.am.tenant_metadata.v1~"):

GTS validation error ... trait validation failed: failed to compile trait schema:
Unknown meta-schema: 'gts://gts.cf.core.am.tenant_metadata.v1~'. Custom meta-schemas
must be registered in the registry before use

i.e. the convention AM's own comments describe is not supported by types-registry.

Expected

A documented, working way to register a derived tenant-metadata schema with typed, validated payload properties — as promised by AM PRD §5.7 / FR "extensible tenant metadata".

Actual

Every typed derived schema is rejected at registration; only free-form type: object (no payload validation) registers.

Root cause (narrowed down while preparing the fix)

The envelope reaches types-registry with an explicit additionalProperties: false: the upstream gts-macros emitter closes every generated object unconditionally, and its attribute surface offers no override (a #[schemars(extend("additionalProperties" = true))] on the struct is overwritten by the emitter — verified by test). OP#12 itself is fine: it honours an explicit additionalProperties: true (verified by an integration test registering a typed derived schema under an open envelope).

AndrejK666:
### Description
The `gts.cf.core.am.tenant_metadata.v1~` envelope reached types-registry with
`additionalProperties: false`: the upstream gts-macros emitter closes the generated
object unconditionally, and its attribute surface offers no override (a
`#[schemars(extend(...))]` on the struct is overwritten by the emitter — we tried).
The OP#12 chain-narrowing check therefore rejected **any** derived metadata schema
declaring a typed payload property (`switch_to_ready` fails, gear init aborts). That
made the PRD §5.7 promise — "extensible tenant metadata with GTS-validated payloads
(branding, contacts)" — unrealisable: only free-form `type: object` derived schemas
could register. Found while building the Constructor Studio backend; confirmed as a
bug on Discord (2026-07-29). Fixes #<issue>.

The fix, entirely within this repo:

- **toolkit-gts-macros**: new wrapper-only `gts_type_schema` argument
  `open_payload = true` — stripped before forwarding to upstream (which rejects
  unknown attributes); the inventory `schema_fn` post-processes the emitted JSON and
  sets `additionalProperties: true` at the schema root. `toolkit-gts` re-exports
  `serde_json` for the generated code.
- **account-management-sdk**: `TenantMetadataEnvelopeV1` declares
  `open_payload = true`; `docs/schemas/tenant_metadata.v1.schema.json` synced.

Semantically the envelope payload was always meant to be open: derived schemas own
the payload shape, and `MetadataService` validates entries against the **derived**
schema at PUT time (`metadata_schema_registry`) — the base is an envelope, not a
payload contract. Traits stay strict (`x-gts-traits-schema` unchanged,
`additionalProperties: false` inside it). Verified along the way that OP#12 honours
an explicit `additionalProperties: true` — the narrowing rule itself needs no change.

### Type of Change

### Testing

- [x] New tests added:
  - `types-registry/tests/abstract_envelope_tests.rs` — a typed derived schema under
    an open abstract envelope registers and survives `switch_to_ready`;
  - `account-management-sdk` sync test `tenant_metadata_envelope_payload_is_open` —
    asserts the **inventory entry** (what types-registry actually registers) and the
    docs artifact both carry `additionalProperties: true` (the raw
    `gts_schema_with_refs()` accessor intentionally still reflects the upstream
    emitter's closed default — documented in the test).
- [x] Unit tests pass (`cargo test -p cf-gears-toolkit-gts-macros -p cf-gears-toolkit-gts -p cf-gears-account-management-sdk -p cf-gears-types-registry`)
- [x] Manual testing completed — original repro (one typed entry in
  `types-registry.config.entities`) previously aborted post-init on a 20-gear
  assembly.

### Documentation

- [x] API documentation updated (`docs/schemas/tenant_metadata.v1.schema.json`)

### Checklist

- [x] Code follows project style guidelines (conventional commit, DCO signed)
- [x] Self-review completed

### Related Issues

Fixes #4338 

## Summary by CodeRabbit

* **New Features**
  * Tenant metadata envelope schemas now allow additional, unspecified payload properties.
  * Typed tenant metadata schemas can be registered and validated when extending the open envelope.

* **Bug Fixes**
  * Improved schema consistency checks to ensure registered and documented schemas both support open payloads.

* **Tests**
  * Added regression coverage for schema registration and ready-mode validation of derived tenant metadata schemas.

### DIFF
--- a/gears/system/account-management/account-management-sdk/src/gts.rs
+++ b/gears/system/account-management/account-management-sdk/src/gts.rs
@@ -411,7 +411,18 @@ pub struct TenantMetadataTraits {
     // the OP#13 rationale. Guarded by
     // `sync_tests::tenant_metadata_envelope_trait_contract_matches_docs`.
     traits = serde_json::json!({ "inheritance_policy": "override_only" }),
-    gts_abstract = true
+    gts_abstract = true,
+    // The envelope's payload is OPEN by contract: derived metadata schemas
+    // own the payload shape (PRD §5.7 — "branding, contacts") and
+    // MetadataService validates entries against the DERIVED schema at PUT
+    // time. Without an explicit `additionalProperties: true` the OP#12
+    // chain-narrowing check treats the envelope as a closed empty object and
+    // rejects every derived schema that declares a single typed property —
+    // making typed tenant metadata unregistrable. The upstream emitter
+    // closes the object unconditionally, so the toolkit wrapper opens it in
+    // the inventory `schema_fn`. Guarded by
+    // `sync_tests::tenant_metadata_envelope_payload_is_open`.
+    open_payload = true
 )]
 pub struct TenantMetadataEnvelopeV1 {
     /// Required by the `gts-macros` base-struct contract; envelopes carry
@@ -608,10 +619,12 @@ mod sync_tests {
     //! (`x-gts-traits-schema` shape + the envelope's `x-gts-traits`
     //! defaults) MUST agree with the documented
     //! `docs/schemas/<name>.v1.schema.json`. The macro additionally emits
-    //! a data-type top-level shape (`id`/`properties`/`required`/
-    //! `additionalProperties`) that the docs omit by design and that is
-    //! inert at runtime (see gear docs), so only the trait subtrees are
-    //! compared.
+    //! a data-type top-level shape (`id`/`properties`/`required`) that the
+    //! docs omit by design, so only the trait subtrees are compared —
+    //! EXCEPT `additionalProperties`, which is NOT inert: the OP#12
+    //! chain-narrowing check reads it to decide whether derived schemas may
+    //! declare payload properties (see
+    //! `tenant_metadata_envelope_payload_is_open`).
 
     use std::fs;
     use std::path::PathBuf;
@@ -708,6 +721,43 @@ mod sync_tests {
             &read_disk_schema("tenant_metadata.v1.schema.json"),
             TenantMetadataEnvelopeV1::TYPE_ID,
             &serde_json::json!({ "inheritance_policy": "override_only" }),
+        );
+    }
+
+    /// The metadata envelope must reach types-registry with an OPEN payload
+    /// (`additionalProperties: true`), otherwise the OP#12 chain-narrowing
+    /// check rejects every derived schema that declares a typed payload
+    /// property — see the typed-tenant-metadata registration bug.
+    ///
+    /// Asserted on the INVENTORY entry (what types-registry actually
+    /// registers): `open_payload = true` opens the schema in the wrapper's
+    /// `schema_fn`, so the raw `gts_schema_with_refs()` accessor still
+    /// reflects the upstream emitter's closed default.
+    #[test]
+    fn tenant_metadata_envelope_payload_is_open() {
+        let registered = toolkit_gts::all_inventory_type_schemas()
+            .expect("collect inventory type schemas")
+            .into_iter()
+            .find(|s| {
+                s.pointer("/$id").and_then(Value::as_str)
+                    == Some(&format!("gts://{}", TenantMetadataEnvelopeV1::TYPE_ID))
+            })
+            .expect("tenant_metadata envelope must be in the type-schema inventory");
+        assert_eq!(
+            registered
+                .pointer("/additionalProperties")
+                .and_then(Value::as_bool),
+            Some(true),
+            "the inventory-registered tenant_metadata envelope must declare \
+             additionalProperties: true so derived schemas can add typed \
+             payload properties",
+        );
+        assert_eq!(
+            read_disk_schema("tenant_metadata.v1.schema.json")
+                .pointer("/additionalProperties")
+                .and_then(Value::as_bool),
+            Some(true),
+            "docs/schemas/tenant_metadata.v1.schema.json must stay in sync",
         );
     }
 }
--- a/gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
+++ b/gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
@@ -1,0 +1,77 @@
+#![allow(clippy::unwrap_used, clippy::expect_used)]
+
+//! Regression tests for typed derived schemas under abstract envelope bases.
+//!
+//! An `x-gts-abstract` envelope base (the account-management
+//! `tenant_metadata.v1~` pattern: no payload properties of its own, traits in
+//! `x-gts-traits-schema`, payload shape owned by derived schemas) must allow
+//! derived schemas to declare typed payload properties when the envelope
+//! explicitly opens its payload with `additionalProperties: true`. Without
+//! the explicit opt-in, the OP#12 chain-narrowing check treats the envelope
+//! as a closed empty object and rejects every typed derived schema, which
+//! made typed tenant metadata unregistrable end-to-end.
+
+mod common;
+
+use common::create_service;
+use serde_json::json;
+
+#[test]
+fn typed_derived_schema_registers_under_open_abstract_envelope() {
+    let service = create_service();
+
+    // Mirrors gts.cf.core.am.tenant_metadata.v1~ with the payload explicitly
+    // opened (additionalProperties: true).
+    let envelope = json!({
+        "$id": "gts://gts.x.test.am.meta_envelope.v1~",
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "description": "Abstract metadata envelope; derived schemas own the payload shape.",
+        "type": "object",
+        "additionalProperties": true,
+        "x-gts-abstract": true,
+        "x-gts-traits-schema": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "inheritance_policy": {
+                    "type": "string",
+                    "enum": ["override_only", "inherit"],
+                    "default": "override_only"
+                }
+            }
+        }
+    });
+
+    // A typed derived schema — the exact shape adopters need for validated
+    // per-tenant settings payloads (AM PRD §5.7).
+    let derived = json!({
+        "$id": "gts://gts.x.test.am.meta_envelope.v1~x.test.am.settings.v1~",
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "description": "Typed settings payload derived from the envelope.",
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+            "automation_level": {
+                "type": "string",
+                "enum": ["manual", "recommendations", "autonomous"]
+            },
+            "approved_worker_categories": {
+                "type": "array",
+                "items": { "type": "string" }
+            }
+        },
+        "x-gts-traits": { "inheritance_policy": "override_only" }
+    });
+
+    let results = service.register(vec![envelope, derived]);
+    for result in &results {
+        assert!(
+            matches!(result, types_registry_sdk::RegisterResult::Ok { .. }),
+            "registration must succeed for an open envelope chain, got: {result:?}",
+        );
+    }
+
+    service
+        .switch_to_ready()
+        .expect("typed derived schema under an open abstract envelope must survive ready-mode validation");
+}
--- a/libs/toolkit-gts-macros/src/lib.rs
+++ b/libs/toolkit-gts-macros/src/lib.rs
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
     }
 }
 
+/// Strips the wrapper-only `open_payload = true|false` argument from the
+/// attribute token stream before it is forwarded to the upstream
+/// `struct_to_gts_schema` macro (which rejects unknown attributes).
+///
+/// `open_payload = true` marks the generated schema's payload as OPEN: the
+/// inventory `schema_fn` post-processes the upstream JSON and sets
+/// `additionalProperties: true` at the schema root. Envelope bases (e.g.
+/// account-management's `tenant_metadata.v1~`) need this so the OP#12
+/// chain-narrowing check lets derived schemas declare typed payload
+/// properties — the upstream emitter otherwise closes the object, which the
+/// attribute surface gives no way to override.
+fn strip_open_payload(attr: &TokenStream2) -> (TokenStream2, bool) {
+    use proc_macro2::TokenTree;
+    let mut out = TokenStream2::new();
+    let mut open = false;
+    let mut iter = attr.clone().into_iter().peekable();
+    while let Some(tt) = iter.next() {
+        if let TokenTree::Ident(id) = &tt {
+            if id == "open_payload" {
+                let eq = iter.next();
+                let val = iter.next();
+                let is_eq = matches!(&eq, Some(TokenTree::Punct(p)) if p.as_char() == '=');
+                let truthy = matches!(&val, Some(TokenTree::Ident(v)) if v == "true");
+                let falsy = matches!(&val, Some(TokenTree::Ident(v)) if v == "false");
+                if is_eq && (truthy || falsy) {
+                    open = truthy;
+                    // Swallow the following comma so upstream sees a clean list.
+                    if matches!(iter.peek(), Some(TokenTree::Punct(p)) if p.as_char() == ',') {
+                        iter.next();
+                    }
+                    continue;
+                }
+                // Malformed usage: re-emit consumed tokens verbatim so the
+                // upstream parser reports its canonical error.
+                out.extend([tt]);
+                if let Some(e) = eq {
+                    out.extend([e]);
+                }
+                if let Some(v) = val {
+                    out.extend([v]);
+                }
+                continue;
+            }
+        }
+        out.extend([tt]);
+    }
+    (out, open)
+}
+
 fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Result<TokenStream2> {
     let crate_path = resolve_crate_path()?;
+    let (attr, open_payload) = strip_open_payload(attr);
+    let attr = &attr;
     let type_id_input = extract_type_id(attr)?;
     let type_id = type_id_input.emitted_expr(&crate_path);
     let struct_name = &input.ident;
@@ -266,10 +317,23 @@ fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Resul
         ));
     }
     let has_generics = type_param_count == 1;
-    let schema_fn_body = if has_generics {
+    let raw_schema_call = if has_generics {
         quote! { <#struct_name::<()>>::gts_schema_with_refs_as_string() }
     } else {
         quote! { <#struct_name>::gts_schema_with_refs_as_string() }
+    };
+    let schema_fn_body = if open_payload {
+        // Post-process the upstream schema: open the payload at the root so
+        // OP#12 chain-narrowing admits typed properties in derived schemas.
+        quote! {{
+            let mut schema: #crate_path::serde_json::Value =
+                #crate_path::serde_json::from_str(&#raw_schema_call)
+                    .expect("gts_type_schema: macro-generated schema must be valid JSON");
+            schema["additionalProperties"] = #crate_path::serde_json::Value::Bool(true);
+            schema.to_string()
+        }}
+    } else {
+        raw_schema_call
     };
 
     Ok(quote! {
--- a/libs/toolkit-gts/src/lib.rs
+++ b/libs/toolkit-gts/src/lib.rs
@@ -43,6 +43,9 @@ pub use gts::{GTS_ID_PREFIX, GTS_ID_URI_PREFIX, GtsId, GtsInstanceId, GtsSchema}
 // `inventory` as a direct dep.
 #[doc(hidden)]
 pub use inventory;
+// Re-exported for `#[gts_type_schema(open_payload = true)]`-generated code,
+// which post-processes the upstream schema JSON before inventory submission.
+pub use serde_json;
 
 // Re-export the companion proc-macros so consumers need only one crate dep.
 pub use gts_macros::GtsTraitsSchema;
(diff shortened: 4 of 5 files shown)


aviator5:
@AndrejK666 I've added open container to `TenantMetadataEnvelopeV1` and `TenantTypeEnvelopeV1` types in https://github.com/constructorfabric/gears-rust/commit/6e34080f30bbc6c64e3a2a05283f3939eb2706ef#diff-72188ec38fba1a414db39576874feadbc8e5616bef2cea45b34191596e7a5838, so I'm going to close this PR. Please feel free to reach me if you have any questions.

MikeFalcon77 (commented):
23 findings from nonstop-reviewer (run 22).

gears/system/account-management/account-management-sdk/src/gts.rs
```diff
@@ -411,7 +411,18 @@ pub struct TenantMetadataTraits {
...
+    // own the payload shape (PRD §5.7 — "branding, contacts") and
+    // MetadataService validates entries against the DERIVED schema at PUT
+    // time. Without an explicit `additionalProperties: true` the OP#12
+    // chain-narrowing check treats the envelope as a closed empty object and
+    // rejects every derived schema that declares a single typed property —
+    // making typed tenant metadata unregistrable. The upstream emitter
+    // closes the object unconditionally, so the toolkit wrapper opens it in
+    // the inventory `schema_fn`. Guarded by
+    // `sync_tests::tenant_metadata_envelope_payload_is_open`.
+    open_payload = true
```
MikeFalcon77, line 425: **HIGH**

Nothing tests that a typed derived schema actually registers against this envelope. The AM metadata suites go through a stubbed `MetadataSchemaRegistry` (`tests/common/mod.rs`) and never register a derived JSON schema, and the new `abstract_envelope_tests.rs` uses a hand-written envelope rather than this one, so the "unregistrable" failure described here has no regression test on the real chain.

MikeFalcon77, line 425: **HIGH**

`open_payload = true` rewrites only the inventory copy, so `TenantMetadataEnvelopeV1::gts_schema_with_refs()` keeps the upstream emitter's closed object while `docs/schemas/tenant_metadata.v1.schema.json` now declares `additionalProperties: true`. `tenant_metadata_envelope_trait_contract_matches_docs` cannot catch that divergence because it compares only the trait subtrees, so the generated schema and the published one can drift further with no guard.

MikeFalcon77, line 425: **HIGH**

This opens the envelope's root object, so nothing in the `tenant_metadata.v1~` chain constrains top-level keys any more. `guidelines/GTS.md` 5.5 keeps the base root at `additionalProperties: false` and exposes an open `payload` object that derived schemas narrow, which clears the OP#12 rejection without disabling closed-object validation for the whole family. As written, a derived metadata schema that omits `additionalProperties: false` accepts and persists arbitrary unvalidated keys on PUT, and the base is not a backstop either: `effective_schema()` inlines this envelope only when the derived schema declares `allOf: [{"$ref": ...}]` (types-registry-sdk `models.rs:444-478`), which the new regression test's derived schema does not.

gears/system/account-management/account-management-sdk/src/gts.rs
```diff
@@ -608,10 +619,12 @@ mod sync_tests {
...
     //! defaults) MUST agree with the documented
     //! `docs/schemas/<name>.v1.schema.json`. The macro additionally emits
-    //! a data-type top-level shape (`id`/`properties`/`required`/
-    //! `additionalProperties`) that the docs omit by design and that is
-    //! inert at runtime (see gear docs), so only the trait subtrees are
-    //! compared.
+    //! a data-type top-level shape (`id`/`properties`/`required`) that the
+    //! docs omit by design, so only the trait subtrees are compared —
+    //! EXCEPT `additionalProperties`, which is NOT inert: the OP#12
+    //! chain-narrowing check reads it to decide whether derived schemas may
```
MikeFalcon77, line 625: **HIGH**

This makes `additionalProperties` part of the contract the module guards, but only `tenant_metadata` gets an assertion. `tenant_type.v1~` is the same abstract-envelope shape (and `docs/schemas/tenant_type.v1.schema.json` declares no `additionalProperties`), so if OP#12 blocks a typed derived tenant type nothing in `sync_tests` catches it. Can `assert_trait_contract` take the expected `additionalProperties` and assert it for both envelopes?

gears/system/account-management/account-management-sdk/src/gts.rs
```diff
@@ -710,4 +723,41 @@ mod sync_tests {
...
+    /// (`additionalProperties: true`), otherwise the OP#12 chain-narrowing
+    /// check rejects every derived schema that declares a typed payload
+    /// property — see the typed-tenant-metadata registration bug.
+    ///
+    /// Asserted on the INVENTORY entry (what types-registry actually
+    /// registers): `open_payload = true` opens the schema in the wrapper's
+    /// `schema_fn`, so the raw `gts_schema_with_refs()` accessor still
+    /// reflects the upstream emitter's closed default.
+    #[test]
+    fn tenant_metadata_envelope_payload_is_open() {
```
MikeFalcon77, line 737: **HIGH**

`tenant_metadata_envelope_trait_contract_matches_docs` still compares `gts_schema_with_refs()`, which is no longer what gets registered: the inventory entry is the post-processed JSON built by the macro's `schema_fn`. This test has that document in hand but checks only `additionalProperties`, so nothing asserts the registered envelope still carries `x-gts-abstract`, `x-gts-traits-schema` and `x-gts-traits` after the post-processing step.

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
```diff
@@ -0,0 +1,77 @@
...
+use common::create_service;
+use serde_json::json;
+
+#[test]
+fn typed_derived_schema_registers_under_open_abstract_envelope() {
+    let service = create_service();
+
+    // Mirrors gts.cf.core.am.tenant_metadata.v1~ with the payload explicitly
+    // opened (additionalProperties: true).
+    let envelope = json!({
```
MikeFalcon77, line 25: **HIGH**

Nothing here is tied to the code this PR changes. The envelope is a hand-written literal with `additionalProperties: true` baked in, so removing `open_payload = true` from `TenantMetadataEnvelopeV1` or deleting `strip_open_payload` from the macro leaves this test green. Can we drive the envelope from the schema the macro actually emits (the inventory entry) so a revert of the fix fails here?

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
```diff
@@ -0,0 +1,77 @@
...
+        "$id": "gts://gts.x.test.am.meta_envelope.v1~",
```
MikeFalcon77, line 26: **HIGH**

Both `$id`s hard-code the `gts://gts.` prefix. Every other test in this directory builds ids with `gts_uri!` / `GTS_ID_PREFIX` (see `rg_gts_type_system_tests.rs`), and the prefix is a compile-time knob that `libs/toolkit-gts/tests/prefix_customization.rs` actually exercises, so with a non-default `GTS_ID_PREFIX` these ids get rejected by `register` and the test fails for an unrelated reason.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
     }
 }
 
+/// Strips the wrapper-only `open_payload = true|false` argument from the
```
MikeFalcon77, line 241: **HIGH**

`open_payload` is a new public argument of `#[gts_type_schema]`, but it is documented only on this private helper, so it never appears in rustdoc. The macro's own doc comment still says it forwards every attribute verbatim and takes no opinions on the attrs, which is no longer true.

MikeFalcon77, line 241: **HIGH**

This puts a workaround for a `gts` 0.11 chain-narrowing semantic into the shared macro wrapper, whose module docs and `gts_type_schema` rustdoc still state it forwards every attribute verbatim and takes no opinions on the attrs. `open_payload` is a generic root-level `additionalProperties` override now available to every `#[gts_type_schema]` user even though one envelope needs it, and nothing links it to the upstream behaviour, so there is no signal to drop it when the chain-narrowing check changes. Why not express the openness through the envelope's own upstream schema attributes, or keep the override inside account-management?

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// `struct_to_gts_schema` macro (which rejects unknown attributes).
+///
+/// `open_payload = true` marks the generated schema's payload as OPEN: the
+/// inventory `schema_fn` post-processes the upstream JSON and sets
+/// `additionalProperties: true` at the schema root. Envelope bases (e.g.
+/// account-management's `tenant_metadata.v1~`) need this so the OP#12
+/// chain-narrowing check lets derived schemas declare typed payload
+/// properties — the upstream emitter otherwise closes the object, which the
+/// attribute surface gives no way to override.
+fn strip_open_payload(attr: &TokenStream2) -> (TokenStream2, bool) {
```
MikeFalcon77, line 252: **HIGH**

`strip_open_payload` and the `open_payload` branch of `schema_fn` ship with no test. `libs/toolkit-gts/tests/macro_integration.rs` is where this wrapper's behavior is pinned and it is untouched: can we add an `open_payload = true` fixture there asserting `(entry.schema_fn)()` parses to `additionalProperties: true`, a control type without the flag that stays closed, and an `open_payload = false` case so the argument is proven stripped for both values?

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    use proc_macro2::TokenTree;
+    let mut out = TokenStream2::new();
+    let mut open = false;
+    let mut iter = attr.clone().into_iter().peekable();
+    while let Some(tt) = iter.next() {
+        if let TokenTree::Ident(id) = &tt {
```
MikeFalcon77, line 258: **HIGH**

This scan does not model the attribute grammar: `open_payload = true, open_payload = false` is accepted silently with the last occurrence winning, and any top-level `open_payload` ident is taken as the argument name wherever it appears, including a value position. Parsing the list as `Punctuated<Meta, Comma>` gives duplicate detection and value validation from the type instead of from caller discipline.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                let falsy = matches!(&val, Some(TokenTree::Ident(v)) if v == "false");
+                if is_eq && (truthy || falsy) {
+                    open = truthy;
+                    // Swallow the following comma so upstream sees a clean list.
+                    if matches!(iter.peek(), Some(TokenTree::Punct(p)) if p.as_char() == ',') {
+                        iter.next();
+                    }
+                    continue;
+                }
+                // Malformed usage: re-emit consumed tokens verbatim so the
```
MikeFalcon77, line 273: **HIGH**

Upstream has no canonical error for `open_payload`: it is a wrapper-only argument that upstream rejects as unknown. So `open_payload = yes` gets the user an "unknown attribute" error blaming the argument name, when the argument is fine and only the value is wrong.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -266,11 +317,24 @@ fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Resul
         ));
     }
     let has_generics = type_param_count == 1;
-    let schema_fn_body = if has_generics {
+    let raw_schema_call = if has_generics {
         quote! { <#struct_name::<()>>::gts_schema_with_refs_as_string() }
     } else {
         quote! { <#struct_name>::gts_schema_with_refs_as_string() }
     };
+    let schema_fn_body = if open_payload {
```
MikeFalcon77, line 325: **HIGH**

The open-payload rewrite only lands in the inventory `schema_fn`. Upstream's generated `gts_schema_with_refs()` / `gts_schema_with_refs_as_string()` still reports `additionalProperties: false`, so one `$id` now has two different schema documents, and the DE0110/DE0902 lints point every in-process consumer at the closed one. Both should come out of the same place.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -266,11 +317,24 @@ fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Resul
...
+        // Post-process the upstream schema: open the payload at the root so
+        // OP#12 chain-narrowing admits typed properties in derived schemas.
+        quote! {{
+            let mut schema: #crate_path::serde_json::Value =
+                #crate_path::serde_json::from_str(&#raw_schema_call)
+                    .expect("gts_type_schema: macro-generated schema must be valid JSON");
```
MikeFalcon77, line 331: **HIGH**

This `expect` lands in the consumer's generated `schema_fn`, so it runs at inventory-collection time in the running process, not at macro expansion. The crate-level `expect_used` allow is justified with "proc macros run at compile time, so panics become compile errors", which does not hold for emitted code. `all_inventory_type_schemas()` already returns `anyhow::Result` and names the offending `type_id` on a parse failure, and `TypesRegistryGear::init` turns that into a startup error, so this replaces a clean boot failure with a process abort.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -266,11 +317,24 @@ fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Resul
...
+            schema["additionalProperties"] = #crate_path::serde_json::Value::Bool(true);
```
MikeFalcon77, line 332: **HIGH**

Assigning through a `&str` key panics when the parsed schema root is not an object, and it silently turns a `null` root into `{"additionalProperties": true}`. Neither case reaches the `Err` path in `all_inventory_type_schemas`: one aborts the process, the other registers a fabricated document as this type's schema. The invariant "upstream always emits a JSON object" lives in another crate, so it is worth checking here.

MikeFalcon77, line 332: **HIGH**

This overwrites the root `additionalProperties` unconditionally, whatever the upstream emitter produced, and nothing checks that the annotated struct is actually a property-free abstract envelope. Put `open_payload = true` on a struct with real typed fields and `#[serde(deny_unknown_fields)]` and the registered schema silently accepts unknown top-level keys that the type's own `Deserialize` still rejects, so the registry and the Rust type disagree about what is valid input.

libs/toolkit-gts/src/lib.rs
```diff
@@ -43,6 +43,9 @@ pub use gts::{GTS_ID_PREFIX, GTS_ID_URI_PREFIX, GtsId, GtsInstanceId, GtsSchema}
 // `inventory` as a direct dep.
 #[doc(hidden)]
 pub use inventory;
+// Re-exported for `#[gts_type_schema(open_payload = true)]`-generated code,
+// which post-processes the upstream schema JSON before inventory submission.
+pub use serde_json;
```
MikeFalcon77, line 48: **HIGH**

This makes `toolkit_gts::serde_json` part of this crate's public API, so a serde_json major bump turns into a breaking change here. The `inventory` re-export directly above is `#[doc(hidden)]` and every other macro-support re-export lives in `__private`; this one is macro-expansion-only as well.

gears/system/account-management/account-management-sdk/src/gts.rs
```diff
@@ -608,10 +619,12 @@ mod sync_tests {
     //! (`x-gts-traits-schema` shape + the envelope's `x-gts-traits`
     //! defaults) MUST agree with the documented
     //! `docs/schemas/<name>.v1.schema.json`. The macro additionally emits
-    //! a data-type top-level shape (`id`/`properties`/`required`/
-    //! `additionalProperties`) that the docs omit by design and that is
-    //! inert at runtime (see gear docs), so only the trait subtrees are
-    //! compared.
+    //! a data-type top-level shape (`id`/`properties`/`required`) that the
```
MikeFalcon77, line 622: This now says `additionalProperties` is not inert, but `assert_trait_contract` still compares only the trait subtrees, so the generated schema's `additionalProperties` is never compared with the disk schema. The type emits the closed form while the docs file now says open, which means `docs/schemas/tenant_metadata.v1.schema.json` describes the post-processed inventory output rather than what `gts_schema_with_refs()` returns.

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
```diff
@@ -0,0 +1,77 @@
+#![allow(clippy::unwrap_used, clippy::expect_used)]
```
MikeFalcon77, line 1: This suppression carries no `reason`, and the file never calls `unwrap`, so half of it is dead. `#![expect(clippy::expect_used, reason = "...")]` would self-remove once the suppression stops being needed.

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
```diff
@@ -0,0 +1,77 @@
...
+//! as a closed empty object and rejects every typed derived schema, which
+//! made typed tenant metadata unregistrable end-to-end.
+
+mod common;
+
+use common::create_service;
+use serde_json::json;
+
+#[test]
+fn typed_derived_schema_registers_under_open_abstract_envelope() {
```
MikeFalcon77, line 20: There is no closed-envelope case, so nothing pins the OP#12 narrowing this whole fix rests on. If types-registry ever stops rejecting typed derived schemas under a closed base, this test still passes and the `open_payload` plumbing becomes dead code with no signal. Can we add the same chain with `additionalProperties: true` removed from the envelope and assert `switch_to_ready()` fails?

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
MikeFalcon77, line 25: This envelope drops the `properties: { id }` / `required: ["id"]` top-level shape that the macro-generated `tenant_metadata.v1~` actually emits (see the `sync_tests` module doc in `account-management-sdk/src/gts.rs`). With no payload properties and no required key on the base, the chain-narrowing check has much less to compare, so this test can pass while the real envelope chain still rejects the derived schema.

gears/system/types-registry/types-registry/tests/abstract_envelope_tests.rs
```diff
@@ -0,0 +1,77 @@
...
+            },
+            "approved_worker_categories": {
+                "type": "array",
+                "items": { "type": "string" }
+            }
+        },
+        "x-gts-traits": { "inheritance_policy": "override_only" }
+    });
+
+    let results = service.register(vec![envelope, derived]);
```
MikeFalcon77, line 66: `create_service()` returns a registry in configuration mode and `register` only validates when `repo.is_ready()` (`register_internal(entities, validate)` in `domain/service.rs`), so these `RegisterResult::Ok` assertions also pass for a closed envelope. The OP#12 check only runs inside `switch_to_ready()` below, which makes the loop's "registration must succeed for an open envelope chain" message claim something it never checked.

libs/toolkit-gts-macros/src/lib.rs
```diff
@@ -238,8 +238,59 @@ pub fn gts_type_schema(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        }
+        out.extend([tt]);
+    }
+    (out, open)
+}
+
 fn expand_gts_type_schema(attr: &TokenStream2, input: &ItemStruct) -> syn::Result<TokenStream2> {
     let crate_path = resolve_crate_path()?;
+    let (attr, open_payload) = strip_open_payload(attr);
+    let attr = &attr;
```
MikeFalcon77, line 293: `let attr = &attr;` exists only to keep the old `&TokenStream2` call shape after the shadowing. `extract_type_id(&attr)` and interpolating the owned stream in `quote!` both work, so this re-borrow can go.

Issue #4287, opened 2026-07-29
CORE - Policy Engine for Admission Control


Artifizer:
**Description**: a gear that defines authorization policies for the Admission Control gear
**Is Plugin:** YES
**Has Plugins**: NO
**Has Extension Points**: NO

PR #4396, opened 2026-08-03
docs(policy-engine): PRD

fluiderson:
#4287

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive product requirements documentation for the Policy Engine.
  * Documented policy lifecycle, tenant assignment, evaluation behavior, administration, integrations, observability, configuration, and reliability expectations.
  * Included use cases, acceptance criteria, risks, assumptions, dependencies, and open questions.

### DIFF
--- a/gears/system/policy-engine/docs/PRD.md
+++ b/gears/system/policy-engine/docs/PRD.md
@@ -1,0 +1,1054 @@
+# PRD — Policy Engine
+
+
+<!-- toc -->
+
+- [1. Overview](#1-overview)
+  - [1.1 Purpose](#11-purpose)
+  - [1.2 Background / Problem Statement](#12-background--problem-statement)
+  - [Relationship to existing policy subsystems](#relationship-to-existing-policy-subsystems)
+  - [Relationship to ADR-0001](#relationship-to-adr-0001)
+  - [1.3 Goals (Business Outcomes)](#13-goals-business-outcomes)
+  - [1.4 Glossary](#14-glossary)
+- [2. Actors](#2-actors)
+  - [2.1 Human Actors](#21-human-actors)
+  - [2.2 System Actors](#22-system-actors)
+- [3. Operational Concept & Environment](#3-operational-concept--environment)
+  - [3.1 Gear-Specific Environment Constraints](#31-gear-specific-environment-constraints)
+- [4. Scope](#4-scope)
+  - [4.1 In Scope](#41-in-scope)
+  - [4.2 Out of Scope](#42-out-of-scope)
+- [5. Functional Requirements](#5-functional-requirements)
+  - [5.1 Policy Content Model](#51-policy-content-model)
+  - [5.2 Bundle Lifecycle](#52-bundle-lifecycle)
+  - [5.3 Assignment and Inheritance](#53-assignment-and-inheritance)
+  - [5.4 Policy Matching](#54-policy-matching)
+  - [5.5 Decision Semantics](#55-decision-semantics)
+  - [5.6 Decision Outputs](#56-decision-outputs)
+  - [5.7 Evaluation Backend](#57-evaluation-backend)
+  - [5.8 Multi-Tenancy](#58-multi-tenancy)
+  - [5.9 Observability](#59-observability)
+  - [5.10 Configuration](#510-configuration)
+- [6. Non-Functional Requirements](#6-non-functional-requirements)
+  - [6.1 Gear-Specific NFRs](#61-gear-specific-nfrs)
+  - [6.2 NFR Exclusions](#62-nfr-exclusions)
+- [7. Public Library Interfaces](#7-public-library-interfaces)
+  - [7.1 Public API Surface](#71-public-api-surface)
+  - [7.2 External Integration Contracts](#72-external-integration-contracts)
+- [8. Use Cases](#8-use-cases)
+- [9. Acceptance Criteria](#9-acceptance-criteria)
+- [10. Dependencies](#10-dependencies)
+- [11. Assumptions](#11-assumptions)
+- [12. Risks](#12-risks)
+- [13. Open Questions](#13-open-questions)
+- [14. Traceability](#14-traceability)
+
+<!-- /toc -->
+
+## 1. Overview
+
+### 1.1 Purpose
+
+The Policy Engine is the Gears platform gear that owns authorization and governance policy: it stores policy content, manages its lifecycle, evaluates it against operations and resources, and returns decisions together with the row-level constraints that Policy Enforcement Points compile into database queries. It is the platform's Policy Decision Point implementation, and the single place where policy content is authored, versioned, and audited.
+
+The gear follows the pattern [CredStore](../../../credstore/docs/PRD.md) established for credential storage: the gear owns the domain model, lifecycle, and contracts, while a pluggable backend performs the actual policy evaluation. This keeps Gears free of a mandated policy language — the evaluation backend is replaceable — while giving the platform a first-class home for policy content that no single consumer owns.
+
+### 1.2 Background / Problem Statement
+
+Gears authorization runs a PEP to PDP to `AccessScope` pipeline described in [ARCHITECTURE_MANIFEST §10.3](../../../../docs/ARCHITECTURE_MANIFEST.md) and [arch/authorization/DESIGN.md](../../../../docs/arch/authorization/DESIGN.md). The enforcement half of that pipeline is complete: `authz-resolver` abstracts the PDP, `PolicyEnforcer` wraps it for domain gears, and the secure database layer compiles returned constraints into scoped queries. The decision half is not. The two plugins in the tree are a development stub whose own documentation says it must not be used in production, and a tenant-scoping plugin that implements a fixed set of hierarchy access rules — it emits real constraints and fails closed correctly, but it has no role or permission model and does not differentiate by action, reading the requested action only for logging. Four documents record the same gap: the `authz-resolver` README plans a production PDP plugin, the stub plugin promises its own replacement, the authorization integration test plan states that a real plugin must resolve access from an external policy source, and [PERMISSION_GTS_TYPE.md](../../../../docs/arch/authorization/PERMISSION_GTS_TYPE.md) lists both the evaluating plugin and an AuthZ management gear owning grants and role bindings as future work.
+
+The platform already defines the permission vocabulary those documents point at. `AuthzPermissionV1` gives permissions a canonical type, declared at compile time. What does not exist is anywhere to store which subject holds which permission, under what conditions, or to version and audit that content as it changes. Authorization policy in Gears today is either hard-coded in a plugin or absent.
+
+### Relationship to existing policy subsystems
+
+Two gears have already met a narrower version of this need, and this gear does not displace either. `quota-enforcement` defines `QuotaResolutionEngineV1`, a capability-based contract with pluggable arbitration engines and no mandated engine technology. `event-broker` defines a `FilterEngine` plugin trait over a GTS-typed registry with a built-in expression engine. Both are accepted specifications, both own the state their policy governs, and both are the right shape for their domain.
+
+The pattern they share — a GTS-typed registry of pluggable evaluation engines — is the pattern this gear adopts, which is evidence it is the platform's idiom rather than an import. The difference is ownership: those registries are local to a gear and evaluate expressions against state that gear holds. Authorization has no such owner, because the state it governs is spread across every gear that enforces. First release therefore addresses authorization only. Whether governance policy consolidates here later is a question for the consumers to raise when they have a need their own registry cannot meet, not one this document should answer on their behalf.
+
+### Relationship to ADR-0001
+
+[ADR-0001](../../../../docs/arch/authorization/ADR/0001-pdp-pep-authorization-model.md) is accepted and states two positions this gear contradicts: that the Policy Administration Point is entirely vendor-controlled and Gears never sees or stores policies, and that the PDP is the vendor's authorization service reached through `authz-resolver`. This gear is both a PAP and a PDP inside the platform.
+
+That is a deliberate departure, and it is narrower than it looks. ADR-0001's reasoning is that Gears must integrate with a vendor's existing authorization infrastructure without demanding resource synchronisation or a policy format. That reasoning holds, and this gear does not weaken it: `authz-resolver` remains the abstraction, a vendor PDP remains selectable, and a deployment that already has a policy manager can ignore this gear entirely. What changes is that a deployment *without* one is no longer required to build a plugin before it can authorise anything.
+
+A superseding ADR recording that change is a prerequisite deliverable for this gear, not an artefact of it. Until it exists, this document's positioning is provisional.
+
+### 1.3 Goals (Business Outcomes)
+
+Milestones below refer to this document's own priority tiers — "p1 complete" means every `p1` requirement in Sections 5 and 6 is met and verified. The repository has no separate release-maturity vocabulary.
+
+| Outcome | Baseline | Target | By |
+|---|---|---|---|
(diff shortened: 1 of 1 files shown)


nonameffh:
## Summary

The PRD's central mechanism — a pluggable evaluation backend so the platform is not locked into one
policy language — is the right thing to build. This review argues it should ship as a shared crate under
`libs/` consumed by each gear that evaluates policy, rather than as a gear that owns policy content and
evaluates on behalf of other gears.

Four claims, with evidence, then a proposal:

1. A shared evaluation abstraction is justified today. It should be engine-agnostic — one contract with CEL,
   Rego, or anything chosen later behind it — with the consumer's result type kept as a parameter so results
   stay typed instead of collapsing to opaque JSON.
2. How the applicable policy set is obtained differs in kind between consumers, and resolution is a query
   over content — so content has to live with the consumer that resolves it.
3. As a gear, this overlaps substantially with serverless-runtime's already-specified registry and
   lifecycle.
4. Evaluation should not be routed through the serverless-runtime host — a separate question from whether
   serverless-runtime's own embedded backends could later build on the same crate (§5).

---

## 1. A shared evaluation crate is justified, and it should be engine-agnostic

Two accepted specs already define embedded pluggable evaluation — the evaluator called directly, in the
consumer's own process — each with a CEL built-in and each naming Rego as a possible second engine:

- **`QuotaResolutionEngineV1`** — `gears/system/quota-enforcement/docs/ADR/0005-cpt-cf-quota-enforcement-adr-evaluation-engine.md`.
  Rust-native in-process linkage; sandboxed (no I/O, no nondeterminism beyond `ctx.time`); cost-bounded
  (default 5 ms per policy); pre-compilable with a cache keyed by `(policy_id, policy_version)`;
  deterministic; closed `Decision` / `EngineError` contract. Reference engine: `cel` via `cel-interpreter`.
- **`FilterEngine`** — `gears/system/event-broker/docs/ADR/0005-subscription-filter-typing.md:164`.
  `compile()` + `eval()`; opaque `CompiledFilter` handle; GTS-typed engine identity; closed `FilterError`
  including `EvalTimeout`.

The two share no code: QE's trait sits in `quota-enforcement-sdk` behind a private `EngineRegistry`
(`quota-enforcement/docs/DESIGN.md:616`), event-broker's in its own SDK. There is no shared crate in
`libs/`, and `cel-interpreter` is not yet a workspace dependency. This PRD adds a third instance of the
same pattern.

**What a shared crate should own.** Not the interpreter — that is a third-party crate and Cargo keeps one
copy of it. What gets re-written per module is the layer around it: cost/step cap, wall-clock timeout, the
sandbox (defined as the absence of host bindings), a compile cache keyed by version, a closed error enum,
the diagnostics shape, GTS engine identity, expression size limits, and metric names. Since the sandbox is
a security control, having that layer in one reviewable place matters more than the code volume: N call
sites configuring an evaluator independently means N places where a cost limit or a timeout can be
omitted. This holds with a single language, so it does not depend on Rego ever arriving.

**One abstraction, but not one fixed shape.** The abstraction should be engine-agnostic — CEL, Rego, or
whatever is chosen later, behind the same contract. The trap is a contract with a *fixed result type*, which
is what a naively generic "evaluation backend" degrades into: it has to be the union of what every consumer
needs and the intersection of what every language provides, so the result collapses to
`serde_json::Value` and validation is pushed back into each consumer. That is precisely what QE's
`principle-strict-engine-boundary` exists to prevent — the orchestrator validates a *typed* `Decision`
against Debit-Plan invariants before mutating a counter.

Both properties are obtainable at once by keeping the result type a parameter of each *evaluation* — a
consumer needs several result shapes from one engine — and confining language-specific detail to the
implementation. What differs per language — source form, compiled artifact, how a context is presented,
diagnostic detail — stays inside the engine; what does not differ — budget, cache, error taxonomy, engine
identity, metrics — is shared. Rego's `data.*` document, for which CEL has no
equivalent, then lives in what the Rego engine accepts as its context, rather than widening a surface every
other engine has to carry.

## 2. Resolution of the applicable policy set is consumer-specific

Each consumer determines *which* policies apply by a different mechanism:

| Consumer | How the applicable set is determined |
|---|---|
| quota-enforcement | applicable Quotas from QE's own state and configuration, then arbitration over that set |
| event-broker | nothing is resolved — the expression is a field on the subscription interest, compiled once at JOIN |
| authorization (the role this PRD describes) | assignments along the tenant ancestry chain, ordered by proximity then priority |
| operation-time validation (the operation trigger / `before` phase in `fr-target-binding`) | bindings matched by selector against the operation, the resource GTS type, and tenant context |

A shared resolution interface would need a shared `ResolutionContext`, which is the part that differs. It
can only be an untyped bag — losing the typing that makes these contracts safe — or a type parameter,
which carries no behaviour.

There is a further consequence worth stating: **resolution is a query over policy content.** If resolution
belongs to the consumer, the content it queries has to sit with it. Otherwise every evaluation either ships
the full content set to the consumer, or the consumer's domain model is pulled into the policy gear. The
PRD as written takes the second path — the gear understands tenant trees, resource types, trigger types,
phases, and barrier semantics (§5.3, §5.4). That is workable for one consumer, but each additional consumer
widens the gear by its own domain.

What *is* common across consumers is not the matching but the **contract of the resolution step**:
deterministic total ordering, a bound on applicable-set size, pinning to a content version so a decision is
reproducible, caching keyed by `(context fingerprint, content version)` with a bounded propagation window,
matched-versus-evaluated accounting, and the short-circuit rule. This PRD already specifies all six
(`fr-deterministic-ordering`, `fr-operational-limits`, `fr-short-circuit`, `nfr-activation-propagation`,
`nfr-cache-safety`, `fr-decision-records`), and none of them is language- or consumer-specific. They are the
most reusable material in the document.

## 3. As a gear, this overlaps with serverless-runtime

serverless-runtime specifies a registry of versioned, GTS-typed definitions dispatched to pluggable
backends. A policy gear built as described here re-specifies much of the same machinery:

| Concern | Already specified in serverless-runtime |
|---|---|
| Definition registry with GTS schema validation | `docs/DESIGN.md` — Function Registry |
| Versioning; in-flight work pinned to the version it started with | `docs/DESIGN.md` — Versioning Model |
| `draft → validate → activate`, with validation errors carrying location and suggested corrections | `docs/DESIGN.md:485-496` |
| Dry-run: validate only, no durable record, no side effects, no quota or rate-limit consumption | `docs/DESIGN.md:913-939` |
| Per-tenant governance — quotas, retention, allowlist of permitted backends | `TenantRuntimePolicy` |
| Per-execution records split into a queryable host index and plugin-owned detail | `docs/DESIGN.md:764-772` |
| Audit events on every definition and execution mutation | `docs/DESIGN.md` — Audit Events |

Maintaining two parallel specifications of the same lifecycle is a long-term cost. A gear consulted on the
write path of every mutating operation also adds a cross-gear dependency to every gear that enforces — paid
on each mutation, and hard to remove once callers depend on it.

If the intended first consumer is operation-time validation of mutating requests — the operation trigger in
the `before` phase, which is what `fr-target-binding` prioritises for first release — then a gear is
defensible, but its distinctive value is the interception point, the content model (policy, binding,
parameters), a mutating output with patch semantics, and a per-policy failure policy. None of those is the
evaluator. Worth noting that no interception point exists in ToolKit today: `OperationBuilder` builds
routes, and the operation pipeline has no pre-commit hook a policy could attach to. That gap, not the
evaluation contract, is what blocks the `before` phase.

## 4. Evaluation should not be routed through the serverless-runtime host

1. **The host exposes `execute()`; policy evaluation needs `compile()` + `eval()`.** QE caches compiled
   artifacts by `(policy_id, policy_version)`; event-broker compiles at JOIN and holds the handle for the
   subscription's lifetime. serverless-runtime has no compile phase — each invocation is a cold dispatch.
2. **Host mediation applies to every call.** Authentication and tenant resolution, transport decode, GTS
   validation, tenant policy, the rate-limit plugin, idempotency deduplication, audit, then
   `ClientHub::get_scoped` (`docs/DESIGN.md:782-784`); the plugin then creates a durable `InvocationRecord`
   in `queued` and emits an event that lands a row in `invocation_index` (`docs/DESIGN.md:1088`). That is two
   durable writes before user code runs. serverless-runtime targets start latency p95 ≤ 100 ms and ≥ 1 000
   starts/sec (`docs/PRD.md:813`, `:844`); QE's per-policy engine budget is 5 ms.
3. **The lifecycle model does not fit.** The invocation state machine covers retry, `compensating`,
   `dead_lettered`, and suspend/resume (`docs/DESIGN.md:441-469`). A policy evaluation is instantaneous and
   side-effect-free; none of those states apply, and by the time a retry could help, the caller has already
   received an answer.

## 5. Proposal

**A platform crate.** `libs/toolkit-expr` (name to be decided) — an **embedded evaluator**, meaning: the
expression is compiled and evaluated by a direct Rust call inside the caller's own process, with no IPC hop
and no FFI on the eval path. Isolation is language-level — the sandbox is the absence of host bindings,
verifiable by reviewing what is bound — and explicitly *not* memory- or process-level: a fault in the
evaluator crate is a fault in the caller's address space. That is adequate for data-only expression
languages such as CEL, and it is the property QE's engine contract already requires. Anything needing
stronger isolation — a large language surface, or authors who are not trusted — is a different mechanism
(separate process, memory-isolated VM) and outside this crate's scope.

The contract is engine-agnostic. The consumer's context and result types are parameters; everything
language-specific lives inside the implementation:

```rust
/// Implemented once per engine — CEL, Rego, or whatever is added later.
/// `Ctx` is the consumer's context shape; the result shape is per evaluation, not per engine.
trait Engine<Ctx>: Send + Sync {
    fn id(&self) -> GtsEngineId;
    fn compile(&self, src: &Source, expects: &ResultShape, limits: &Limits)
        -> Result<Compiled, CompileError>;
    fn eval(&self, c: &Compiled, ctx: &Ctx, budget: Budget) -> Result<Value, EvalError>;
}

/// Typed extraction, written once in the core rather than in every consumer.
/// Blanket-implemented for every `Engine` including `dyn Engine`, so the generic
/// method stays out of the object-safe trait. Nobody implements this by hand.
trait EngineExt<Ctx>: Engine<Ctx> {
    fn eval_as<Out: FromValue>(&self, c: &Compiled, ctx: &Ctx, budget: Budget)
        -> Result<Out, EvalError>
    {
        Out::from_value(self.eval(c, ctx, budget)?)
    }
}

impl<Ctx, E: Engine<Ctx> + ?Sized> EngineExt<Ctx> for E {}
```

`Source` covers both a single expression and a module set with an entrypoint, so Rego fits without CEL
carrying the concept. `Compiled` is an opaque, core-owned handle with size accounting — the shape
event-broker already specifies — which is what lets the core own the cache rather than each engine
rebuilding one. Which context shapes a given engine accepts is decided by that engine's crate: the Rego
engine implements the trait for a context carrying `input` plus a `data` document, the CEL engine for a plain
variable set, and neither imposes its shape on the other.

**The result shape belongs to the evaluation, not to the engine.** One consumer routinely needs more than
one: admission validation returns allow/deny with a reason while admission mutation returns a patch, both
from the same engine; an explain or dry-run call (`fr-explanation`, `fr-dry-run`) wants a richer result from
the same compiled policy than the enforcing call does. So `eval` returns the core's closed `Value` and typed
extraction happens through `eval_as`, whose `FromValue` impls are written once per result type in the core.
That keeps the trait object-safe — engines stay selectable at runtime by GTS engine ID, the mechanism both
existing specs use — without pushing result handling back into each consumer, which is the failure mode §1
warns about. Extraction sits in an extension trait rather than on `Engine` because a generic method would
break object safety; the blanket impl covers `dyn Engine`, so callers still write
`engine.eval_as::<Decision>(…)`. Keeping it a named operation rather than an inline conversion also earns
extraction its own error variant and its own counter: a result of the wrong shape is an authoring defect,
not a failed evaluation, and since `compile` was given the expected shape it should be rare enough to alert
on. The expected shape passed to `compile` is what makes that possible — an expression whose result cannot
satisfy it is rejected at authoring time rather than at evaluation time, which is what
`fr-content-validation` asks for.

Selecting and constructing an engine is a second, separate concern — and the two existing specs already
diverge on it: QE resolves through a private static `EngineRegistry` with engines linked at build time
(`quota-enforcement/docs/DESIGN.md:616`), event-broker resolves through `ClientHub` keyed by
`interest.filter.engine` at JOIN. That belongs in the core as well:

```rust
/// One instance per consumer context shape, built once at gear init from deployment config.
trait EngineProvider<Ctx>: Send + Sync {
    fn get(&self, id: &GtsEngineId) -> Result<Arc<dyn Engine<Ctx>>, EngineLookupError>;
    fn enabled(&self) -> &[GtsEngineId];
}
```

This is where four things stop being per-consumer decisions: an unknown engine ID becomes one typed error
rather than each consumer inventing its own (event-broker currently defines `400 UnknownFilterEngine`);
operator-configured limits and budgets are applied uniformly at construction instead of at each call site;
the compile cache is attached to a shared engine instance rather than duplicated per caller; and the enabled
engine set — plus its GTS type registration — is declared in one place. Whether a given deployment resolves
statically or through `ClientHub` becomes an implementation detail of the provider, so neither existing
consumer has to change its public contract to adopt this.

The core owns the parts that must be identical everywhere: cost/step cap and wall-clock budget, the sandbox,
the compile cache keyed by a caller-supplied version key and handed out as a leased handle (so a consumer
pinned to a version can safely hold a compiled artifact), the closed `EvalError`, the diagnostics shape, the
GTS engine-identity convention, metric names, and a conformance harness every engine runs. It owns no
persistence, no HTTP, and no async orchestration — the same constraint serverless-runtime's SDK already
states for itself. The engine owns language semantics and nothing else, which is the same split this PRD
already draws in `fr-backend-boundary`.

The capability set — what evaluated code may reach outside itself — is fixed when an engine instance is
constructed, not stored with the content. The pure case binds nothing, and making it a type parameter turns
determinism and absence of side effects into compiler-checked properties instead of review conventions.
That is what lets one contract serve policy evaluation, where any I/O would be unsound, and embedded
function execution, where host bindings are the entire point.

gears/system/policy-engine/docs/PRD.md
```diff
@@ -0,0 +1,1054 @@
...
+
+## 1. Overview
+
+### 1.1 Purpose
+
+The Policy Engine is the Gears platform gear that owns authorization and governance policy: it stores policy content, manages its lifecycle, evaluates it against operations and resources, and returns decisions together with the row-level constraints that Policy Enforcement Points compile into database queries. It is the platform's Policy Decision Point implementation, and the single place where policy content is authored, versioned, and audited.
+
+The gear follows the pattern [CredStore](../../../credstore/docs/PRD.md) established for credential storage: the gear owns the domain model, lifecycle, and contracts, while a pluggable backend performs the actual policy evaluation. This keeps Gears free of a mandated policy language — the evaluation backend is replaceable — while giving the platform a first-class home for policy content that no single consumer owns.
+
+### 1.2 Background / Problem Statement
```
aviator5, line 56: Currently, the PRD focuses on authorization-related problems - for example, the lack of a Gear for authorization policy management. However, as we recently discussed, the policy engine is intended to manage admission control policies rather than authorization policies.

I suggest refining the "Background / Problem Statement" section to clearly define the policy engine's scope and explicitly state that authorization is outside its area of responsibility.

gears/system/policy-engine/docs/PRD.md
aviator5, line 1: How about creating `gears/oss` directory and place all OSS-related gears (including `policy-engine`) there?
So, we will have the following structure:
```
gears
|- bss
|- oss
|- genai
|- ....
|- system // perhaps should be renamed to "core"
```
@nonameffh, @MikeFalcon77, @Artifizer, @entropyshift what do you think?

At the same time, `admission-control` gateway gear will be within `geras/system`.

fluiderson:
Superseded by #4498.

PR #4397, opened 2026-08-03
test(usage-collector): enforce OpenAPI contract

capybutler:
## Description

The gear publishes `docs/usage-collector-v1.yaml` by hand. Nothing failed when the
document and the code disagreed, so the document drifted.

This PR adds an in-tree test that fails on that disagreement, plus the document
corrections the test surfaced. The test builds the route registry from the
`OperationBuilder` registrars. It needs no server, no database, and no `Service`.
It compares the document against the code on:

- the set of operations, with their methods, operation ids, and tags
- each parameter, by name, location, and required flag
- each request body, by media type, required flag, and referenced schema
- each 2xx response, by status, media type, and referenced schema
- the canonical error set on each operation
- the authentication requirement of each operation
- that the document describes every component the runtime publishes
- that every `$ref` in the document resolves

Descriptions, summaries, examples, the contents of a field schema, response
headers, `info`, and `servers` stay under manual control. The module header states
which fields the test enforces. The test reads the document with `serde-saphyr`,
which is already in the lockfile, so no new crate enters the tree.

The test found four disagreements, all corrected here:

1. Schema names did not match the code. They are now `CreateUsageRecords*`,
   `QueryAggregatedUsageRecordsRequest`, and `Page_UsageType`.
2. The `metadata.<key>`, `$filter`, `$orderby`, and `$select` parameters were
   absent from the routes that accept them.
3. The `MetadataFilter` schema was published, but no route takes that body. It is
   removed.
4. Page size was published as `$top` alone, while the `OperationBuilder` registry
   declared `limit` alone. Both spellings are now declared on both sides — see
   below.

### Page size: `$top` and `limit` are both accepted

Since #4422, `toolkit::api::odata::ODataParams` declares `limit` with
`#[serde(alias = "$top")]`. The `OData` extractor folds `$top` — canonical OData,
OASIS OData 4.01 Part 2: URL Conventions, section 5.1.6 — and `limit` onto one
`ODataQuery.limit` slot, and rejects a request carrying both as a duplicate field.

The document and the `OperationBuilder` registry therefore declare both spellings.
Declaring either alone would under-report the accepted surface, which is the drift
class this test exists to catch. The list allowlist admits both, the aggregate
allowlist rejects both (aggregation is not paginated), and the `MAX_PAGE_SIZE`
violation names `$top` with the alias in its detail: the parsed `ODataQuery` does
not carry the spelling the caller sent, and naming `$top` matches the convention
`toolkit_odata` already uses for `InvalidLimit`.

**No breaking change and no version bump.** An earlier revision of this PR rejected
`$top` with `400 InvalidArgument` and bumped `cf-gears-usage-collector` to 0.3.0,
on the finding that no code path bound the parameter. #4422 bound it, so that
rejection and its version bump are both reverted. The crate stays at 0.2.5.

Two wire-visible deltas remain, neither of them breaking. The served OpenAPI
document now declares `$top` next to `limit`, which is additive. The over-cap
page-size violation carries a new `detail` string naming both spellings; its
`field` (`$top`), its `reason` (`VALIDATION`), and its HTTP status (400) are
unchanged. No request is accepted or rejected differently than before this PR.

Known gap, not fixed here: #4422 also bound `$skiptoken` as an alias of `cursor`,
and this gear's list allowlist does not admit it, so `?$skiptoken=…` answers 400
on a parameter the extractor honours. That is the same defect class as the `$top`
one and belongs in its own change.

### TimescaleDB plugin

`effective_page_size` documented its floor-to-1 as a guard against a page size of
zero that "the core gateway passes through unclamped". The gateway does not: the
toolkit `OData` extractor rejects a zero page size with `InvalidLimit` before any
handler runs. The floor still earns its place — an in-process SDK caller builds its
own `ODataQuery` and reaches neither that check nor `prepare_list_query` — so the
documentation and tests now give the reason that holds. The plugin change is two
doc comments and four test-function renames, with no production code touched, so
`cf-gears-timescaledb-usage-collector-plugin` is unbumped. `cf-gears-usage-collector-sdk`
has no source change in this PR.

## Type of Change
- [x] Documentation update

## Testing
- [x] Unit tests pass
- [ ] Integration tests pass

`cargo test -p cf-gears-usage-collector` — 380 pass, including the 8 new contract
tests and 4 new allowlist tests. `cargo test -p cf-gears-timescaledb-usage-collector-plugin`
— 133 pass.

The plugin's `--features postgres` integration suite was not re-run, because this
PR's plugin change is doc comments and test-function names with no production code
touched. No manual run either: the change has no behavior a test does not already
cover.

## Documentation

No README change applies. `docs/usage-collector-v1.yaml` carries the four
corrections above. `DECOMPOSITION.md`, `DESIGN.md`, `domain-model.md`, `ADR-0012`,
and the usage-emission and usage-query feature documents carry the same
corrections.

## Checklist
- [x] Self-review completed

## Related Issues
Builds on #4422, which bound `$top` and `$skiptoken` as aliases in the toolkit
`OData` extractor. #4413 tracked the toolkit-wide `$top` assumption that this PR
originally described as a follow-up; #4422 addressed it.

## Summary by CodeRabbit

* **API Contract**
  * Updated the Usage Collector OpenAPI contract to reflect current routes, schemas, identifiers, pagination, filtering, aggregation, and error definitions.
  * Separated accepted and rejected usage-record submissions.
  * Usage-type identifiers now require valid GTS instance IDs.

* **Querying**
  * Added equivalent `$top` and `limit` page-size parameters.
  * Added validation for page-size limits, ordering rules, unsupported parameters, metadata filters, and aggregation queries.
  * Clarified that unsupported projections are rejected.

* **Testing**
  * Added comprehensive contract validation to keep documented and served APIs aligned.


gears/system/usage-collector/docs/features/usage-query.md
```diff
@@ -160,7 +167,7 @@ User-facing interactions that start with an actor (human or external system) and
...
...ence is resolved by a pre-dispatch `get_usage_type` (Method 7) call, an unregistered `gts_id` surfacing as canonical `NotFound` (404) before dispatch; that resolution also yields the usage `kind`, which feeds a pre-dispatch op-kind check that rejects a mismatched `(op, kind)` pair — `SUM` on a gauge, or `MIN` / `MAX` / `AVG` on a counter — with `InvalidArgument` (`400`, `.reason="OP_NOT_ALLOWED_FOR_KIND"`; counter admits `{SUM, COUNT}`, gauge admits `{MIN, MAX, AVG, COUNT}`)), `cpt-cf-usage-collector-algo-usage-query-pdp-constraint-composition-v2` intersects the PDP constraint set with the user-supplied filters under intersection-only (narrowing) semantics, `cpt-cf-usage-collector-algo-usage-query-plugin-spi-aggregate-dispatch-v2` invokes the Plugin SPI `query_aggregated_usage_records` capability so the storage plugin executes the chosen aggregation and any `group_by` dimensions server-side, `cpt-cf-usage-collector-algo-usage-query-active-and-inactive-record-visibility` enforces that both `active` and `inactive` rows within the authorized scope contribute to the result, and the gateway returns a `AggregationResult` (`gts_id`, `aggregation`, `buckets`) per `usage-collector-v1.yaml`.
```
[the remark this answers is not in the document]
capybutler: Already dispositioned at the section level, so no change here.

§1.6 (lines 121-125) states:

> Both surfaces are wired end-to-end; the legacy aggregation-request body shape described below in **§§2-5** (with mandatory `time_range`, single-UsageType arity check, typed `aggregation` operator on the wire) is **not** the shape actually implemented. Treat those sections as forward-looking intent that has since been superseded by the SDK-aligned shape described here.

and the paragraph after it names the replacements explicitly (`QueryAggregatedUsageRecordsRequest` with `op` + `group_by`; `gts_id` / `$filter` / `metadata.<key>` as query parameters) and records that `AggregationRequest`, `AggregationSpec` and `MetadataFilter` no longer exist in `usage-collector-v1.yaml`.

All four cited lines (170, 186, 434, 859) plus 728 fall inside §§2-5, so they are covered by that statement. Adding a per-passage legacy marker would repeat it across most paragraphs of three sections without adding information a reader does not already have from §1.6.

gears/system/usage-collector/docs/features/usage-query.md
```diff
@@ -205,7 +212,7 @@ User-facing interactions that start with an actor (human or external system) and
...
...lugin dispatch, `cpt-cf-usage-collector-algo-usage-query-attribution-and-pdp-authorization-on-read` resolves the caller into a `SecurityContext` and binds the `(PdpDecision, PdpConstraint set)` envelope to the request, `cpt-cf-usage-collector-algo-usage-query-pdp-constraint-composition-v2` AND-merges the PDP constraint set into the parsed `FilterNode<UsageRecordFilterField>` under intersection-only (narrowing) semantics, `cpt-cf-usage-collector-algo-usage-query-cursor-pagination-orchestration-v2` projects the validated cursor to the plugin keyset `(created_at, id)`, `cpt-cf-usage-collector-algo-usage-query-plugin-spi-raw-page-dispatch-v2` invokes the Plugin SPI `list_usage_records` capability with the structured tuple `(filter_ast, order_keys, page_after, limit)` bounded by `limit`, `cpt-cf-usage-collector-algo-usage-query-active-and-inactive-record-visibility` enforces that both `active` and `inactive` rows within the authorized scope are returned, and the gateway mints the next `CursorV1` from `last_keyset` (when present) bound to the current `$filter` / `$orderby` and returns a `toolkit_odata::Page<UsageRecord>` envelope with `@nextLink` cursor URL per `usage-collector-v1.yaml`.
```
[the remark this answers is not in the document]
capybutler: Applied, with one correction to the finding and one deliberate deferral.

**Applied:**

- **`gts_id` removed from the optional `$filter` predicate lists** (4 occurrences). Confirmed against `usage-collector-sdk/src/models.rs:1028-1034` — `gts_id` is intentionally absent from `UsageRecordQuery`, so `parse_odata_filter::<UsageRecordFilterField>` rejects a `gts_id`-touching predicate at parse time as `FilterError::UnknownField`. A caller following the old text got a 400. The algorithm-input passage that described "the `gts_id` field on `UsageRecordFilterField` is reserved" was also wrong — there is no such field — and now says so.
- **`$orderby` normalization.** The seven "MUST project the canonical keyset `(created_at, id)`" claims now describe what the gateway actually does: `prepare_list_query` appends the canonical unique suffix via `ODataOrderBy::ensure_tiebreaker` in the caller order's direction (`handlers/usage_records.rs:529`), and an omitted `$orderby` normalizes to `created_at asc, id asc`. Callers do not supply it verbatim. The same stale claim in `DECOMPOSITION.md` is fixed as well.
- **Mixed-direction / nullable-key rejections reattributed to `VALIDATION`**, matching the two guards at `handlers/usage_records.rs:484` and `:517`.

**Correction to the finding:** `INVALID_ORDERBY_FIELD` *is* a real wire reason — `libs/toolkit-odata/src/problem_mapping.rs:33-39` emits it for `Error::InvalidOrderByField`, which is reachable both from `$orderby` parse failures at the toolkit extractor and from a malformed signed-token order via `ODataOrderBy::from_signed_tokens` during cursor validation. So it is not a blanket rename: the mentions in the cursor-validation bullet and in the two downstream-contract error-category sets are correct and stayed. Only the three passages that attributed *canonical-keyset projection* failures to it were wrong. The cursor bullet now names both paths so the distinction is explicit.

**Deferred:** the `timestamp` → `created_at` rename (16 occurrences). §1.6's "Naming reconciliation" note already declares `timestamp` a legacy synonym for `created_at` pending a full spec refresh, so the meaning is unambiguous today; doing the rename here would add a large diff to a PR scoped at doc-vs-code alignment without changing what the document says.

gears/system/usage-collector/docs/usage-collector-v1.yaml
```diff
@@ -396,27 +459,23 @@ components:
...
       required: false
-      description: Number of items per page (1..200).
+      description: Number of items per page (1..1000).
       schema:
         type: integer
         minimum: 1
-        maximum: 200
-        default: 50
+        maximum: 1000
+        default: 100
```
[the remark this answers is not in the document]
capybutler: Not changing — the premise does not hold.

`#/components/parameters/Limit` is referenced exactly once, by `list_usage_types` (line 314). `list_usage_records` declares its own `limit` inline at line 148 with no `default`, so it is not governed by this schema and the `MAX_PAGE_SIZE` normalization on the raw path is irrelevant to it.

On the one operation that does reference `Limit`, `default: 100` is accurate:

- `handle_list_usage_types` forwards the extracted `ODataQuery` unchanged — it never calls `prepare_list_query`, so there is no gateway-side `MAX_PAGE_SIZE` fill-in on this path.
- `CatalogStore::list` resolves the page size through `effective_page_size(query.limit, DEFAULT_PAGE_SIZE)` with `DEFAULT_PAGE_SIZE = 100` (`timescaledb-usage-collector-plugin/src/infra/storage/catalog_store.rs:54,351`).

An omitted `limit` therefore does return 100 items, which is exactly what an OpenAPI parameter-level `default` documents.

One genuine adjacent concern, though not the one raised: that 100 is *plugin-owned*, so a different storage plugin could resolve an absent `limit` differently while the gear's contract keeps promising 100. Either pinning the default in the core (normalizing an absent `limit` on the usage-type list path, mirroring the raw path) or dropping `default` because the contract cannot guarantee it would be defensible. That is a behavior/contract decision independent of this PR's doc-vs-code alignment, so it is not being made here.

gears/system/usage-collector/usage-collector/src/api/rest/handlers/usage_records.rs
```diff
@@ -377,10 +377,16 @@ pub const MAX_METADATA_FILTER_VALUES: usize = 32;
...
+///
+/// `limit` is the page-size parameter, and the only one. Canonical
+/// `OData` spells it `$top`, but [`toolkit::api::odata::ODataParams`]
+/// declares no `$top` field, so nothing would bind it — admitting it
+/// here would accept the caller's page size and silently ignore it,
+/// returning a full [`MAX_PAGE_SIZE`] page they would read as their
+/// requested one. It is therefore rejected as unrecognised, which tells
+/// the caller to send `limit` instead of leaving them to infer it from
+/// a page that looks complete.
+const OUR_ODATA_PARAMS: &[&str] = &["$filter", "$orderby", "$select", "limit", "cursor"];
```
parviz-93: **Mark the `$top` rejection as a breaking change, and bump the crate to 0.3.0.**

This commit removes `"$top"` from `OUR_ODATA_PARAMS`. The change is wire-visible: `GET /v1/records?$top=50` returned 200 before this commit and returns 400 after it. The old published document named `$top` as the page-size parameter.

The commit type is `test(...)`, with no `!` marker and no `BREAKING CHANGE:` footer. The crate version stays 0.2.3. As a result, release-plz ships the break with no version signal.

CONTRIBUTING.md gives the rules:
- `:270` — the `test` type is for tests only.
- `:279` and `:299` — a breaking change must carry the marker or the footer.
- `:421-431` — a breaking change must get a MAJOR bump (0.3.0 before 1.0).
- `:536-538` — the PR description must give the reason for the bump category.

Also make the squash-merge title shorter. It has 81 characters. The guideline at `:295` gives a limit of 50.

gears/system/usage-collector/docs/usage-collector-v1.yaml
```diff
@@ -139,14 +145,29 @@ paths:
...
+        - name: metadata.<key>
+          in: query
+          required: false
+          description: |
+            Repeated metadata-filter entries. Values for the same key are
+            OR-ed; distinct keys are AND-ed. Each `<key>` MUST be a member
+            of the queried usage type's declared `metadata_fields`.
+          schema:
+            type: string
+        - name: $select
```
parviz-93: **Remove the new `$select` entry, or implement the projection.**

This commit adds `$select` to the document as a projection that works. The same entry exists on the aggregate operation. No code applies it: the toolkit extractor parses `ODataQuery.select`, the handlers return full DTOs, and the plugin SQL uses fixed column lists. No code calls `apply_select`.

`GET /records?$select=id` returns 200 with complete records. `$select=nonexistent_field` also returns 200.

This is the accept-and-ignore behavior that the new `$top` comment in this PR names as the reason for rejection. The document did not contain `$select` before this commit.

gears/system/usage-collector/docs/usage-collector-v1.yaml
```diff
@@ -139,14 +145,29 @@ paths:
             pagination requires a uniform-direction, never-null keyset.
           schema:
             type: string
-        - name: $top
+        - name: limit
```
parviz-93, line 169: **Record the decision to rename `$top` to `limit` in place, or accept `$top` as a deprecated alias for one release.**

CONTRIBUTING.md:461-463 states the policy: a breaking REST change requires a new API version and a deprecation period.

The server never honored `$top`. Callers got a full page without an error. So this change can count as a document fix plus stricter validation. Write that decision in the PR description or the changelog.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+use toolkit::api::operation_builder::{OperationSpec, ParamLocation, RequestBodySchema};
+
+use super::{usage_records, usage_types};
+
+/// The contract document, embedded at compile time. Moving or deleting it
+/// is a build error rather than a silently skipped test.
+const YAML: &str = include_str!("../../../../../docs/usage-collector-v1.yaml");
+
+/// Build a registry populated by the gear's full REST surface.
+fn registry() -> OpenApiRegistryImpl {
```
parviz-93, line 71: **Use one registrar composition for production and for this registry.**

`registry()` calls the two current registrars by hand. This duplicates `register_routes` (`routes/mod.rs:24-26`).

If a developer wires a third registrar only into production, no test in this file sees it. If the YAML also misses the new endpoints, the set-equality check compares two incomplete views and stays green. This is the drift class that this suite exists to prevent.

Extract a Service-free `register_api_routes(router, openapi)`. Call it from `register_routes` before `.layer(Extension(service))`, and from this helper. The registrars already take `(Router, &dyn OpenApiRegistry)`, so the extraction compiles as-is.

After that change, the hardcoded count of 9 below only guards against symmetric shrinkage. Keep it or drop it, but decide.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+    serde_saphyr::from_str(YAML).expect("usage-collector-v1.yaml must be valid YAML")
+}
+
+/// Canonical operation key shared by both sides, e.g.
+/// `GET /usage-collector/v1/records`.
+fn op_key(method: &str, path: &str) -> String {
+    format!("{} {}", method.to_uppercase(), path)
+}
+
+fn registry_ops(reg: &OpenApiRegistryImpl) -> BTreeMap<String, OperationSpec> {
```
parviz-93, line 101: **Compare the YAML against the `build_openapi()` output, not against hand-projected `OperationSpec` fields.**

The functions `registry_ops`, `registry_params`, `registry_request_body`, `registry_success_responses`, and `missing_standard_errors` re-implement the projection that `OpenApiRegistryImpl::build_openapi` (`libs/toolkit/src/api/openapi_registry.rs:114`) already owns. As a result, the tests check the pre-serialization structs, not the served document.

Three blind spots exist today:
- `operationId` falls back to `handler_id` (`openapi_registry.rs:126`). If a route drops `.operation_id()`, this suite forces the YAML to disagree with the served `/openapi.json`.
- `build_openapi` forces `required: true` on path parameters (`:171-176`). `registry_params` uses `param.required` as-is.
- `build_openapi` rewrites paths with `axum_to_openapi_path` (`:281`). This file keys on raw axum paths.

Serialize `registry().build_openapi(&OpenApiInfo::default())` to `serde_json::Value`. Then run the existing YAML-side extractors over both documents. One extraction family remains instead of two.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+            yaml_success_responses(&doc, &key, &op),
+            registry_success_responses(spec),
+            "{key}: 2xx response mismatch (status, media type, schema)",
+        );
+    }
+}
+
+/// Kept in sync with `OperationBuilder::standard_errors` in
+/// `libs/toolkit/src/api/operation_builder.rs`.
+const STANDARD_ERROR_STATUSES: &[u16] = &[400, 401, 403, 404, 409, 429, 500];
```
parviz-93: **Take the canonical error set from the toolkit. Do not keep a hand-written copy.**

`STANDARD_ERROR_STATUSES` duplicates `OperationBuilder::standard_errors` (`operation_builder.rs:1436-1444`). Only a comment keeps the two lists equal.

`missing_standard_errors` checks one direction only. Extra statuses on the registry side are invisible. If the toolkit adds a status, no assertion fires. The toolkit rustdoc already drifted once: `operation_builder.rs:1390` lists 422, and the body denies it.

Fix: register a probe `OperationBuilder::get("/probe").standard_errors(&reg)` into a throwaway registry. Then read the statuses back from `reg.operation_specs`. `OperationBuilder.spec` is private.

If the `build_openapi` comparison lands, this item is not necessary.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+/// without it is wrong, not merely undocumented. Phrasing it as
+/// agreement would leave the check disarmed by the one edit most likely
+/// to happen: dropping `.standard_errors(openapi)` from a route and
+/// `default:` from its operation together, after which "both sides
+/// agree" holds vacuously and the document promises nothing.
+///
+/// A route that genuinely cannot produce these has to say so by editing
+/// this test.
+#[test]
+fn every_operation_declares_the_standard_error_set() {
```
parviz-93, line 765: **Add the 503 that every route registers to the enforced set.**

All 9 routes call `.error_503(openapi)`. The checked status set does not contain 503. If a route drops `.error_503`, every test stays green.

The YAML makes no explicit 503 promise: the status lives under `default`. So the missed drift degrades the runtime document, not a stated YAML promise. It is still the type of silent contract loss that this suite exists to catch.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+    let reg = registry();
+    let doc = spec_doc();
+    let registered = registry_ops(&reg);
+
+    for (key, op) in yaml_ops(&doc) {
+        let spec = spec_for(&registered, &key);
+
+        assert!(
```
parviz-93: **Make this check look inside the `default` response, not only at its presence.**

A mutation test proves the gap: replace `default: $ref: '#/components/responses/Problem'` with an inline `{description: ...}`, and all 8 tests stay green. The module header claims that the canonical error surface is enforced on both sides.

The explicit non-2xx blocks in the document get no comparison at all: aggregate 400/404, create-type 400/409, delete-type 404/409. `yaml_success_responses` filters to 2xx.

Fix: deref `default` and check for `application/problem+json` with a `Problem` schema ref. Apply the same check to the explicit non-2xx blocks.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+                doc.pointer("/components/securitySchemes")
+                    .and_then(|schemes| schemes.get(scheme))
+                    .is_some(),
+                "{key}: `security` names `{scheme}`, which \
+                 `components.securitySchemes` does not declare",
+            );
+        }
+    }
+
+    !requirements.is_empty()
```
parviz-93: **Treat `security: [{}]` as unauthenticated.**

The check is `!requirements.is_empty()`. In OpenAPI 3.x, an empty security-requirement object grants anonymous access.

A mutation test proves the gap: add `security: [{}]` to an `.authenticated()` operation, and all tests stay green. The document then advertises anonymous access. This is the drift that `security_matches_authenticated_routes` exists to catch.

Fix: treat an array that contains `{}` as unauthenticated, or panic on it as an unsupported construct.

parviz-93 (commented):
The High findings are in the inline comments. The cleanup and minor items are in a separate comment below.

parviz-93:
The inline review covers the High items. This list holds the Cleanup and Minor items. None of them block the merge.

Paths are relative to `gears/system/usage-collector/`. `tests` = `usage-collector/src/api/rest/routes/openapi_contract_tests.rs`.

## Cleanup

1. **Hoist the duplicated field-violation extractors and the `p()` helpers to file scope** (`handlers/usage_records_tests.rs:1127`, `:1760`, `:1867`). This PR adds two identical extractors under two names, plus an inline copy. The file also gains a fourth `p()` copy (`:1497`, `:1663`, `:1756`, `:1826`). One file-level helper for each is enough.

2. **Fold the field-name assertion into the existing limit-above-cap test** (`handlers/usage_records_tests.rs:1157`). The new test repeats the setup of the test at `:1140` to add one assertion.

3. **Collapse the five repeated walk prologues into one `for_each_documented_op` helper** (`tests:137/293/470/527/635`). If the `build_openapi` comparison lands, this item is not necessary.

4. **Compare parameter schema types too** (`tests:187`). `ParamTriple` holds (name, in, required) only. A document that declares `limit: {type: string}` passes while the runtime publishes integer. The module header discloses this scope, so this is an extension, not a broken promise.

5. **Replace the parameter override merge with concat plus a no-duplicate assertion** (`tests:97`). The keyed merge implements the OpenAPI collision rule for a collision that occurs zero times in this document. The change deletes `param_key`.

6. **Name the comparison structs** (`tests:365`). `RequestBodyTriple` puts a bool between two Strings. Assert output prints unlabeled tuples. Three small derive structs fit the same BTreeSets.

## Minor / hardening

7. **Model the `metadata.<key>` family honestly on both sides** (`docs/usage-collector-v1.yaml:156`, `:209`; `routes/usage_records.rs:57`, `:95`). OpenAPI has no syntax for dynamic parameter names. A generated client emits a parameter that no caller can send as written. The shape predates this PR, but the contract test now forces both sides to move together, so this is the right moment.

8. **Document the default page size on GET /records** (`docs/usage-collector-v1.yaml:151`). An omitted `limit` becomes 1000 (`handlers/usage_records.rs:441`). The usage-types `Limit` component documents default 100.

9. **Reconcile the over-limit behavior of the two list endpoints, and correct the plugin comment** (`handlers/usage_types.rs:74`; plugin `query.rs:18`). `GET /usage-types?limit=5000` returns 200 with a silent clamp to 1000. The same value on `GET /records` returns 400. The rewritten comment in `query.rs:18` claims that the gateway rejects `limit > 1000` before any plugin call. That claim is true only for the records path.

10. **Correct the `AggregationBucket.key` description or the DTO** (`docs/usage-collector-v1.yaml:865`). The text says that `key` is omitted when `group_by` was empty. The DTO (`dto.rs:411-413`) has no `skip_serializing_if`, so the wire always carries `"key": []`.

11. **Correct the toolkit `InvalidLimit` mapping that names `$top`** (`libs/toolkit-odata/src/problem_mapping.rs:106`). `GET /records?limit=0` returns a Problem whose violation field is `$top`. The caller never sent that parameter and can no longer send it. This needs a toolkit change.

12. **Extend unknown-parameter rejection to the usage-types list, or record the asymmetry as deliberate** (`handlers/usage_types.rs:74`). The rationale in the new `OUR_ODATA_PARAMS` comment applies to this endpoint too. It has no parameter guard at all.

13. **Bump `info.version`** (`docs/usage-collector-v1.yaml:18`). The document renames eight schemas, removes `$top`, changes the `Limit` bounds, and reshapes the per-record ack. The version stays 1.1.0.

14. **Split the commit.** Per `CONTRIBUTING.md:297`, one logical change per commit. The behavior change, the 657-line harness, and the document rewrite revert as one unit.

15. **Tie the documented page-size bound to the constant** (`docs/usage-collector-v1.yaml:148-155`). `maximum: 1000` restates `MAX_PAGE_SIZE` (`handlers/usage_records.rs:365`), and no test connects them.

16. **Disclose or enforce the `x-odata-filter` / `x-odata-orderby` extensions** (`tests:29`). The runtime document publishes `allowedFields`. The test compares nothing there, and the "deliberately NOT enforced" list does not name this axis.

17. **Add an orphan-component check** (`tests:600`). Containment is one-way. A documented schema that loses its last `$ref` stays in the document forever with every test green.

18. **Harden the `$ref` walker edges** (`tests:194`, `:210`). `resolve_ref` reports an external `other.yaml#/...` reference as a dangling target. `collect_refs` skips a non-string `$ref:` value without an error, and it would flag literal `"$ref"` keys inside `example` data. None of these occur in the current document.

19. **Add operation context to the `name_and_location` panics** (`tests:253`). A bare `expect("parameter name")` names neither the operation nor the parameter.

20. **Tighten the `UsageTypeGtsId` pattern with care** (`docs/usage-collector-v1.yaml:913`). `[^~]+` accepts tails that the server rejects with 400. `maxLength: 512` is enforced by no layer. A tightened pattern must accept both the segment grammar and UUID instance tails.

21. **Resolve `discriminator.mapping` targets in the ref audit** (`tests:214`). `collect_refs` reads only literal `$ref` keys. A dangling mapping target in `CreateUsageRecordResult` (`docs/usage-collector-v1.yaml:750-754`) stays green.

22. **Restore aggregate-path coverage for the `$top` rejection** (`handlers/usage_records_tests.rs:1852`). The commit drops `"$top"` from the forbidden-params loop. No remaining test checks that path.

23. **Optional: extract the gear-agnostic harness core before a second gear copies it.** Only `registry()` and the `include_str!` path are gear-specific. `account-management` carries the same unenforced contract YAML.

gears/system/usage-collector/docs/usage-collector-v1.yaml
```diff
@@ -108,9 +116,7 @@ paths:
           required: true
           description: Single UsageType GTS id being queried.
           schema:
-            type: string
-            minLength: 1
-            maxLength: 512
+            $ref: "#/components/schemas/UsageTypeGtsId"
```
parviz-93, line 119: **Correct the operand list in the `$filter` description of GET /records** (line 127, eight lines below — the PR does not change that line, so GitHub anchors this thread here).

The text claims that the declared `metadata_fields` of the queried usage type are valid operands. The filter grammar is the closed `UsageRecordFilterField` set (`usage-collector-sdk/src/models.rs:1051-1106`). A comment in that file states that metadata filtering does not flow through OData. If a client obeys the document and sends `region eq 'eu'` in `$filter`, the server returns `400 INVALID_FILTER`.

The text also contradicts the `metadata.<key>` parameter below it, which is the real metadata channel. The list also omits `id` and `corrects_id`, which are valid operands.

gears/system/usage-collector/usage-collector/src/api/rest/handlers/usage_records.rs
capybutler: Done in 7acbcfa4. `cf-gears-usage-collector` goes 0.2.3 → 0.3.0, the commit carries `fix(usage-collector)!:` plus a `BREAKING CHANGE:` footer so release-plz gets the signal, and the PR title now carries the same marker at 53 chars, down from 81.

The bump landed as a separate commit rather than an amend of the `test(...)` one, so both the `!` marker and the footer survive a squash-merge. The bump justification is in the PR description per `CONTRIBUTING.md:536-538`.

One verification detail worth having on the record, because it decides the deprecation question in your other comment: the wire binding for this parameter family lives in exactly one place, `toolkit::api::odata::ODataParams` (`libs/toolkit/src/api/odata.rs:12-22`), which declares `$filter`, `$orderby`, `$select`, `limit` (unprefixed), and `cursor`. There is no `$top` field and no `#[serde(alias = "$top")]`, and `toolkit-odata` does no HTTP query parsing at all — it operates on already-parsed values. So `?$top=50` left `limit` as `None`, the gateway filled `MAX_PAGE_SIZE`, and the caller received 200 with up to 1000 records.

To be clear about what was wrong: `$top` is canonical OData (OASIS OData 4.01 Part 2: URL Conventions, section 5.1.6, "System Query Options $top and $skip"). The parameter was not invented — it was documented as the page-size parameter and never wired to anything.

gears/system/usage-collector/docs/usage-collector-v1.yaml
capybutler: Recorded in the PR description, under "Breaking change: `$top` is rejected, `limit` is the page-size parameter". The decision is written out: rename in place, no `/v2` and no deprecated alias.

The reasoning, since you cite `CONTRIBUTING.md:461-463`: that policy protects callers who depend on a working parameter, and there was none to protect here. A deprecated `$top` alias would have to either keep ignoring the value — preserving the bug under a deprecation notice — or start honoring it, which is a larger behavior change than the rejection. The gear is pre-release with no installations, so there is no deployed client to migrate.

Your read is the one the description now states, with the evidence attached: `ODataParams` (`libs/toolkit/src/api/odata.rs:12-22`) declares no `$top` field and no serde alias, so no code path ever bound it. Document fix plus stricter validation.

Listed in the description as follow-up and deliberately not fixed here: the same "the toolkit binds `$top`" assumption is written into four places outside this gear, all latent rather than live wire bugs.

- `libs/toolkit-odata/src/problem_mapping.rs:106` emits a `field_violations` entry named `"$top"` for the `limit` parameter, pinned by the toolkit test `invalid_limit_keys_to_top`.
- `libs/toolkit-odata/src/limits.rs` carries `ODataLimits::max_top` / `validate_top` as dead code — `ODataLimits` is re-exported and never constructed anywhere in the repo.
- `gears/system/account-management/account-management/tests/api_children_test.rs:80-93` sends `?$top=999999` and asserts a clamp to `max_top`. It passes vacuously: `$top` never binds, `limit` stays `None`, and the default fills `max_top`, so the assertion holds whether or not the clamp works. Its sibling at `api_users_test.rs:234-239` has a comment reading `$top=1` over a URL that sends `limit=1`.
- `gears/chat-engine/chat-engine/src/domain/search.rs:49-52` states that "the `toolkit-odata`-derived `ODataQuery` lifts these field names verbatim, hence the leading `$` on the wire". It does not; chat-engine works because it hand-rolls its own `#[serde(rename = "$top")]` DTO, which makes it the one gear where `$top` is honored.

A toolkit-wide wire change affecting every gear that uses the `OData` extractor does not belong in a contract-test PR.

gears/system/usage-collector/usage-collector/src/api/rest/handlers/usage_records.rs
capybutler: Withdrawing my earlier reply on this thread — the change it described is no longer in the branch.

#4422 landed after it and bound `$top`: `toolkit::api::odata::ODataParams.limit` now carries `#[serde(alias = "$top")]`, so the extractor folds `$top` and `limit` onto one `ODataQuery.limit` slot. The premise of the rejection — that nothing binds the parameter — stopped holding, so the rejection and the `0.3.0` bump that signalled it are both reverted. `cf-gears-usage-collector` stays at `0.2.5`, and the commit is `test(...)` with no `!` and no footer, which is now the correct classification.

One wire-visible change does remain, from your `$select` comment: it is rejected `400` where it previously returned `200` and ignored the projection. `$select` was not in the published document at `0.2.5`, so no documented promise breaks — but if you read that as triggering `:421-431`, say so and I will add the bump.

The title is fixed independently: `test(usage-collector): enforce OpenAPI contract`, 47 characters against the ≤50 guideline (now `CONTRIBUTING.md:309`).

gears/system/usage-collector/docs/usage-collector-v1.yaml
capybutler: Dropped, in all three places it lived.

`.with_odata_select()` is off both list registrars and the `$select` entries are out of the document. The allowlist entry went too — `OUR_ODATA_PARAMS` no longer contains it, so `GET /records?$select=id` answers `400` naming the parameter instead of `200` with every field. Leaving the entry would have preserved exactly the behaviour you describe, just undocumented; the allowlist's own doc comment argues that an entry with no binding behind it is worse than a rejection.

Pinned by `select_is_rejected_because_no_projection_is_applied` and an explicit "MUST NOT declare `$select`" assertion on the usage-types route spec, so re-adding `.with_odata_select()` without implementing the projection fails.

Not covered: `/usage-types` has no parameter allowlist at all, so it still accepts and ignores `$select` — along with every other unrecognised parameter. Rejecting there means adding validation to a route that has none. The parameter is now undocumented on both routes but refused only on `/records`.

gears/system/usage-collector/docs/usage-collector-v1.yaml
capybutler: Withdrawing my earlier reply here too — it recorded a decision the branch no longer implements.

Your alternative is what ships. #4422 bound `$top` as an alias of `limit`, so both spellings work and neither is deprecated: no rename, no `/v2`, and `CONTRIBUTING.md:461-463` is not engaged because nothing is removed. The document and the `OperationBuilder` registry declare both spellings — declaring one alone would under-report the accepted surface, which is the drift class this PR's test exists to catch.

The follow-up list in that old reply is closed: #4413's four items were all resolved by #4422.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: Extracted. `register_api_routes(router, openapi)` in `routes/mod.rs` composes both registrars; `register_routes` is that call plus `.layer(Extension(service))`, and `openapi_contract_tests::registry()` goes through the same function.

Verified by mutation: registering an extra route inside `register_api_routes` and leaving the document alone now fails `harness_sees_the_whole_rest_surface` and `operation_identity_matches`. With the duplicated composition it was invisible.

On the count of 9 — kept, with the reasoning in a doc comment. `operation_identity_matches` already compares the two sets, so the count adds exactly one thing: it fires when a route leaves both sides in the same commit, where set equality holds vacuously. That makes deleting a route a deliberate edit rather than a silent shrink.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: Taking the gaps, not the rewrite. Each one first:

- **`operationId` fallback — real.** Fixed by putting it out of reach: every route must now declare `.operation_id(…)`, so the `handler_id` default cannot apply. Worth naming the exact failure mode, because it is not a silent pass — the suite goes *red*, comparing the documented id against `None`. The hazard is the repair: deleting `operationId` from the document turns it green while `/cf/openapi.json` still serves `get:_usage-collector_v1_records`. The new assertion prints that derived id so the intended fix is unambiguous.
- **Path-parameter `required` — not reachable.** `build_openapi:192-193` forces it, but `path_param()` hardcodes `required: true` (`operation_builder.rs:587`) and is the only way to push a `ParamSpec` with `location: Path`. Mirrored in `registry_params` anyway.
- **`axum_to_openapi_path` — not reachable here.** It is `path.replace("{*", "{")` (`:59-63`) and no route uses a wildcard. `registry_ops` keys on the converted path regardless.

On the larger claim — that the suite should read the served document rather than `OperationSpec` — I generated the document and compared shapes before deciding. The two artefacts differ deliberately in most of what a document-to-document diff would touch: root-level `security` versus per-operation; `default: $ref: Problem` versus explicit `400,401,403,404,409,429,500,503`; hand-written `$filter` prose versus the auto-derived `"OData v4 filter expression\n- id: eq|ne|in…"`; `$ref: UsageTypeGtsId` versus inline `{"type":"string"}` on `gts_id`. A raw comparison flags all of it, so both sides need a normalizer — two extraction families again, plus a new exemption list. What survives normalization is the tuple set the suite already compares.

Where the idea does pay is narrow, so it is in: `registry_keys_match_the_generated_document` builds a real document from the registry and compares its operation keys against the ones every other check keys on. That covers the path rewrite and the `_ => HttpMethod::Get` collapse at `:279-286` (also unreachable — no constructor outside get/post/put/delete/patch) without dragging in the rest.

Mutation-checked both directions: with a `/files/{*rest}` route added, the key check passes with the conversion and fails without it (`{*rest}` vs `{rest}`).

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: `STANDARD_ERROR_STATUSES` is gone. `canonical_error_statuses()` registers a throwaway probe making the two calls every route here makes — `.standard_errors()` and `.error_503()` — into a scratch registry and reads the `application/problem+json` statuses back off the spec, as you suggested, since `OperationBuilder.spec` is private. It resolves to `{400,401,403,404,409,429,500,503}`.

That also settles the direction concern: a status added to either toolkit method now enters the expected set automatically and every route is held to it — the case a transcribed list could not catch.

The rustdoc drift you cite is still there, for the record: `operation_builder.rs:1495` and `:1529` advertise a 422 that `:1533` deliberately omits and the array at `:1541-1549` does not contain. Left alone as a toolkit change outside this PR.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: In the enforced set, via the same probe as the previous comment — it calls `.error_503()` alongside `.standard_errors()`, so 503 is derived rather than hand-added.

Verified by mutation: dropping `.error_503(openapi)` from one route now fails with `missing [503]`. It stayed green before.

Your reading of the document side is right and is recorded in the test's doc comment: the yaml makes no explicit 503 promise — the status rides `default:` — so this stays a registry-side requirement, which the test already was by design.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: Both halves fixed.

`assert_problem_envelope` derefs the response and requires a single `application/problem+json` media type whose schema is a `$ref` onto the `Problem` component. It runs on `default:` and on every explicit non-2xx block, so the aggregate 400/404, create-type 400/409 and delete-type 404/409 are compared now instead of being filtered out by the 2xx predicate. Each explicit status must also be one the route registers, or the document promises an error the runtime cannot emit.

Your mutation now fails: replacing the `$ref` with an inline `{description: …}` gives *"the `default` response must carry a `content` block holding the `Problem` envelope"*. A second one — pointing an explicit 404 at `application/json` / `UsageRecord` — gives *"the `404` response must be `application/problem+json`"*.

gears/system/usage-collector/usage-collector/src/api/rest/routes/openapi_contract_tests.rs
capybutler: Fixed. The reader returns `!requirements.is_empty() && !anonymous_alternative`, where an empty requirement object anywhere in the array sets the second flag — the entries are OR-ed, so one `{}` makes the operation public whatever the others name (OpenAPI 3.1.0, Security Requirement Object: *"To make Security optional, an empty Security Requirement (`{}`) can be included in the array"*).

Your mutation now fails: `security: [{}]` on the `.authenticated()` list route gives `left: false, right: true`. Added `an_empty_security_requirement_object_reads_as_unauthenticated` covering `[]`, `[{}]` and `[{bearerAuth: []}, {}]` directly, since nothing in the document exercises the construct.

gears/system/usage-collector/docs/usage-collector-v1.yaml
capybutler: Corrected. The description now enumerates the closed `UsageRecordFilterField` set — `id`, `created_at`, `tenant_id`, `resource_id`, `resource_type`, `subject_id`, `subject_type`, `corrects_id`, `status` — which adds the `id` and `corrects_id` you noted, and states that anything else is rejected `400 INVALID_FILTER` at parse time. Metadata keys are named explicitly as *not* operands, pointing at `metadata.<key>`, and so is `gts_id`, which is the required typed parameter.

Confirmed against `usage-collector-sdk/src/models.rs` (the `UsageRecordQuery` field list, plus the file comment stating metadata filtering does not flow through OData) and `toolkit-odata/src/problem_mapping.rs:25-31` for the reason code.

Two adjacent items I have not changed, same class but outside this thread:

- `DECOMPOSITION.md:274` repeats the metadata-keys-as-operands claim and adds that the `$filter` AST is type-checked against `metadata_fields` via a `get_usage_type` dispatch. `Service::list_usage_records` does authz → `require_bounded_time_window` → dispatch; there is no `get_usage_type` call on the raw path.
- The `metadata.<key>` description promises each key must be a declared member of `metadata_fields`, and nothing on the raw list path enforces it — the same accept-and-ignore shape as `$select`.

parviz-93 (approved):

gears/system/usage-collector/usage-collector/src/api/rest/routes/usage_records.rs
```diff
@@ -75,12 +104,16 @@ pub(super) fn register_usage_record_routes(
...
+        // projection. The handler returns whole `UsageRecordDto`s and the
+        // plugin selects a fixed column list, so declaring `$select` would
+        // advertise a parameter that is read and discarded.
         .with_odata_filter::<UsageRecordFilterField>()
         .with_odata_orderby::<UsageRecordFilterField>()
-        .with_odata_select()
         .standard_errors(openapi)
         .error_503(openapi)
         .register(router, openapi);
+    router = list_records_route;
```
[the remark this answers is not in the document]
capybutler: The sentence at `usage-query.md:149` is scoped to the two **record** paths, and both of those do reject `$select`:

- `handle_list_usage_records` → `reject_unknown_list_params` (`handlers/usage_records.rs:577`) against `OUR_ODATA_PARAMS = ["$filter", "$orderby", "$top", "limit", "cursor"]` (`:395`)
- `handle_query_aggregated_usage_records` (`:261`) → `reject_unknown_aggregate_params` (`:319`) against `AGGREGATE_ODATA_PARAMS = ["$filter"]` (`:313`)

Neither allowlist contains `$select`, so `$select` on either record path is a `400 InvalidArgument` with `field_violations[].reason="VALIDATION"` before the query pipeline runs. To remove the ambiguity that let the sentence read as covering the catalog list, it now says "rejected on both **record** paths".

On `routes/usage_types_tests.rs:176-182`: that assertion is that the list route does not **declare** `$select` in its registered `OperationSpec` — the parameter is absent from the published contract. It is not an assertion that a `$select` request is refused, so nothing there is contradicted by the handler's behaviour.

The underlying observation is correct, and I am not disputing it. `handle_list_usage_types` has no `reject_unknown_*` guard, so the toolkit parses `$select`, the handler forwards it inside `ODataQuery`, and `CatalogStore::list` never reads `query.select` — it selects a fixed `TYPE_COLUMNS` list. The caller gets `200` with whole objects and no signal that the parameter was dropped.

I am not changing that here, for three reasons:

1. It is a production behaviour change on a different endpoint — requests that return `200` today would return `400` — while this PR is scoped to aligning `usage-collector-v1.yaml` with the registered routes and adding the conformance test that keeps them aligned.
2. It is not drift the new test can detect: `$select` is declared for `usage_collector.list_usage_types` on neither side (no `.with_odata_select()` in the registrar, no `$select` parameter in the document), so the registry and the document agree.
3. The catalog list belongs to the usage-type-lifecycle feature, not the query feature whose document carries the flagged sentence.

Tracked as a follow-up instead: add `reject_unknown_list_params`-equivalent rejection to `handle_list_usage_types`, with coverage asserting the `400`, rather than widening this PR.

PR #4403, opened 2026-08-03
feat(toolkit/oop): gateway provider + reverse-proxy edge registration

### LINKED ISSUE #4107: [OoP-5] GatewayProvider + ToolKitGatewayProvider
**Parent**: [OoP](https://github.com/constructorfabric/gears-rust/issues/4102)

**Blocked by**: [OoP-4](https://github.com/constructorfabric/gears-rust/issues/4106)

### Description

Define the `GatewayProvider` trait abstraction and implement `ToolKitGatewayProvider` for reverse-proxying public routes through api-gateway.

**PRD**: `cpt-cf-fr-gateway-registration`, `cpt-cf-fr-gateway-abstraction`, `cpt-cf-fr-api-visibility`, `cpt-cf-fr-gateway-minimal`, `cpt-cf-fr-edge-modes` (Mode A)
**Design**: `cpt-cf-component-gateway-provider`, `cpt-cf-component-edge-architecture` (Mode A only)
**ADR**: `0003-cpt-cf-adr-gateway-abstraction`, `0007-cpt-cf-adr-edge-architecture`
**Effort**: Medium–Large

### Tasks

- [x] **5.1** Define `GatewayProvider` trait with typed wrappers (per DESIGN.md § 3.2) — mark as `unstable` per PRD § 7.1 `cpt-cf-interface-gateway-provider` (breaking change policy: minor version bump during unstable phase):
  - `GearName(String)` — gear identity
  - `OpenApiSpec<'a>` — enum: `Owned`, `Borrowed`, `SerializedJson`
  - `Endpoint { scheme, authority }` — validated network endpoint
  - `GatewayError` — typed error
  - `register_routes(&self, gear, spec, endpoint) -> Result<(), GatewayError>`
  - `deregister_routes(&self, gear) -> Result<(), GatewayError>`
- [x] **5.2** Implement `ToolKitGatewayProvider`:
  - Parse OpenAPI spec for public routes (`is_public`)
  - Add reverse-proxy routes to api-gateway's Axum router using `toolkit-http::HttpClient`
  - Propagate `Authorization: Bearer <jwt>` on proxied requests (no `x-secctx-bin`; SecurityContext is reconstructed from JWT at each hop per ADR-0008)
- [x] **5.3** Add reverse-proxy capability to `api-gateway` gear (currently missing entirely):
  - Accept route registrations from `ToolKitGatewayProvider`
  - Forward matching requests to OoP gear endpoints
- [x] **5.4** Wire into OoP bootstrap (OoP-4): call `register_routes` after HTTP server starts, `deregister_routes` on shutdown
- [x] **5.5** Validate Mode A (Embedded Edge) works end-to-end: api-gateway + authn-resolver + reverse-proxy to OoP gear, with the api-gateway scope constraint (DESIGN § 3.10) enforced
- [x] **5.6** Tests: register routes, proxy a request end-to-end, verify SecurityContext propagation through proxy, deregister

### Acceptance Criteria

- External request → api-gateway → reverse-proxy → OoP gear works with full SecurityContext propagation
- Only `is_public` routes are exposed through the gateway

### LINKED ISSUE #4401: [OoP-2.5] Close the platform-plane gRPC loop: SA-token reader + shape-discriminating validation
**Parent:** [OoP](https://github.com/constructorfabric/gears-rust/issues/4102)
 
### Context
 
OoP-2 (#4104) landed the platform-plane primitives and **inbound HTTP** enforcement:
`InternalCredential`/`PlatformIdentity`/`PlatformSecurityContext`/`PeerAuthenticated`,
the `InternalAuthenticator` trait, `attach_/extract_internal_token_{http,grpc}`,
`internal_auth_middleware`, and the `K8sTokenReviewAuthenticator`. All five OoP-2
acceptance criteria are met and the inbound HTTP plane is wired via
`build_internal_authenticator` → `OopServeOptions.internal_authenticator`.
 
Two OoP-2 task items were intentionally **not** completed because they had no
consumer in P1. This ticket tracks them:
 
- **2.6** `KubeServiceAccountToken` reader (watch projected volume, reload on rotation)
- **2.3** `validate_internal_auth` explicit token-shape discriminator (JWT vs. opaque)
 
### Current gaps
 
- **Outbound gRPC credential is not wired.** `InternalAuthInterceptor`
  (`libs/toolkit-transport-grpc/src/internal_auth.rs` exposes a `TokenProvider`
  seam but is never attached to any client (`with_interceptor`). Gears do **not**
  present an SA token on outbound `DirectoryService` calls.
- **No SA-token reader exists.** `InternalCredential::KubeServiceAccountToken.token_path`
  is declared but never read; there is no file-watch / periodic re-read / rotation
  handling. (The interceptor doc-comment already assumes "a reader at the bootstrap
  layer" that does not exist.)
- **Inbound gRPC validation is not enforced.** No `extract_internal_token_grpc`
  consumer exists in `gears/`; the orchestrator / `DirectoryService` RPCs do not
  validate an internal token. So the platform-plane gRPC loop is open on both ends.
- **No JWT-vs-opaque dispatch.** `K8sTokenReviewAuthenticator` submits every
  token to `TokenReview` unconditionally. 2.3's discriminator only becomes meaningful
  once a second (opaque / Profile-2 bootstrap) branch exists — which is deferred to P2.
 
### Scope
 
**In scope**
- [x] **2.6 — SA-token reader.** Read `KubeServiceAccountToken.token_path`, reload on
      rotation (inotify watch or TTL re-read), expose a `TokenProvider` that yields the
      current token.
- [x] **4.1 — Wire outbound.** Feed the reader's provider into `InternalAuthInterceptor::new`
      at the bootstrap layer and attach it to the `DirectoryService` gRPC client so
      system calls (register/heartbeat/resolve) carry `x-toolkit-internal-token`.
- [x] **Enforce inbound gRPC.** Validate the internal token on the orchestrator /
      `DirectoryService` RPCs via `extract_internal_token_grpc` + `InternalAuthenticator`,
      setting `PeerAuthenticated` for workload-policy checks (async — installed at the
      wiring layer, not the sync tonic `Interceptor`).
- [x] Unit/integration tests: rotation re-read, outbound attach round-trip, inbound
      reject-on-missing/invalid, `PeerAuthenticated` propagation.
 
**Deferred to P2 (track, do not implement here)**
- [ ] **2.3 — token-shape discriminator.** Add JWT-vs-opaque dispatch only when the
      opaque / Profile-2 `BootstrapToken` validation path lands; today it would be a
      single-branch pass-through.
 
### Acceptance criteria
- A Profile-3 gear attaches a rotating SA token to outbound `DirectoryService` calls,
  and the token survives projected-volume rotation without restart.
- `DirectoryService` RPCs reject calls lacking a valid internal token and set
  `PeerAuthenticated` on success.
- `cargo test` green for the affected crates; clippy clean under workspace lints.
 
### Non-goals
- Profile-2 bootstrap-token or mTLS validation/wiring (separate P2 work).
- Any change to the already-shipped inbound **HTTP** plane.

mattgarmon:
Closes: https://github.com/constructorfabric/gears-rust/issues/4107
Closes: https://github.com/constructorfabric/gears-rust/issues/4401

OoP gears advertise their REST API and the built-in `api-gateway` reverse-proxies external traffic to them via directory discovery.
 
- **Breaking:** removed `OperationBuilder::public()`. Split into two axes — `.anonymous()`/`.authenticated()` (auth) and `.exposed()`/internal-by-default (edge visibility). Migrate `.public()` → `.anonymous()` (+`.exposed()` if edge-facing). Updated all callers, examples, and docs.
- **New crate `libs/toolkit-gateway`** (leaf, no toolkit dep): `GatewayProvider` trait, `ProxyRegistry` (matchit route table + per-route auth), `Forwarder`/`proxy_fallback` reverse proxy, `ToolKitGatewayProvider`, plus E2E test.
- **Directory discovery:** new `ListAllInstances` RPC + `InstanceInfo.openapi_spec` field; SDK, gRPC client, and `gear-orchestrator` server handler/mappers updated.
- **`api-gateway` embedded edge:** new `proxy.rs` directory-sync loop (poll → reconcile → register/prune), `GatewayProxyConfig` (default off), proxy fallback mounted before middleware, auth policy consults registry.
- **Platform-plane internal auth:** new shared-secret + K8s `TokenReview` providers, `InternalAuthInterceptor` (`x-toolkit-internal-token`), and `AuthNResolverBearerAuthenticator` auto-wired via `ClientHub`.
- **Toolkit OoP runtime:** OpenAPI emits `x-cf-api-visibility` + bearer security; workers register REST endpoint/OpenAPI on serve and deregister on shutdown.
- **Docs:** updated ADR 0003 (gateway abstraction), `DESIGN.md`, `PRD.md`, web-docs, and slides.

## Summary by CodeRabbit

* **New Features**
  * Added directory-driven API gateway routing with embedded reverse-proxy support.
  * Added cross-service instance discovery, OpenAPI metadata sharing, and automatic route synchronization.
  * Added shared-secret and Kubernetes-based internal authentication options.
  * Added bearer authentication integration through the authentication resolver.
* **Enhancements**
  * Separated route visibility from authentication, supporting exposed, authenticated, and anonymous routes.
  * Improved proxy route authentication and handling of unavailable upstream services.
* **Documentation**
  * Updated architecture, guides, examples, and security guidance for the revised routing and authentication model.


libs/toolkit-gateway/src/forward.rs
```diff
@@ -0,0 +1,241 @@
...
+    for (name, value) in headers {
+        if strip_from_request(name) {
+            continue;
+        }
+        if let Ok(value) = value.to_str() {
+            builder = builder.header(name.as_str(), value);
+        }
+    }
+    builder
+}
```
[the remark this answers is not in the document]
mattgarmon: Forwarded headers not needed in this case, it is internal network

libs/toolkit-gateway/src/toolkit_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+            routes = templates.len(),
+            "registering gear proxy routes",
+        );
+        self.registry
+            .register(gear.clone(), endpoint.clone(), templates);
+        Ok(())
+    }
+
+    async fn deregister_routes(&self, gear: &GearName) -> Result<(), GatewayError> {
+        let removed = self.registry.deregister(gear);
```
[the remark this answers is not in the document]
mattgarmon: Fixed

gears/system/gear-orchestrator/src/server.rs
```diff
@@ -3,25 +3,76 @@
...
+                tracing::debug!(peer = %peer.name, "platform-plane call authenticated");
+                Ok(Some(peer))
+            }
+            Err(InternalAuthNError::InvalidToken) => {
+                Err(Status::unauthenticated("invalid internal token"))
+            }
+            Err(InternalAuthNError::Unavailable) => {
+                Err(Status::unavailable("internal-auth backend unavailable"))
+            }
+            Err(_) => Err(Status::internal("internal authentication failure")),
```
MikeFalcon77: **MEDIUM**

The catch-all arm maps every non-`InvalidToken`/`Unavailable` `InternalAuthNError` (including `Other(String)`, which carries a diagnostic message) to `Status::internal` while discarding the message and logging nothing.

Unlike the HTTP middleware's `log_authn_error`, a real platform-plane auth-backend failure on the gRPC path leaves no server-side trace, so the failure is undiagnosable in production.

Log the source error before returning: add a `tracing::warn!(error = %err, "platform-plane authentication failed")` in the catch-all arm so the underlying cause is preserved.

libs/toolkit-gateway/src/toolkit_provider.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// router rebuild: `instances` are registered (public routes extracted from
+    /// each spec) and `removals` are dropped.
+    ///
+    /// An instance whose spec cannot be parsed is logged and skipped so one bad
+    /// gear cannot stall the rest; its previous routes (if any) are retained
+    /// because it is neither re-registered nor removed.
+    pub fn apply_snapshot(&self, instances: Vec<InstanceSpec<'_>>, removals: &[GearInstance]) {
+        let mut additions = Vec::with_capacity(instances.len());
+        for inst in instances {
+            match extract_public_routes(&inst.spec) {
```
MikeFalcon77: **MEDIUM**

`apply_snapshot` calls `extract_public_routes` once per instance, so N co-located instances of the same gear re-parse the identical OpenAPI document N times every poll, even though `fetch_or_get_spec` already deduped the fetch to one `Bytes` per gear.

For the explicitly-supported multi-instance case (Profile 2, Host + Workers) this re-runs `serde_json` parsing and route extraction over the same document on every sync tick, wasting background-poll CPU proportional to instance count rather than gear count.

Cache the extracted `Vec<RouteTemplate>` per gear (keyed by `GearName`, mirroring the `Bytes` cache) and reuse it for all instances of that gear instead of parsing per instance.

libs/toolkit-gateway/tests/proxy_e2e.rs
```diff
@@ -0,0 +1,156 @@
...
+        RouteTemplate::new(http::Method::GET, "/calc/v1/items/{id}", true),
+        RouteTemplate::new(http::Method::POST, "/calc/v1/items/{id}", true),
+    ]
+}
+
+#[tokio::test]
+async fn proxies_get_and_post_and_propagates_bearer() {
+    let endpoint = start_upstream().await;
+    let registry = Arc::new(ProxyRegistry::new());
+    registry.register(
```
MikeFalcon77, line 103: **LOW**

The reverse-proxy e2e tests only ever register a single upstream, so the forward path is never exercised with multiple registered gears.

End-to-end "routes to the correct upstream among several" is verified only by the non-I/O `ProxyRegistry` unit tests, so a misrouting regression in the forwarder/router wiring would not be caught at the HTTP layer.

Add an e2e case that starts two distinct echo upstreams under different gears/paths, forwards to each, and asserts each response comes from the correct upstream.

gears/system/gear-orchestrator/src/server.rs
mattgarmon: added

libs/toolkit-gateway/src/toolkit_provider.rs
mattgarmon: done

libs/toolkit-gateway/tests/proxy_e2e.rs
mattgarmon: added test

libs/toolkit-gateway/src/forward.rs
```diff
@@ -0,0 +1,410 @@
+//! Reverse-proxy request forwarder over a shared [`ProxyRegistry`].
```
striped-zebra-dev, line 1: Just for future consideration: I used Pingora in Event Broker to build a custom proxy. It might make sense to generalize this functionality into a `toolkit-proxy` or `toolkit-gateway` library. This looks like a common pattern for complex HA services/gears.

striped-zebra-dev (approved):

libs/toolkit-gateway/src/provider.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// The trait is object-safe (via [`macro@async_trait`]) so the `OoP` bootstrap
+/// can hold it as `Arc<dyn GatewayProvider>` and inject the concrete provider at
+/// startup based on the deployment profile.
+///
+/// # Stability
+/// This trait is **unstable**: it may change in a minor release while the `OoP`
+/// gateway story stabilizes (see PRD § 7.1).
+#[async_trait]
+pub trait GatewayProvider: Send + Sync {
```
MikeFalcon77, line 36: The trait is not what the edge actually drives. `proxy.rs` constructs a concrete `ToolKitGatewayProvider` and calls `apply_snapshot`, which is an inherent method, not part of this trait. So a Kong/Tyk provider cannot be plugged into the sync loop without rewriting `proxy.rs` — which is the thing ADR-0003 promised the trait would prevent.

Suggestion: move the batched reconcile onto the trait (`apply_snapshot(&self, desired, removals)`) with a default impl over `register_routes`/`deregister_routes`, and have `spawn_directory_sync` take `Arc<dyn GatewayProvider>`.

gears/system/api-gateway/src/proxy.rs
```diff
@@ -0,0 +1,656 @@
...
+const MIN_SYNC_INTERVAL: Duration = Duration::from_secs(1);
+
+/// Upper bound on a single reconcile pass. A hung `list_all_instances` or
+/// `get_openapi_spec` call is abandoned after this so it can neither block
+/// subsequent ticks nor stall shutdown.
+const SYNC_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Spawn the background directory-sync loop. Returns immediately; the task runs
+/// until `cancel` fires.
+pub fn spawn_directory_sync(
```
MikeFalcon77, line 45: This takes `Arc<ProxyRegistry>` and builds the concrete provider internally, so the whole sync loop is hard-wired to the built-in edge. Taking `Arc<dyn GatewayProvider>` here is what makes Mode B (ADR-0007) reachable later.

gears/system/api-gateway/src/proxy.rs
```diff
@@ -0,0 +1,656 @@
...
+        gear: gear.clone(),
+        instance_id: inst.instance_id.clone(),
+        spec: OpenApiSpec::SerializedJson(spec),
+        endpoint,
+    })
+}
+
+/// Fetch the `OpenAPI` spec for `gear` once per sync pass, returning a cheap
+/// (ref-counted) clone of the cached [`Bytes`] if already fetched this pass.
+async fn fetch_or_get_spec(
```
MikeFalcon77, line 317: The spec cache is per pass, so every 10s poll re-fetches the full OpenAPI document of every gear over gRPC, re-parses it into `serde_json::Value`, and rebuilds the whole matchit router under a write lock — even when nothing changed. At 30 gears x ~200KB that is steady control-plane traffic for no reason.

A spec hash/version in `ListAllInstances` would let the edge skip the fetch and the rebuild when unchanged (same idea as k8s `resourceVersion` or xDS `version_info`).

libs/toolkit-gateway/src/forward.rs
```diff
@@ -0,0 +1,410 @@
...
+    match err {
+        HttpError::Timeout(_) | HttpError::DeadlineExceeded(_) => StatusCode::GATEWAY_TIMEOUT,
+        HttpError::Overloaded | HttpError::ServiceClosed => StatusCode::SERVICE_UNAVAILABLE,
+        _ => StatusCode::BAD_GATEWAY,
+    }
+}
+
+/// Hop-by-hop request headers (RFC 9110 §7.6.1) plus `host`/`content-length`,
+/// which the proxy client sets itself.
+fn strip_from_request(name: &HeaderName) -> bool {
```
MikeFalcon77, line 188: `x-toolkit-internal-token` is not stripped, so an external client can send it and the proxy forwards it to the gear — where `internal_auth_middleware` sits on the same routes that are being proxied. It needs the shared secret to be exploitable, but this removes the network boundary the platform plane relies on (ADR-0008 wants it on a separate listener). Platform-plane headers should be dropped unconditionally at the edge.

Also missing: `X-Forwarded-For` / `X-Forwarded-Proto` are neither stripped from the client nor set by the proxy.

libs/toolkit-gateway/src/forward.rs
```diff
@@ -0,0 +1,410 @@
...
+use axum::extract::{Request, State};
+use axum::response::{IntoResponse, Response};
+use http::{HeaderMap, HeaderName, Method, StatusCode};
+use toolkit_canonical_errors::Problem;
+use toolkit_http::{HttpClient, HttpError, HttpResponse, RequestBuilder};
+
+use crate::registry::{ProxyRegistry, RouteMatch};
+
+/// Maximum request body buffered before forwarding (10 MiB).
+const MAX_PROXY_BODY_BYTES: usize = 10 * 1024 * 1024;
```
MikeFalcon77: The request body is fully buffered before forwarding, with a hard 10 MiB cap. file-storage does multipart uploads well past that, and they will get a 413 from the gateway rather than from the gear that owns the limit. A reverse proxy should stream the request body and let the upstream enforce its own limit.

gears/system/api-gateway/src/gear.rs
```diff
@@ -682,6 +729,121 @@ impl ApiGateway {
...
+    /// Mount the reverse-proxy [`Forwarder`](toolkit_gateway::Forwarder) as the
+    /// router fallback, so any request not matched by a gateway-owned route is
+    /// proxied to the owning out-of-process gear pod.
+    ///
+    /// # Errors
+    /// Returns an error if the outbound HTTP client cannot be constructed.
+    fn mount_proxy_fallback(&self, router: Router) -> Result<Router> {
+        let forwarder = toolkit_gateway::Forwarder::new(
+            Arc::clone(&self.proxy_registry),
+            toolkit_http::HttpClient::new()?,
```
MikeFalcon77: `HttpClient::new()` uses the general-purpose defaults, which are wrong for a data plane:

- `request_timeout: 30s` — SSE endpoints (`sse_json`, chat-engine) get cut at 30s
- `max_body_size: 10MB` — also caps the *response*, so large downloads break
- `retry: 3 attempts` incl. 500/502/503 for idempotent methods — the edge amplifies load during an upstream incident
- `max_concurrent_requests: 100` — a single semaphore for the entire gateway; request 101 gets a 503 `Overloaded`

This needs a dedicated proxy profile: no retries, no client-side rate limit, no body cap, no blanket request timeout.

MikeFalcon77 (commented):
.

libs/toolkit-gateway/src/registry.rs
```diff
@@ -0,0 +1,679 @@
...
+        // Deterministic insertion order: sort by gear name then path.
+        let mut path_entries: Vec<_> = path_map.into_iter().collect();
+        path_entries.sort_by(|((a_gear, a_path), _), ((b_gear, b_path), _)| {
+            a_gear
+                .as_str()
+                .cmp(b_gear.as_str())
+                .then(a_path.cmp(b_path))
+        });
+
+        // First-gear-winner for duplicate paths across gears.
```
MikeFalcon77, line 190: Nothing stops a gear from claiming another gear's path. If a gear publishes `/account-management/v1/...` as public, whoever sorts first alphabetically wins and the loser only gets a `warn!`. Combined with the auth flag coming from the gear's own OpenAPI `security`, a gear can claim a path and declare it anonymous.

Two options: validate that exposed paths start with the owning gear's prefix, or keep an explicit owner registry. For the conflict rule itself, Gateway API precedence (most specific match wins) is a better default than alphabetical order.

libs/toolkit-gateway/src/registry.rs
```diff
@@ -0,0 +1,679 @@
...
+    /// matches it.
+    ///
+    /// **Profile 3 selection:** a matched path resolves to a single, stable
+    /// endpoint (the first instance in deterministic order); cross-replica load
+    /// balancing is delegated to the gear's stable k8s Service DNS. The
+    /// instance-keyed registry retains every instance, so Profile 2 (Host +
+    /// Workers) multi-worker selection is a future change to `rebuild` +
+    /// `match_path`, not a storage change.
+    #[must_use]
+    pub fn match_path(&self, path: &str) -> Option<RouteMatch> {
```
MikeFalcon77, line 326: Fine for Profile 3, where k8s Service DNS balances. In Profile 2 (Host + Workers) the proxy *is* the only load balancer, so all traffic goes to whichever worker sorts first and the rest sit idle. The comments call this a seam, which is fair, but Profile 2 multi-worker is effectively not working yet — worth saying so in the PR description.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -251,10 +251,17 @@ pub struct OperationSpec {
...
+    /// Independent of [`is_public`](Self::is_public); maps 1:1 to the
+    /// `PublicRoute` marker in the `OoP` per-gear middleware (`!authenticated`).
     pub authenticated: bool,
-    /// Explicitly mark route as public (no auth required)
+    /// Visibility axis: whether this route is registered in the gateway for
+    /// external access (`true`) or is internal-only, reachable only via
+    /// inter-gear communication (`false`). Defaults to `false` (internal).
+    /// Independent of [`authenticated`](Self::authenticated) — an exposed route
+    /// may still require a JWT.
     pub is_public: bool,
```
MikeFalcon77: The two axes are the right call, but there are now three vocabularies for them: the field is `is_public`, the builder is `.exposed()`, the extension value is `visibility: public` — while `public` *also* still means "no auth" in the OoP `PublicRoute` marker. Renaming this field to `is_exposed` (and `PublicRoute` to `AnonymousRoute`) would remove the ambiguity that the doc comments currently have to explain away.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -1000,14 +1023,13 @@ where
...
-    ///     .public()
+    ///     .anonymous()
     ///     .handler(health_check)
     ///     .json_response(StatusCode::OK, "OK")
     ///     .register(router, &registry);
     /// # let _ = router;
     /// ```
-    pub fn public(mut self) -> OperationBuilder<H, R, S, AuthSet, LicenseSet> {
-        self.spec.is_public = true;
+    pub fn anonymous(mut self) -> OperationBuilder<H, R, S, AuthSet, LicenseSet> {
```
MikeFalcon77, line 1032: The semantic shift here is easy to miss during migration: old `.public()` meant anonymous **and** edge-visible, new `.anonymous()` means anonymous **and** internal. A mechanical `.public()` -> `.anonymous()` replacement silently removes the endpoint from the edge, and it only shows up at runtime.

For out-of-tree gears, keeping `.public()` as `#[deprecated]` for one release (mapping to `.anonymous().exposed()`) would make this a compile-time warning instead of a runtime surprise.

libs/toolkit/src/api/openapi_registry.rs
```diff
@@ -177,6 +177,17 @@ impl OpenApiRegistryImpl {
                 ext.insert("x-odata-orderby".to_owned(), value);
             }
 
+            // Visibility axis (`OperationSpec.is_public`): mark routes that are
+            // registered in the gateway for external access. The `GatewayProvider`
+            // reads this vendor extension to select which routes to reverse-proxy.
+            // The key is mirrored as a constant in `cf-gears-toolkit-gateway`.
+            if spec.is_public {
+                ext.insert(
+                    "x-cf-api-visibility".to_owned(),
```
MikeFalcon77: Prefix is inconsistent with the rest of the repo — we already use `x-toolkit-*` (`x-toolkit-spec-scope`, `x-toolkit-problem-bin`) and `x-odata-*`. `x-toolkit-visibility` would fit; `x-cf-*` is a new namespace with one member.

gears/system/api-gateway/src/config.rs
```diff
@@ -69,6 +73,50 @@ pub struct ApiGatewayConfig {
...
+/// gears' public routes are forwarded to the owning gear pod.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, default)]
+pub struct GatewayProxyConfig {
+    /// Enable the directory-driven reverse proxy.
+    pub enabled: bool,
+    /// gRPC endpoint of the `DirectoryService` (e.g. the grpc-hub endpoint,
+    /// `http://gear-orchestrator:50051`). Required when `enabled` is true.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub directory_endpoint: Option<String>,
```
MikeFalcon77, line 99: `enabled: true` with `directory_endpoint: None` produces one `warn!` and a silently dead proxy — the gateway serves 404s for every OoP route and looks healthy. Given the rest of these configs use `deny_unknown_fields`, this should fail at config validation instead.

libs/toolkit-gateway/src/forward.rs
mattgarmon: Updated

gears/system/api-gateway/src/config.rs
mattgarmon: Added validation

libs/toolkit-gateway/src/registry.rs
mattgarmon: Created follow-up issue: https://github.com/constructorfabric/gears-rust/issues/4487

gears/system/api-gateway/src/gear.rs
mattgarmon: Added dedicated profile

libs/toolkit-gateway/src/forward.rs
mattgarmon: - The response side is now uncapped via the proxy profile.
- Raised the in-memory request-body buffer cap from 10 MiB to 100 MiB, documented as an interim memory-safety bound (owning gear enforces the real limit).

True streaming (the correct fix, requiring a toolkit-http streaming body API - blast radius is the whole Full<Bytes> tower stack) is deferred to GitHub issue #4488

libs/toolkit-gateway/src/provider.rs
mattgarmon: Implemented suggestion

gears/system/api-gateway/src/proxy.rs
mattgarmon: Added `openapi_spec_hash`

gears/system/api-gateway/src/proxy.rs
mattgarmon: Fixed

libs/toolkit/src/api/operation_builder.rs
mattgarmon: Good call, updated as suggested

libs/toolkit/src/api/operation_builder.rs
mattgarmon: Kept `.public()` as shorthand for `.anonymous().exposed()`

libs/toolkit/src/api/openapi_registry.rs
mattgarmon: Fixed

MikeFalcon77 (approved):
Merged. 

Two issues that stood out: putting `apply_snapshot` on the trait with a default impl makes the abstraction real instead of decorative, and keeping `.public()` as a deprecated shorthand with its **original** semantics (anonymous + exposed) means a mechanical migration cannot silently drop an endpoint from the edge — better than the shim I suggested. The fingerprint invalidation is also correct: caching only on `complete && !empty_snapshot_valve` is the part that is easy to get wrong, and a partial pass or a fired safety valve properly forces a retry next tick.

Approving on the basis that `gateway_proxy.enabled` defaults to false, so this lands as a dormant capability and Profile 1 is untouched.

**One nit, fine as follow-up if you are merging now:**

1. `docs/arch/toolkit-oop/PRD.md` says auth-skip is "surfaced as the `AnonymousRoute` extension in the per-gear middleware", but the type is still `PublicRoute` (`libs/toolkit-http-middleware/src/auth.rs:73`) — only the api-gateway matcher was renamed. Doc/code drift.

**Before `gateway_proxy.enabled` is turned on anywhere**, these need to land — none of them block the merge, but all of them block enabling the flag:

1. #4487 and #4488. 

Unrelated to this PR: contract-declared endpoints still cannot be exposed at the edge (`rest_contract.rs` hardcodes `.authenticated()`, no visibility axis). That one is mine — handling it in the toolkit-contracts branch.

PR #4407, opened 2026-08-04
docs(serverless-sdk): scope the SDK to the consumer contract

vgprod:
## Summary by CodeRabbit

- **Documentation**
  - Clarified the separation between consumer and plugin SDK contracts.
  - Documented `ServerlessRuntimeClientV1` for invoking functions and workflows.
  - Added concepts for invocation outcomes, summaries, statuses, controls, validation, idempotency, replay, and refusal errors.
  - Defined host, consumer, and plugin responsibilities, including notification-only updates.
  - Added ADRs covering asynchronous APIs, environments, API stability, and workflow handling.
  - Updated planning, feature, and architecture documentation to reflect revised SDK boundaries.


gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
```diff
@@ -22,922 +22,420 @@ Updated: 2026-03-30 by Constructor Tech
...
+|---|---|
+| `InvokeRequest` | What to run and how: `callable_id`, `mode`, `params`, `dry_run`, `idempotency_key`. |
+| `InvocationMode` | `Sync` or `Async`. Adopted from the gear's model unchanged. |
+| `InvocationSummary` | One row of the gear's invocation index — see below. |
+| `InvocationErrorSummary` | Stable failure type and retry category stored in the gear's index. |
+| `InvocationErrorCategory` | Bounded classification a caller uses when deciding whether to replay. |
+| `InvocationOutcome` | What starting or replaying a run returns — see below. |
+| `InvocationStatus` | The nine run states, adopted verbatim from the gear's `gts.cf.core.sless.status.v1~` schema: `Queued`, `Running`, `Suspended`, `Succeeded`, `Failed`, `Canceled`, `Compensating`, `Compensated`, `DeadLettered`. |
+| `ControlAction` | `Cancel`, `Suspend`, `Resume`, `Retry` — the in-place interventions. |
+| `InvocationId` | Identity assigned by the gear. |
```
MikeFalcon77, line 204: Host DESIGN §Invocation Flow step 7 says the *plugin* creates the record and assigns the `invocation_id`, and `InvocationRequest` carries no id field. So this line and PRD §1.4 ("an invocation identifier the `serverless-runtime` gear assigns") contradict the host design. This also breaks idempotency: dedup is host-owned and applied before dispatch (host DESIGN §Host-side mediation), so the host must persist `key -> invocation_id` for an id that does not exist until the plugin returns, and nothing states who guarantees the ids are unique across plugins that all write into `invocation_index`.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
```diff
@@ -81,689 +60,816 @@ SCOPE:
...
+1. The consuming gear obtains the client contract from the platform's client registry.
+2. It starts the callable, supplying inputs and choosing not to wait, marking the request with a
+   key identifying the work.
+3. The runtime accepts the request and returns the new run's identity and initial state.
+4. The consuming gear stores that identity against its own record of the work.
 
 **Postconditions**:
-- `InvocationRecord.result` contains the serialised output.
-- An invocation span with `succeeded` event is emitted.
+- A run exists and is progressing; the caller can retrieve it by identifier.
```
MikeFalcon77, line 727: The host index is populated only from plugin-emitted events and "the host never writes to it directly" (host DESIGN §Invocation index). On the async path the plugin returns the id immediately and emits the first status notification separately, so nothing orders that notification before `invoke` returns. What does `get_invocation` return in that window — `NotFound`, or is the host required to hold an index row before it returns from `invoke`? As written this postcondition is not guaranteed by the defined protocol.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
```diff
@@ -81,689 +60,816 @@ SCOPE:
...
+- **Architecture Allocation**: See [DESIGN.md](./DESIGN.md) §2.2.
 
+#### Reads never reach the execution backend
 
-#### Zero Unsafe Code
+- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-nfr-read-locality`
 
-- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-nfr-no-unsafe`
+Retrieving a run or querying runs **MUST** be answerable from what the runtime has itself
+recorded, without contacting the backend that executed the callable.
```
MikeFalcon77, line 577: This makes the host index authoritative for everything a consumer can read while the plugin stays authoritative for execution state, and it forbids the host from closing the gap on the read path. If one status notification is lost, or the plugin restarts mid-invocation, the index keeps reporting a stale status indefinitely and a caller sees a run stuck in `Running` that finished hours ago. No document defines a reconciliation path or a port operation the host could use to resync, and G-01..G-06 do not cover this.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
```diff
@@ -756,22 +834,34 @@ pub trait ServerlessRuntimeClient: Send + Sync {
...
+/// This is a **notification surface only**. It is not a general-purpose callback
+/// API: the host accepts notifications, updates the host-owned invocation index and
+/// re-emits timeline events to subscribers. Nothing here returns data to the plugin,
+/// and the host never calls back into the plugin through it. See ADR
+/// `cpt-cf-serverless-runtime-adr-thin-host` (Consequences, and the 2026-07-30
+/// amendment recording the SDK split).
+#[async_trait]
+pub trait InvocationIndexPort: Send + Sync {
     /// Notify the host of an invocation status transition. Plugins call this on
     /// every status change so the host invocation index stays current.
```
MikeFalcon77, line 884: This notification has no sequence number, timestamp or event id, and it returns `Result`, so a plugin whose call times out will re-send. Who deduplicates, and what stops a replayed or out-of-order notification from moving the index backwards from `Succeeded` to `Running`? Consumer reads are served from this index, so the caller observes the regression as real state.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
```diff
@@ -756,22 +834,34 @@ pub trait ServerlessRuntimeClient: Send + Sync {
...
-    /// `error_summary` carries the human-readable failure reason for terminal
-    /// failure statuses; it is `None` for non-terminal transitions and successes.
+    /// `error` is present for failure statuses; full error data remains plugin-local.
     async fn publish_invocation_status(
```
MikeFalcon77, line 886: DESIGN §Invocation index states that id, function_id, adapter, status, timestamps and error summary all originate in plugin-emitted events, and §Invocation Flow step 7 says the first notification "lands a row in `invocation_index`". This signature carries none of `function_id`, `adapter` or the timestamps, so the row cannot be created or timestamped from it. Either the port needs those fields, or the host writes part of the row itself — which the stated invariant forbids.

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
```diff
@@ -22,922 +22,420 @@ Updated: 2026-03-30 by Constructor Tech
...
+HTTP status on the gear's REST surface, so in-process and over-HTTP behaviour agree. Gear error
+types below are relative to `gts.cf.core.sless.err.v1~cf.core.sless.err.…`.
+
+| Variant | Gear error type | HTTP | Source |
+|---|---|---|---|
+| `NotFound` | `not_found.v1~` | 404 | host DESIGN §Dry-Run validations |
+| `NotActive` | `not_active.v1~` | 409 | callable is draft, disabled or archived |
+| `InvalidInput` | `validation.v1~` | 422 | inputs fail the callable's schema |
+| `QuotaExceeded` | `quota_exceeded.v1~` | 429 | tenant concurrency exhausted |
+| `SyncSuspension` | `sync_suspension.v1~` | 409 | a synchronous run reached a suspension point |
```
MikeFalcon77: This contradicts `cpt-cf-serverless-runtime-sdk-fr-failure-vs-refusal` and §2.1: the request was accepted, the run started, and it then reached a suspension point, so representing it as `Err` uses a refusal for something the runtime did not refuse. The variant also carries no `InvocationId`, so a caller cannot observe, resume or cancel the run it just created. Host DESIGN §Workflow does not say what happens to that invocation either (does it stay `suspended`, or get cancelled?).

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
```diff
@@ -22,922 +22,420 @@ Updated: 2026-03-30 by Constructor Tech
...
-**Invariants**:
-- `compensate` may be called more than once for the same `original_workflow_invocation_id`;
-  implementations must be idempotent.
-- Returning `Err(_)` transitions the original invocation to `dead_lettered`.
+| Operation | Input | Returns |
+|---|---|---|
+| `invoke` | `InvokeRequest` | `InvocationOutcome` |
+| `get_invocation` | `&InvocationId` | `InvocationSummary` |
+| `list_invocations` | `&ODataQuery` | `Page<InvocationSummary>` |
+| `control_invocation` | `&InvocationId`, `ControlAction` | `()` |
```
MikeFalcon77, line 297: What does `Ok(())` mean here — the action was accepted, or applied? Host DESIGN contradicts itself: §Invocation Control Actions (Generic) says "the host executes these directly — they never reach the plugin", while §Plugin Control Actions (Passthrough), the Invocation Engine component and `RuntimeAdapter::control_invocation` route cancel/suspend/resume/retry into the plugin, which then reports the transition asynchronously through the event port. Under the second reading `Ok(())` cannot mean the run is cancelled, and `usecase-cancel` step 3 ("confirms it reached a cancelled state") has no bound on when that becomes visible.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
```diff
@@ -81,689 +60,816 @@ SCOPE:
...
 
-#### ServerlessSdkError Variants
+- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-fr-refusal-reasons`
 
-- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-fr-error-model`
+When the runtime refuses a request, the system **MUST** report the reason in a form the caller
+can act on directly, without inspecting message text. The reasons **MUST** separately cover: the
+callable is unknown; the callable is not currently accepting work; the inputs are invalid; the
+caller is not permitted; the tenant's quota is exhausted; the requested intervention is not
+valid for the run's state; and the runtime or its backend is unavailable.
```
MikeFalcon77: Two reachable refusals are missing from this list and from gap G-01. Host DESIGN §Workflow requires rejecting a synchronous invocation of an async-only workflow, and `FunctionTraits.supported_invocations` lets a callable refuse a mode in general — no reason here covers "mode not supported". Separately, suspend and resume are workflow-only (host DESIGN §Invocation Control Actions), which is a callable-kind refusal, not the state-based condition `UnsupportedControl` is defined as.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
```diff
@@ -81,689 +60,816 @@ SCOPE:
...
-by adapters running on multi-threaded async runtimes.
+#### Suppress duplicate runs
 
-- **Rationale**: Adapters run on multi-threaded async runtimes; handler instances must be safely shared.
-- **Actors**: `cpt-cf-serverless-runtime-sdk-actor-adapter-dev`
+- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-fr-idempotency`
 
-### 5.2 WorkflowHandler Trait
+The system **MUST** allow a caller to mark a request with a key identifying the work it
+represents, so that repeating the request does not run the callable a second time. When a stored
```
MikeFalcon77, line 350: This does not define the idempotency scope for deduplication. Host DESIGN gives explicit key tuples for response caching but for dedup only a retention window (`deduplication_window_seconds`), so it is unspecified whether a key is unique per tenant, per subject, per callable or globally — a caller cannot tell whether reusing one key for two different callables collides. Note also that the caching tuple includes `function_version`, so the same key re-executes after the callable is republished, which "repeating the request does not run the callable a second time" does not allow for. G-05 covers only the response shape, not the key scope.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
```diff
@@ -81,689 +60,816 @@ SCOPE:
...
-- **Actors**: `cpt-cf-serverless-runtime-sdk-actor-adapter-dev`
+- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-fr-dry-run`
 
-#### FunctionHandler Concurrent-Use Guarantee
+The system **MUST** allow a caller to submit a request for validation only. Nothing **SHALL**
+execute, no run **SHALL** be recorded, and the caller **MUST** be able to tell from the response
+that this was a validation and not a real run.
 
-- [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-sdk-fr-handler-send-sync`
+- **Rationale**: Callers that assemble automation from configuration need to know a request will
```
MikeFalcon77: A passing dry run does not establish that the request would be accepted. Host DESIGN §Dry-Run Behavior performs four checks (callable exists, callable status, params schema, tenant quota) and short-circuits before rate limiting, dedup and plugin dispatch, so the same request can still be refused for `NoPluginAvailable`, `ServiceUnavailable`, rate limit, or an unsupported invocation mode. State what a successful dry run does and does not guarantee, otherwise callers will use it as pre-flight acceptance.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
```diff
@@ -1,21 +1,37 @@
...
 > This file contains the complete Rust type definitions and trait interfaces for the
-> Serverless Runtime domain model. These types are transport-agnostic and live in the
-> `serverless-runtime-sdk` crate. Per ADR `cpt-cf-serverless-runtime-adr-thin-host`, the
-> public surface is split across two traits: `ServerlessRuntimeClient` (host-implemented,
-> consumer- and plugin-callable; carries the public CRUD surface plus a thin
-> plugin-to-host event port) and `RuntimeAdapter` (Runtime Plugin-implemented,
-> host-callable; carries identity, lifecycle hooks, invocation, schedule, and
-> event-trigger methods). Each Runtime Plugin implementation (Temporal, Starlark, cloud
-> FaaS, etc.) provides a `RuntimeAdapter` impl bound to a single GTS adapter type.
+> Serverless Runtime domain model. These types are transport-agnostic.
```
MikeFalcon77, line 11: This sentence previously said these types live in `serverless-runtime-sdk`; with four traits across three crates the file no longer says which crate declares each type. `InvocationStatus`, `InvocationTimestamps`, `InvocationId` and `OwnerRef` are all needed by both SDKs, and `serverless-sdk/docs/DESIGN.md` requires that neither SDK depend on the other. So either each declares its own copy — and nothing states the two status enums must stay identical, while the plugin conformance harness only locks down the plugin-side one — or one crate depends on the other. Which?

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
```diff
@@ -326,6 +345,43 @@ pub struct InvocationRecord {
...
 
+/// One row of the host invocation index — the lightweight, queryable projection the
+/// host persists from plugin-emitted events (ADR `cpt-cf-serverless-runtime-adr-thin-host`).
+/// This is what host-trait reads return: answering them never requires a plugin
+/// round-trip. Fields absent here (`params`, `result`, full observability, the timeline)
+/// live only with the owning plugin.
+#[derive(Clone, Debug)]
+pub struct InvocationSummary {
+    pub invocation_id: InvocationId,
+    pub function_id: FunctionId,
```
MikeFalcon77: ADR-0001 as amended in this PR mandates `callable_id` for the SDK models because the field accepts workflows too, and `InvokeRequest` uses it. `InvocationSummary` is published to consumers as "exactly the gear's index row" (`serverless-sdk/docs/DESIGN.md` §3.1) yet keeps `function_id`, so one published surface carries two names for the same concept — or the SDK renames the field and the "exactly the index row" claim stops holding.

MikeFalcon77 (commented):
.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
```diff
@@ -326,6 +345,43 @@ pub struct InvocationRecord {
...
+
+/// Redacted failure data stored in the consumer-readable host index.
+/// Backend messages and arbitrary details remain plugin-local.
+#[derive(Clone, Debug)]
+pub struct InvocationErrorSummary {
+    pub error_type_id: GtsId,
+    pub category: InvocationErrorCategory,
+}
+
+/// Consumer-facing retry classification.
```
MikeFalcon77: These are the plugin's retry-decision categories (`RuntimeErrorCategory`, consumed by `RetryPolicy`), and the state machine reaches `failed` only when retries are exhausted. What is a consumer meant to conclude from `Failed` plus `Retryable`, given the plugin has already retried to policy? `serverless-sdk/docs/DESIGN.md` describes this as the classification "a caller uses when deciding whether to replay", which is a different decision from the one the category was defined for.

gears/serverless-runtime/serverless-runtime/docs/features/gear-scaffold.md
```diff
@@ -112,7 +111,9 @@ Not applicable — the scaffold itself has no entity lifecycle. Gear lifecycle i
 
 - [ ] `p1` - **ID**: `cpt-cf-serverless-runtime-dod-gear-scaffold-cargo-wired`
 
-The system **MUST** add `gears/serverless-runtime/serverless-runtime/Cargo.toml` as a member of the root Cargo workspace, with the crate name `serverless-runtime` and dependencies limited to ToolKit core, the SDK crate (`serverless-runtime-sdk`), and any other crates already required by the ToolKit `#[toolkit::gear]` macro. **MUST NOT** depend on any plugin crate at compile time (per `cpt-cf-serverless-runtime-adr-thin-host`).
+The system **MUST** add `gears/serverless-runtime/serverless-runtime/Cargo.toml` as a member of the root Cargo workspace, with the crate name `serverless-runtime` and dependencies limited to ToolKit core and any other crates the ToolKit `#[toolkit::gear]` macro already requires. **MUST NOT** depend on any plugin crate at compile time (per `cpt-cf-serverless-runtime-adr-thin-host`).
+
+Both contract crates are deferred: the consumer SDK until its trait is implemented, and the plugin SDK until F-04.
```
MikeFalcon77, line 116: Deferred to which feature? No feature in the host DECOMPOSITION (F-01..F-08) implements `ServerlessRuntimeClientV1` or registers it in `ClientHub`, and the host DECOMPOSITION defers consumer-SDK allocation to `serverless-sdk/docs/DECOMPOSITION.md`, which scopes itself to declaring the contract and excludes runtime implementation behaviour. The published consumer contract therefore has no implementation owner in either decomposition, and neither owns the integration test that SDK PRD §1.3 uses to verify its first goal.

.cf-studio/config/artifacts.toml
```diff
@@ -84,6 +84,10 @@ patterns = ["gears/simple-user-settings/*"]
 reason = "Ignore module 'types-registry' (not SDLC-documented yet: missing required PRD/DESIGN)."
 patterns = ["gears/types-registry/*"]
 
+[[ignore]]
+reason = "Ignore gears/serverless-runtime/serverless-plugin-sdk/BACKUP-pre-split-sdk-docs — a backup of the SDK documents as they stood before the SDK was split by audience, kept only so the prior work (notably four ADRs) is not lost. It is not current design, is not maintained, and must not be validated: it has 6 dangling feature links (its DECOMPOSITION lists features 0001-0007, only 0001 exists) and a README matching no artifact pattern. When the plugin-facing SDK is designed, write fresh documents in that crate's own docs/ directory and register that instead — do not lift this ignore to cover them."
+patterns = ["gears/serverless-runtime/serverless-plugin-sdk/*"]
```
MikeFalcon77: This glob covers the whole crate directory, so a future `serverless-plugin-sdk/docs/` tree is silently excluded from validation too — precisely what the reason text says must not happen ("do not lift this ignore to cover them"). Scope the pattern to `gears/serverless-runtime/serverless-plugin-sdk/BACKUP-pre-split-sdk-docs/*` so real plugin-SDK documents fail validation until they are registered instead of passing unnoticed.

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
vgprod: Glossary fixed: the backend assigns the id, the gear records it. The ordering problem is real. Dedup is step 2, id assignment step 7, so concurrent retries can both start a run. Logged as G-07 with a proposed fix (host mints the id), which would also close G-08 and most of G-09. Not fixed here.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
vgprod: Now as written it returns NotFound but its wrong. Host must hold the index row before invoke returns. That falls out of G-07: if the host creates the id, it can write the row before returning, and the plugin's first notification updates it instead of creating it.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
vgprod: Confirmed. Logged as G-09.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
vgprod: Today: nobody, and nothing. Both should be the host, since it owns the index and the plugin cannot know whether its timed-out call landed. G-09 records the intended rule: a monotonic sequence per invocation so stale updates are dropped, plus rejecting any transition the state machine does not allow from the stored state. It lands with the plugin SDK, since it needs a field on publish_invocation_status.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
vgprod: Right. The host already knows function_id and adapter from dispatch, so those can be host-derived like tenant and owner. Only timestamps need adding to the port. Covered in G-09.

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
vgprod: Fixed. SyncSuspension { invocation_id }, and the principle now reads "no result was delivered" rather than "refused". The run's fate was already answerable: the state machine routes suspension points to suspended, and the host forbids dropping the suspension. So the run stays suspended. Now stated explicitly.

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
vgprod: Fixed. Ok(()) means accepted, not applied. ADR-0005 settles the contradiction, since the plugin trait "includes … control … methods". Line 881 was wrong and is corrected.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
vgprod: Both added. supported_invocations can exclude either mode, so the reason is mode-neutral. UnsupportedControl now covers state or callable kind. Both join G-01.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
vgprod: Scope now stated as the runtime's to define, and G-05 extended. The version point is in the host register but not asserted here, because that tuple is documented for caching only.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
vgprod: Fixed. Now states it rules out NotFound, NotActive, InvalidInput and QuotaExceeded, and nothing else.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
vgprod: Stated. Only InvocationId and InvocationStatus actually span both surfaces. InvocationTimestamps and OwnerRef are consumer and host only. Each SDK declares its own, and the GTS schema is the contract, with conformance now required on both sides.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
vgprod: Renamed on the published field, per ADR-0001's "the Rust SDK models". The registry's function_id stays. The "exactly the index row" claim is reworded.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
vgprod: Renamed to Transient, Permanent, ResourceLimit, Timeout, Canceled. Named for the failure, not an action already spent. The plugin keeps RuntimeErrorCategory for its retry policy.

gears/serverless-runtime/serverless-runtime/docs/features/gear-scaffold.md
vgprod: F-04 now owns implementing and registering the trait, plus the §1.3 integration test. Every operation on it is dispatch or an index read, so nothing earlier could own it.

.cf-studio/config/artifacts.toml
vgprod: Narrowed to BACKUP-pre-split-sdk-docs/*.

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
MikeFalcon77: The PRD glossary is fixed, but the line this comment sits on is not: `serverless-sdk/docs/DESIGN.md:204` still reads `| InvocationId | Identity assigned by the gear. |`.

So this PR now says both things — the backend assigns it (PRD §1.4) and the gear assigns it (SDK DESIGN §3.1 entity table). Same one-line fix as the glossary got.

gears/serverless-runtime/serverless-sdk/docs/PRD.md
MikeFalcon77: G-09 is logged, but its proposed repair collides with the NFR this comment is on.

G-09 asks for "a way to resync a run from the plugin when the index is suspected stale". `nfr-read-locality` says a read **MUST** be answerable without contacting the backend, threshold "Zero backend round-trips for read operations".

If resync is only ever out-of-band (background job, admin action) the two can coexist — but nothing says that. Worth one sentence in G-09 or in the NFR, otherwise resolving G-09 runs into a MUST added by this same PR.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
MikeFalcon77: Answer is fine and G-09 captures it. What is left: nothing at the port itself says so.

`DESIGN_RUST_TYPES.md` `InvocationIndexPort` is unchanged — no note on ordering, none on idempotency, no pointer to G-07/G-08/G-09. Whoever implements the plugin SDK reads that trait, not `NEXT_ADR_SCOPE.md`. One sentence in the doc comment saying the ordering and dedup rules are still open would cover it.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
MikeFalcon77: Agreed on the answer, but the text it contradicts was not touched. `docs/DESIGN.md:776-777` still says:

- "The index is **populated only from plugin-emitted events** ... The host never writes to it directly."
- "Tenant and owner are host-derived from request context; all other columns (id, `function_id`, `adapter`, status, timestamps, error summary) originate in plugin-emitted events."

Those two bullets already contradict each other, and the second still names `function_id` and `adapter` as plugin-originated — the opposite of this reply and of G-09. G-09's resolution column mentions retiring the invariant, but a reader of §Invocation index gets the old rule with no marker on it. Either fix the two bullets or tag them with "see G-07/G-09".

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
MikeFalcon77: Variant and the run's fate are fixed. The new definition of a refusal is wider than you want, though.

`fr-failure-vs-refusal` now reads: a refusal means "either the runtime declined the request, or it started the run and could not complete it synchronously". A sync run that exceeds its execution timeout also started and also could not complete synchronously, so by that wording it is a refusal — but there is no variant for it in the §3.3 table, and by the same principle a run that ran and failed must come back as `Ok` with `status: Failed` (`InvocationErrorCategory::Timeout` exists for exactly that case).

Narrow the second clause to the suspension case instead of the general "could not complete it synchronously".

gears/serverless-runtime/serverless-sdk/docs/DESIGN.md
MikeFalcon77: `Ok(())` = accepted is clear now, and the host DESIGN contradiction is gone.

`usecase-cancel` still overclaims. Step 3 is "re-reads the run until it reports a cancelled state" with no bound, and the postcondition is "The run reaches a cancelled state". With cooperative cancellation an accepted cancel can lose the race and the run finishes in `Succeeded` — that outcome is in neither the postcondition nor the alternative flows (the one there covers a run that had already finished when the request was made, which the runtime refuses).

Postcondition should be closer to "the run reaches a terminal state, cancelled if the cancel won the race", and step 3 needs either a bound or an explicit note that there is none.

gears/serverless-runtime/docs/DESIGN_RUST_TYPES.md
MikeFalcon77: Ownership is answered clearly. The new conformance rule has no owner, though.

`DESIGN_RUST_TYPES.md` now says both crates' mirrors "MUST be verified against that schema" and "Conformance is required on **both** sides" — but that MUST exists only in that prose. There is no matching requirement in `serverless-sdk/docs/PRD.md`, no feature in `serverless-sdk/docs/DECOMPOSITION.md` owns it, and that same file lists "Runtime-plugin conformance testing" under out of scope.

So the consumer-side check is as unenforced as before, now with a MUST written against it. It needs an FR/NFR in the SDK PRD and a feature to carry it.

gears/serverless-runtime/serverless-runtime/docs/features/gear-scaffold.md
MikeFalcon77: F-04 owning it is the right answer. Two loose ends:

1. F-04's **Depends On** still reads "Cross-crate dependency on the plugin trait surface (defined in `gears/serverless-runtime/serverless-plugin-sdk/`)". Implementing and registering `ServerlessRuntimeClientV1` also requires a dependency on `serverless-sdk`, which is not listed.
2. The line this comment is on — `features/gear-scaffold.md:116` — is unchanged: "the consumer SDK until its trait is implemented, and the plugin SDK until F-04". Now that F-04 owns both, it can name F-04 for both instead of leaving the consumer one open-ended.

MikeFalcon77 (changes requested):

PR #4408, opened 2026-08-04
docs(resource-management): add Resource Management PRD

parviz-93:
## Description
Add the Product Requirements Document for the Resource Management (RMS) gear at `gears/resource-management/docs/PRD.md` (v1.0.0, DRAFT — open for contributor review), following the standard `gears/<system>/docs/` layout picked up by the Constructor Studio registry autodetect. The gear name follows the platform's suffix-free convention (cf. `account-management`).

The document went through three review passes before this PR (full per-layer semantic review against the PRD checklist, a closure-verification pass, and a focused inbound/outbound dependency review). All findings are resolved or explicitly dispositioned; open decisions are tracked in §16 with owners and dates.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## Testing
- [ ] Unit tests pass (N/A — docs-only change)
- [ ] Integration tests pass (N/A — docs-only change)
- [x] Manual testing completed (`make cfs-validate`: 0 errors; `cfs validate --artifact`, `validate-toc`, `check-language`: clean; 121 `cpt-cf-resource-management-*` IDs, traceability verified with no dangling references)
- [ ] New tests added for new functionality (N/A)

## Documentation
- [ ] Code is documented with rustdoc comments (N/A)
- [ ] README updated (N/A)
- [ ] API documentation updated (N/A)

## Checklist
- [x] Code follows project style guidelines (PRD template structure incl. §18 Traceability; stable CPT IDs; RFC-2119 usage; kebab-case gear name)
- [x] Self-review completed (three-pass semantic review, findings dispositioned)
- [ ] No linting errors (`cargo clippy`) (N/A — docs-only change)
- [ ] Code is properly formatted (`cargo fmt`) (N/A — docs-only change)
- [ ] Tests pass (`cargo test`) (N/A — docs-only change)

## Related Issues
—

Issue #4319, opened 2026-07-29
OSS - Infrastructure Resource Manager


claudedigon:
**Description**: Infrastructure Resource Manager — unified resource state mgmt, versioning, and resource lifecycle. Promoted to Fabric gear per meeting #4.
**Is Plugin**: NO
**Has Plugins**: NO
**Has Extension Points**: YES (Infrastructure Adapters framework: S3, V, IS, V, AM)

> Source: review-meeting decision #4 ("Infrastructure Resource Manager").
> VZ scope: Resource State Mgmt & Versioning, Unified Resource Mgmt API (Q2); Custom resource lifecycle actions, Resources Explorer (Q3); Diff Engine (Q4); Adapter lifecycle mgmt, Public Adapter Rust SDK (2027 Q1).

Issue #4321, opened 2026-07-29
CORE - Token Issuer plugin


claudedigon:
**Description**: Service-to-service token issuer (S2S signing).
**Is Plugin**: YES
**Has Plugins**: NO
**Has Extension Points**: NO

> Source: milestones doc §2 (vhp-core); VZ Q2 (inf).

fdlockgraf:
closing as it's done as part of the IRM gear

PR #4409, opened 2026-08-04
docs(infrastructure-resource-manager): add Infrastructure Resource Manager PRD

parviz-93:
## Description
Add the Product Requirements Document for the Infrastructure Resource Manager (IRM) gear at `gears/infrastructure-resource-manager/docs/PRD.md` (v1.0.0, DRAFT — open for contributor review). The gear name follows the review-meeting decision recorded in #4319 ("Infrastructure Resource Manager") and the platform's kebab-case convention; the artifact is picked up by the Constructor Studio registry autodetect.

The document went through multiple review passes (full per-layer semantic review against the PRD checklist, closure verification, a focused inbound/outbound dependency review, an adversarial spec review, and a maintainer validation round). All findings are resolved or explicitly dispositioned; open decisions are tracked in §16 with owners and dates.

Refs #4319. Supersedes #4408.

## Type of Change
- [x] Documentation update

## Testing
- [ ] Unit tests pass (N/A — docs-only change)
- [ ] Integration tests pass (N/A — docs-only change)
- [x] Manual testing completed (`make cfs-validate`: 0 errors; `cfs validate --artifact`, `validate-toc`, `check-language`: clean; 122 `cpt-cf-infrastructure-resource-manager-*` IDs, traceability verified with no dangling references)
- [ ] New tests added for new functionality (N/A)

## Documentation
- [ ] Code is documented with rustdoc comments (N/A)
- [ ] README updated (N/A)
- [ ] API documentation updated (N/A)

## Checklist
- [x] Code follows project style guidelines (PRD template structure incl. §18 Traceability; stable CPT IDs; RFC-2119 usage; kebab-case gear name)
- [x] Self-review completed (multi-pass semantic review, findings dispositioned)
- [ ] No linting errors (`cargo clippy`) (N/A — docs-only change)
- [ ] Code is properly formatted (`cargo fmt`) (N/A — docs-only change)
- [ ] Tests pass (`cargo test`) (N/A — docs-only change)

## Related Issues
Refs #4319

### DIFF
--- a/gears/infrastructure-resource-manager/docs/PRD.md
+++ b/gears/infrastructure-resource-manager/docs/PRD.md
@@ -1,0 +1,2219 @@
+---
+refs: []
+---
+
+# PRD — Infrastructure Resource Manager (IRM)
+
+
+<!-- toc -->
+
+- [Document Information](#document-information)
+  - [Change Log](#change-log)
+- [1. Overview](#1-overview)
+  - [1.1 Purpose](#11-purpose)
+  - [1.2 Background / Problem Statement](#12-background--problem-statement)
+  - [1.3 Goals (Business Outcomes)](#13-goals-business-outcomes)
+  - [1.4 Glossary](#14-glossary)
+- [2. Architecture Alignment](#2-architecture-alignment)
+- [3. Actors](#3-actors)
+  - [3.1 Human Actors](#31-human-actors)
+  - [3.2 System Actors](#32-system-actors)
+- [4. Operational Concept & Environment](#4-operational-concept--environment)
+  - [4.1 Module-Specific Environment Constraints](#41-module-specific-environment-constraints)
+- [5. Scope](#5-scope)
+  - [5.1 In Scope](#51-in-scope)
+  - [5.2 Out of Scope](#52-out-of-scope)
+- [6. Functional Requirements](#6-functional-requirements)
+  - [6.1 Type System and Adapters](#61-type-system-and-adapters)
+  - [6.2 Resource Lifecycle](#62-resource-lifecycle)
+  - [6.3 Declarative Deployments and Reconciliation](#63-declarative-deployments-and-reconciliation)
+  - [6.4 Lifecycle Actions](#64-lifecycle-actions)
+  - [6.5 Relationships and Topology](#65-relationships-and-topology)
+  - [6.6 Discovery and Inventory](#66-discovery-and-inventory)
+  - [6.7 Resource Groups and Organization](#67-resource-groups-and-organization)
+  - [6.8 Governance and Security](#68-governance-and-security)
+  - [6.9 API Contract and Platform Hardening](#69-api-contract-and-platform-hardening)
+- [7. Non-Functional Requirements](#7-non-functional-requirements)
+  - [7.1 NFR Inclusions](#71-nfr-inclusions)
+  - [7.2 NFR Exclusions](#72-nfr-exclusions)
+- [8. Five Quality Vectors Analysis](#8-five-quality-vectors-analysis)
+- [9. Public Library Interfaces](#9-public-library-interfaces)
+  - [9.1 Public API Surface](#91-public-api-surface)
+  - [9.2 External Integration Contracts](#92-external-integration-contracts)
+- [10. Use Cases](#10-use-cases)
+- [11. User Interaction and Design](#11-user-interaction-and-design)
+- [12. Acceptance Criteria](#12-acceptance-criteria)
+  - [Governance Cross-Cut](#governance-cross-cut)
+  - [Type System and Adapters](#type-system-and-adapters)
+  - [Declarative Change Management](#declarative-change-management)
+  - [Day-2 and Topology](#day-2-and-topology)
+  - [Discovery](#discovery)
+  - [Placement and Groups](#placement-and-groups)
+  - [Authorization Granularity](#authorization-granularity)
+  - [Destructive Operations](#destructive-operations)
+  - [Type System and Adapter Onboarding](#type-system-and-adapter-onboarding)
+  - [Resource and Deployment Lifecycle](#resource-and-deployment-lifecycle)
+  - [Access Control](#access-control)
+  - [Secret Handling](#secret-handling)
+  - [Adapter Trust Boundary](#adapter-trust-boundary)
+  - [Placement Propagation](#placement-propagation)
+  - [Destructive Operation Disclosure](#destructive-operation-disclosure)
+  - [Discovery and Inventory](#discovery-and-inventory)
+  - [Capabilities and Tags](#capabilities-and-tags)
+  - [Retention and Accounting](#retention-and-accounting)
+  - [History and Migration](#history-and-migration)
+  - [Dependency Unavailability](#dependency-unavailability)
+  - [Non-Functional Requirements (Show-Stoppers)](#non-functional-requirements-show-stoppers)
+- [13. Dependencies](#13-dependencies)
+- [14. Assumptions](#14-assumptions)
+- [15. Risks](#15-risks)
+- [16. Open Questions](#16-open-questions)
+- [17. Reference Materials](#17-reference-materials)
+- [18. Traceability](#18-traceability)
+
+<!-- /toc -->
+
+## Document Information
+
+| **Field** | **Value** |
+|-----------|----------|
+| **Version** | 1.0.0 |
+| **Last review** | 2026-08-05 |
+| **Target release** | To be set — project planning tracks delivery sequencing |
+| **Document status** | DRAFT — open for contributor review |
+| **Lifecycle position** | First artifact for the component — no implementation exists yet. A technical design follows from this PRD. Implementation follows the design. |
+| **Self-containment** | This document states every requirement, term, actor, limit, and acceptance criterion in full. Reviewers need no other document. |
+
+### Change Log
+
+| **Date** | **Version** | **Change** |
+|----------|-------------|------------|
+| 2026-07-31 | 1.0.0 | Initial component PRD, opened for contributor review. It consolidates the platform's earlier IRM requirement material into one self-contained specification. |
+
+## 1. Overview
+
+### 1.1 Purpose
+
+The Infrastructure Resource Manager (IRM) is the central orchestration layer for all infrastructure and application resources on the platform. It provides one consistent management surface and type system for every resource class. It provides a declarative deployment model with safe reconciliation (diff, preview, apply, rollback). It provides day-2 lifecycle actions and automated discovery of existing estates. It provides a virtual resource graph for topology and dependency analysis, and a scope hierarchy for governance and multi-tenancy.
+
+This document is the single requirements source for IRM. It describes what IRM must do, not how to build it. The technical design that follows will settle the how.
+
+### 1.2 Background / Problem Statement
+
+Resource management on the platform was historically fragmented. Multiple interfaces had inconsistent behavior per resource class. Provisioning was manual and error-prone. There was no desired-state tracking (environments diverge silently). Policy enforcement was inconsistent, and there were audit gaps. Every integration required custom work.
+
+IRM addresses this with a single governed management surface. A registered type describes every resource class. IRM classifies each change and makes it previewable before it happens. IRM records every successful change as an immutable revision that can be rolled back. Every operation is tenant-scoped, policy-gated, and audited.
+
+### 1.3 Goals (Business Outcomes)
+
+- Single pane of glass: one API surface and one graph model across virtualization, multi-cloud, and on-prem resources.
+- Zero-surprise changes: every apply is previewable, deterministic, reversible, and duplicate-safe.
+- Governance built in: tenant isolation, role-based access, policy and quota gating, and complete audit on every operation.
+- Ecosystem and revenue: a versioned resource type registry and adapter model let third parties integrate without core changes. Per-resource attribution enables usage-based billing.
+- Less manual work: automated discovery, standardized day-2 actions, and desired-state reconciliation replace manual provisioning.
+
+Each goal is tracked by a metric with a defined data source. Baselines and targets are open (§16).
+
+| **Metric** | **Data Source** |
+|------------|------------------|
+| Single pane: % of resource classes managed through the IRM API | Type registry |
+| Zero-surprise: % of applies preceded by a preview | Audit events |
+| Governance: % of operations with complete audit correlation; audit-gap count | Audit sink |
+| Ecosystem: third-party adapters onboarded without core changes | Adapter registry |
+| Less manual: share of the estate adopted through discovery versus manual re-creation | Discovery jobs |
+
+### 1.4 Glossary
+
+| Term | Definition |
+|------|------------|
+| Resource | A managed infrastructure or application entity (VM, volume, network, container, service) with a type, properties, outputs, and lifecycle state. |
+| Resource Type | Versioned schema defining a class of resources (properties, actions, capabilities), registered in the type registry under a GTS identifier. |
+| GTS | Global Type System — platform-wide identifier and versioning scheme for types and events. |
+| Adapter | Deployable component that manages resources for a specific provider or backend. It registers resource types and executes provisioning, reads, and deletes. |
+| Deployment | Declarative definition of a set of resources with parameters, variables, dependencies, and outputs. It is the unit of apply, history, and rollback. |
+| Anonymous Deployment | Implicit single-resource deployment that IRM auto-creates when a caller manages a resource directly. |
+| Diff Engine | Deterministic reconciler that classifies each resource change as one of five operations (no change, create, update, replace, delete) and determines the execution order that the workflow executor carries out. |
+| Plan | In-memory result of diff classification, bound to its inputs by a canonical fingerprint. A dry-run previews it, and IRM re-validates it at apply. |
+| Revision | Immutable record of a successful apply. It is the authoritative target for history and rollback. |
+| Lineage | Identity thread that survives resource replacement, keeping history and rollback reachable across re-provisioning. |
+| Management Policy | Per-resource protection level: **full**, **no-delete** or **no-touch**. |
+| Capability | Optional feature (backup, monitoring, encryption) that a resource type offers. It is enabled and configured per resource instance. |
+| Virtual Resource Graph | Typed nodes (resources) and edges — ownership (parent-child), dependency, attachment — that support traversal, impact analysis, and visualization. |
+| Discovery | Automated detection and synchronization of existing provider resources into IRM inventory. |
+| Resource Group | Lifecycle container for related resources. Every resource belongs to exactly one group within the scope hierarchy. |
+| Tenant | Isolated customer or organizational boundary. All IRM operations carry tenant context. |
+| Deployment Address | The tuple (tenant, resource group, deployment name) that uniquely identifies a deployment. |
+| Default Group | Per-tenant resource group that IRM uses as the implicit placement when the caller gives none. |
+| Membership Convergence | Asynchronous reconciliation that propagates placement decisions to the resource-group service. |
+| Adapter Package | Declarative package that describes an adapter, its resource types, data-plane operations, delegation scopes, and policy bundles. It is the single-call onboarding input. Requirement names use "manifest" as a synonym for this package. |
+| Data-Plane Operation Catalog | Per-type registry of provider operations that IRM publishes so that capability grants can be issued and discovered. |
+| Capability Token | Short-lived, single-purpose credential minted per outbound adapter call. |
+| Owned Subtree | A parent resource together with the resources it owns. When the parent is deleted, IRM tears the subtree down to completion. |
+| Trusted System Actor | The internal identity that IRM uses for its own maintenance work, clamped to the tenant being served. |
+| SRE | Site Reliability Engineer — the operator persona that runs day-2 actions, rollbacks, and cleanup. |
+| RBAC | Role-Based Access Control — the access model that the platform RBAC engine resolves. |
+| IdP | Identity Provider — the platform component that authenticates callers and supplies subject identity. |
+| AM | Account Management — the platform component that owns tenants and accounts. |
+| Policy Decision Service | Abstract decision-service role that answers admission, policy, quota, and license-entitlement questions fail-closed. A capability, not a component; which engine implements it is a design decision. |
+| IRM | The short name for the Infrastructure Resource Manager gear, used throughout this document. |
+
+## 2. Architecture Alignment
+
+| **Field** | **Value** |
+|-----------|----------|
+| **Applicable Manifest(s)** | Not referenced — this PRD is self-contained. The rows below state the architecture context in its own terms. |
+| **Platform position** | IRM is the resource-management component of the platform's operations layer. It owns resource types and adapters, resource lifecycle, declarative deployments and reconciliation, day-2 actions, resource relationships, discovery, and the scope hierarchy. |
+| **Adjacent components** | Separate platform components own policy decisions, role definitions, identity, durable execution, durable storage, resource-group membership, and event and audit delivery. §3.2 declares each as a system actor. §13 records the criticality of each. |
+| **Deliberate scope decision** | Continuous drift detection and reconciliation loops are **not** an IRM responsibility in this scope. Adapters own continuous reconciliation. IRM provides on-demand refresh and preview (§5.2). §16 records the question of revisiting this. |
+| **Deliberate scope decision (multi-region)** | Multi-region management is out of scope for this release (§7.2), but it is a known platform direction: the technical design **MUST NOT** preclude a later placement dimension (such as a region) in deployment addressing, identifiers, or group semantics. §16 records the question. |
+| **Safety applicability** | Safety (ISO/IEC 25010 §4.2.9) is not applicable: IRM is a control plane for IT resources operated through API and CLI; it does not actuate physical equipment. Destructive-operation risk to managed infrastructure is governed by `cpt-cf-infrastructure-resource-manager-fr-guardrails`, `cpt-cf-infrastructure-resource-manager-fr-cascade-admission`, `cpt-cf-infrastructure-resource-manager-fr-cascade-disclosure`, and `cpt-cf-infrastructure-resource-manager-fr-operation-cancel`. |
+| **Recorded platform conventions** | CloudEvents (event-broker ADR-0003) for the event envelope. RFC 9457 (ToolKit `05_errors_rfc9457.md`) for error responses. The Idempotency-Key header (toolkit-http) for duplicate-safe mutations. CEL (quota-enforcement / serverless-runtime precedents) for declarative expressions. AuthZEN-based authorization resolution (authz-resolver, `docs/arch/authorization/`) for access decisions. |
+| **IRM-level recorded choices** | UUID v7 (RFC 9562) for time-sortable identifiers. Salted per-tenant digests for secret-field change detection without cleartext exposure. A canonical plan fingerprint that binds an apply to the exact inputs it was previewed against. |
+
+## 3. Actors
+
+All human actors are technical professionals who work through the API, CLI, and operator consoles. The interfaces optimize for expert efficiency and scriptability rather than first-use guidance. Business sponsors are stakeholders, not actors: their needs are expressed through the §1.3 goals and metric table, not through a dedicated actor entry here.
+
+### 3.1 Human Actors
+
+#### Platform Engineer
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`
+
+**Role**: Defines and maintains resource types and platform standards. Operates the management surface for automation and CI/CD integration.
+**Needs**: One consistent, versioned interface for every resource class. Deterministic previews before changes reach production.
+
+#### Automation Engineer
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-automation-engineer`
+
+**Role**: Builds integrations and self-service automation on top of IRM. Registers custom resource types for internal or partner use.
+**Needs**: Stable contracts, idempotent operations, and machine-readable previews and results.
+
+#### SRE / Operator
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-sre-operator`
+
+**Role**: Troubleshoots resources, executes day-2 actions, performs rollbacks and orphan cleanup during incident response.
+**Needs**: Reliable history of what changed and when. Safe rollback. Visibility into detached and degraded resources.
+
+#### System Administrator
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-system-administrator`
+
+**Role**: Plans maintenance and capacity using the resource topology. Administers discovery across providers.
+**Needs**: Accurate dependency graph. Discovery health status. Control over discovery scheduling and failure handling.
+
+#### Tenant Administrator
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-tenant-administrator`
+
+**Role**: Manages the tenant's resource estate: hierarchy, resource groups, tags, quotas, and access for tenant users.
+**Needs**: Strict isolation of the tenant's data. Scope-level access control. Clear quota and policy feedback.
+
+#### Adapter Developer
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-adapter-developer`
+
+**Role**: Authors adapters for providers (virtualization platforms, public clouds, custom backends). Registers adapters and their resource types. Onboarding requests enter IRM inbound through the management API.
+**Needs**: A provider-agnostic contract that requires no IRM core changes. A controlled onboarding lifecycle.
+
+### 3.2 System Actors
+
+#### Infrastructure Adapter
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-infrastructure-adapter`
+
+**Role**: Executes provisioning, updates, reads, and deletes against the concrete provider. Supplies discovery inventory and health signals. Semi-trusted: IRM validates responses and bounds their size.
+**Direction**: Outbound — IRM invokes the adapter.
+
+#### Policy Decision Service
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-policy-engine`
+
+**Role**: Evaluates admission, policy, quota, and license-entitlement decisions for IRM operations. Decisions are fail-closed. This is an abstract decision-service role: IRM depends on the capability, not on a concrete component. Which engine implements it — including the runtime that evaluates adapter-registered policy bundles — is a design decision (§16).
+**Direction**: Outbound — IRM requests decisions.
+
+#### AM and IdP
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-identity-provider`
+
+**Role**: Account Management (AM) and the Identity Provider (IdP) authenticate callers and supply subject identity and tenant context.
+**Direction**: Inbound — identity and tenant context arrive with each request.
+
+#### Workflow Executor
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-workflow-executor`
+
+**Role**: Durable execution substrate for long-running operations (apply, actions, discovery). It survives process crashes and resumes work.
+**Direction**: Bidirectional — IRM dispatches work outbound. Status callbacks arrive inbound.
+
+#### Event & Audit Consumers
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-event-consumer`
+
+**Role**: Downstream systems (audit sink, billing, notification, graph projections) that consume IRM domain and audit events.
+**Direction**: Outbound — IRM publishes.
+
+#### Resource Group Service
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-resource-group-service`
+
+**Role**: Owns groups and their membership. IRM validates group references against it, propagates membership asynchronously, and treats it as the authorization truth that the policy decision service reads.
+**Direction**: Outbound — IRM validates references and writes membership.
+
+#### Trusted System Actor
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-system-trusted`
+
+**Role**: The internal identity under which IRM performs its own maintenance work. This work includes creating a tenant's default group, converging membership, repairing drift, and registering its types at start-up. Every elevation is clamped to the tenant being served and is individually attributable.
+**Direction**: Internal — not an external integration.
+
+#### Grant Issuance Service
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-grant-service`
+
+**Role**: Consumes the data-plane operation catalog and resource resolution that IRM publishes, to issue and scope capability grants for direct data-plane access.
+**Direction**: Inbound — the service calls the IRM catalog and resolution APIs.
+
+#### RBAC Engine
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-rbac-engine`
+
+**Role**: Resolves role definitions and scope-based access. The caller's effective authority is compiled from it.
+**Direction**: Outbound — IRM consults it through the platform authorization resolution path.
+
+#### Type Identifier Service
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-type-identifier-service`
+
+**Role**: Allocates and resolves platform-wide type identifiers. IRM registers its schemas and per-type authorization identities with it.
+**Direction**: Outbound — IRM registers and resolves identifiers.
+
+#### Token Issuer
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-token-issuer`
+
+**Role**: Mints the per-call capability tokens that IRM attaches to outbound adapter traffic.
+**Direction**: Outbound — IRM requests a token before each adapter call.
+
+#### Persistence Layer
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-persistence`
+
+**Role**: Durable storage substrate providing atomic reservations, consistency guards, and cursor pagination.
+**Direction**: Outbound — IRM reads and writes through the platform data layer.
+
+## 4. Operational Concept & Environment
+
+### 4.1 Module-Specific Environment Constraints
+
+- IRM is pre-GA: one-time breaking migrations (event namespace normalization to the platform vendor token, migration to the deployment-scoped resource model) MAY be executed without a dual-publish compatibility window.
+- Deployment and release mechanics (build, packaging, rollout) are owned by the platform ToolKit lifecycle, not by this PRD.
+- This PRD imposes no further runtime, OS, or lifecycle constraints. The technical design settles anything beyond the NFRs in §7.
+
+## 5. Scope
+
+### 5.1 In Scope
+
+Priority below is the strongest requirement priority within the row, on the same `p1`–`p4` scale that §6 and §7 use. `p1` must ship — the capability already exists in a reference implementation. `p2` should ship — planned. `p3` is deferred — blocked on a platform dependency that is not yet available. `p4` is unused.
+
+| **#** | **Feature** | **Priority** | **Requirements** | **Notes** |
+|-------|-------------|--------------|------------------|-----------|
+| 1 | Unified resource lifecycle management | p1 | `cpt-cf-infrastructure-resource-manager-fr-resource-crud`, `cpt-cf-infrastructure-resource-manager-fr-lifecycle-states` | One consistent interface for create, read, update, delete across all resource types. Scoped listing with filtering and pagination. |
+| 2 | Resource type registry (GTS) | p1 | `cpt-cf-infrastructure-resource-manager-fr-type-registry`, `cpt-cf-infrastructure-resource-manager-fr-type-evolution` | Register, version, query, and retire resource types with schemas, actions, and capabilities. IRM rejects invalid definitions. |
+| 3 | Adapter onboarding and registry | p1 | `cpt-cf-infrastructure-resource-manager-fr-adapter-onboarding`, `cpt-cf-infrastructure-resource-manager-fr-adapter-health`, `cpt-cf-infrastructure-resource-manager-fr-adapter-retirement` | Adapter registration lifecycle (pending → active), type contribution, and health visibility. Activation requires at least one registered type. |
+| 4 | Resource capabilities | p2 | `cpt-cf-infrastructure-resource-manager-fr-capabilities` | Optional per-resource features (backup, monitoring, encryption): discover, enable, configure, disable. These operations are fully audited. |
+| 5 | Declarative deployments | p1 | `cpt-cf-infrastructure-resource-manager-fr-declarative-definitions`, `cpt-cf-infrastructure-resource-manager-fr-conditional-resources`, `cpt-cf-infrastructure-resource-manager-fr-parameters` | Multi-resource definitions with parameters, variables, dependencies, outputs, dynamic expressions, and conditional inclusion. Validation returns actionable errors. |
+| 6 | Change classification (diff engine) | p1 | `cpt-cf-infrastructure-resource-manager-fr-change-classification` | Five-operation classification per resource (no change / create / update / replace / delete) driven by type metadata (immutable, computed, secret fields). |
+| 7 | Preview (dry-run) | p1 | `cpt-cf-infrastructure-resource-manager-fr-preview` | Human-readable and machine-readable preview of every planned change with zero side effects. Secrets are always redacted. |
+| 8 | Plan binding and concurrency safety | p1 | `cpt-cf-infrastructure-resource-manager-fr-plan-binding` | Apply executes exactly the previewed change, or IRM rejects the apply when definition, state, or type metadata drifted since preview. |
+| 9 | Ordered durable execution | p1 | `cpt-cf-infrastructure-resource-manager-fr-ordered-execution`, `cpt-cf-infrastructure-resource-manager-fr-deployment-status` | Dependency-ordered execution, parallel only where no dependency relates, durable and resumable, with compensation on failure. |
+| 10 | Replacement strategies | p1 | `cpt-cf-infrastructure-resource-manager-fr-replace-strategies` | Delete-before-create (default) and create-before-destroy per type with per-resource override. IRM re-wires dependent resources safely. |
+| 11 | Management policy | p1 | `cpt-cf-infrastructure-resource-manager-fr-guardrails` | One per-resource protection mechanism with three levels (full / no-delete / no-touch). no-delete detaches the provider object instead of destroying it. |
+| 12 | Duplicate-safe writes (idempotency) | p1 | `cpt-cf-infrastructure-resource-manager-fr-idempotent-writes` | Mandatory caller-supplied idempotency keys make resource and deployment mutations safely retryable with verbatim replay. `cpt-cf-infrastructure-resource-manager-fr-idempotent-writes` lists the operations exempt by construction. |
+| 13 | Revisions, history, rollback | p1 | `cpt-cf-infrastructure-resource-manager-fr-revisions-history`, `cpt-cf-infrastructure-resource-manager-fr-rollback` | Immutable revision per successful apply. Unified chronological history. Multi-selector rollback as a fresh reconciliation. Lineage survives replacement. |
+| 14 | Refresh of actual state | p1 | `cpt-cf-infrastructure-resource-manager-fr-refresh` | On-demand re-read of provider state to surface out-of-band changes before the next apply. |
+| 15 | Soft-delete, retention, orphans | p1 | `cpt-cf-infrastructure-resource-manager-fr-soft-delete-retention`, `cpt-cf-infrastructure-resource-manager-fr-delete-uncertainty` | Tombstones with configurable retention and purge. Orphaned provider objects are first-class: visible, capped per tenant, operator-cleanable. |
+| 16 | Secret hygiene | p1 | `cpt-cf-infrastructure-resource-manager-fr-secret-hygiene` | Secret values are never persisted or emitted in cleartext anywhere (state, history, previews, logs, events). Comparison detects a changed secret value without storing or exposing cleartext. |
+| 17 | Lifecycle actions (day-2) | p2 | `cpt-cf-infrastructure-resource-manager-fr-action-framework`, `cpt-cf-infrastructure-resource-manager-fr-action-execution` | Provider-defined actions (start, stop, snapshot, resize, migrate) with state validation, asynchronous tracking, and full audit. |
+| 18 | Virtual resource graph | p1 | `cpt-cf-infrastructure-resource-manager-fr-relationship-model`, `cpt-cf-infrastructure-resource-manager-fr-graph-query`, `cpt-cf-infrastructure-resource-manager-fr-relationship-backfill` | Typed relationships derived from resource data. Traversal and impact queries. Consistency maintenance (cascade, orphan-edge cleanup, migration archiving). |
+| 19 | Topology data surface | p2 | `cpt-cf-infrastructure-resource-manager-fr-visualization` | Machine-readable topology surface for the frontend visualization: scoped queries, path computation, export. |
+| 20 | Discovery and inventory | p2 | `cpt-cf-infrastructure-resource-manager-fr-discovery-jobs`, `cpt-cf-infrastructure-resource-manager-fr-discovery-sync`, `cpt-cf-infrastructure-resource-manager-fr-tenant-assignment`, `cpt-cf-infrastructure-resource-manager-fr-discovery-compliance` | Manual, scheduled, and event-driven discovery. Idempotent bulk sync. Stale handling. Tenant assignment with discovery pool. Circuit breaker. Non-blocking compliance flagging. |
+| 21 | Resource groups | p1 | `cpt-cf-infrastructure-resource-manager-fr-resource-groups` | Scope model is tenant → resource group → resource. Every managed resource is in exactly one group. Group placement and move. IRM rejects deletion of non-empty groups. |
+| 22 | Tags | p2 | `cpt-cf-infrastructure-resource-manager-fr-tags` | Key-value tags on groups and resources with downward inheritance. Filtering and policy targeting. |
+| 23 | Governance cross-cut | p1 | `cpt-cf-infrastructure-resource-manager-fr-tenant-isolation`, `cpt-cf-infrastructure-resource-manager-fr-rbac`, `cpt-cf-infrastructure-resource-manager-fr-policy-gating`, `cpt-cf-infrastructure-resource-manager-fr-audit-events`, `cpt-cf-infrastructure-resource-manager-fr-admission-pipeline`, `cpt-cf-infrastructure-resource-manager-fr-data-classification` | Tenant isolation, role-based access, policy and quota gating, audit events with correlation on every operation. |
+| 24 | Migration to the deployment-scoped model | p2 | `cpt-cf-infrastructure-resource-manager-fr-model-migration` | Per-tenant migration (expand → backfill → verify → contract) with abort and resume. Legacy data stays readable until completion. |
+| 25 | Group placement and deployment addressing | p1 | `cpt-cf-infrastructure-resource-manager-fr-group-addressing`, `cpt-cf-infrastructure-resource-manager-fr-default-group`, `cpt-cf-infrastructure-resource-manager-fr-group-validation`, `cpt-cf-infrastructure-resource-manager-fr-deployment-scoped` | Deployment identity is (tenant, group, name). Submitting a definition creates-or-updates at that address. Default group is implicit when the caller gives none. |
+| 26 | Explicit group move | p1 | `cpt-cf-infrastructure-resource-manager-fr-group-move`, `cpt-cf-infrastructure-resource-manager-fr-group-move-concurrency` | Relocating a deployment between groups is a separate, synchronous, optimistically-concurrent operation. Apply never moves anything. |
+| 27 | Membership convergence and drift repair | p1 | `cpt-cf-infrastructure-resource-manager-fr-membership-convergence`, `cpt-cf-infrastructure-resource-manager-fr-membership-ordering`, `cpt-cf-infrastructure-resource-manager-fr-membership-durability`, `cpt-cf-infrastructure-resource-manager-fr-membership-failure-handling`, `cpt-cf-infrastructure-resource-manager-fr-placement-drift` | IRM commits placement locally and propagates it asynchronously with bounded staleness. A periodic sweep reconciles out-of-band changes in both directions. |
+| 28 | Manifest-based adapter onboarding | p1 | `cpt-cf-infrastructure-resource-manager-fr-manifest-onboarding`, `cpt-cf-infrastructure-resource-manager-fr-manifest-policy`, `cpt-cf-infrastructure-resource-manager-fr-adapter-delegation` | One call ingests an adapter package and atomically registers the adapter, its types, data-plane operations, delegation scopes, and policy bundles, then activates it. |
+| 29 | Data-plane operation catalog | p1 | `cpt-cf-infrastructure-resource-manager-fr-data-plane-catalog`, `cpt-cf-infrastructure-resource-manager-fr-grantable-types` | Per-type catalog of provider operations published for capability-grant issuance, discovery, and per-instance availability. |
+| 30 | Per-resource-type authorization with response masking | p1 | `cpt-cf-infrastructure-resource-manager-fr-per-type-authz`, `cpt-cf-infrastructure-resource-manager-fr-write-admission`, `cpt-cf-infrastructure-resource-manager-fr-authz-list-union`, `cpt-cf-infrastructure-resource-manager-fr-authz-payload-masking`, `cpt-cf-infrastructure-resource-manager-fr-authz-topology-narrowing` | IRM decides access per resource type. Unreadable members stay listed, but IRM withholds and marks their payloads. IRM silently narrows topology neighbors. |
+| 31 | Mid-flight re-authorization | p2 | `cpt-cf-infrastructure-resource-manager-fr-midflight-reauth` | A running deployment re-validates the caller's live authority before each side-effecting stage and cancels when authority was revoked. |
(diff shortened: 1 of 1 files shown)


parviz-93:
@coderabbitai full review

parviz-93:
@coderabbitai full review

parviz-93:
@coderabbitai full review

parviz-93:
@coderabbitai full review

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+| IdP | Identity Provider — the platform component that authenticates callers and supplies subject identity. |
+| AM | Account Management — the platform component that owns tenants and accounts. |
+| Policy Decision Service | Abstract decision-service role that answers admission, policy, quota, and license-entitlement questions fail-closed. A capability, not a component; which engine implements it is a design decision. |
+| IRM | The short name for the Infrastructure Resource Manager gear, used throughout this document. |
+
+## 2. Architecture Alignment
+
+| **Field** | **Value** |
+|-----------|----------|
+| **Applicable Manifest(s)** | Not referenced — this PRD is self-contained. The rows below state the architecture context in its own terms. |
```
MikeFalcon77, line 165: Just a note: **Gear is not registered in the platform architecture map.** `docs/GEARS.md` is the architecture index — every gear has a section there with a category

MikeFalcon77 (commented):
.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+| **Platform position** | IRM is the resource-management component of the platform's operations layer. It owns resource types and adapters, resource lifecycle, declarative deployments and reconciliation, day-2 actions, resource relationships, discovery, and the scope hierarchy. |
```
MikeFalcon77, line 166: **Gear category / directory placement needs a decision.**

The artifact is at `gears/infrastructure-resource-manager/`, which by convention means **Business Logic Gear**. But this row describes a platform control plane, and §3.2 has the Grant Issuance Service calling *into* IRM (inbound, `p1`).

`docs/GEARS.md` dependency rules say no cross-category sideways deps, and core platform services must not depend on a business gear. Comparable components — `resource-group`, `types-registry` — live in `gears/system/`.

Either move to `gears/system/infrastructure-resource-manager`, or state here why it belongs in the business tier.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+### 3.2 System Actors
+
+#### Infrastructure Adapter
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-infrastructure-adapter`
+
+**Role**: Executes provisioning, updates, reads, and deletes against the concrete provider. Supplies discovery inventory and health signals. Semi-trusted: IRM validates responses and bounds their size.
+**Direction**: Outbound — IRM invokes the adapter.
+
+#### Policy Decision Service
```
MikeFalcon77, line 231: **This actor collapses three separate platform gears into one.**

Admission, policy, quota and license entitlement are handled today by three distinct gears: `authz-resolver`, `quota-enforcement`, `license-resolver`. `docs/GEARS.md` also keeps Policy Manager, License Resolver and Quota Enforcer as separate hops in the request path.

Abstracting for portability is fine, but §13 should carry the mapping to the concrete gears. Without it the design can't be started and the `p1` criticality means nothing.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+**Direction**: Inbound — the service calls the IRM catalog and resolution APIs.
+
+#### RBAC Engine
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-rbac-engine`
+
+**Role**: Resolves role definitions and scope-based access. The caller's effective authority is compiled from it.
+**Direction**: Outbound — IRM consults it through the platform authorization resolution path.
+
+#### Type Identifier Service
```
MikeFalcon77, line 287: **types-registry does not do this today.**

`gears/system/types-registry/docs/PRD.md` scopes the gear to read-only GTS schema lookup for LLM Gateway tool definitions: get-by-id, batch-get, id validation. No registration API.

Per-type authz identities are worse: `docs/arch/authorization/PERMISSION_GTS_TYPE.md` builds the permission catalog from a **link-time** `inventory` collector seeded by `types-registry::init()`. Runtime registration is documented only as a fallback *"during init()"*.

IRM needs post-startup, persistent registration of new resource types **and** new authz permissions by third-party adapters. That capability does not exist. It should be a §15 risk with a named owner, not an unremarked `p1` dependency.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+**ID**: `cpt-cf-infrastructure-resource-manager-actor-type-identifier-service`
+
+**Role**: Allocates and resolves platform-wide type identifiers. IRM registers its schemas and per-type authorization identities with it.
+**Direction**: Outbound — IRM registers and resolves identifiers.
+
+#### Token Issuer
```
MikeFalcon77, line 294: **"Token Issuer" has no counterpart in the platform.**

There is no gear that mints per-call capability tokens. The closest existing components are `credstore` (tenant-scoped secret storage) and `oagw` (credential injection on outbound calls).

Say which one is intended. If this is a genuinely new gear, it becomes a blocking `p1` dependency and belongs in §15.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+- IRM is pre-GA: one-time breaking migrations (event namespace normalization to the platform vendor token, migration to the deployment-scoped resource model) MAY be executed without a dual-publish compatibility window.
+- Deployment and release mechanics (build, packaging, rollout) are owned by the platform ToolKit lifecycle, not by this PRD.
+- This PRD imposes no further runtime, OS, or lifecycle constraints. The technical design settles anything beyond the NFRs in §7.
+
+## 5. Scope
+
+### 5.1 In Scope
+
+Priority below is the strongest requirement priority within the row, on the same `p1`–`p4` scale that §6 and §7 use. `p1` must ship — the capability already exists in a reference implementation. `p2` should ship — planned. `p3` is deferred — blocked on a platform dependency that is not yet available. `p4` is unused.
```
MikeFalcon77: **`p1` here does not mean "ships first".**

`p1` is defined as *"the capability already exists in a reference implementation"* — but §18 says no implementation is bound to this document. So `p1` means "exists somewhere else", which is not a delivery signal.

24 of 35 scope rows are `p1`, target release is "to be set". Missing: an MVP cut — the smallest slice that provisions one resource type end to end, and which adapter goes first.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+- Billing, rating, and metering — IRM exposes resource and scope data for attribution only.
+- Capacity pools — allocatable-capacity containers are out of scope. Resource groups are lifecycle and authorization containers only and carry no capacity or allocation semantics; a future pool concept, if introduced, is a separate abstraction, not a kind of group.
+- Lightweight typed object storage — the `simple-resource-registry` gear owns schema-validated CRUD of simple typed objects; IRM owns provider-backed orchestration. Provider projections belong to IRM inventory, not the registry.
+- Continuous drift detection and reconciliation loops — infrastructure adapters own these. IRM provides on-demand refresh and preview.
+- Interface schemas, wire contracts, data models, and error-code taxonomies — this PRD states required behavior and the distinctions that callers must be able to make. The component's technical design and its published interface description settle the concrete schemas and codes.
```
MikeFalcon77, line 368: **Deferring the error taxonomy removes most of the testable contract.**

Roughly 25 requirements in §6 demand a "distinct machine-readable reason" or that two conditions be "distinguishable". Those distinctions *are* the contract; §12 is not verifiable without them.

The PRD doesn't need concrete codes, but it should enumerate the axes it requires callers to tell apart. At minimum: plan drift vs revision conflict vs precondition failure vs occupied address vs stale view vs duplicate-in-flight vs duplicate-with-different-body vs malformed cursor.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+- Single pane of glass: one API surface and one graph model across virtualization, multi-cloud, and on-prem resources.
+- Zero-surprise changes: every apply is previewable, deterministic, reversible, and duplicate-safe.
+- Governance built in: tenant isolation, role-based access, policy and quota gating, and complete audit on every operation.
+- Ecosystem and revenue: a versioned resource type registry and adapter model let third parties integrate without core changes. Per-resource attribution enables usage-based billing.
+- Less manual work: automated discovery, standardized day-2 actions, and desired-state reconciliation replace manual provisioning.
+
+Each goal is tracked by a metric with a defined data source. Baselines and targets are open (§16).
+
+| **Metric** | **Data Source** |
```
MikeFalcon77, line 118: **Nothing in §6/§7 requires these metrics to be emitted.**

Data sources are named (type registry, audit events, adapter registry, discovery jobs) but there is no requirement that the system publishes them.

Same problem in §15: several mitigations rest on *"the parked count is an alertable metric"*, audit-gap count, adapter health, unrepaired drift. §5.2 correctly excludes dashboards — but metrics are not dashboards. Add explicit "MUST expose" requirements for the signals the risks depend on.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+## 6. Functional Requirements
+
+> **Testing strategy**: Automated tests (unit, integration, e2e) cover all requirements. The target is 90%+ code coverage, unless specified otherwise. Record the verification method only for non-test approaches (analysis, inspection, demonstration).
+
+### 6.1 Type System and Adapters
+
+#### Resource Type Registry
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-type-registry`
```
MikeFalcon77, line 385: **Runtime registration of third-party types is the hard part and it is assumed away.**

The platform's GTS integration (`libs/toolkit-gts`) is a link-time inventory populated by proc-macros and read once by `types-registry::init()`. Registering a new type — and its `AuthzPermissionV1` instance — after the process is running, persistently, from an externally supplied package, is not something the current model supports.

This requirement plus `fr-manifest-onboarding` (which publishes per-type authz identity before the adapter serves traffic) are the two that depend on it. Worth an explicit note that this is new platform capability, not reuse.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+**Rationale**: Type changes must never silently break running estates.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`
+
+### 6.2 Resource Lifecycle
+
+#### Unified Resource Management
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-resource-crud`
```
MikeFalcon77, line 467: **Listing reinvents OData instead of using it.**

"caller-selected ordering over a published, bounded field set with opaque cursor pagination" plus type/status/group/tag filters is what `libs/toolkit-odata` already provides, and it's the standard listing surface for gears in this repo.

Either reference OData as the mechanism or state why IRM needs a different one.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+The system **MUST** produce a preview of every planned change, in both human-readable and machine-readable form. The preview **MUST** persist nothing and touch no provider. The system **MUST** redact secret values in every preview form. Preview **MUST** work before the first apply with no caller-supplied parameters. In that case, preview **MUST** resolve declared defaults. The preview output **MUST** be deterministic and carry totals per operation class. The system **MUST** present values that only exist after provisioning (cross-resource references) as unresolved, not guessed.
+
+**Rationale**: Zero-surprise change management: reviewers approve exactly what will happen.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`, `cpt-cf-infrastructure-resource-manager-actor-automation-engineer`
+
+#### Plan Binding (Zero-Surprise Apply)
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-plan-binding`
```
MikeFalcon77, line 589: **Plan binding rejects on state drift, but the age of recorded state is never bounded. This is the biggest hole in the model.**

Chain of decisions:
- §2 assigns continuous drift detection to adapters, not IRM.
- §9.2 makes the adapter drift-report channel **optional**.
- `fr-refresh` is on-demand only.
- `fr-preview` must touch no provider.

So the "current state" a plan binds to can be arbitrarily stale, and nothing detects that. The apply is then rejected-or-accepted against a reality nobody has looked at. "Zero-surprise" does not hold.

Needs one of: preview implies a refresh, or a stated maximum age of recorded actual state at which apply is admitted, or an explicit statement that IRM guarantees nothing about actual-state freshness (and then the drift-rejection wording should be softened).

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+An apply **MUST** execute exactly the change that was previewed, or refuse when the definition, current state, or type metadata drifted since preview. To do this, the system **MUST** bind the plan to its inputs (definition, current state, type metadata, tenant, options). As a result, definitions that differ only in ordering, or that explicitly state a declared default, bind to the same plan. When any input drifted since preview, the system **MUST** reject the apply with a distinct, actionable reason. Concurrent submissions against the same deployment **MUST** admit against its current revision under a consistency guard. The system **MUST** refuse a submission that lost the race as a conflict.
+
+**Rationale**: If the executed change can differ from the reviewed one, approval is meaningless.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`
+
+#### Ordered, Durable Execution
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-ordered-execution`
```
MikeFalcon77, line 599: **"The system MUST NOT revert a resource that was updated rather than created" contradicts the reversibility goal.**

After a failed apply the estate matches neither the old definition nor the new one. §1.3 promises *"every apply is previewable, deterministic, reversible"*.

This is a defensible choice — Terraform behaves the same way — but say it plainly: a partially applied change is recovered by an explicit rollback, not automatically reverted. Otherwise §1.3 and §8 (Reliability) overclaim.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Where a change requires re-provisioning, the system **MUST** support both delete-before-create (default) and create-before-destroy strategies. The strategy is selected per type, with a per-resource override. The system **MUST** re-wire dependent resources to the replacement safely. Create-before-destroy **MUST** bring the replacement into service and re-point dependents before the system tears down the replaced instance.
+
+**Rationale**: Different resource classes have different availability and uniqueness constraints during replacement.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`
+
+#### Guardrails and Management Policy
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-guardrails`
```
MikeFalcon77, line 629: **Interaction between `no-delete` detach and the orphan cap is unspecified.**

`no-delete` turns a delete into a detach, which produces an orphan. `fr-soft-delete-retention` caps orphans per tenant. What happens when an apply removes several `no-delete` resources from the definition and the cap is reached mid-apply?

Options — fail the whole apply at admission, refuse the individual change, or exceed the cap — all have different blast radius. None is stated. Admission-time evaluation would be consistent with how `fr-cascade-admission` handles blast radius.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+The system **MUST** support rollback to any retained revision selected by identifier, timestamp, or relative position. The selectors include a previous-meaningful selector that skips no-change revisions. Rollback **MUST** execute as a fresh reconciliation against current actual state, never a replay of a stored plan. Rollback **MUST** use the revision's frozen type metadata and policies. The system **MUST** reject targets outside the resource's lineage. Type-evolution compatibility **MUST** be a graded verdict. Identical and additively-compatible targets proceed, the latter with a warning. The system **MUST** refuse an incompatible target and name the offending types. A rollback that revives a deleted resource **MUST** re-derive its relationships and ownership ancestry. This re-derivation restores topology, not only the record.
+
+**Rationale**: Reversibility is a first-class promise. Rollback must be as safe and previewable as forward change.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-sre-operator`
+
+#### Actual-State Refresh
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-refresh`
```
MikeFalcon77, line 724: **On-demand-only refresh leaves drift ownership with nobody for the first release.**

§2 assigns continuous reconciliation to adapters. §9.2 makes the adapter's drift-report channel optional and says drift SLAs are per-adapter and out of scope. This requirement is manual.

Net effect: in a deployment with adapters that don't implement the optional channel, out-of-band drift is invisible until someone runs a refresh by hand. That's a product-level gap worth stating in §15, not just a scope boundary.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Deletion **MUST** tombstone resources with configurable retention before permanent removal. Every tombstone **MUST** record why it was created. At minimum, the record **MUST** distinguish removal from the definition from removal through a cascade. For a cascade, the record **MUST** name the originating parent. Provider objects detached by policy (orphans) **MUST** remain queryable with their provider identity preserved. Orphans **MUST** count against a per-tenant orphan capacity. Orphans **MUST** be cleanable only through an explicit operator action with confirmation.
+
+**Rationale**: This gives recoverability after deletion and controlled handling of intentionally preserved provider objects.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-sre-operator`, `cpt-cf-infrastructure-resource-manager-actor-tenant-administrator`
+
+#### Secret Hygiene
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-secret-hygiene`
```
MikeFalcon77, line 744: **A per-tenant salt does not deliver "no offline recovery of the value".**

The requirement asks for two properties. A salted digest gives the first (no cross-tenant correlation of equal values). It does not give the second: for short or enumerable secrets, an attacker holding the state store and the tenant salt brute-forces the value cheaply.

Either name a KDF with a work factor here, or narrow the requirement to cross-tenant correlation and record in §15 that Phase-1 digests are not a defence against state-store compromise (Phase-2 envelope encryption is the actual mitigation, and it's deferred).

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+The system **MUST NOT** persist or emit secret values in cleartext in any artifact. Artifacts include state, revisions, previews, history, logs, metrics, events, and error messages. The system **MUST** detect a change to a secret field without storing, exposing, or reconstructing its cleartext. The comparison artifacts the system derives from a secret value **MUST NOT** enable cross-tenant correlation of equal values or offline recovery of the value. When a field becomes secret through type re-registration, the system **MUST** re-protect existing persisted values before further changes on affected types proceed.
+
+**Rationale**: A single cleartext leak in any persisted artifact defeats all other secret handling.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-platform-engineer`
+
+#### Model Migration with Abort Path
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-model-migration`
```
MikeFalcon77: **Is there any data to migrate?**

§18 says no implementation is bound to this document. §4.1 says IRM is pre-GA and one-shot breaking migrations are acceptable.

If nothing is deployed on the legacy model, this requirement plus scope row #24, the `usecase-migrate-tenant` use case and a §15 risk row are all carried over from the reference implementation for nothing. Either name the system whose data is being migrated, or drop the workstream.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+The system **MUST** let providers define day-2 actions per resource type, with parameters, allowed source states, and an execution contract. The system **MUST** make defined actions discoverable per type.
+
+**Rationale**: Standardized day-2 operations across heterogeneous providers are a core platform differentiator.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-adapter-developer`
+
+#### Validated Asynchronous Action Execution
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-action-execution`
```
MikeFalcon77, line 766: **Action vs concurrent apply is not specified.**

`fr-refresh` explicitly forbids running concurrently with an apply on the same scope. Nothing says what happens when an action is invoked on a resource that a running apply is currently changing.

Two more pairs are also open: cancel vs an in-progress cascade teardown, and group move vs a running apply on the same deployment. Given the 2-hour operation lifetime in `nfr-limits`, these overlaps are not rare.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Discovery synchronization **MUST** idempotently create and update IRM inventory from provider output at the volume stated in `cpt-cf-infrastructure-resource-manager-nfr-discovery-throughput`. Synchronization **MUST** support full and incremental modes where the provider allows them. Synchronization **MUST** handle resources missing from the source per a configurable policy (default: flag only). Synchronized inventory **MUST** carry the reported configuration and operational state of each resource, not only its existence. Discovered resources **MUST** take their place in the resource graph like any managed resource.
+
+**Rationale**: This gives inventory accuracy without destructive surprises. A repeated sync run must always be safe.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-infrastructure-adapter`
+
+#### Tenant Assignment and Discovery Pool
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-tenant-assignment`
```
MikeFalcon77, line 830: **Which deployment does an assigned discovered resource land in?**

This says the group and deployment invariants apply from tenant assignment onward. `fr-deployment-scoped` says every managed resource must belong to a deployment.

Not stated: whether assignment creates an anonymous deployment per resource, adopts the resource into an existing one, or something else. Also unclear what the resource's recorded *definition* is, since it was never declared — which matters for the next `fr-change-classification` run against it.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+**Rationale**: An accurate inventory that includes violations is worth more than a blocked sync. The system enforces compliance on change, not on observation.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-tenant-administrator`, `cpt-cf-infrastructure-resource-manager-actor-policy-engine`
+
+### 6.7 Resource Groups and Organization
+
+#### Resource Group Containment
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-resource-groups`
```
MikeFalcon77, line 852: **"Exactly one resource group" conflicts with what the `resource-group` gear actually does.**

`gears/system/resource-group/docs/PRD.md` states the opposite: *"Within its own tenant (or related tenants per hierarchy rules), a resource may be a member of multiple RG Entities simultaneously"* — that's a deliberate feature (same resource visible in several organizational views).

Two more mismatches:
- RG is a multi-level forest (University -> Branch -> Faculty -> ...). This model is flat tenant -> group -> resource.
- RG allows a vendor-specific provider behind the same **read** contracts. IRM depends on membership **writes** and default-group creation, which the extensibility model doesn't guarantee.

IRM can enforce this invariant in its own store, but not in RG. Suggest rewording: placement is IRM-owned and strongly consistent locally; RG membership is a derived projection. And raise the multi-membership question with the RG owners.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Propagation **MUST** distinguish transient failure from permanent failure. The system **MUST** retry transient failures. The system **MUST** park permanent failures for operator attention, and not retry them indefinitely. Parked work **MUST** be observable and **MUST** have a stated operator action that resumes it.
+
+**Rationale**: Silent infinite retry hides a broken placement. A parked placement without an observable count and a documented resume path is stranded.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-sre-operator`, `cpt-cf-infrastructure-resource-manager-actor-system-trusted`
+
+#### Placement Drift Reconciliation
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-placement-drift`
```
MikeFalcon77, line 959: **This deletes rows in another gear's store.**

"remove membership records for resource types that IRM manages that no longer correspond to a managed resource; membership records of other platform components MUST NOT be touched" — that's IRM reaching into RG's data and partitioning it by ownership.

RG has no contract for that. It needs either an RG-side scoped-delete operation with an owner tag, or a request to the RG owners to add one. As written it assumes RG membership rows carry a "who wrote this" attribute.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Resource creation **MUST** pass through an ordered, extensible admission pipeline before anything is persisted. The first rejecting check aborts the creation, with nothing persisted and the remaining checks skipped. An admission extension **MAY** enrich the request (defaults for name, labels, or properties). The enriched values are what the system validates and persists. The type of the resource **MUST** be resolved before any extension runs and **MUST NOT** be changeable by an extension.
+
+**Rationale**: Policy enforcement and platform defaults need one sanctioned interception point. An extension that redirects a request to a different type bypasses every type-scoped decision made upstream.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-policy-engine`
+
+#### Policy and Quota Gating
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-policy-gating`
```
MikeFalcon77, line 1071: **Quota is evaluated, not reserved.**

Evaluation happens before execution, including the peak for create-before-destroy. But an apply can run up to 2 hours (`nfr-limits`), and nothing stops a second apply from consuming the headroom in between. Both pass the gate, both exceed quota.

This needs a reservation held for the life of the operation, released on terminal state — not a point-in-time check.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+**Rationale**: The tenant owns the data of its resources; the platform owns the audit trail. Audit records carry the minimum identity attributes needed for attribution — subject, tenant, and operation context — and nothing more. Data-protection obligations for caller-supplied content rest with the deployment operator: the regime-layering answer recorded in §16 (2026-08-03) lets an operator apply regime-specific obligations on top of the primitives this requirement states, without IRM encoding any regime itself.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-tenant-administrator`, `cpt-cf-infrastructure-resource-manager-actor-sre-operator`
+
+### 6.9 API Contract and Platform Hardening
+
+#### Per-Call Adapter Credentials
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-adapter-credential`
```
MikeFalcon77, line 1103: **Map this to `credstore` / `oagw` or declare it as new.**

Per-call credentials with a short lifetime and no reuse across long-running work is the right requirement. But the platform already has `credstore` for secret material and `oagw` for injecting credentials into outbound calls (`docs/GEARS.md`: *"credential injection via Credentials Store"*).

See the comment on `fr-adapter-egress` — the whole outbound-call block should be expressed as requirements *on* OAGW rather than a second implementation inside IRM.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Every outbound call to an adapter **MUST** carry a capability token. This token is a credential usable only for that adapter and that operation. The credential expires well within the duration of the work that it authorizes. Long-running work **MUST** obtain a fresh credential rather than extend or reuse an expiring one.
+
+**Rationale**: A credential can outlive its call, or can be replayable against a different adapter or operation. Such a credential turns one leaked value into general access to the provider estate.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-infrastructure-adapter`, `cpt-cf-infrastructure-resource-manager-actor-token-issuer`
+
+#### Outbound Traffic Confinement
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-adapter-egress`
```
MikeFalcon77, line 1113: **This duplicates the Outbound API Gateway.**

`gears/system/oagw` is the platform's centralized gateway for external calls. Per `docs/GEARS.md` it owns credential injection, retry with backoff, circuit breaker, per-target rate limiting, timeouts and audit. `docs/GEARS.md` also states the rule: *"Only integration/adapters talk to external components."*

Building a second egress path inside IRM means SSRF defence has to be correct in two places, and only one of them gets the platform's security review.

Recommend: adapter traffic goes through OAGW; this requirement becomes "IRM MUST route all adapter calls through the platform outbound gateway" plus the list of guarantees IRM needs from it (per-call destination revalidation, no redirect following, fail-closed on unresolvable destination). Same applies to `fr-adapter-async-protocol` and the discovery circuit breaker.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Adapter responses **MUST** be treated as untrusted input. Responses **MUST** be size-bounded before parsing. A malformed response **MUST** be rejected. For a creation, a response without the identity of the newly provisioned resource **MUST** be rejected. Responses **MUST NOT** be able to impersonate internal protocol markers. Responses **MUST** be validated against the declared output shape of the type. Provider error text surfaced to users **MUST** be truncated. Ambiguous provider state **MUST** be treated as not-yet-ready rather than ready.
+
+**Rationale**: A hostile or broken adapter must not be able to corrupt platform state, exhaust memory, or trick the engine into a different protocol path.
+
+**Actors**: `cpt-cf-infrastructure-resource-manager-actor-infrastructure-adapter`
+
+#### Asynchronous Adapter Protocol
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-fr-adapter-async-protocol`
```
MikeFalcon77, line 1133: Backoff polling, terminal-vs-transient error classification and per-target retry are also OAGW responsibilities (see the `fr-adapter-egress` comment). What is genuinely IRM's here is the *protocol semantics*: accepted-without-location is fatal, the poll location must belong to the same adapter, retries carry the same duplicate-safety key, and cancel attempts provider-side cancellation. Worth separating those two layers explicitly.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+Change preview **MUST** complete within 2 s at p95 for definitions of up to 100 resources. Change preview **MUST** complete within 10 s at p95 for definitions of up to 1000 resources. Single-resource preview **MUST** complete within 200 ms at p95. These resource counts are measurement bands, not limits. The enforced bound on definition size is the request-body limit in Declared Limits.
+
+**Threshold**: p95 per definition size band.
+
+**Rationale**: Preview sits in every review loop. Slow previews push users to skip them.
+
+#### Availability and Durability
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-nfr-availability`
```
MikeFalcon77, line 1225: **Nothing covers reconciliation after a restore.**

RPO of 1 hour means the control plane can come back an hour behind, while providers moved on. For a control plane that is a correctness problem, not just an operational one: recorded state, revisions and idempotency records are all stale, and the next apply binds a plan to that stale state (see the `fr-plan-binding` comment).

Add a requirement: after a restore, affected scopes are marked as requiring refresh before an apply is admitted — or state explicitly that operators must run refresh manually.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+A retried mutation that carries a previously used key **MUST NOT** produce a second set of side effects. This also applies when the retry arrives during the original execution, or after a crash in the middle of the original execution.
+
+**Threshold**: Zero duplicate side effects across the retry and crash-recovery test matrix, including concurrent duplicate submission.
+
+**Rationale**: A single double-provision breaks billing and trust.
+
+#### Placement Convergence
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-nfr-placement-convergence`
```
MikeFalcon77, line 1285: **These numbers conflict with the `resource-group` gear's own targets.**

`gears/system/resource-group/docs/PRD.md` sets `<= 250 ms p95` for hierarchy queries in the default profile (`max_depth = 10`). This requires group reference validation at `<= 50 ms p95` and default-group provisioning at `<= 100 ms p95`.

Either the numbers get renegotiated with the RG owners, or IRM has to cache/denormalize the group view — which is a design decision this NFR would then be forcing. Flag it in §16 either way.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+| Running-operation maximum lifetime | 2 hours |
+
+**Threshold**: As tabulated. Deployment size and definition size are deliberately expressed as a request-body bound, not as an independent resource count. As a result, one enforceable check covers both. The type-identifier bound is derived, not arbitrary. The per-type authorization identity of the identifier (the platform prefix plus the identifier) must stay grantable in the authorization system of the platform. Otherwise an accepted type can never receive a narrow per-type grant.
+
+**Rationale**: A published but unenforced limit is worse than none. Callers plan against it and discover the real bound in production. Every value here is one that the system rejects on.
+
+> **Scope note.** Earlier requirement drafts declared three limits. These limits are deliberately **not** introduced: a maximum resource count per deployment, a maximum dependency depth, and a cap on retained deployment history per tenant. The request-body bound, cycle detection, and retention provide the operative bounds. A change request backed by production data is required to introduce these limits anew.
```
MikeFalcon77, line 1332: **Revision retention is never defined, so history growth is unbounded.**

`fr-rollback` operates on "any **retained** revision", and §16 justifies dropping the history cap by saying *"retention bounds history"*. But the only retention defined anywhere is tombstone retention for deleted resources and 24h for completed operations. Revisions have neither a retention nor a cap.

Every apply writes an immutable revision. At 100k resources per tenant and CI/CD-driven applies, that's the fastest-growing table in the system with no stated bound. Either define revision retention here or reinstate the per-tenant history cap.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+| **Reliability** | Duplicate-safe writes, crash-resumable execution with compensation, idempotent discovery sync, and rollback available for every retained revision. | The control plane must never leave estates in an unrecoverable half-state. |
+| **Performance** | Interactive p95 targets (§7.1) MUST hold at declared scale (100 k+ resources/tenant, 1 M+ topology nodes). | Latency regressions at scale silently kill automation and UX. |
+| **Security** | Mandatory tenant context everywhere. Fail-closed policy gating. Zero cleartext secrets in any persisted or emitted artifact. Complete correlated audit. | One isolation or secret leak is an existential compliance failure. |
+| **Versatility** | The type registry + adapter contract MUST allow new providers and resource classes with zero core changes. Replacement strategies, management policies, and capabilities are configurable per type and per resource. | Ecosystem growth across heterogeneous infrastructure is the strategic bet. |
+
+## 9. Public Library Interfaces
+
+### 9.1 Public API Surface
+
+#### Unified Management API
```
MikeFalcon77, line 1358: **The public surface is under-specified against the toolkit.**

"Type: REST API" only. The gear contract in this repo is REST (via `api-gateway`) + gRPC (via `grpc-hub`) + in-process client registered in `ClientHub`. §9.1 has an "In-Process Service Client" entry but no gRPC.

Also not mentioned anywhere in the PRD, though all are mandatory for gears here:
- `SecurityContext` propagation and SecureORM scoping (a hard rule in `docs/GEARS.md`)
- `OperationBuilder` / `OpenApiRegistry` as the path to the published OpenAPI document
- `toolkit-contract` parallel contract versioning (ADR-0007) — that's the mechanism behind the "Breaking Change Policy" line, and it changes what "major version bump" costs

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+**Stability**: stable
+
+**Description**: The in-process client contract for platform services that consume IRM (grants, tooling) without the network edge.
+
+**Breaking Change Policy**: Versioned contract. New majors ship alongside old ones until consumers migrate.
+
+### 9.2 External Integration Contracts
+
+#### Adapter Contract
```
MikeFalcon77, line 1396: **No compatibility story for the adapter package format itself.**

"Versioned contract" covers the call contract. Two cases are open:
- What happens to already-onboarded adapters when the package schema changes?
- What happens to resources provisioned under an older data-plane operation catalog after the adapter re-submits a package that removes or changes those operations?

`fr-type-evolution` covers type schemas but not the operations catalog or delegation scopes.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+- [ ] `p1` - **ID**: `cpt-cf-infrastructure-resource-manager-contract-adapter`
+
+**Direction**: required from client (adapter implementations)
+
+**Protocol/Format**: HTTP/REST (provider-agnostic, any implementation stack)
+
+**Compatibility**: Versioned contract. IRM validates and size-bounds all adapter responses. Long-running provider operations are trackable to completion. Data-handling obligations for transmitted resource properties and secret material are part of this contract; the contract specification settles them. The contract includes an optional drift-report channel through which an adapter **MAY** surface detected out-of-band divergence to IRM; drift-detection service levels are per-adapter and outside this PRD's scope.
+
+#### Workflow Executor Contract
```
MikeFalcon77, line 1406: **The executor plugin has no available implementation today.**

In `docs/GEARS.md`: Serverless Gateway *"define and execute workflows"* is `p1` and unchecked; Durable Objects *"integration for durable execution"* is `p3` and unchecked; there is no Jobs Manager gear in the tree.

§16 mentions *"today one plugin implementation exists"* — name it here. Durable execution is a `p1` bidirectional dependency for the entire apply/action/discovery path, so this is the largest schedule risk in the document and it currently sits in §14 Assumptions rather than §15 Risks.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+
+- [ ] `p2` - **ID**: `cpt-cf-infrastructure-resource-manager-contract-events`
+
+**Direction**: provided by library
+
+**Protocol/Format**: CloudEvents envelope, versioned event names under the platform vendor namespace
+
+**Compatibility**: Additive schema evolution within a major version. Consumers deduplicate by event identity. A breaking rename of the event namespace is permitted at most once before general availability, and MUST be announced to consumers in advance. After that, a breaking rename requires a major version.
+
+## 10. Use Cases
```
MikeFalcon77, line 1426: **Not a single concrete provider appears anywhere in the document.**

All 11 use cases are generic. Without one worked example — Kubernetes, vSphere, AWS, anything — the adapter contract, the resource-type schema shape and the day-2 action vocabulary are unvalidated.

Suggest adding one end-to-end walkthrough against a real type: package contents, registered schema, a preview output, one day-2 action, one discovery run.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+- **Then** the management surface MUST meet ≥ 99.9 % availability and ≥ 99.999 % durability
+- **And** policy and authorization failures MUST fail closed
+
+**71. Crash-safe execution**
+- **Given** a process failure during a multi-resource apply
+- **When** execution resumes
+- **Then** the system MUST NOT apply any resource twice
+- **And** the operation MUST reach a terminal state visible in history
+
+## 13. Dependencies
```
MikeFalcon77, line 2097: **Add a readiness column.**

Criticality tells us how much IRM needs each dependency. It doesn't tell us whether the dependency exists.

By my reading of the repo, at least four `p1` rows have no usable implementation today: Workflow Executor (see §9.2 comment), Token Issuer (no counterpart gear), Type Identifier Service (types-registry is read-only lookup today), Grant Issuance Service (§16 itself notes it has no gear or design documentation).

A `p1` dependency on something that doesn't exist is a §15 risk with an owner and a date, not a §14 assumption.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+| Event & Audit Consumers | Delivery of domain and audit events to consumers | outbound | p1 |
+| Resource Group Service | Group existence, membership, and default-group semantics. The decision point compiles group-scoped access from the membership it holds | outbound | p1 |
+| Token Issuer | Mints the per-call credentials used for outbound adapter traffic | outbound | p1 |
+| Grant Issuance Service | Consumes the data-plane operation catalog and resource resolution that IRM publishes | inbound | p1 |
+
+## 14. Assumptions
+
+- The Policy Decision Service, RBAC Engine, AM, and IdP are operational and expose stable contracts.
+- Infrastructure adapters are semi-trusted external components. Their responses are validated and bounded.
+- A durable workflow engine is available through the executor plugin contract.
```
MikeFalcon77, line 2120: This is not an assumption — no such engine is wired up today (see the comment on `contract-workflow-executor`). Move it to §15 with an owner.

gears/infrastructure-resource-manager/docs/PRD.md
```diff
@@ -0,0 +1,2219 @@
...
+- Persistence supports atomic reservations, consistency guards, and cursor pagination as required by idempotency and history.
+- Tenants receive a provisioned secret salt at creation for digest-based secret comparison, per the mechanism recorded in the §2 "IRM-level recorded choices" row.
```
MikeFalcon77: **This is an unfunded requirement on `account-management`.**

Provisioning a per-tenant secret salt at tenant creation is a change to AM. It is not in AM's PRD, not in §13 as a dependency, and not in §15 as a risk.

Either raise it with the AM owners and add the dependency row, or drop the salt mechanism in favour of `credstore` references (which is where Phase-2 is heading anyway).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The observability requirements for the section 15 signals already exist: nfr-placement-convergence:1295 with its threshold line :1297, fr-membership-failure-handling:971, and fr-dependency-unavailability:1211. The real gap was the section 1.3 metrics: no requirement made IRM publish them. The section 16 business-metrics question now also settles how each section 1.3 metric is emitted or derived, with its owner and date kept. We did not add per-metric MUST-expose requirements: fr-adapter-health:1171 defines health on demand, and the section 1.3 metrics are product roll-ups over surfaces that section 6 already requires.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Not fixed in this commit: docs/GEARS.md is a separate file. IRM is already in the backend roadmap: issue #4319 (milestone 26.09). We will add the IRM section (category, scenarios, PRD link) and link #4319. One question: can we edit docs/GEARS.md ourselves in this PR, or does that file have an owner and a separate process? The category will follow the placement decision (see the line-165 thread).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 16 now records final placement (gears/ vs gears/system/) as an open question with an owner, and marks the current path as interim. We did not move the directory now: that is the open decision. One correction to the premise: top-level gears/ is not only business logic; chat-engine, credstore, and file-parser also live there. A later move stays cheap: the tree holds only docs, and the IDs are name-based.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The section 13 Policy Decision Service row now carries an informative mapping: admission and policy decisions map to authz-resolver (the "Policy Manager" hop), quota maps to quota-enforcement (specified, not built), and license maps to license-resolver or the gateway license middleware. The actor itself stays abstract: PRD.md:157 and the section 16 engine question route the engine choice to design, with an owner and a date. Two corrections: quota-enforcement and license-resolver are docs-only today with zero code, and Quota Enforcer is not a hop in the GEARS.md:60 request path.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 15 now carries a risk row for the platform type-publication path. One correction first: a registration API exists today (TypesRegistryClient::register and POST /types-registry/v1/entities, routes.rs:37-40, with full validation in the ready phase). The true gap is different: the store is in-memory, each start re-seeds it from the link-time inventory, and PDP use of runtime-registered permissions is not designed. The risk row records that gap. Mitigation: IRM re-publishes adapter types and permissions from its own durable store.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 13 now references the planned Token Issuer gear (#4321, milestone 26.08), section 15 tracks its readiness, and section 16 asks whether the #4321 token formats cover per-call capability tokens. On the mapping question: neither credstore nor oagw mints tokens today. S2S authentication exists under other names (exchange_client_credentials per ADR-0008, platform-plane internal tokens per ADR-0006), but in both an external party mints the token. IRM needs per-call tokens scoped to one adapter and one operation, and that maps to #4321.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The legend now defines p1 as "must ship in the first release", rows 10, 25, and 28 moved to p2, and a section 16 question assigns the MVP first slice to project planning. One correction: 27 of 35 rows were p1, not 24. We did not add a full MVP cut with a named adapter inside section 5.1, because section 5.2 puts adapter selection out of scope. The section 16 adapter-validation question names VHI/OpenStack as the first adapter to validate the contract against.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Section 5.2:368 defers only the concrete codes and the wire schemas; its second sentence keeps the distinctions normative. Each axis on your list is already a MUST: malformed cursor :469, plan drift vs revision conflict :591, duplicate-in-flight vs different-body :649, occupied address vs stale view :904, precondition failure :1181. Section 12 restates them as testable criteria that do not depend on code strings. Question: is any distinction missing from that set? If yes, name it and we add the requirement. If the concern is navigation, we can add a non-normative index of the distinctness requirements.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The section 15 type-publication risk row (shared with the line-286 thread) and a note at the section 13 Type Identifier Service entry now record this, with fr-adapter-onboarding and fr-grantable-types named as the dependents. One boundary stays: fr-type-registry itself governs the registry that IRM owns. The unproven part is the platform path: durable identity for adapter-contributed types after a restart, and PDP use of their permissions.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The section 2 recorded-conventions row now names OData ($filter and $orderby with opaque cursor pagination, via toolkit-odata) next to RFC 9457 and Idempotency-Key. The text of fr-resource-crud stays mechanism-agnostic, per section 5.2.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8, with your third option. fr-plan-binding now defines "current state" as the recorded actual state, detects drift of recorded inputs only, and makes no promise about provider-side freshness. Freshness comes from fr-refresh and the optional section 9.2 drift channel, and the section 15 drift row tracks the residual risk. The first two remedies conflict with recorded decisions: preview-implies-refresh breaks the no-provider MUST in fr-preview:581 and the section 8 efficiency show-stopper (:1355), and a maximum state age is a freshness SLA that :1411 puts out of scope.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. fr-ordered-execution now states: an explicit rollback (fr-rollback) recovers a partly applied change, and compensation removes only the resources that the failed change created. Sections 1.3 and 8 stay unchanged: section 8 already says "unrecoverable half-state" and lists rollback as the mechanism.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Orphan capacity is now evaluated at plan admission over the aggregate detaches, and an apply that takes the tenant past capacity is refused whole, like fr-cascade-admission. The fix is partial against your comment because acceptance criterion 64 (:2077-2080) already excluded cap excess and defined the refusal report; the open part was only the granularity inside a multi-detach apply.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. Section 15 now carries the residual-risk row, like the Phase-2 secret precedent at :370. The row cites the mitigations: operator refresh before change windows, the adapter drift channel where present, and scheduled discovery sync (fr-discovery-sync:842 carries state, not only existence). The scope decision itself stays behind the section 16 change request.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8, with your second option. The requirement is narrowed to the property the mechanism delivers: no cross-tenant correlation of equal values. The section 15 secret-hardening row now records that Phase-1 digests are not a defense against state-store compromise, and envelope encryption stays the Phase-2 mitigation per the recorded phase split (:370).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. The migration workstream is removed: scope row 24, fr-model-migration, usecase-migrate-tenant, acceptance criterion 66, the section 15 migration row, fr-relationship-backfill, and the section 4.1 and event-namespace clauses. No deployed estate exists on the legacy model. IRM starts from scratch with an empty store, and existing provider resources enter through discovery and adoption (fr-discovery-sync, fr-tenant-assignment).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8, with rules for the three pairs. (a) Action admission runs under the same consistency guard as apply admission, and the system refuses the action in a transitional state unless the action lists that state as an allowed source. (b) A committed cascade teardown is not cancellable; only the pre-commit window accepts a cancel. (c) The system refuses a group move while an apply runs on the deployment, like fr-refresh:736.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. fr-tenant-assignment now states: assignment wraps each resource in an anonymous deployment, placement goes to the group chosen at assignment or to the tenant default group, and the recorded desired state is seeded from the last observed configuration. An unchanged estate then classifies as no change on the next fr-change-classification run.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The invariant is now scoped to IRM placement records ("IRM records exactly one placement per managed resource"), fr-placement-drift states the reconciler rule, and a section 16 item raises multi-membership tolerance and the write contract with the RG owners. We did not adopt the projection rewording wholesale, because fr-membership-convergence:941 already makes placement IRM-owned, locally transactional, and asynchronously propagated to RG.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: The partition key already exists: the RG membership key contains resource_type as a per-row field (RG PRD composite key), the public API queries and removes by it, and IRM owns the resource-type registry, so "types that IRM manages" is decidable on the IRM side. In e702a7c8, fr-placement-drift also names resource_type as the partition key and requires GTS-qualified values. Question: does that close this point, or do you still see a case that needs an RG-side owner tag or a new RG operation? The one residue we see is that RG's resource_type is a free-form string (their own open question, RG PRD:1157); we support GTS validation of that field on the RG side.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. fr-policy-gating now holds admitted capacity from admission to the terminal state and commits or releases it there, so concurrent admissions cannot jointly exceed quota. The requirement stays behavioral; the rationale cites the platform Quota Enforcement reserve/commit/release lease (GEARS.md:586).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. The section 13 Token Issuer row now references the planned gear (#4321, milestone 26.08). On the mapping: neither component carries this today. credstore stores durable secrets and mints nothing (credstore PRD:429), and the oagw auth plugins inject stored or per-upstream credentials (oagw PRD:279-285), not per-call operation-scoped tokens. If adapter egress routes through the gateway, token injection lands as a new oagw auth plugin fed by the issuer.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. fr-adapter-egress is rewritten: adapter traffic routes through the platform's central egress path (an abstract section 13 role, with OAGW as the current implementation), and IRM requires per-attempt destination revalidation, no redirect following, and fail-closed behavior on unresolvable destinations from that path; acceptance criterion 52 matches. Not done yet: filing the missing guarantees against the oagw PRD; that is a follow-up outside this document. Two corrections: the oagw PRD puts request retries out of scope ("client responsibility", :186, :228), and the discovery breaker is job-scheduling suspension with operator alerts, not a request-path breaker.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: The oagw PRD assigns retries to the client (:186, :228), and the gateway has no polling capability. So backoff polling (:1158), the terminal-versus-transient decision (:1159), the one-hour cap, and the duplicate-safety key stay in IRM; the duplicate-safety key binds to IRM operation idempotency and cannot live in a stateless proxy. In e702a7c8 a cross-reference sentence now separates the layers: transport mechanics belong to the egress path, and operation-level protocol semantics stay in IRM. Question: given those oagw constraints, where does the poll loop live if not in IRM? GEARS.md:788-789 lists retry and breaker as unchecked p2 OAGW scenarios; if OAGW implements them, the transport-level classification moves with them, and we will flag the document discrepancy to the OAGW owners.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8, in two parts. First, your gate: a restore marks affected scopes as refresh-required, and the system refuses apply admission until refresh (new requirement nfr-restore-gate). Second, beyond it: a section 15 row records the residual duplicate-side-effect exposure from idempotency records lost inside the recovery point, bounded by the RPO. The refresh gate alone cannot repair the idempotency half; that is why the risk row exists.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. A section 16 question with an owner now records the gap, and nfr-placement-convergence carries a threshold note that points at it. Precise framing: RG publishes two read SLOs (hierarchy queries p95 <= 250 ms, membership reads p95 < 30 ms) and none for a group-entity point read or for any write. The IRM budgets depend on exactly the unpublished operations. The question lists the options: renegotiated RG targets for these two operations, an IRM-side cache of the group view, or a budget that excludes the remote hop.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Fixed in e702a7c8. Revision retention is now a configurable per-tenant window with a stated default and floor, and it bounds "any retained revision" in fr-rollback. Acceptance criterion 63's "expired revisions" is now well-founded. We chose the window over a count cap because a cap requires a rewrite of criterion 63 and of the section 16 decision record.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8: section 9.1 now states that no gRPC projection ships in this release. On the rest, the facts differ from the premise. GEARS.md:28 lists the interfaces as alternatives ("Rust-native interfaces, HTTP APIs, or gRPC"), and exactly one gear implements GrpcServiceCapability (gear-orchestrator). No toolkit-contract crate exists, and the v1/v2 coexistence strategy is explicitly deferred to a future ADR (toolkit-contract-binding DESIGN.md:734-744). :1373 already mandates a published machine-readable interface description; OperationBuilder and OpenApiRegistry are the mechanism that satisfies it, and SecurityContext and SecureORM are the mechanism behind fr-tenant-isolation, which section 5.2 leaves to design. Question: with the section 9.1 note in place, which concrete surface requirement do you still see missing at PRD level?

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. fr-data-plane-catalog now defines re-submission semantics: if a re-submitted package removes or changes an operation with outstanding grants, the system at minimum marks the operation deprecated and refuses or flags its removal. The section 9.2 adapter-contract entry now carries a package-format compatibility sentence. One correction: delegation scopes were already covered; fr-adapter-delegation:447 narrows grants automatically on re-submission (criterion :1846 restates this).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 15 now tracks the durable-execution substrate, like the event-broker row, with a reference to the section 16 question (owner, 2026-10-31). Section 16 names the implementation: the Temporal-based executor plugin of the vhp-core reference implementation. Section 9.2 stays implementation-agnostic on purpose: the no-op default clause is contract, and the plugin name is not.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8, in two parts. Appendix A (informative, non-normative) walks a VHI/OpenStack virtual machine through package, registered schema, preview, one day-2 action, and one discovery run. A section 16 question binds the technical design to end-to-end validation of the adapter contract against VHI/OpenStack as the named first adapter, with an owner and a date. Concrete schemas stay out of the normative sections per section 5.2:368. One count correction: section 10 has 10 use cases, not 11.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 13 now has a Readiness column, and section 15 rows with owners track Workflow Executor, Token Issuer (#4321, milestone 26.08), and Grant Issuance Service. One correction: types-registry is not read-only; it has a batch registration endpoint today (POST /types-registry/v1/entities, routes.rs:38). Its real gap is the one from the line-286 thread: the store is in-memory and PDP use of runtime-registered permissions is undesigned. The readiness column points at that section 15 row, so all four of your p1 rows carry a tracked risk. Section 16 also now notes that the types-registry PRD understates the implemented surface.

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. Section 15 now carries the risk row. The assumption stays in section 14 because it states the planning basis, and the document's own pattern backs an assumption with a risk row (event-broker: :2159 plus :2166). Section 16 already assigns this topic an owner and a date (2026-10-31).

gears/infrastructure-resource-manager/docs/PRD.md
parviz-93: Partially fixed in e702a7c8. fr-secret-hygiene now states that IRM provisions and stores its own per-tenant key lazily on first use, and the section 14 salt-provisioning assumption is dropped. We did not take either proposed fix. The AM PRD forbids non-identity provisioning and secret material in AM metadata (:124, :130) and defers tenant lifecycle events until the platform event bus lands (:129), so no tenant-created signal exists as a trigger. credstore references reverse the recorded Phase-1/Phase-2 split (:370) and break the inline-secret caller contract that fr-secret-hygiene serves. No AM change, no new section 13 row.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: `docs/GEARS.md` has no CODEOWNERS entry — from `docs/` only `/docs/arch/authorization` is owned (@aviator5). Recent edits to it come from three different people, so there is no gate. You can add the section yourselves.

I'd put it in a separate small PR so this one stays docs-only for a single gear, but that's preference, not process.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: Two of the three pairs are clean.

The cascade rule leaves a contradiction. `fr-cascade-delete` now says a committed teardown **MUST NOT** be cancellable, but `fr-operation-cancel` (:694) is still unconditional: *"let an authorized caller request cancellation of a running operation"*, *"cancellation MUST take effect at a change boundary"*. As written the two requirements disagree. Add the carve-out as a cross-reference in `fr-operation-cancel`.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: The `resource_type` partition is confirmed — RG PRD:138 has the composite key `(group_id, resource_type, resource_id)`. That part of my comment is closed.

The new sentence is the problem:

> When a managed resource holds more group memberships than its recorded placement, the reconciler **MUST** remove the extra memberships and keep the recorded placement.

RG supports multi-membership deliberately — *"a resource may be a member of multiple RG Entities simultaneously, enabling the platform to present the same data set in multiple organizational views"*. The `resource_type` partition protects **other components'** rows. It does not protect a legitimate second membership of a resource that IRM owns. A tenant admin who adds a managed VM to a reporting group loses that membership on the next sweep, silently.

Second issue: §16 asks the RG owners to confirm that multi-membership is tolerated, while this requirement has already decided the answer — and decided it by deleting data.

Suggestion: until §16 is answered, weaken to "the reconciler **MUST** report the divergence", or scope the removal to memberships IRM itself created.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: Hold semantics are right.

One gap: the rationale cites the platform quota-enforcement reserve/commit/release lease as the precedent, but that gear is docs-only — `gears/system/quota-enforcement` contains no `.rs` files at all. §13 Readiness says "Partial", which is honest, but a hard MUST inside a `p1` requirement now rests on an unbuilt primitive.

Token Issuer got its own §15 row for exactly this shape. This deserves one too.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: Gate plus the separate idempotency risk row — good. The idempotency half is something I missed.

One follow-up, which also answers the question in the L368 thread: your list of distinctness axes was right and I withdraw that ask. But the new gate adds an axis that isn't covered. A refusal because a scope is refresh-required after a restore has to be distinguishable from other admission refusals — otherwise a caller can't tell "run refresh" apart from "your plan is stale" or "you lack authority".

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: Nothing missing — with the gRPC note in place I'm satisfied.

You're also right that no `toolkit-contract` crate exists; `libs/` has no such directory, so that reference in my comment was wrong. Your point about v1/v2 coexistence being deferred (`toolkit-contract-binding/DESIGN.md:734`) holds.

gears/infrastructure-resource-manager/docs/PRD.md
MikeFalcon77: Readiness column is good.

One fallout from the priority changes in the L319 thread. §13 defines criticality as *"the highest §5.1 scope priority that requires the integration"*. `fr-data-plane-catalog` and scope row 28 both moved to `p2`, but the Grant Issuance Service row (:2114) is still `p1` — and that catalog is its only input. Should be `p2`.

Token Issuer is fine: `fr-adapter-credential` sits in `p1` row 34.

MikeFalcon77 (approved):

PR #4411, opened 2026-08-04
refactor(cluster): arbitrate locks by row, not one connection per lock

asmith987:
## Summary by CodeRabbit

* **New Features**
  * Improved PostgreSQL lock coordination with lease-based ownership and liveness detection.
  * Added recovery for expired, orphaned, or unavailable locks.
  * Added batched change notifications and more reliable watcher shutdown behavior.
  * Added bounded shutdown handling to prevent resources from hanging indefinitely.

* **Bug Fixes**
  * Improved cleanup and waiter notifications when locks expire or are released.
  * Added startup validation for supported transaction isolation settings.

* **Documentation**
  * Clarified connection-pool limits, lock behavior, lifecycle, and testing expectations.


gears/system/cluster/plugins/postgres-cluster-plugin/src/config.rs
```diff
@@ -88,6 +88,14 @@ pub struct PostgresClusterConfig {
...
 
     /// Maximum pool size (write pool). Default: 5.
+    ///
+    /// Sizes the *write pool* only — cache operations, `cluster_lock` row
+    /// writes, `pg_notify`, migrations, and the reapers' queries. It does not
+    /// bound how many locks this instance can hold: every advisory-lock
+    /// statement runs on one dedicated connection outside this pool
+    /// (DESIGN.md §3.3), so a held lock consumes no pool connection. The
+    /// dedicated lock session and the LISTEN connections are each additional to
+    /// this number.
```
MikeFalcon77: **[RUST-NO-007] Stale doc comment — references old per-lock dedicated session architecture**

This doc comment for `pool_max_size` says:
> every advisory-lock statement runs on one dedicated connection outside this pool (DESIGN.md §3.3), so a held lock consumes no pool connection. The dedicated lock session and the LISTEN connections are each additional to this number.

With this PR's changes, there is no longer a per-lock dedicated session. Lock operations (acquire, renew, release) now run through the write pool via `try_acquire`/`renew`/`release`. The only dedicated connection outside the pool is the **beacon**, which carries no lock traffic at all.

The claim "a held lock consumes no pool connection" is no longer accurate — a held lock's guard task uses pool connections for renew/release operations (though it doesn't hold one permanently).

Suggested update: replace "dedicated lock session" with "beacon connection" and remove the claim that held locks consume no pool connections.

gears/system/cluster/plugins/postgres-cluster-plugin/src/config.rs
```diff
@@ -109,8 +117,9 @@ pub struct PostgresClusterConfig {
     /// its `cluster_postgres_lock_active_names` gauge — an imminent `expires_at`
     /// shortens an individual sleep so an expired lock is reclaimed near its
     /// actual deadline (`lock::reaper`'s "Wake schedule", DESIGN.md §5.2). Also
-    /// the cadence at which each lock's guard task re-asserts
-    /// `synchronous_commit = on` on its pinned connection (DESIGN.md §3.4).
+    /// the cadence at which the dedicated lock session re-asserts
+    /// `synchronous_commit = on` on its own connection (DESIGN.md §3.4), and at
+    /// which the reaper's orphan audit runs (§5.2).
```
MikeFalcon77: **[RUST-NO-007] Stale doc comment — references non-existent `synchronous_commit` re-assertion on dedicated lock session**

This doc comment for `lock_reaper_interval_ms` says:
> Also the cadence at which the dedicated lock session re-asserts `synchronous_commit = on` on its own connection (DESIGN.md §3.4)

There is no longer a dedicated lock session that re-asserts `synchronous_commit`. The beacon connection writes nothing and needs no `synchronous_commit` assertion (documented in `pg_setup.rs` and `beacon.rs`). The pool's `before_acquire` hook handles `synchronous_commit` enforcement for all pooled connections.

Suggested fix: remove the sentence about the dedicated lock session re-asserting `synchronous_commit`, or replace it with a reference to the pool's `before_acquire` hook.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
```diff
@@ -910,16 +1267,38 @@ impl DistributedLockBackend for PostgresLock {
             let deadline = started + timeout;
 
             let mut first_attempt = true;
+            // The most recent beacon outage, if the last attempt never
```
MikeFalcon77, line 1278: Stale inline comment a few lines above — references `pg_try_advisory_lock` as the source of truth

MikeFalcon77 (commented):
.

MikeFalcon77:
btw, should we take into account this PR?   https://github.com/constructorfabric/gears-rust/pull/4237

striped-zebra-dev (approved):

gears/system/cluster/plugins/postgres-cluster-plugin/src/config.rs
asmith987: Fixed the doc.

gears/system/cluster/plugins/postgres-cluster-plugin/src/config.rs
asmith987: Fixed the doc.

gears/system/cluster/plugins/postgres-cluster-plugin/src/lock/mod.rs
asmith987: Fixed the doc.

asmith987:
> btw, should we take into account this PR? #4237

@MikeFalcon77 I took a look at the PR and here is an analysis of the two implementations. They differ in implementation, but I do believe it makes sense for both of them to coexist.

|Dimension|#4237 — session advisory lock|#4411 — arbitrate by row|
|---|---|---|
|**Contract**|Hold the key while my session lives; no deadline|Lease with a TTL against the DB clock; renew or lose it|
|**Arbitrator**|`pg_try_advisory_lock` / `GET_LOCK(_, 0)` on one session|`cluster_lock` row: PK detects, tuple lock serializes, `WHERE` re-read decides (`READ COMMITTED`)|
|**Connections per held lock**|0 — one shared session (`max=1/min=1` pool) remembers all keys; guards hold `Arc<Session>` + key|0 — statements ride the shared write pool and return it|
|**Concurrent locks bounded by**|Nothing|`cluster_lock` cardinality (WARN threshold), not `pool_max_size`|
|**In-process exclusivity**|Needs a `HeldKeys` mutex registry — one session makes advisory locks re-entrant, so a second acquire would hand out a second guard|Free — two tasks in one process race exactly as two instances; `local_holders` exists only to feed the orphan sweep and is documented as non-authoritative|
|**TTL / renew**|None|Per-acquisition TTL, `renew()`, `LockExpired`|
|**Holder learns it lost the lock?**|No — the generation epoch knows but is only consulted at unlock; a long holder believes it holds a dropped key|Yes — `renew()` → `LockExpired`; beacon ping (1 s) bounds how long the instance can be wrong|
|**Reclaim after crash**|Server drops keys on session end. SQLite file backend: no automatic reclaim at all|Beacon absent → stealable by any acquirer immediately; TTL as worst case; reaper GC + orphan sweep + shutdown drain + post-reconnect hand-over|
|**Who may reclaim**|Only the session that took it|Any instance — a row DELETE needs no cooperation from the holder|
|**Wedged (not dead) holder**|Wedges the name until the process dies|Reclaimed at TTL|
|**Waiting**|Client-side exponential backoff + jitter, poll only|`LISTEN/NOTIFY` release-wake + 250 ms heartbeat safety net; release, expiry, and reclaim all NOTIFY|
|**Reconnect handling**|`AtomicU64` generation epoch invalidates every belief|New per-incarnation beacon key; `renew` fences on it, so pre-disconnect rows can never match|
|**Drop semantics**|Enqueues `(key, gen)` to a maintenance task; key stays claimed until drained → same-key re-acquire fails meanwhile|Row DELETE + `pg_notify` in one CTE; immediately re-acquirable|
|**Session affinity needed for**|Every lock operation|Beacon + 2 `LISTEN` connections only; all lock ops work under transaction pooling (config rejects `pgbouncer_transaction_mode`)|
|**Per-op cost**|One round trip, no table, no `pg_locks` scan — but every op in the process serializes on one connection (unmeasured)|One round trip on the shared pool; contended path scans `pg_locks` (un-indexed function scan) → linear in cluster-wide advisory-lock count: ~0.6 ms at a few, ~2.7 ms at 5 k (`PG-SPEC-014`). Uncontended path short-circuits via `CASE`, held to `EXPLAIN ANALYZE` by `PG-SPEC-012`|
|**Introspection**|`pg_locks` only — key is an XXH3 hash, irreversible|`cluster_lock` carries names, holders, deadlines; documented operator queries|
|**Observability**|Tracing; keepalive-failure escalation|Metrics (`cluster_postgres_lock_active_names`, cardinality WARN), beacon key logged at INFO for log correlation|
|**Backends**|PG, MySQL, SQLite (file-marker fallback)|Postgres; other backends via the SDK's `DistributedLockBackend`|
|**Schema footprint**|None|One table + migration|

Worth stating once in the thread so the comparison isn't read as "old vs. new": the design #4411 replaced was neither of these — it pinned **one `PoolConnection` per held lock** in a `DashMap`, so concurrent locks were bounded by `pool_max_size` and competed with cache traffic. #4237 and #4411 are two independent fixes for that same defect, differing in where exclusivity lives.

## Why arbitrate by row

- **No in-process map is needed to know what we hold.** Both `pg_try_advisory_lock` and `GET_LOCK` are session-scoped and *re-entrant* — re-locking a key the same session already holds succeeds and just bumps the server's hold count, with no error. Once one session serves many locks, the database can no longer answer "does this process already hold this key," so exclusivity has to move into a mutex-guarded registry in application memory. A row makes the primary key answer it: the `PRIMARY KEY (name)` conflict plus the tuple lock arbitrates two tasks in one process identically to two instances in a fleet.
- **The registry isn't just extra code, it's extra failure modes.** Claims have to be generation-stamped, retracted on unlock, retained on a *failed* unlock (re-acquiring would raise the hold count to two, a leak no release can balance), and drained by a background task — which is why re-acquiring a key right after `Drop` fails until the drain runs. None of that exists when the row is the claim.
- **"Steal if expired" is only expressible against a row.** A session-scoped lock can be released only by the session that took it, so no third party can reclaim it on a deadline. TTL leasing and per-lock advisory locks are mutually exclusive by construction — not awkward to combine, impossible.
- **A wedged holder is as dangerous as a dead one.** Session death releases advisory locks, but a process that is starved, GC-paused, or behind a half-open socket holds its keys indefinitely. The lease deadline covers that case; session scope does not.
- **Reclamation doesn't have to route back to the owner.** A row DELETE is issuable by any instance, so the acquire predicate itself reclaims — no healthy reaper on the owning instance, no recovery path per lock.
- **Session affinity shrinks to two things instead of everything.** With every lock statement on the pool, transaction-mode pooling would serve acquire, renew, release, and both sweeps correctly. Only the beacon and the two `LISTEN` connections still need a pinned session, and they're opened outside the pool deliberately.
- **The one advisory lock left is the one that composes with a lease.** The beacon is per-*instance* liveness, never per-*lock*: established by one statement, deleted by the server the instant the session ends, readable by every other session in SQL, and unfiltered by privilege (unlike `pg_stat_activity`). That last property is what lets "is this row's holder still alive?" be decided *inside* the acquire statement rather than in a second round trip that would have to be raced. Nobody ever contends for it, so nobody ever needs to release someone else's.
- **Operators get names, not hashes.** `pg_locks` only ever exposed the two halves of a name hash, which is irreversible — identifying a lock meant enumerating candidate names and hashing each. `cluster_lock` is a supported inspection surface with holders and deadlines.
- **A held lock costs no connection, and that's now true rather than asserted.** The predecessor pinned one per lock; #4237 avoids it with a shared session at the cost of the registry; the row avoids it with no in-memory ownership state at all.
- **Cost is bounded and measured where it's real.** The `pg_locks` probe only runs on the contended path (`CASE`, not `OR` — SQL doesn't guarantee `OR` operand order), and its scaling is characterized in `PG-SPEC-014` rather than assumed.

PR #4421, opened 2026-08-05
Fix(event-broker): Gate test-util-only helpers behind cfg(feature = "test-util")

maurolacy:
Fixes clippy `dead_code` errors when building with `--no-default-features --all-targets`.
Test helpers used exclusively by `#[cfg(feature = "test-util")]` tests are now gated behind the same feature flag.

## Summary by CodeRabbit

* **Tests**
  * Improved test organization by conditionally compiling test-only utilities when the `test-util` feature is enabled.
  * Centralized shared test helpers for event dispatching, offset handling, runtime events, and handler behavior.
  * Cleaned up unused test imports and declarations without changing public functionality.

### DIFF
--- a/gears/system/event-broker/event-broker-sdk/src/consumer/builder_tests.rs
+++ b/gears/system/event-broker/event-broker-sdk/src/consumer/builder_tests.rs
@@ -1,5 +1,3 @@
-#[cfg(all(feature = "db", feature = "test-util"))]
-use crate::consumer::ConsumerCommitMode;
 use crate::consumer::{
     BatchHandlerOutcome, ConsumerBuffering, ConsumerBuilder, ConsumerGroupRef, ConsumerHandle,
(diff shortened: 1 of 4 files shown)


striped-zebra-dev:
Doesn't really seem to be the right solution.
I would move the code rather then squeezing feature macrosses all over the place, does it block anyone?

maurolacy:
Tried moving all these to their own module, but it added clutter and required more use statements. So we ended up choosing this solution.

I’ll push that change, so that you can take a look at it.

I don’t think this is blocking anyone; it’s just CI that’s flagging it AFAIK.

striped-zebra-dev (approved):

striped-zebra-dev:
@maurolacy thank you.

MikeFalcon77 (approved):

Issue #4413, opened 2026-08-05
toolkit-odata: $top is documented, tested and named in errors but never bound on the wire


capybutler:
Split out of #4397, where the usage-collector side of this was fixed. None of the items below is a live wire bug; all four are latent, and each one will mislead the next person who reads it.

## The fact

`$top` is canonical OData — OASIS OData 4.01 Part 2: URL Conventions, section 5.1.6, "System Query Options $top and $skip". But nothing in this repo's toolkit binds it.

The wire binding for the OData parameter family lives in exactly one place, `toolkit::api::odata::ODataParams` (`libs/toolkit/src/api/odata.rs:12-22`):

```rust
#[serde(rename = "$filter")]  pub filter:  Option<String>,
#[serde(rename = "$orderby")] pub orderby: Option<String>,
#[serde(rename = "$select")]  pub select:  Option<String>,
pub limit:  Option<u64>,      // unprefixed — no $top field, no serde alias
pub cursor: Option<String>,
```

`toolkit-odata` does no HTTP query parsing at all; it operates on already-parsed values. So `?$top=<n>` binds nothing, and `limit` is the only page-size parameter any gear using the `OData` extractor actually accepts.

## The four places that assume otherwise

1. **`libs/toolkit-odata/src/problem_mapping.rs:106`** — `InvalidLimit` emits a `field_violations` entry named `"$top"`. A caller who sends `?limit=0` gets a 400 naming a parameter they did not send and cannot send. Pinned by the toolkit test `invalid_limit_keys_to_top` (`:261`), so fixing the name means renaming the test.

2. **`libs/toolkit-odata/src/limits.rs`** — `ODataLimits::max_top` and `validate_top` are dead code. `ODataLimits` is re-exported at `lib.rs:14` and never constructed anywhere in the repo, so the documented "maximum `$top` value (default 1000)" is enforced by nothing. `limits.rs:4`, `:15`, `:45`, `:70` and `lib.rs:279` all describe `$top` as a live parameter.

[...]

PR #4422, opened 2026-08-05
feat(toolkit)!: bind $top/$skiptoken, reject other $ options

capybutler:
`$top` is canonical OData (OASIS OData 4.01 Part 2: URL Conventions, §5.1.6 "System Query Options
$top and $skip"), and this repo already documented it, tested it, and named it in error payloads.
Nothing bound it. The wire binding for the whole OData family lives in exactly one place,
`toolkit::api::odata::ODataParams`, which declared an unprefixed `limit` with no `$top` field and no
alias — and `toolkit-odata` does no HTTP query parsing at all. So `?$top=<n>` was accepted and
discarded on every endpoint using the `OData` extractor, and `limit` was the only page-size
parameter that actually worked.

This binds `$top` as `#[serde(alias = "$top")]` on `ODataParams.limit`: one slot, two accepted
spellings. `limit` keeps working for every current caller, the platform matches canonical OData, and
the `$top` field violation already emitted by `toolkit-odata::problem_mapping` becomes truthful
without a rename.

**Binding `$top` alone would have made the `$` namespace worse, not better** — raised in review, and
the second half of this PR. `ODataParams` claims every `$`-prefixed query key by convention but
bound four of them, and axum drops query keys no extractor claimed. A client paginating the
canonical way sends `$top` with `$skip`: before, both were dropped and the wrong page size was an
immediate tell; with `$top` bound and `$skip` still dropped, the page size is honored, the offset
silently is not, and the client re-reads page one while believing it advanced.

OData does not leave this to taste. Part 1: Protocol, §6.1 "Query Option Extensibility":

> OData services SHOULD NOT require any query options to be specified in a request. Services SHOULD
> fail any request that contains query options that they do not understand and MUST fail any request
> that contains unsupported OData query options defined in the version of this specification
> supported by the service.

`$skip` and `$count` are OData-defined and unsupported here, so refusing them is a **MUST**; a typo
like `$filtre` is the SHOULD case. The same section reserves the `$` and `@` prefixes for OData,
which is why this guard polices `$` keys only: an unprefixed key belongs to the handler's own params
struct (chat-engine's `q`, file-storage's `offset`), and refusing those stays each gear's business
(account-management has `reject_non_odata_params`).

**Changes**

- `libs/toolkit/src/api/odata.rs`
  - `#[serde(alias = "$top")]` on `ODataParams.limit`, plus rustdoc on the struct stating that this
    is the single wire-binding point for OData query parameters.
  - `#[serde(alias = "$skiptoken")]` on `ODataParams.cursor` — the same one-slot-two-spellings shape
    as `$top`. `$skiptoken` is OData's opaque continuation token for server-driven paging, which is
    what `PageInfo.next_cursor` already is, and resource-group's own docs advertise the spelling.
  - `ACCEPTED_SYSTEM_QUERY_OPTIONS` (`$filter`, `$orderby`, `$select`, `$top`, `$skiptoken`) and a
    guard that answers `400` for every other `$`-prefixed key, before any value parsing. One
    `UNSUPPORTED_QUERY_PARAM` field violation per offending key, deduplicated, in wire order, so one
    round trip names every mistake.
  - The violation description separates the three cases a caller can hit: `$skip` / `$count` /
    `$format` are OData-defined but unsupported and each says what to send instead; `$Top` is a
    supported option in a spelling that does not bind, and names the one that does; `$filtre` is not
    an OData option at all and reads as `unknown query option`, not `unsupported`.
- `libs/toolkit-odata/src/problem_mapping.rs` — the `InvalidLimit` violation keeps `$top` as the
  field name and now describes both spellings: `"Invalid page size parameter ($top, alias limit)"`.
- `libs/toolkit-odata/src/limits.rs` — docs only. `ODataLimits::max_top` was documented as an
  enforced "maximum $top value (default 1000)"; nothing in the repo constructs `ODataLimits`, so
  nothing enforced it. The module docs now say it is an opt-in validator and point at what actually
  caps each request. Deleting the type was the alternative and was rejected: it is a `pub`
  re-export, so removal breaks downstream compilation for no gain.
- `docs/toolkit_unified_system/07_odata_pagination_select_filter.md` — the page-size section (both
  spellings, the ambiguity rule, `$top=0`, and the note that upper-bound capping is the handler's
  job), a continuation-token section, and an **Unsupported system query options** section: the
  accepted set, the refusal table, the §6.1 quote behind it, and the two boundaries it implies —
  unprefixed keys are out of scope, and case-sensitivity is a stated deviation from Part 2 §5.1
  (which wants option names accepted case-insensitively and with or without `$`) rather than an
  oversight. Also corrects that file's error table, which listed HTTP `422` for every `OData` error
  variant: all of them except `Db` map through `OdataError::invalid_argument()`, which renders as
  `400`, and the canonical error taxonomy has no `422` status at all — so no `OData` error could ever
  have produced one. That is a pre-existing documentation error independent of `$top`; five of the
  six corrected rows concern `$filter`, `$orderby`, and cursors. Fixed here (raised in review)
  because the same table names `InvalidLimit`.
- `gears/system/account-management/.../handlers/common.rs` — `reject_non_odata_params` claimed
  `$`-prefixed keys were "out of scope: the `OData` extractor parses them and rejects unknown ones
  (`$filtre` etc.)". That was false when written — it is exactly the hole this PR closes — and the
  gear built its non-`$` guard on top of the assumption. The comment now points at
  `ACCEPTED_SYSTEM_QUERY_OPTIONS`, and its test no longer pins `$skip` / `$count` as keys AM accepts.

**Why `$skip` is refused rather than implemented.** Offset paging is not this platform's model:
`ODataQuery` carries no offset field and `paginate_odata` is keyset-only, so honoring `$skip` would
mean adding a pagination mode, not binding a parameter. `$count` has no truthful answer either —
`PageInfo` is `{next_cursor, prev_cursor, limit}`, and serving a total would mean the `COUNT(*)` per
page that keyset pagination exists to avoid.

**Out of scope.** Toolkit crates plus the one account-management comment this PR falsifies. The
gear-side items from #4413 — the vacuous `$top` clamp test in account-management, and the
chat-engine `SearchQuery` doc comment that credits `toolkit-odata` for its own serde renames — are
left to a separate change. Accepting `$Top` case-insensitively and the unprefixed canonical spellings
(`top`, `filter`) that Part 2 §5.1 asks for is a separate conformance change; this PR answers `400`
naming the spelling that binds instead of dropping the request.

## Type of Change
- [x] Documentation update

### Breaking change details

Two wire-behavior changes, both on every endpoint using the `OData` extractor (ledger, chat-engine,
mini-chat, account-management, resource-group, usage-collector).

**1. `?$top=<n>` now sets the page size** instead of being silently ignored.

- A caller that sent `$top` and received the endpoint default now receives the page size it asked
  for.
- Gears that reject an out-of-range page size rather than clamping now answer `400` where they
  answered `200`. Concretely: usage-collector rejects `$top > 1000`
  (`gears/system/usage-collector/.../usage_records.rs`) — that branch was unreachable from the wire
  before and is reachable now.
- Sending `$top` and `limit` in one request is ambiguous and answers `400 INVALID_QUERY_PARAMS`
  (serde reports the second spelling as a duplicate field).

**2. Unsupported `$`-prefixed keys now answer `400`** where they previously answered `200` with the
key discarded. Anything outside `$filter`, `$orderby`, `$select`, `$top`, `$skiptoken`: `$skip`,
`$count`, `$expand`, `$search`, `$format`, `$compute`, `$index`, `$schemaversion`, miscased
spellings such as `$Top`, and typos such as `$filtre`. The response is
`400 UNSUPPORTED_QUERY_PARAM`, one violation per offending key, each naming the key and what to
send instead.

Migration:

- Callers using `limit` and `cursor` — none.
- Callers that sent `$top` expecting it to be ignored must stop sending it, or send a value within
  the endpoint's cap. Do not send both spellings in one request.
- Callers that sent `$skip` must drop it and page with the previous response's
  `page_info.next_cursor`, sent as `$skiptoken` or `cursor`.
- Callers that sent any other `$` option must drop it.

Not affected: unprefixed query parameters of any name, and chat-engine's `/search`, which binds its
own `$top` / `$skip` in `SearchQuery` outside this extractor and is unchanged. That leaves `$skip`
working on `/search` and refused elsewhere — a platform inconsistency this PR documents but does not
resolve.

### Version bump justification

Required by CONTRIBUTING §Enforcement. No `Cargo.toml` versions are touched in this PR —
release-plz derives them from the conventional commit (`feat(toolkit)!:` + `BREAKING CHANGE:`
footer).

| Crate | Current | Expected | Why |
|---|---|---|---|
| `cf-gears-toolkit` | 0.6.18 | 0.7.0 | Breaking wire-behavior change (two of them), pre-1.0 → bump `0.(x+1).0` per the CONTRIBUTING table. No Rust-API break: the `ODataParams` field set is unchanged (serde aliases only) and `ACCEPTED_SYSTEM_QUERY_OPTIONS` is additive, so `semver_check` has nothing to flag and downstream compilation is unaffected. |
| `cf-gears-toolkit-odata` | 0.8.4 | 0.8.5 (PATCH) | Error-description string and rustdoc only. No public API or type change; `ODataLimits` intentionally kept. |
| `cf-gears-account-management` | 0.5.0 | 0.5.1 (PATCH) | One rustdoc comment corrected and one unit test's key set updated. No API, behavior, or wire change of its own. Flagged because release-plz keys off the commit, whose `!` marker describes the toolkit wire contract — if it proposes a breaking bump for AM, that is the marker leaking, not a real break. |

## Testing
- [x] Unit tests pass
  - `cargo test -p cf-gears-toolkit --lib odata` → 37 passed, 0 failed
  - `cargo test -p cf-gears-toolkit --lib` → 261 passed, 0 failed
  - `cargo test -p cf-gears-toolkit-odata --lib` → 111 passed, 0 failed
  - Blast radius, gears using the extractor: `cf-gears-account-management` → 800 passed,
    `cf-gears-usage-collector` → 368 passed
- [ ] Integration tests pass — not run locally (AM's `tests/api_*.rs` need Postgres). All six
      extractor gears were grepped for requests carrying a now-rejected option; there are none, so
      no integration expectation flips. Left to CI.
- [ ] Manual testing completed — behavior is covered by extractor-level tests that build real
      `axum` requests, so no manual pass was made.

Twelve tests in `libs/toolkit/src/api/odata_tests.rs`, each watched fail before the code existed:

| Test | Asserts |
|---|---|
| `test_extract_odata_query_top_binds_limit` | `?$top=10` reaches `ODataQuery.limit` |
| `test_extract_odata_query_top_zero_error` | `?$top=0` → `400`, `field_violations[0].field == "$top"`, code `INVALID_LIMIT` |
| `test_extract_odata_query_top_and_limit_conflict` | `?$top=20&limit=50` → `400 INVALID_QUERY_PARAMS` |
| `test_odata_extractor_top_alias` | End-to-end through the `OData` extractor with `$filter` + `$top` |
| `test_extract_odata_query_skiptoken_binds_cursor` | `?$skiptoken=<encoded CursorV1>` reaches `ODataQuery.cursor` |
| `test_extract_odata_query_skip_rejected` | `?$skip=20` → `400`, field `$skip`, reason `UNSUPPORTED_QUERY_PARAM`, detail names `$skiptoken` |
| `test_extract_odata_query_count_rejected` | `?$count=true` → `400`, field `$count`, detail says a page carries no total |
| `test_extract_odata_query_misspelled_option_rejected` | `?$filtre=…` → `400`, described as `unknown`, not `unsupported` |
| `test_extract_odata_query_miscased_option_rejected` | `?$Top=10` → `400`, detail names the spelling that binds |
| `test_extract_odata_query_reports_every_unsupported_option` | `?$skip&$count&$expand` → three violations, in wire order |
| `test_extract_odata_query_repeated_unsupported_option_reported_once` | `?$skip=20&$skip=40` → one violation, not two |
| `test_extract_odata_query_non_odata_params_pass_through` | `?q=needle&context=2&limit=5` → `200`; the guard must not police the non-`$` namespace |

The last one is the regression guard that matters most: it pins that handlers sharing the query
string with the extractor (chat-engine's `q` / `context`, file-storage's `offset`) keep working.

## Documentation
- [x] Code is documented with rustdoc comments — `ODataParams` (wire-binding role),
      `ODataParams.limit` and `.cursor` (both spellings each, ambiguity rule),
      `ACCEPTED_SYSTEM_QUERY_OPTIONS` and `ODATA_SYSTEM_QUERY_OPTIONS` (why a `$` key is refused as
      unsupported vs unknown), the guard itself (§6.1, and what a silent drop looks like to a
      caller), `ODataLimits` / `max_top` / `validate_top` (opt-in, not automatic), and the
      `InvalidLimit` arm in `problem_mapping.rs`
- [ ] README updated (if applicable) — not applicable
- [x] API documentation updated (if applicable) —
      `docs/toolkit_unified_system/07_odata_pagination_select_filter.md`: page size, continuation
      token, unsupported system query options, plus the error-table status correction (`422` → `400`)
      described under **Changes**. The correction changes no behavior and no crate version — the file
      lives in `docs/`, outside both crates.
      Note for gear owners: OpenAPI cannot express one parameter under two names — publish the
      spelling your gear treats as canonical and mention the other in its `description`. A gear that
      wants unprefixed accidents like `?status=approved` refused rather than ignored needs its own
      guard; the extractor only polices the `$` namespace.

## Checklist
- [x] Self-review completed
- [x] No linting errors (`cargo clippy -p cf-gears-toolkit -p cf-gears-toolkit-odata -p cf-gears-account-management --all-targets` → clean)
- [x] Code is properly formatted (`cargo fmt -p cf-gears-toolkit -p cf-gears-toolkit-odata -p cf-gears-account-management -- --check` → clean)
- [x] Tests pass (see **Testing**)

## Summary by CodeRabbit

- **New Features**
  - OData pagination now accepts `$top` alongside `limit`.
  - Continuation requests can use `$skiptoken` alongside `cursor`.
  - Duplicate aliases are rejected, and invalid page sizes return clear validation errors.
  - Unknown, unsupported, repeated, or incorrectly cased OData system options are rejected with HTTP 400 responses.

- **Documentation**
  - Expanded guidance on pagination aliases, supported query options, validation behavior, and error responses.
  - Clarified that non-OData query parameters continue to pass through.

### DIFF
--- a/docs/toolkit_unified_system/07_odata_pagination_select_filter.md
+++ b/docs/toolkit_unified_system/07_odata_pagination_select_filter.md
@@ -349,6 +349,101 @@ pub fn page_to_projected_json<T: serde::Serialize>(
 
 ## Cursor-based pagination
 
+### Page size (`$top` / `limit`)
+
+The page size binds off the query string in exactly one place —
+`toolkit::api::odata::ODataParams` — and lands in `ODataQuery.limit`. Two
+spellings reach that slot:
+
+| Spelling | Notes |
+|----------|-------|
+| `$top` | Canonical OData (OASIS OData 4.01 Part 2: URL Conventions, §5.1.6). Serde alias. |
+| `limit` | The unprefixed spelling most gears publish in their OpenAPI documents. |
+
+```bash
+# Equivalent
+/users-info/v1/users?$top=20
+/users-info/v1/users?limit=20
+
+# Ambiguous — 400 InvalidArgument (duplicate field)
+/users-info/v1/users?$top=20&limit=50
+```
+
+Both spellings are one parameter, so a gear needs no per-endpoint handling to
+honor either. Two consequences worth knowing:
+
+- OpenAPI cannot express one parameter under two names. Publish the spelling
+  your gear treats as canonical and mention the other in its `description`.
+- `$top=0` is rejected (`InvalidLimit` → `400`, `field_violations[0].field` is
+  `$top`). Enforcing an upper bound is the handler's job — the extractor caps
+  filter/orderby/select size, not page size.
+- **`$skip` is not supported.** Offset paging is not the platform's model, and
+  a request that pairs `$top` with `$skip` is refused rather than served with a
+  page size it asked for at an offset it did not get. Continue a page with
+  `$skiptoken` (alias `cursor`); see [Unsupported system query
+  options](#unsupported-system-query-options).
+
+### Continuation token (`$skiptoken` / `cursor`)
+
+The same one-slot-two-spellings rule applies to the continuation token: both
+`$skiptoken` and `cursor` land in `ODataQuery.cursor`. `$skiptoken` is OData's
+opaque continuation token for server-driven paging, which is exactly what
+`PageInfo.next_cursor` is — pass the previous page's `next_cursor` back under
+either spelling.
+
+```bash
+# Equivalent
+/users-info/v1/users?$top=20&$skiptoken=eyJ2IjoxLC…
+/users-info/v1/users?limit=20&cursor=eyJ2IjoxLC…
+```
+
+### Unsupported system query options
+
+Every `$`-prefixed query key the extractor does not bind is **rejected**, not
+ignored:
+
+```bash
+# 400 InvalidArgument, field_violations[0] = { field: "$skip",
+#   reason: "UNSUPPORTED_QUERY_PARAM" }
+/users-info/v1/users?$top=20&$skip=20
+```
+
+The accepted set is `toolkit::api::odata::ACCEPTED_SYSTEM_QUERY_OPTIONS` —
+`$filter`, `$orderby`, `$select`, `$top`, `$skiptoken`. Anything else in the
+`$` namespace answers `400` with one `UNSUPPORTED_QUERY_PARAM` violation per
+offending key, so one round trip names every problem:
+
+| Option | Why it is refused |
+|--------|-------------------|
+| `$skip` | Offset paging is not the model. Use `$skiptoken` / `cursor`. |
+| `$count` | A page carries no total; `PageInfo` is `{next_cursor, prev_cursor, limit}`. Serving a total would mean a `COUNT(*)` per page, which keyset pagination exists to avoid. |
+| `$expand`, `$search`, `$compute`, `$index`, `$schemaversion` | Not implemented. |
+| `$format` | Responses are JSON; negotiate the media type with `Accept`. |
+| `$filtre`, `$topp`, … | Not OData options at all, so the violation reads `unknown query option` rather than `unsupported`. A dropped `$filtre` returns the whole unfiltered collection with a `200`. |
+
+This is required behavior, not a house rule — OASIS OData 4.01 Part 1:
+Protocol, §6.1 "Query Option Extensibility":
+
+> OData services SHOULD NOT require any query options to be specified in a
+> request. Services SHOULD fail any request that contains query options that
+> they do not understand and MUST fail any request that contains unsupported
+> OData query options defined in the version of this specification supported by
+> the service.
+
+Two boundaries follow from the same section:
+
+- **Unprefixed keys are out of scope.** §6.1 reserves the `$` prefix for
+  OData, so a non-`$` key belongs to the handler's own params struct — the
+  extractor never rejects one. A gear that wants `?status=approved` refused
+  instead of ignored needs its own guard; account-management has one
+  (`reject_non_odata_params`).
+- **Spelling is case-sensitive, deliberately.** OData 4.01 Part 2 §5.1 asks a
+  4.01 service to accept system query option names case-insensitively and with
+  or without the `$` prefix. This platform accepts neither `$Top` nor the
+  unprefixed canonical `top`; it answers `400` naming the spelling that binds
+  rather than dropping the request. Closing that gap is a separate conformance
+  change.
+
 ### Page structure
 
 ```rust
@@ -475,15 +570,21 @@ OData errors are defined in `toolkit_odata::Error` (aliased as `ODataError` in `
 
 | Variant | Description | HTTP status |
 |---------|-------------|-------------|
-| `InvalidFilter(String)` | Malformed `$filter` expression | 422 |
-| `InvalidOrderByField(String)` | Unsupported `$orderby` field | 422 |
-| `InvalidCursor` / `CursorInvalid*` | Malformed or expired cursor | 422 |
-| `OrderMismatch` | Cursor/query order conflict | 422 |
-| `FilterMismatch` | Cursor/query filter conflict | 422 |
-| `InvalidLimit` | Invalid limit parameter | 422 |
+| `InvalidFilter(String)` | Malformed `$filter` expression | 400 |
+| `InvalidOrderByField(String)` | Unsupported `$orderby` field | 400 |
+| `InvalidCursor` / `CursorInvalid*` | Malformed or expired cursor | 400 |
+| `OrderMismatch` | Cursor/query order conflict | 400 |
+| `FilterMismatch` | Cursor/query filter conflict | 400 |
+| `InvalidLimit` | Invalid page size parameter (`$top`, alias `limit`) | 400 |
 | `Db(String)` | Database error (logged, generic message returned) | 500 |
 
+Every variant above except `Db` maps through `OdataError::invalid_argument()`, and
+`CanonicalError::InvalidArgument` renders as `400`. The canonical error taxonomy has no `422`
+status, so no `OData` error can produce one.
+
 `$select` validation errors (too long, too many fields, duplicates) are caught during parsing in the `OData` extractor and returned as `400 Bad Request` with RFC 9457 Problem Details before reaching the handler.
+
+Unsupported system query options are not `toolkit_odata::Error` variants — the extractor builds the canonical `InvalidArgument` itself, before any value parsing, with one `UNSUPPORTED_QUERY_PARAM` field violation per offending key. See [Unsupported system query options](#unsupported-system-query-options).
 
 ### Error conversion
 
--- a/libs/toolkit/src/api/odata.rs
+++ b/libs/toolkit/src/api/odata.rs
@@ -9,6 +9,11 @@ use toolkit_odata::{CursorV1, Error as ODataError, ODataOrderBy, OrderKey, SortD
 pub use toolkit_odata::ODataQuery;
 // CursorV1 is available through the private import above for internal use
 
+/// Wire binding for the `OData` query-parameter family.
+///
+/// This struct is the single place where `OData` query parameters are bound
+/// off the URL: `toolkit-odata` operates on already-parsed values and does no
+/// HTTP query parsing of its own.
 #[derive(Deserialize, Default)]
 pub struct ODataParams {
     #[serde(rename = "$filter")]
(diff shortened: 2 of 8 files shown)


libs/toolkit/src/api/odata.rs
```diff
@@ -17,6 +22,15 @@ pub struct ODataParams {
...
+    /// Page size. Accepted as `limit` or as the canonical `OData` spelling
+    /// `$top` (OASIS `OData` 4.01 Part 2: URL Conventions, §5.1.6 "System
+    /// Query Options $top and $skip") — both spellings fold onto this one
+    /// slot, so a gear needs no per-endpoint handling to honor either.
+    ///
+    /// Sending both in one request is ambiguous and is rejected with
+    /// `400 InvalidArgument` (serde reports the second spelling as a
+    /// duplicate field).
+    #[serde(alias = "$top")]
     pub limit: Option<u64>,
```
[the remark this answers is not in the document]
capybutler: Correct that the extractor has no upper bound — `extract_odata_query` rejects `limit == 0` and passes everything else to `with_limit` unchecked. But that predates this PR and the diff doesn't move it: `limit` was already bound and already uncapped on `main`. The alias adds a second spelling onto the same slot, so the accepted range is identical before and after. No new unbounded path is introduced here.

Capping in `ODataParams` is also the wrong layer. That struct is the shared extractor for every gear, and page-size policy is deliberately per-endpoint — the repo already carries two different policies with two different numbers:

- `usage-collector` rejects `> 1000` as `InvalidArgument` (`gears/system/usage-collector/.../handlers/usage_records.rs`, `MAX_PAGE_SIZE`)
- `chat-engine` clamps to `50` silently (`gears/chat-engine/.../domain/search.rs`, `MAX_PAGE_SIZE`)

A toolkit-wide constant would be too low for one and too high to protect the other, and it would change behavior for existing `limit` callers on every endpoint using the extractor — turning a one-line breaking change (`$top` now binds) into a platform-wide page-size break. It would also make usage-collector's own 400 branch unreachable and flip chat-engine from clamping to rejecting.

`ODataLimits::max_top` is the opt-in validator for gears that want one; this PR documents it as caller-driven rather than ambient precisely so this distinction is written down (`libs/toolkit-odata/src/limits.rs`).

On the integration test: there is no endpoint-level limit at this layer to test against, so the test would have to invent the cap it verifies. The per-endpoint caps above are already covered by their own gears' tests.

Out of scope for this PR — not planned.

docs/toolkit_unified_system/07_odata_pagination_select_filter.md
```diff
@@ -349,6 +349,35 @@ pub fn page_to_projected_json<T: serde::Serialize>(
...
+# Ambiguous — 400 InvalidArgument (duplicate field)
+/users-info/v1/users?$top=20&limit=50
+```
+
+Both spellings are one parameter, so a gear needs no per-endpoint handling to
+honor either. Two consequences worth knowing:
+
+- OpenAPI cannot express one parameter under two names. Publish the spelling
+  your gear treats as canonical and mention the other in its `description`.
+- `$top=0` is rejected (`InvalidLimit` → `400`, `field_violations[0].field` is
```
MikeFalcon77, line 377: This section needs a third bullet about $skip, and probably a code change too.

OData clients send $top and $skip as a pair. Before this PR both were dropped, so a client that tried to paginate got the same page every time and noticed on request two. After this PR $top works and $skip is still dropped,  so the page size changes, the offset never does, and the client gets page one forever while believing it is paging through the set. That is harder to debug than the old behavior, not easier.

Two options:

Reject $skip with 400 InvalidArgument, field $skip, detail pointing at cursor. Costs one more field on ODataParams and one check in the extractor. A client that sends it finds out immediately.

Document it here: state that $skip is not supported, that offset paging is not the model, and that cursor is the replacement.
I would do both. Option 2 alone still leaves the silent-drop behavior in place for anyone who doesn't read the docs, and the whole point of binding $top is that clients now expect canonical OData paging to work.

Same question applies to $count, though it's less urgent — it has no working equivalent under another name, so nobody is half-served by it.

capybutler: Both, and the reject generalized past `$skip` / `$count`.

`ODataParams` claimed the whole `$` namespace but bound four options, so every other `$` key was accepted and discarded — not only `$skip` and `$count`, but `$expand`, `$search`, `$format`, miscased `$Top`, and typos like `$filtre` (a dropped `$filtre` returns the unfiltered collection with a 200). A dedicated `$skip` field would have closed two holes out of a class. `extract_odata_query` now refuses any `$` key outside `ACCEPTED_SYSTEM_QUERY_OPTIONS` — `$filter`, `$orderby`, `$select`, `$top`, `$skiptoken` — before any value parsing: 400, one `UNSUPPORTED_QUERY_PARAM` violation per offending key, deduplicated, in wire order, so one round trip names every mistake.

The spec makes this obligatory rather than a preference, which is worth having in the PR — OASIS OData 4.01 Part 1: Protocol, §6.1 "Query Option Extensibility":

> OData services SHOULD NOT require any query options to be specified in a request. Services SHOULD fail any request that contains query options that they do not understand and MUST fail any request that contains unsupported OData query options defined in the version of this specification supported by the service.

`$skip` and `$count` are OData-defined and unsupported here, so refusing them is the MUST; `$filtre` is the SHOULD case. §6.1 also reserves the `$` and `@` prefixes for OData, and that is what keeps this guard off the unprefixed namespace: `q`, `context`, `offset` belong to the handler's own params struct, with a test pinning that they still pass. Refusing unprefixed accidents like `?status=approved` stays each gear's business — AM's `reject_non_odata_params`.

One correction to the framing, which does not change the conclusion. The observable symptom is the same before and after: identical rows on request two. What binding `$top` alone changed is the *signal* — pre-PR a wrong page size was an immediate tell that the server ignored OData, and post-PR the page size looks right, so the repeated page is the only tell left. Worse signal, not a new failure mode.

`$skip` is refused rather than implemented because offset paging is not the model at any layer: `ODataQuery` carries no offset field and `paginate_odata` is keyset-only, so honoring it means adding a pagination mode, not binding a parameter. Its violation detail points at `$skiptoken` / `cursor`.

On `$count`: same treatment, and not planned. `PageInfo` is `{next_cursor, prev_cursor, limit}`, so there is no total to report, and adding one means the `COUNT(*)` per page that keyset pagination exists to avoid. The violation detail says exactly that.

Two things that came out of this and are worth flagging:

- **`$skiptoken` is now a `cursor` alias**, same one-slot-two-spellings shape as `$top`. It is OData's opaque continuation token for server-driven paging, which is what `PageInfo.next_cursor` already is, and `resource-group`'s own docs advertise the spelling — without the alias the new catch-all would have started 400ing it.
- **AM's comment was already wrong about this.** `reject_non_odata_params` documented that "`$`-prefixed keys are intentionally out of scope: the `OData` extractor parses them and rejects unknown ones (`$filtre` etc.)". That was false when written, and the gear built its non-`$` guard on top of the assumption; its test pinned `$skip` / `$count` as keys AM accepts. Both now match the extractor.

Two decisions I would rather you weigh in on than settle myself:

1. **Binding stays case-sensitive.** Part 2 §5.1 says a 4.01 service MUST accept system query option names case-insensitively and with or without the `$` prefix — so strictly, `$Top` and an unprefixed `top` should both bind. Neither does; both now answer 400 naming the spelling that binds, which is at least not a silent drop. Full conformance (normalize casing, accept `top` / `filter` / `orderby` / `select`) is a larger change and is documented as a separate one. Reasonable line to draw?
2. **Cursor violations are still keyed to `cursor`, not `$skiptoken`**, while page-size violations are keyed to the canonical `$top` with both spellings in the description. Renaming the cursor key would change error payloads for every existing cursor caller with no functional gain, so I left the asymmetry. Say if you would rather it match `$top`'s treatment.

Also worth knowing: `chat-engine`'s `/search` binds its own `$top` / `$skip` in `SearchQuery` outside this extractor, so `$skip` keeps working there and answers 400 everywhere else. The PR documents that inconsistency without resolving it.

Twelve extractor tests cover this (each watched fail first), including one that pins the non-`$` namespace as untouched. The third bullet you asked for plus a continuation-token section and an "Unsupported system query options" section — accepted set, refusal table, the §6.1 quote, and the two boundaries above — are in `docs/toolkit_unified_system/07_odata_pagination_select_filter.md`.

PR #4425, opened 2026-08-05
docs(cluster): draft design to support deployable independent gear process

asmith987:
## Summary

Adds `gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md` — a **proposed** design
(no implementation) for running `cluster` as its own deployable out-of-process
gear, instead of a library every consumer links and hosts in-process.

Today cluster registers backends into the consumer's **in-process** `ClientHub`
under `cluster:{profile}`. That cannot serve Profile 3 (K8s Native): an OoP
consumer has its own hub (every resolve fails `ProfileNotBound`), needs its own
backend credentials and pool per replica, and duplicates renewal loops, reapers
and watch fan-out per process.

## The central decision

The process boundary is cut **exactly at the four backend traits** (proposed
ADR-011). The remote client isn't a new facade — it's four ordinary backend impls
that satisfy their trait with a remote call. One `Arc<dyn ClusterClient>` per
process is a factory for them; the profile rides on each request and resolves
server-side; local impl wins over remote.

Consumer-visible machinery therefore keeps its shape across profiles, and a
consuming gear writes `deps = [cluster]` plus its existing profile marker — no
mode flag, endpoint or client config. **One deliberate SDK break:** `resolve()`
becomes `async` (all 73 call sites are inside the cluster tree).

Transport is a gRPC data plane plus a REST lifecycle plane. Performance is the
largest consequence (§7.2) — compound ops like `increment` are the answer, and a
Profile 1 win too; phase 0a benchmarks before anything is committed.

## Scope and follow-up

Docs only — no code, build or test impact. §0 is a reading guide; §2 and §3 carry
the decisions that govern the rest, and §9 lists the blocking open questions.

Once reviewed, this document gets **synced into `DESIGN.md`** and the **required
ADRs amended or added** (§10 lists the deltas: ADR-002 amendment, new ADR-011
through ADR-015, plus PRD and platform-doc changes).

### DIFF
--- a/gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
+++ b/gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
@@ -1,0 +1,3505 @@
+# Technical Design — Cluster as a Deployable (Out-of-Process) Gear
+
+**Status**: Proposed — design only, no implementation.
+
+**Amends**: [cluster DESIGN.md](./DESIGN.md) §3.15 ("Cluster is an in-process Rust library SDK; it has no deployment topology of its own") and extends [cluster PRD.md](./PRD.md) §3.1 with the platform's OoP deployment profiles.
+
+**Conforms to**: [`docs/arch/toolkit-oop/PRD.md`](../../../../docs/arch/toolkit-oop/PRD.md) (ToolKit Distributed Gears) and its ADRs — chiefly [ADR-0002 REST-first OoP](../../../../docs/arch/toolkit-oop/ADR/0002-cpt-cf-adr-rest-first-oop.md), [ADR-0005 Eventual Readiness](../../../../docs/arch/toolkit-oop/ADR/0005-cpt-cf-adr-eventual-readiness.md), [ADR-0006 Platform-Plane Auth](../../../../docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md), [ADR-0008 Two-Plane Auth](../../../../docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md).
+
+**Change driver**: the team decided cluster must be a gear of its own — a deployable process/pod that other gears reach over a network — rather than a library every consumer links and hosts in-process.
+
+<!-- toc -->
+
+- [0. Reading Guide](#0-reading-guide)
+- [1. Problem Statement & Goals](#1-problem-statement--goals)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+  - [The primitive set: cache, distributed lock, leader election](#the-primitive-set-cache-distributed-lock-leader-election)
+- [2. Alignment with the Platform OoP Model](#2-alignment-with-the-platform-oop-model)
+  - [2.0 Framework state as of this writing](#20-framework-state-as-of-this-writing)
+  - [2.1 What the platform PRD dictates](#21-what-the-platform-prd-dictates)
+  - [2.2 Transport decision — contract-first, gRPC data plane, REST lifecycle plane](#22-transport-decision--contract-first-grpc-data-plane-rest-lifecycle-plane)
+    - [2.2.1 Live option — split the transport by primitive](#221-live-option--split-the-transport-by-primitive)
+    - [2.2.2 What cluster reuses, and what it adds](#222-what-cluster-reuses-and-what-it-adds)
+  - [2.3 Where cluster does not fit the standard mould](#23-where-cluster-does-not-fit-the-standard-mould)
+- [3. The Central Architectural Decision — the Remote Backend Seam](#3-the-central-architectural-decision--the-remote-backend-seam)
+  - [3.1 Where the boundary goes](#31-where-the-boundary-goes)
+  - [3.2 Deployment profiles, one consumer API](#32-deployment-profiles-one-consumer-api)
+  - [3.3 Revised layer diagram](#33-revised-layer-diagram)
+  - [3.4 Crate layout changes](#34-crate-layout-changes)
+  - [3.5 Rejected alternatives](#35-rejected-alternatives)
+- [4. Making Cluster a Deployable Gear](#4-making-cluster-a-deployable-gear)
+  - [4.1 Current state audit](#41-current-state-audit)
+  - [4.2 Capability set changes](#42-capability-set-changes)
+  - [4.3 Binary target and OoP bootstrap](#43-binary-target-and-oop-bootstrap)
+  - [4.4 Readiness — the four-state model](#44-readiness--the-four-state-model)
+  - [4.5 Discovery: name resolution vs. finding the name](#45-discovery-name-resolution-vs-finding-the-name)
+  - [4.6 Platform-plane authentication](#46-platform-plane-authentication)
+  - [4.7 Startup: eventual readiness vs. loud capability validation](#47-startup-eventual-readiness-vs-loud-capability-validation)
+    - [4.7.1 `resolve()` is async, and validates inline whenever it can](#471-resolve-is-async-and-validates-inline-whenever-it-can)
+  - [4.8 Shutdown and the reverse-dependency drain rule](#48-shutdown-and-the-reverse-dependency-drain-rule)
+  - [4.9 Consumer-side wiring](#49-consumer-side-wiring)
+    - [4.9.1 What is wired, and what is resolved per call](#491-what-is-wired-and-what-is-resolved-per-call)
+    - [4.9.2 But it must not be hand-written per consumer](#492-but-it-must-not-be-hand-written-per-consumer)
+    - [4.9.3 The registration itself](#493-the-registration-itself)
+  - [4.10 Deployment artifacts and operator config](#410-deployment-artifacts-and-operator-config)
+    - [4.10.1 Schema migrations](#4101-schema-migrations)
+  - [4.11 Modification checklist](#411-modification-checklist)
+- [5. Runtime Profile Management](#5-runtime-profile-management)
+  - [5.1 What must become runtime state](#51-what-must-become-runtime-state)
+  - [5.2 The `ProfileRegistry`](#52-the-profileregistry)
+  - [5.3 Backend instance sharing — the N-connections problem](#53-backend-instance-sharing--the-n-connections-problem)
+  - [5.4 Per-client profile binding and session tracking](#54-per-client-profile-binding-and-session-tracking)
+  - [5.5 Profile discovery and capability validation over the wire](#55-profile-discovery-and-capability-validation-over-the-wire)
+  - [5.6 Dynamic profile reload](#56-dynamic-profile-reload)
+  - [5.7 Capacity, isolation and failure modes](#57-capacity-isolation-and-failure-modes)
+  - [5.8 Replication — leases live in the store, not in a replica](#58-replication--leases-live-in-the-store-not-in-a-replica)
+    - [5.8.1 The model](#581-the-model)
+    - [5.8.2 Restarts and upgrades revoke nothing](#582-restarts-and-upgrades-revoke-nothing)
+    - [5.8.3 Replica count: the model now, the count after conformance](#583-replica-count-the-model-now-the-count-after-conformance)
+- [6. API Surface](#6-api-surface)
+  - [6.1 Contract-first, three projections](#61-contract-first-three-projections)
+  - [6.2 The contract traits](#62-the-contract-traits)
+  - [6.3 Complete facade → wire mapping](#63-complete-facade--wire-mapping)
+  - [6.4 Cache contract](#64-cache-contract)
+  - [6.5 Lock contract — unary against a store-owned lease](#65-lock-contract--unary-against-a-store-owned-lease)
+  - [6.6 Leader-election contract — unary join/resign plus one watch](#66-leader-election-contract--unary-joinresign-plus-one-watch)
+  - [6.7 Profile / admin contract](#67-profile--admin-contract)
+  - [6.8 Watch stream semantics](#68-watch-stream-semantics)
+  - [6.9 Error model over the wire](#69-error-model-over-the-wire)
+  - [6.10 Idempotency and retry](#610-idempotency-and-retry)
+  - [6.11 Versioning](#611-versioning)
+- [7. Cross-Cutting Consequences](#7-cross-cutting-consequences)
+  - [7.1 Security](#71-security)
+  - [7.2 Performance — the primary consequence](#72-performance--the-primary-consequence)
+    - [7.2.1 The hop arithmetic](#721-the-hop-arithmetic)
+    - [7.2.2 The cost is concentrated, not spread](#722-the-cost-is-concentrated-not-spread)
+    - [7.2.3 The hot pattern, priced](#723-the-hot-pattern-priced)
+    - [7.2.4 The binding constraint is pool fan-in, not gRPC](#724-the-binding-constraint-is-pool-fan-in-not-grpc)
+    - [7.2.5 Other per-operation costs](#725-other-per-operation-costs)
+    - [7.2.6 What gets better](#726-what-gets-better)
+    - [7.2.7 Escape hatches, in order of preference](#727-escape-hatches-in-order-of-preference)
+    - [7.2.8 The critical-section rule — ADR-002 is amended](#728-the-critical-section-rule--adr-002-is-amended)
+  - [7.3 Leadership liveness tracks the consumer, not the transport](#73-leadership-liveness-tracks-the-consumer-not-the-transport)
+  - [7.4 Scoping across the boundary](#74-scoping-across-the-boundary)
+  - [7.5 Observability](#75-observability)
+  - [7.6 Testing strategy](#76-testing-strategy)
+- [8. Phasing](#8-phasing)
+- [9. Open Questions and Decisions Needed](#9-open-questions-and-decisions-needed)
+- [10. Documentation Deltas](#10-documentation-deltas)
+- [11. Appendix — A Consumer and the Gear, End to End](#11-appendix--a-consumer-and-the-gear-end-to-end)
+  - [11.1 Consumer side](#111-consumer-side)
+  - [11.2 Gear side](#112-gear-side)
+  - [11.3 Operator config](#113-operator-config)
+  - [11.4 One call, end to end](#114-one-call-end-to-end)
+  - [11.5 What actually differs between the profiles](#115-what-actually-differs-between-the-profiles)
+- [12. Appendix — Implementation Sketch](#12-appendix--implementation-sketch)
+  - [12.1 The contract and its DTOs — `cluster-sdk/src/{contract,dto}.rs`](#121-the-contract-and-its-dtos--cluster-sdksrccontractdtors)
+  - [12.2 The error codec — `cluster-sdk/src/convert.rs`](#122-the-error-codec--cluster-sdksrcconvertrs)
+  - [12.3 Gear: the profile registry — `cluster/src/registry.rs`](#123-gear-the-profile-registry--clustersrcregistryrs)
+  - [12.4 Gear: `LocalClusterClient` — `cluster/src/local_client.rs`](#124-gear-localclusterclient--clustersrclocal_clientrs)
+  - [12.5 Gear: the session index — `cluster/src/session.rs`](#125-gear-the-session-index--clustersrcsessionrs)
+  - [12.6 Gear: the service impls — `cluster/src/api/grpc/`](#126-gear-the-service-impls--clustersrcapigrpc)
+  - [12.7 Gear: the gear — `cluster/src/gear.rs`](#127-gear-the-gear--clustersrcgearrs)
+  - [12.8 Gear: the binary — `cluster/src/{main,registered_gears}.rs`](#128-gear-the-binary--clustersrcmainregistered_gearsrs)
+  - [12.9 Client: `RemoteClusterClient` — `cluster-sdk/src/client/remote.rs`](#129-client-remoteclusterclient--cluster-sdksrcclientremoters)
+  - [12.10 Client: the cache backend — the simple case](#1210-client-the-cache-backend--the-simple-case)
+  - [12.11 Client: the lock backend](#1211-client-the-lock-backend)
+  - [12.12 Client: leader election — event pump plus resign](#1212-client-leader-election--event-pump-plus-resign)
+  - [12.13 Client: the requirement registry and readiness — `cluster-sdk/src/requirements.rs`](#1213-client-the-requirement-registry-and-readiness--cluster-sdksrcrequirementsrs)
+  - [12.14 Client: the resolver — `cluster-sdk/src/cache/resolver.rs`](#1214-client-the-resolver--cluster-sdksrccacheresolverrs)
+  - [12.15 Client: the registration — `cluster-sdk/src/wiring.rs`](#1215-client-the-registration--cluster-sdksrcwiringrs)
+  - [12.16 Consumer: the whole surface, once](#1216-consumer-the-whole-surface-once)
+  - [12.17 Constraints the SDK's existing types impose](#1217-constraints-the-sdks-existing-types-impose)
+
+<!-- /toc -->
+
+## 0. Reading Guide
+
+| Question | Section |
+|---|---|
+| How does this fit the platform's OoP model, and which transport? | §2 |
+| What must change so cluster is a deployable gear? | §4 (premise in §3) |
+| How are deployment profiles read and managed at runtime, and how does a request route to the right backend connection? | §5 |
+| What is the API surface, and how does every facade method map onto it? | §6 |
+
+§2 and §3 come first because two decisions govern everything else: which transport the platform allows, and where the process boundary is cut. §7 collects consequences, including three that change committed contracts (§7.1 last row, §7.2, §7.3).
+
+## 1. Problem Statement & Goals
+
+Today the cluster gear (`cf-gears-cluster`, crate `cluster`) is a `RunnableCapability` that, during `start`, instantiates each profile's backends from operator config and registers them in the **in-process** `ClientHub` under the scope `cluster:{profile}`. Consumer gears resolve `Arc<dyn ClusterCacheBackend>` (and the three siblings) straight out of that same hub. The Postgres pool, the TTL reapers, the leader-renewal loops and the watch fan-out all live in the consumer's own process.
+
+That model cannot serve Profile 3 (K8s Native):
+
+- A gear in its own pod has its own `ClientHub`. Nothing is registered under `cluster:{profile}` there, so every resolve fails with `ProfileNotBound`.
+- Every OoP consumer would need its own Postgres/Redis credentials and its own pool, multiplying infrastructure connections by replica count and pushing credential distribution out to every gear.
+- Coordination state that ought to be shared — a renewal loop, a watch subscription, a lock reaper — would be duplicated per consumer process.
+
+### Goals
+
+1. **`cluster` runs as its own pod/process** and serves the three coordination primitives to gears in other processes, within the platform's OoP model (§2).
+2. **Consumer code does not change *between deployment profiles*** — one source file compiles and behaves identically whether cluster is in-process or in another pod. The four facades, the typed-profile resolver, capability requirements, scoping, watch-union events and the auto-restart combinator all keep their shapes and semantics. This is the platform's own `cpt-cf-fr-developer-transparency` and `cpt-cf-fr-client-transparency` applied to cluster.
+
+   One **one-time SDK change** is required to get there, and it is deliberate: `resolve()` becomes `async` (§4.7.1). A remote resolution must be able to fetch a profile descriptor before it can validate declared capabilities, and a sync signature cannot. The migration is `resolve()` → `resolve().await` at each call site, inside `init`/`start`, which are already `async fn` (`libs/toolkit/src/contracts.rs:38,162`). All 73 current call sites are inside the cluster tree — tests, examples and SDK internals — with none in any consuming gear, so this is the cheapest it will ever be. Profile *transparency* is the invariant worth protecting here; never touching the SDK signature is not.
+3. **Profile 1 (Embedded) stays fully supported.** Cluster PRD §1.3 requires zero-infrastructure dev/test; platform PRD Profile 1 requires the same code to work in-process.
+4. **Backend connections are owned once, by the cluster gear**, and shared across profiles pointing at the same infrastructure.
+5. **Profiles are runtime state**, queryable over the wire, so a client can be told what it is bound to and capability validation still fails loudly.
+6. **Internal-only.** Cluster has no externally-visible API. Under `cpt-cf-fr-api-visibility` — as rewritten by [PR #4403](https://github.com/constructorfabric/gears-rust/pull/4403) — visibility and authentication are **orthogonal axes**: cluster's routes stay at the default visibility (`is_public = false`, i.e. no `.exposed()`) *and* are `.authenticated()`. `OperationBuilder::public()`, which conflated the two, is removed. Nothing is registered with the gateway (§6.7).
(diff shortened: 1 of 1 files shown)


gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+| **3 — specified but unimplemented** | Present only in platform docs; cluster must not assume it | `InternalAuthMiddleware` as a server-side middleware (ADR-0006 / toolkit-oop DESIGN); `ApiContractsConfig.remote_grpc_endpoints` (DESIGN §"OoP bootstrap" — no Rust type exists); a dependency-resolution loop that *wires typed clients* (see the reality check below); gRPC endpoint resolution inside that loop |
+
+> **Reality check on dependency resolution (tier 3), because three sections depend on it.** The shipped loop —
+> `resolve_one_dep`, `libs/toolkit/src/runtime/oop_registration.rs:196-224` — calls
+> `DirectoryClient::resolve_rest_service(dep)`, stores the resulting **base-URL string** in `ResolvedRestEndpoints`,
+> and marks the dep resolved for `/readyz`. It builds no client, invokes no callback, and never resolves a gRPC
+> endpoint. The platform DESIGN's *"on resolved: wire client into ClientHub"* is a specification, not current
+> behaviour. PR #4084's `#[toolkit::consumes]` / `ConsumerRegistration` is the first implementation of it, and its
+> closure is REST-shaped and synchronous — which is precisely the gap §4.9.3 has to close.
```
MikeFalcon77: **The base moved — this reality check cites removed code.**

`resolve_one_dep`, `resolve_deps` and `ResolvedRestEndpoints` are gone from `main`, explicitly retired (`libs/toolkit/src/runtime/oop_registration_tests.rs:143-147`, `host_runtime.rs:1445-1449`). The proxy-wiring phase has landed: `run_proxy_wiring_phase` at `host_runtime.rs:419`, gated on `contract-directory-rest-client`.

So the tier-3 row "a dependency-resolution loop that wires typed clients" is now tier 1. Re-verify §4.9.2 and §4.9.3 against the merged phase — the gap they describe may already be closed, or closed differently.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+
+Async alone does not solve B — it changes how you wait, not whether you may. So `resolve()` has two outcomes, and which one applies depends solely on whether the descriptor is obtainable in time:
+
+| | Descriptor available (Profile 1, or Profile 3 with cluster reachable) | Descriptor unavailable within the timeout (cold start) |
+|---|---|---|
+| Validation | **Inline**, at `resolve()` | Deferred to the SDK's readiness contributor |
+| On failure | `Err(CapabilityNotMet { primitive, capability, provider })` — the full diagnostic, returned to the caller | `Unhealthy` carrying the same triple; `/readyz` stays 503 |
+| Consumer code | Unchanged | Unchanged |
+
+**The reachable case is the normal one.** In Profile 1 the descriptor is intrinsic — the bound object *is* the real backend — so validation is immediate and exactly as it is today. In Profile 3 a consumer restarting into a running fleet gets its descriptor from the prefetch or a single `DescribeProfiles` in milliseconds. So in steady-state operation, remote resolution behaves **identically to embedded**, including for a consumer that branches on `CapabilityNotMet` to degrade or pick another profile.
```
MikeFalcon77: **Not true under the shipped phase order — inline validation never happens in Profile 3.**

`run_proxy_wiring_phase()` runs *after* `run_start_phase()` (`host_runtime.rs:1441-1452`), and §10 tells consumers to resolve facades in `start`. So at `resolve()` time there is no `RemoteClusterClient` in the hub, no channel, no `DescribeProfiles`, and no descriptor — validation always defers to readiness.

Either get the registration to run before `start`, or drop the claim that Profile 3 behaves identically to embedded in steady state.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+**Non-blocking:**
+
+| # | Question |
+|---|---|
+| 16 | **Who registers the client, and when?** (§4.9.2, §4.9.3) — **no longer blocking.** Ownership is settled: the runtime's proxy-wiring phase replays an inventoried `ConsumerRegistration` (#4084, ADR-0004), and cluster's registration fits that closure signature as shipped. Two remaining asks are both preferences rather than requirements: **(a) invoke the registration before `start`**, which would make inline validation the norm and remove the first-call bind on the cold path — the lazy facade binding of §4.9.1 covers late registration correctly, so this is an optimisation; **(b) own the k8s-DNS naming convention**, which buys consistency rather than capability. What *is* worth raising with the #4084 owners: the platform DESIGN's two contradictory statements of ownership (§4.9.2), and whether a gRPC counterpart to `#[toolkit::consumes]`'s REST resolving client is planned — without one, every platform-plane gRPC consumer hand-writes the registration cluster hand-writes |
```
MikeFalcon77: **Decision 16(a) is blocking, not an optimisation.**

Registration running before `start` is the precondition for §4.7.1's entire inline-validation story, not a cold-path improvement. Under the shipped order (proxy wiring after start) the inline path never executes in Profile 3.

Move 16(a) to the blocking table.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+3. **DirectoryService** — `resolve_grpc_service(..)`, shipped at `libs/system-sdks/sdks/directory/src/api.rs:82`.
+
+> **What each source is actually for.** (3) is not a DNS substitute; it answers "where is this thing" when the address is **not derivable by convention** — Profile 2's UDS paths and dynamic ports, non-k8s deployments with no cluster DNS, ephemeral test ports, and enumerating instances that a single VIP hides. In k8s none of those apply, which is exactly why `cpt-cf-fr-k8s-native` forbids *requiring* DirectoryService there.
+
+**Who builds the string.** The platform DESIGN assigns this to the framework — `cpt-cf-fr-k8s-native` and the sequence at `DESIGN.md:1338` (`Discovery: k8s DNS ({gear}.{ns}.svc.cluster.local)`) — but no DNS-derived resolution exists in `libs/toolkit` today (tier 3, §2.0); the shipped dep loop calls `resolve_rest_service`, and the only `svc.cluster.local` reference in the toolkit is a doc comment on `oop_serve.rs:188`'s `advertise_uri`, which is the address a gear publishes *for itself*. So cluster's `ConsumerRegistration` derives the endpoint until the framework does (§4.9.3, decision 16).
```
MikeFalcon77, line 721: **The shipped resolver is directory-based, not DNS — and it hard-gates readiness.**

`run_proxy_wiring_phase` builds a `DirectoryEndpointResolver` from the hub's `DirectoryClient`. With no `DirectoryClient` present it logs an error and leaves every remote dep unresolved, so `/readyz` stays 503 (`host_runtime.rs:436-453`).

A k8s deployment that deliberately runs no DirectoryService — which `cpt-cf-fr-k8s-native` permits — therefore never becomes ready if cluster consumers go through this phase. "DNS-first, DirectoryService-optional" describes the plan, not the framework.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+Those pre-existing command-channel seams are why this is cheap rather than a rewrite: `LockGuard::channel(name, buffer) -> (LockCommandReceiver, LockGuard)` was built so a backend could service `renew`/`release` asynchronously. A remote backend is just another such servicer.
+
+### 3.2 Deployment profiles, one consumer API
+
+Using the platform's profile vocabulary rather than inventing terms:
+
+| Platform profile | Who owns backends | Consumer wiring |
+|---|---|---|
+| **Profile 1 — Embedded** | The consumer's own process (cluster gear or `ClusterWiring` linked in) | Cluster gear's `start` registers the real backends **and** a `LocalClusterClient` under `dyn ClusterClient` into the shared hub. Explicit `deps = [cluster]` gives the topo-sort ordering |
+| **Profile 2 — Host + Workers** (P2) | The Platform Host or a dedicated worker process | Same as Profile 3, over UDS/localhost |
```
MikeFalcon77: **Multi-host Profile 2 is not "same as Profile 3 over UDS" — it is undesigned.**

Across hosts there is no UDS and no k8s DNS. And the topology fork is never answered:

- **One cluster process for the whole deployment** — every coordination op crosses hosts, and it is a single point of failure on one host with no scheduler to restart it.
- **One process per host** — leases become per-host, so locks and elections silently stop being deployment-wide.

Nothing in the document forbids the second. Pick one explicitly and say how it is enforced.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+
+**What violating the constraint looks like** — worth stating, because none of it is loud:
+
+- A `Renew` landing on a non-acquiring replica returns `LockExpired`, so a consumer mid-critical-section is told its lease lapsed while the lock is in fact still held. The lock then lingers to TTL with no live owner.
+- A `Release` landing on a non-acquiring replica returns **success** and deletes nothing.
+- `AwaitChange` against an unknown `election_id` cannot distinguish "wrong replica" from "election gone", so a leader's watch reports loss spuriously.
+
+**Enforcement.** Helm sets `replicaCount: 1` **and** `strategy: Recreate` (§4.10). The strategy is load-bearing: a `Deployment`'s default `RollingUpdate` with `maxSurge: 25%` transiently runs two pods at `replicas: 1`, which is enough to produce every failure above during an ordinary upgrade. `maxSurge: 0` is an equivalent alternative. A startup assertion that the gear is not horizontally scaled is worth adding once a mechanism to observe that exists; until then this is an operator-enforced constraint and belongs in the deployment docs (§10).
```
MikeFalcon77: **Helm cannot enforce this in Profile 2.**

`replicaCount: 1` and `strategy: Recreate` are k8s-only. In Profile 2 the Platform Host spawns worker processes and nothing stops it starting two cluster workers, which reproduces every failure listed above — silently.

The single-instance constraint needs an enforcement mechanism that is not a Helm value: a startup assertion, a lease on a shared backend, or an explicit operator gate.

MikeFalcon77 (commented):
.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+
+**One consequence the constraint does not remove.** Because the beacon dies with its process, a restart of the single replica revokes every Postgres-backed lock held on behalf of *live* consumers immediately, rather than letting them lapse at TTL — the acquire predicate treats an unvouched row as stealable on sight. Consumers learn of it at their next `renew` (`LockExpired`), which DESIGN §3.3's pattern C already establishes as the authoritative loss signal, so this is detectable rather than silent. But it is a shorter and sharper window than the TTL story elsewhere in this document assumes, it applies at every rolling upgrade, and no replica count fixes it. It needs settling with decision 11 rather than by this constraint.
```
MikeFalcon77: **This is a release blocker, not something to settle with decision 11.**

`strategy: Recreate` plus beacon-death-on-restart means **every routine cluster upgrade** revokes every Postgres-held lock across the whole fleet, and closes the coordination window for all consumers. That is a change to the availability contract of every gear, triggered by a deploy rather than a failure.

Deferring it to the multi-replica decision means shipping a platform whose coordination layer has a planned outage per release.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+
+Mechanism 2 is not a defect to fix. It is the cost side of a deliberate trade in the Postgres lock design ([PR #4411](https://github.com/constructorfabric/gears-rust/pull/4411), tier 2 per §2.0): one advisory-lock beacon per *instance* on a dedicated connection, rather than one per held lock, is what removes the per-lock connection pinning that would otherwise cap concurrent held locks at `pool_max_size` fleet-wide (§5.3, §7.2.4). Replica-agnostic lock leases would require a per-lease beacon — that is, a dedicated connection per held lock, exactly what that design removes. **Under the beacon model, connection-count efficiency and replica-agnostic lock leases are mutually exclusive**, and this design takes the former.
```
MikeFalcon77: **A third option is missing, and it is the one the industry picked.**

The trade is framed as connection-count efficiency vs. replica-agnostic leases. There is a third shape: **client-supplied lease tokens fenced in the row**, with no server-side session state at all. etcd lease IDs live in the raft log; Consul session IDs live in Consul's own state; a k8s `Lease` is just `holderIdentity` + `renewTime`. Any replica serves any renew because the store, not a process, owns the lease.

Worth noting that mechanism 1 (in-process session registry) is a design choice, not a consequence of remoting — single-replica follows from holding lease state server-side. Choosing the beacon model locks the platform into a singleton coordination service permanently, so it deserves an evaluated alternative rather than a deferral.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+| **2. Profile 1 for hot consumers** | OAGW links the cluster wiring into its own process and pays only the direct DB round trip. Note this is **not** a correctness compromise: embedding puts the backend *client* in-process while the backend itself stays shared, so cross-replica coordination is unaffected. Note also that this is **deployment composition, not configuration** — the gear links `cluster`, whose `start` registers a `LocalClusterClient`, and local-wins means no remote client is registered alongside it (§4.9.3). **Whole-process**: the gear links the same `cluster` gear and must configure every profile it resolves — see the note below | That gear needs DB credentials and its own pool, reopening §5.3's win for it alone, **and it must link `grpc-hub` and expose a gRPC port it does not otherwise need** (§4.2) |
+| **3. Colocation** — cluster as a DaemonSet, consumers reaching a node-local instance over UDS | Cuts the added hop from ~0.3 ms to ~0.05 ms | Multiplies pools by node count, undoing much of the connection-count win; complicates session affinity (§5.4) and the multi-replica question (§9) |
```
MikeFalcon77: **Node-local is the standard shape for this, not a fallback.**

Consul client agent, Vault Agent, etcd `grpc-proxy` and kubelet are all node-local by design, precisely to avoid a central coordination SPOF and to keep the added hop in microseconds. The industry consistently makes this trade in the opposite direction to the one-line dismissal here.

Given §5.8 forces a single replica anyway, a DaemonSet removes the SPOF and the latency question at the same time. It deserves a real assessment, not "available but not planned".

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+    }
+    fn healthcheck(&self, _ctx: &GearCtx) -> Option<Arc<dyn Healthcheck>> {
+        Some(Arc::new(ClusterReadiness { profiles: self.profiles_weak() }))
+    }
+}
+```
+
+Two points that follow from ADR-0005 and matter here:
+
+- **`Degraded`, not `Unhealthy`, for one bad profile.** Evicting the pod because one DSN is unreachable would take down coordination for *every* profile. The per-component detail lands on `/health`, which is where an operator decides whether to page.
```
MikeFalcon77, line 681: **Right for the pod, but it leaves a consumer-side hole.**

Consumer readiness gates on the `cluster` dep at **gear** granularity — it cannot express "profile `oagw` is unavailable". So a `Degraded` cluster reports `ready: true`, an OAGW consumer resolves, gets `Ok` (validation deferred per §4.7.1), and starts taking traffic against a profile that cannot serve.

Not covered in §9. Either the dep gate needs per-profile granularity, or the SDK's readiness contributor has to fail on a degraded profile it recorded a requirement for.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+- `cluster_backend_pool_saturation` gauge with a WARN on acquire-timeout, so saturation is visible before it is an incident.
+- Per-`ClientId` concurrency caps (§5.7) so one noisy consumer cannot starve the rest.
+
+#### 7.2.5 Other per-operation costs
+
+- **Platform-plane auth is per-request.** An uncached SA-token TokenReview is a call to the API server — tens of milliseconds — which at 10k ops/s would dominate the budget outright and invalidate every figure above. **Requirement**: internal-auth validation must be cached per credential with a bounded TTL, or cluster's gRPC plane must validate once per connection rather than per request. Confirm this against [PR #4403](https://github.com/constructorfabric/gears-rust/pull/4403)'s authenticators (`toolkit-security/src/{authenticator,internal_auth_config,shared_secret}.rs`, `toolkit-transport-grpc/src/sa_token.rs`) — **not** against `InternalAuthMiddleware`, which does not exist in code (§2.0, §4.3). ADR-0006 says TokenReview is cached; that claim needs verifying against the implementation before any of these numbers can be trusted (§9, decision 5).
```
MikeFalcon77: **Per-connection validation is the standard answer — worth stating as the target.**

Per-credential caching with a TTL works, but validating once per connection is what mTLS gives for free and what every high-throughput internal RPC layer relies on. That is an argument for moving the mTLS+SPIFFE phase of ADR-0006 forward rather than treating it as "next phase".

At 10k ops/s a per-request TokenReview is not a tuning problem, it is a design constraint.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+replicaCount: 1                          # correctness constraint, not capacity — §5.8
+strategy: { type: Recreate }             # must not transiently run two pods — §5.8
+grpc: { enabled: true, port: 50051 }     # the sanctioned internal-gRPC shape
+service: { type: ClusterIP, port: 8080 }
+internalAuth: { audience: toolkit-internal }
+```
+
+Note that `default` and `event-broker` name the same DSN — three bindings, and today three independent pools. §5.3 collapses them to one. Also needed: a `NetworkPolicy` restricting :50051 to platform namespaces, and a preStop delay per §4.8.
+
+`replicaCount` and `strategy` are the two values an operator must not override. A `RollingUpdate` at `replicas: 1` runs two pods during every upgrade, which is sufficient to break lease operations (§5.8); `ClusterIP` rather than a headless Service matters for the same reason, since it keeps one client channel pinned to one backend.
```
MikeFalcon77: **Direct conflict with the open proposal in #4426.**

[PR #4426](https://github.com/constructorfabric/gears-rust/pull/4426) §5 requires targeted roles to sit behind a headless Service + StatefulSet and **never** a VIP. This section requires `ClusterIP` and names headless as breaking lease affinity.

#4426 also names the cluster gear as the consumer that must address "a specific backend instance (the shard owner or leader)" — which presumes multiple instances, while §5.8 here mandates exactly one. Two open PRs, one gear, incompatible topologies. Sequence them deliberately.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
```diff
@@ -0,0 +1,3105 @@
...
+| **1b — compound operations** | `increment`, `MultiGet`, and whatever else 0a justifies — on the cache contract and both backends. **Land before the OoP client**, so the hot path is never naively remoted, and so Profile 1 gets the win too (§7.2.3) | 0a, 1 |
+| **2 — runtime profile management** | `ProfileRegistry`, `BackendInstanceCache`, `probe()`, composite readiness, revised pool-sizing guidance + saturation metric (§7.2.4). **Ships value before any OoP work**: the shared-instance dedup fixes real waste in today's in-process gear | — (parallel with 1) |
+| **3 — server** | Five service impls, session registry, `GrpcServiceCapability`, `RestApiCapability` + admin routes, capabilities `+grpc,rest,system`, revoke fan-out | 1, 2 |
+| **4 — client** | `ClusterClient` trait, `RemoteClusterClient` + four `Remote*Backend`s, the `ConsumerRegistration` + profile inventory, lazy-binding facades, transient/permanent readiness classification | 1 |
+| **5 — deployable** | `main.rs`, `registered_gears.rs`, `[[bin]]`, image, Helm values, NetworkPolicy, preStop, migration subcommand + Job (§4.10.1), OoP smoke test | 3, 4 |
```
MikeFalcon77: **Say plainly that the first deployable version is k8s-only.**

Profile 2 is deferred by decision 12, `ApiContractsConfig.remote_grpc_endpoints` does not exist, and the shipped wiring path is directory + REST. So there is currently no working endpoint-resolution mechanism for Profile 2 at all — including single-host, which is the on-prem baseline.

Phase 5 should state that scope limit rather than leaving it to be inferred from a non-blocking decision.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
asmith987: I believe `run_proxy_wiring_phase` is in #4084 branch rather than main. For the initial phase since the PR is currently not merged, I have noted this in the design. To prevent blocking, there will be a two-tier approach where anything missing for cluster will simply be manually done and migrate as soon as #4084 is closed.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
asmith987: Section rewritten to retire the liveness beacon which resolves the blocker outright. Essentially, a processes death shouldn't lead to lock revocation. This means the per-release coordination outage disappears.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
asmith987: Explicitly marked it as undesigned for the time being. I will touch back on this in the coming days.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
asmith987: Moved to store-owned leases to allow any process to serve a lease. This makes replica count a capacity decision in every profile. This shouldn't be an issue anymore

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
asmith987: Decided to adopt this approach. Section rewritten around store-owned leases

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: You're right and I was wrong — I reviewed against #4084's branch, not `main`. Verified: `run_proxy_wiring_phase` is absent from `origin/main` and `resolve_one_dep` / `ResolvedRestEndpoints` are still there. The tier-2 treatment in the rewrite is the correct model. Resolved.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved, and by a better route than I asked for. Having `resolve()` self-construct the client and prefetch descriptors removes the dependency on phase order entirely, instead of chasing the framework to reorder. §4.7.1 + the note at the end of §4.9.3 make the two paths clear.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Withdrawn. Once `resolve()` no longer waits on the phase, the ordering genuinely isn't blocking, so leaving 16 as raise-not-block is right. The two extra questions you folded into it (is the post-`start` order deliberate, is a gRPC counterpart planned) are the useful part — answered in a separate comment.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved, and you found the part I missed: the gate sits earlier than the wiring phase. `run_oop_with_options` connects eagerly and propagates the error, so today no OoP gear starts without a directory at all — `bootstrap/oop.rs:542`, confirmed on `main`. Decision 21 states it correctly.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved. Marking P2 out of scope and splitting decision 12 into transport / endpoint resolution / topology is clearer than the fork I described. The backend-scope reframing is the right one: after store-owned leases, process count stops being the correctness question and shared backend config becomes it.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved — the constraint no longer exists, so there is nothing to enforce. Store-owned leases make this a capacity decision in every profile.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved at the root. Retiring the beacon removes the per-release coordination outage entirely, and the doc states the cost plainly (TTL-bounded reclaim in every profile, where embedded used to reclaim in milliseconds). Worth flagging that checklist 10c changes already-merged code from #4411 — the doc says so, just make sure that author is looped in.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved — adopted as ADR-012, with both rejected alternatives recorded. The fence-guarantee wording (no reuse within `fence_retention`, not global monotonicity) is a good addition and not something I raised.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved. The assessment is the right shape: nothing structural blocks it now that a lease is a row, the real cost is the watch fan-out collapse weakening, and gating on the failover suite is the correct sequencing.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved, and this is the most thorough fix in the batch. Per-profile health in the descriptor, a readiness gate scoped to profiles the process actually recorded, explicitly reversible `Unhealthy` with a 10s poll, and the `probe()` extension all close the hole end to end.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved and strengthened. I re-checked the caching claim: `K8sTokenReviewAuthenticator::authenticate()` is `self.review(token).await` with no cache anywhere in `toolkit-k8s-auth`, and `DynInternalAuthenticator` is only wired into `oop_serve.rs`, so there is no inbound gRPC path on `main`, in #4403, or in #4084. Decision 5's two facts hold.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Addressed — the §4.10 cross-reference, retiring cluster's SD, and the §10 edits owed to #4426 cover it. One leftover from the store-owned-lease rewrite: the replica claims have drifted apart. ADR-012 says "the single-replica constraint disappears", but §4.10 still reads "a single-replica cluster is an entrypoint role", §6.7 says "the cluster gear runs a single replica (§5.8)", and the §10 row for #4426 says "§5.8 mandates a single replica". That last one matters most — it hands #4426's author a premise this document no longer holds. The `ClusterIP`-over-headless conclusion may well survive, but it needs a reason other than single-replica.

gears/system/cluster/docs/DESIGN-DEPLOYABLE-GEAR.md
MikeFalcon77: Resolved. §3.2's table plus decision 12 state the scope limit directly, including that single-host P2 has no working endpoint resolution today.

MikeFalcon77:
Answering the three §9 decisions that point at #4084, since they're mine.

## Decision 21 — endpoint resolution with no DirectoryService

There are two gates here with different owners, and only the second is mine.

**The one that actually blocks you is not in #4084.** `run_oop_with_options` connects to the directory eagerly and propagates the failure (`bootstrap/oop.rs:542`), so no OoP gear starts without one — that predates #4084 and is on `main` today. A DNS fallback in the wiring phase is unreachable until that is lifted, so the two have to move together and the bootstrap half needs the OoP runtime owner, not me.

**The wiring-phase behaviour is deliberate and I'd rather not weaken it.** With no `DirectoryClient` the phase installs a `NullEndpointResolver` so remote deps register as unresolved readiness gates and `/readyz` stays 503. Silently skipping wiring would let a misconfigured consumer report Ready, which is the worse failure.

**There is already a working answer today, with no new code.** The ADR-0004 static override — `gears.<owner>.config.consumer_wiring.<dep> = "http://host:port"` (`StaticEndpointResolver`, `discovery.rs:118`) — lets a directory-less k8s deployment pin endpoints from Helm values. Worth naming in §4.5 as the current mechanism rather than leaving the section describing only the intended state.

**On "it is a `format!`":** the name is, the port isn't. The directory returns a full base URI; DNS returns a name only. A `DnsEndpointResolver` therefore needs a port convention that does not exist — per-dep config, a platform-wide default, or reading it back from the directory, which defeats the purpose. The resolver itself is cheap (`EndpointResolver` is already a trait with three impls) and I'm happy to add it, but the port convention is the actual decision and it belongs in the platform PRD, not in #4084.

## Decision 16 — is the post-`start` ordering deliberate, and is a gRPC counterpart planned?

**Not deliberate.** The two paths genuinely differ: in-process runs init → proxy-wiring → post-init → rest → grpc → start (`host_runtime.rs:1273-1289`), OoP runs init → post-init → grpc → start → proxy-wiring (`:1441-1451`). The asymmetry is incidental rather than required — the phase body only needs a populated `ClientHub` and a known gear registry, neither of which depends on `start`. I'll align OoP with the in-process order so both wire before `start`. That makes your inline-validation path the normal case rather than the lucky one, though I agree §4.7.1 is right not to depend on it.

**No gRPC counterpart in #4084**, and the feature name admits it: `contract-directory-rest-client`. `#[toolkit::consumes]` generates a directory-resolving REST client only. `ConsumerRegistration` itself is transport-agnostic — owner gear, dep, wire closure — so cluster's hand-written gRPC registration rides the same inventory and replay path and will not need rework when a generated gRPC client lands. I'd rather cluster prove the shape in practice first and then lift it into the macro than guess at the codegen now, so §6.1's "hand-rolled, migrated later" is the sequencing I'd pick too.

## Decision 5 — server-side gRPC internal-auth interceptor

**Both facts confirmed, and neither is #4084's to fix.** I checked all three trees: there is no inbound gRPC auth path on `main`, in #4403, or in #4084. `DynInternalAuthenticator` is wired only in `oop_serve.rs` (the HTTP path), and `toolkit-transport-grpc` ships an outbound `InternalAuthInterceptor` plus token extraction. #4084 is contract codegen and does not own the inbound platform-plane path — the right owner is the OoP runtime / ADR-0006 two-plane work, the same area as #4403.

The gate as written is right (interceptor by phase 5, otherwise the coordination port ships behind a NetworkPolicy only). One addition: even once it exists, an uncached per-request TokenReview is unusable at cluster's rate — I verified `K8sTokenReviewAuthenticator::authenticate()` is a bare `self.review(token).await` with no cache in the crate. So the interceptor and pulling ADR-0006's mTLS phase forward are one work item, not two, and cluster is the forcing function for both halves.

striped-zebra-dev (approved):

PR #4426, opened 2026-08-05
docs(toolkit/oop): instance addressable discovery

mattgarmon:
Introduces new design for instance addressable service discovery in toolkit.

References work in other open PRs:
- https://github.com/constructorfabric/gears-rust/pull/4490
- https://github.com/constructorfabric/gears-rust/pull/4403
- https://github.com/constructorfabric/gears-rust/pull/4084

## Documentation

* Documented accepted instance-addressable discovery for targeting service instances by role, shard, or stable labels.
* Clarified that round-robin routing remains the default when no specific instance is requested.
* Added guidance for readiness, routing, deployment, endpoints, stale registrations, and topology refresh.
* Updated architecture, deployment-profile, Helm, product-requirement, and service-decomposition documentation to support role-qualified and sharded workloads.


docs/arch/toolkit-oop/PRD.md
```diff
@@ -142,6 +142,7 @@ application code remains deployment-agnostic. ToolKit Distributed Gears bring th
 - Deployment Profile 3: K8s Native (each gear = pod, external gateway, k8s DNS discovery)
 - Two-plane auth — gateway validates the JWT at the edge; OoP gears **re-validate it per hop** (tenant plane) and use a tenant-less identity for infra calls (platform plane). See [ADR-0008](ADR/0008-cpt-cf-adr-two-plane-auth.md)
 - Direct gear-to-gear REST communication (not routed through gateway)
+- Instance-addressable discovery — target a specific gear instance (role/shard) via additive directory metadata + targeted resolve, with round-robin remaining the default. Targeted selection does **not** expand the Toolkit's load-balancing scope beyond round-robin: the directory only enumerates/pinpoints instances; any pick-one over a matched set (e.g. consistent-hash sharding) is application-owned policy in the consuming gear. See [ADR-0009](ADR/0009-cpt-cf-adr-instance-addressable-discovery.md)
```
striped-zebra-dev, line 145: The relative link is broken, this adr belongs to a different folder (ADR/0009-cpt-cf-adr-instance-addressable-discovery.md)

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+### 1. Modeling rule: two sanctioned shapes
+
+A gear name may map to deployments in one of two shapes, and the platform supports **both**:
+
+**Shape 1 — separate gears (one contract each).** If deployments have genuinely independent contracts and each
+should be discoverable by its own name, model them as separate gears (Option A). Name-based discovery,
+generated clients, readiness, and the edge work unchanged.
+
+**Shape 2 — one gear, multiple role deployments (chosen for `event-broker`).** A single gear name MAY host
+multiple deployments that play different **roles** (`dispatcher`, `ingest`, `delivery`) under that one name.
```
striped-zebra-dev: I'd avoid designing the model around a single use case. We should prefer established industry patterns that remain consistent across services.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+* An **`entrypoint` flag** (**presence-tracked** `optional bool` — three states: *unset*, `true`, `false`) —
+  the **sole directory-interpreted signal**: it marks whether an instance is returned by **bare-name
+  resolution**. It MUST be wire-optional (proto3 `optional`) so a legacy registration that omits the field
+  (*unset*) is distinguishable from an explicit `false`; this distinction is what keeps mixed-version rollouts
+  safe (§3). It is a dedicated typed field (not a reserved label key) precisely to keep the `labels` map purely
+  consumer-owned; edge exposure is governed separately by route visibility (§3).
+
+Add **targeted lookups** to `DirectoryClient`, backed by selection in `GearManager` and the local/gRPC clients:
+
+* `resolve_instance(gear, id)` — pinpoint a single instance.
```
striped-zebra-dev: In this case `id` - instance id?

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+* `resolve_by_label(gear, key, value)` — the set of instances matching a label.
```
striped-zebra-dev: I believe the most common way would be a list of key value.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+
+**Stable identity (decision):** durable targeting MUST key on a **stable label** — a stable `shard` / `role`, or
+a deployment-provided stable per-pod name **registered as a stable label** (e.g. `pod=ingest-0`) — resolved via
+`resolve_by_label`. It **never** keys on the ephemeral `instance_id`, which is minted fresh per process/pod and
+does not survive a restart. There is deliberately **no** stable-name lookup distinct from label resolution: a
+deployment-provided stable name is durable only when the instance re-registers it as a label, so folding it into
+the `labels` axis keeps a single durable-addressing path (`resolve_by_label`) rather than a second field + API.
+`resolve_instance(gear, id)` is therefore **restricted to ephemeral `instance_id` values** and is only valid
+within a single instance lifetime (e.g. a caller already holding a live id); it MUST NOT be used for durable
+targeting. A shard→instance (or stable-name→instance) map MUST be built from stable labels via `resolve_by_label`
```
striped-zebra-dev: `shard→instance` better use `->` in documents.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+and refreshed on membership change.
+
+### 3. Entrypoint resolution semantics
+
+Because a gear may now host multiple roles under one name, **name-based resolution is redefined** to select the
+gear's front door rather than any instance:
+
+* `resolve_rest_service(gear)` / `resolve_grpc_service(...)` round-robin over **eligible** instances (see
```
striped-zebra-dev: `resolve_rest_service` vs `resolve_rest_entrypoint`
I'd prefer consistency with the `entrypoint` terminology used across all documentation and code.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+  eligibility computation below). For `event-broker`, only `dispatcher` instances set the flag, so other gears
+  resolving `event-broker` by name reach the dispatcher and never an internal `ingest` / `delivery` pod.
+  (`resolve_grpc_service` selects by gRPC service name across gears rather than per-gear; entrypoint filtering
+  applies within that candidate set — see Known implementation constraints.)
+* **Eligibility computation (decision):** an instance is an eligible entrypoint if its `entrypoint` is
+  `true` **or** *unset*; only an explicit `false` excludes it. Back-compat for existing single-role gears comes
```
striped-zebra-dev: Do we really need backward compatibility here? I still consider this project to be a greenfield one.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+* **Edge-eligible predicate (decision):** the edge and `get_openapi_spec(gear)` share **one** predicate. An
+  instance is **edge-eligible** iff (a) it is an eligible entrypoint (`entrypoint` `true` **or** *unset*, per
+  the eligibility computation above) **and** (b) it exposes **public** routes (via the `.exposed()` /
+  `x-cf-api-visibility` visibility split #4403 introduces). Clause (a) excludes explicit-`false` internal
+  roles; clause (b) excludes any instance whose routes are all internal-by-default. The `api-gateway`
+  reverse-proxy edge (PR #4403) exposes only the **public** routes of edge-eligible instances, so internal
+  roles never enter the edge's public route table; `entrypoint` additionally keeps them out of the edge's
+  name-keyed single-endpoint selection.
+* **One public contract per gear name (constraint):** all edge-eligible instances of a gear MUST share the
+  same public API. Endpoint selection and OpenAPI generation operate over the **same** edge-eligible set: the
```
striped-zebra-dev: >MUST share the same public API

This seems to contradict the `mixed-version rollout` described above. During a rolling deployment, old and new versions may legitimately expose different API versions. I'd leave backward compatibility to the gear developer rather than enforce API equality at the platform level.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+  shared spec — so the advertised spec and the routed endpoint can never diverge. Role-differentiated *public*
+  APIs under one name are therefore not supported; internal role APIs are private and unconstrained.
+  **Backward-compatibility (all-*unset* legacy gear):** a single-role gear whose instances all leave
+  `entrypoint` *unset* has its **entire** fleet edge-eligible (clause (a) holds for *unset*, and their normal
+  routes are public), so edge exposure and `get_openapi_spec` behave exactly as they do today.
+
+### 4. Mixed consumers: round-robin and targeting against one provider
+
+Different consumers of the **same** provider gear may resolve it differently, with no provider-side
+coordination and no conflict — round-robin and targeting are simply two selectors over the same instance set:
```
striped-zebra-dev: —
Let's use regular hyphens (-) consistently across all documentation.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+
+**ID**: `cpt-cf-adr-instance-addressable-discovery`
+
+## Table of Contents
+
+<!-- generated by `cypilot toc` -->
+
+## Context and Problem Statement
+
+The service directory's unit of discovery is the **gear name**. Everything layered above it keys on that
```
striped-zebra-dev, line 22: This paragraph frames the problem entirely in gear-specific terms (gear name, `DirectoryClient`, PR #4403/#4084) before establishing that it's actually a well-known, industry-general problem: name-based service discovery collapsing to a single endpoint/VIP vs. needing to address a specific instance for partitioned or stateful workloads.

Consider opening with that framing before narrowing into the gear-specific mechanics - it'll help a reader unfamiliar with "gears" terminology see this isn't a bespoke local design problem, and it sets up Option C/D later (Kubernetes-native, service mesh) as the natural resolution of an already-recognized problem rather than options that appear out of nowhere.

```suggestion
Name-based service discovery that collapses a logical service to one endpoint - and can't address a specific instance - is a well-known limitation, not a gap unique to this platform. Kubernetes handles it by pairing headless Services with StatefulSets: a stable per-pod DNS name lets a caller reach one specific pod, while the plain Service name still load-balances across all of them. Consul and mesh proxies like Envoy solve the same need with a service catalog that carries instance-level tags/metadata and routing rules, so a call can be pinned to a matching instance instead of round-robined over the whole set. Kafka and Cassandra hit the same wall from the partition-ownership side: a client or coordinator needs to reach the specific broker/node that owns a partition, not any member of the cluster.

The service directory's unit of discovery is the **gear name**. Everything layered above it keys on that
```

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+Lookups are split by cardinality rather than overloading one method:
+
+* **`resolve_instance(gear, id)` → at most one** (`Option` / not-found). Pinpoint addressing.
+* **`resolve_by_label(gear, key, value)` → a set (array)** — effectively `list_instances(gear)` filtered by
+  metadata, returning each match's `instance_id`, `labels`, `entrypoint`, readiness state, and endpoints (gRPC
+  `endpoint` and/or `rest_endpoint`). This is the enumeration form the broker dispatcher uses to build/refresh a
+  partition→instance map, and that fan-out/broadcast consumers use to reach every match. The API filters a
+  **single** `key`/`value`; matching on multiple labels (e.g. `role` **and** `shard`) is done by the caller
+  intersecting results client-side — a multi-label / expression selector is intentionally out of scope for
+  Layer 1.
```
striped-zebra-dev: Layer 2 extension point:

```rust
struct LabelSelector {
    matches: HashMap<String, String>,
    only_serving: bool,  // default false: matches current full-set behavior
}

fn resolve_by_label(gear: &str, selector: LabelSelector) -> Result<Vec<InstanceInfo>>;
```

A struct here, not a bare `HashMap`, keeps future additions (multi-label match, health filtering) additive fields instead of signature changes.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+
+### 5. Routing mechanism: endpoints, not IPs
+
+Routing to a selected instance reuses the endpoint the directory **already stores** — there is no separate
+IP-tracking subsystem. Each instance advertises its reachable URI at registration (the OoP worker's
+`advertise_uri`, registered as `rest_endpoint`; gRPC services carry their own endpoints), stored on
+`ServiceInstanceInfo` (`endpoint` / `rest_endpoint`). Targeting therefore reduces to: resolve `(gear, selector)`
+→ the instance's advertised URI → dial it with a plain HTTP/gRPC client.
+
+The correctness constraint is not tracking IPs but **what the advertised URI points at**, which differs by
```
striped-zebra-dev, line 255: In Kubernetes, determining the address an instance should advertise isn't always straightforward, and the same may be true for other platforms. Perhaps OOP should define the recommended mechanism for determining and advertising the instance's reachable URI.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,441 @@
...
+    does not yet fall back to unhealthy — is tracked in Known implementation constraints.)*
+  * **Targeted `resolve_instance(gear, id)`** returns the instance **regardless of health** — no fallback and no
+    failover — carrying its readiness state; a down target is a matched-but-not-ready result, distinct from
+    `None` (no such id). Failover is the caller's responsibility (§4).
+  * **`resolve_by_label(gear, key, value)`** returns the **full** matching set **regardless of health**, each
+    entry carrying its readiness state for the caller to filter/select on. Health never drops a match, so an
+    all-unhealthy match set is still `Ok(non-empty)` — kept distinct from `Ok(empty)` (no match) and from a
+    backend error.
+
+### 7. Layering relative to PR #4084
```
striped-zebra-dev: It is top-level document, I don't think it can contain references to any PRs.

mattgarmon: Agreed, fixed

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Good catch, I've left the motivation, but removed specifics from design

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Updated to `instance_id` for clarity

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Agreed, updated to be `resolve_by_labels`, which is analogous to k8s `matchLabels`

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: `entrypoint` is concept introduced by this design, the `resolve_rest_service` is already existing method in DirectoryService.

For most use cases, `service` does make more sense. The method returns a `ServiceEndpoint`, and it matches the "service discovery" language used throughout the codebase. 

I've updated the wording to mention that this resolves the `entrypoint` subset.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Removed backward compatibility details

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: The mixed-version rollout section was removed (greenfield), so that contradiction is gone

I've rescoped the constraint to forbid only role-differentiated public APIs, explicitly leaving cross-version compatibility during rolling deploys to the gear developer.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added LabelSelector, allowing additive additions in the future

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added section recommending mechanism

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,496 @@
...
+  governed separately by route visibility (§3).
+  * ***unset* and explicit `true` are behaviorally identical** - both eligible. Both exist because keeping a
+    **positive** field name (`entrypoint`) means "absent" must be its own state to fail safe on omission; a
+    two-state design would require inverting to a negative field (e.g. `internal`, default-eligible), which we
+    reject for readability. Explicit `true` therefore adds no behavior over *unset* - it only lets an instance
+    assert front-door intent explicitly.
+  * **The safe direction is chosen for single-role gears.** The flip side: for a multi-role gear a *forgotten*
+    `false` leaves an internal role *unset* -> eligible, so it silently joins **in-mesh** `resolve_rest_service`
+    RR. The **edge** still excludes it (edge-eligibility also requires public routes, §3), but in-mesh name
+    resolution does not - this is the accepted trade for never silently dropping a single-role instance.
```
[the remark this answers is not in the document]
mattgarmon: No code implementation yet

striped-zebra-dev (approved):

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+endpoint** (first healthy instance / k8s Service VIP); the caller cannot ask for a *specific* instance.
+
+Two upcoming designs need to route to a *specific* instance rather than "any instance of a gear":
+
+* **Event broker** - `event-broker` is modeled as a **single gear** with multiple **role deployments**
+  (`dispatcher`, `ingest`, `delivery`) registered under that one name. `dispatcher` is the entrypoint
+  other gears reach by name; internally it must route to specific `ingest` / `delivery` instances (partitioned
+  work), not a random one.
+* **Cluster gear** - the cluster coordinator must address a *specific* backend instance (e.g. the shard owner
+  or leader), where all backends share one contract.
```
MikeFalcon77: **blocker — duplicates the shipped cluster gear**

The cluster gear already ships this capability, and in a richer form. In `main` today: `ServiceRegistration.metadata: HashMap<String,String>` (`gears/system/cluster/cluster-sdk/src/discovery/types.rs:44-57`), `DiscoveryFilter` with AND-conjoined per-key predicates and `MetaMatch::{Equals, OneOf}` (`types.rs:75-150`), `InstanceState::{Enabled, Disabled}` as declared serving intent, `ServiceDiscoveryBackend::watch` (`discovery/backend.rs:117`), TTL heartbeat, plus `standalone` and `postgres` backends and a conformance suite.

So this ADR proposes a second, weaker discovery layer (equality-only, no watch) next to a merged one. There may be a valid layering reason — `directory` is toolkit-level, `cluster` is a gear, and the toolkit cannot depend on a gear — but the ADR never states it.

Please add a section answering: why not `cluster-sdk::ServiceDiscoveryV1`, what the boundary between the two layers is, and how they avoid becoming two competing sources of topology truth.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+once REST contract codegen lands - the generated typed contract clients (`<Trait>RestResolvingClient`,
+wired through an `EndpointResolver` bound to a single gear name). Resolution deliberately **collapses a gear to a single
+endpoint** (first healthy instance / k8s Service VIP); the caller cannot ask for a *specific* instance.
+
+Two upcoming designs need to route to a *specific* instance rather than "any instance of a gear":
+
+* **Event broker** - `event-broker` is modeled as a **single gear** with multiple **role deployments**
+  (`dispatcher`, `ingest`, `delivery`) registered under that one name. `dispatcher` is the entrypoint
+  other gears reach by name; internally it must route to specific `ingest` / `delivery` instances (partitioned
+  work), not a random one.
```
MikeFalcon77: **major — the motivating case may already be solved**

Two problems with using event-broker as the driver.

First, `gears/system/event-broker/docs/ADR/0007-service-decomposition.md` (Accepted) already decided that `domain/cluster.rs` resolves `ServiceDiscoveryV1` via `ClientHub` under the `evbk` prefix. The broker has a working targeting primitive today, so §7's claim that Layer 1 "unblocks" it needs revisiting.

Second, that ADR decided **one binary, mode-selected wiring** — not one *directory name*. Nothing stops `mode=cluster_ingest` from registering as `event-broker-ingest`. If it does, the `entrypoint` marker is unnecessary for this case, and the wrong-role failure mode becomes structurally impossible instead of flag-dependent.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+  governed separately by route visibility (§3).
+  * ***unset* and explicit `true` are behaviorally identical** - both eligible. Both exist because keeping a
+    **positive** field name (`entrypoint`) means "absent" must be its own state to fail safe on omission; a
+    two-state design would require inverting to a negative field (e.g. `internal`, default-eligible), which we
+    reject for readability. Explicit `true` therefore adds no behavior over *unset* - it only lets an instance
+    assert front-door intent explicitly.
+  * **The safe direction is chosen for single-role gears.** The flip side: for a multi-role gear a *forgotten*
+    `false` leaves an internal role *unset* -> eligible, so it silently joins **in-mesh** `resolve_rest_service`
+    RR. The **edge** still excludes it (edge-eligibility also requires public routes, §3), but in-mesh name
+    resolution does not - this is the accepted trade for never silently dropping a single-role instance.
```
MikeFalcon77: **blocker — this fails open on the exact case the ADR exists to prevent**

This contradicts the first decision driver ("a role-split or sharded gear must never silently route to the wrong role/shard"). The accepted trade is the failure mode the ADR was written to close, and it fails open on a per-instance config field — the easiest thing to get wrong.

The dilemma is self-inflicted because `entrypoint` is declared per *instance*. Declare the entrypoint role once, at gear/manifest level ("`role=dispatcher` is the front door for `event-broker`"), and a forgotten per-instance flag can no longer misroute: an unlabeled instance of a gear that declared an entrypoint role is simply not that role. The tri-state bool and the whole *unset*-vs-`true` discussion above (L133-137) then disappear.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+**Stable identity (decision):** durable targeting MUST key on a **stable label** - a stable `shard` / `role`, or
+a deployment-provided stable per-pod name **registered as a stable label** (e.g. `pod=ingest-0`) - resolved via
+`resolve_by_labels`. It **never** keys on the ephemeral `instance_id`, which is minted fresh per process/pod and
+does not survive a restart. There is deliberately **no** stable-name lookup distinct from label resolution: a
+deployment-provided stable name is durable only when the instance re-registers it as a label, so folding it into
+the `labels` axis keeps a single durable-addressing path (`resolve_by_labels`) rather than a second field + API.
+`resolve_instance(gear, instance_id)` is therefore **restricted to ephemeral `instance_id` values** and is only valid
+within a single instance lifetime (e.g. a caller already holding a live `instance_id`); it MUST NOT be used for durable
+targeting. A shard->instance (or stable-name->instance) map MUST be built from stable labels via `resolve_by_labels`
+and refreshed on membership change.
```
MikeFalcon77: **minor — `resolve_instance` has no valid caller; drop it**

If it must not be used for durable targeting, it has no callers. The only way to obtain a live `instance_id` is `resolve_by_labels`, which already returns the endpoint.

Dropping it makes the contract change smaller: one new RPC instead of two, and one less method to reconcile with the REST-codegen directory edits.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+deployment profile (ADR-0001):
+
+* **Profile 1 (Embedded)** - all gears share one process and call in-process via `ClientHub`; there is exactly
+  **one** instance per gear name, so targeting is degenerate. The lookups still **apply their selector** to that
+  lone instance, consistent with the uniform semantics in §6: `resolve_instance` returns it only when
+  `instance_id` matches (else `None`), and `resolve_by_labels` returns it only when it carries **every**
+  requested label (an empty selector matches; a non-match yields `Ok(empty)`). A match dispatches through
+  `ClientHub`, not a dialed URI. The multi-instance role/shard model (Shape 2, §1) is **not expressible** in
+  Embedded - it requires an OoP profile - so per-instance targeting is degenerate here (at most one candidate),
+  not a limitation.
```
MikeFalcon77: **major — this is a test-matrix hole, state it as one**

"not a limitation" understates it. The whole justification of ADR-0001 is a small, fully-covered matrix.

Consequence to state explicitly: the broker's cluster mode, and every gear adopting Shape 2, can only be verified in Profile 2/3. Say who owns that coverage.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+  instance is a no-op (the URI routes back through the Service load balancer). Instance targeting therefore
+  requires each targeted instance to advertise a **per-instance-addressable** endpoint: a `StatefulSet` +
+  **headless Service** per-pod DNS name (e.g. `ingest-0.ingest.<ns>.svc`) or the pod IP. This is a
+  deployment/registration concern; the directory carries whatever URI is advertised.
+  Note that a **headless** Service is DNS-only - it does **not** load-balance - and **discovery is the
+  directory's job, not k8s Services'**: cross-gear resolution goes through `DirectoryClient`, so k8s primitives
+  exist only to make advertised URIs *reachable*. Concretely, per role: an **entrypoint** role may sit behind a
+  plain VIP Service (or advertise per-pod DNS and let the toolkit round-robin), while **internal/targeted** roles
+  use a headless Service + `StatefulSet` and are **never** fronted by a VIP (which would defeat targeting). There
+  is no single Service in front of the whole gear - the directory unifies its roles under one name.
```
MikeFalcon77: **blocker — reverses an accepted Profile-3 requirement**

`docs/arch/toolkit-oop/PRD.md:273-275` (`cpt-cf-fr-k8s-native`, p1): "DirectoryService MAY be used alongside k8s DNS but **MUST NOT be required** for basic service resolution in k8s." `DESIGN.md:1344`: "k8s DNS provides service discovery; DirectoryService is optional for metadata." ADR-0001 Profile 3 says the same.

Also, "there is no single Service in front of the whole gear" breaks `DESIGN.md:1343` ("Each gear is an independent k8s Deployment + Service"). One gear name now maps to several Deployments/StatefulSets and several Services.

Both may be the right calls, but they are amendments to accepted decisions and need to be written as such — not as a paragraph in §5.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+
+**Determining the advertised URI (recommendation).** `advertise_uri` MUST resolve to an address reachable by
+*other* gears - never a bind-only address (`localhost` / `0.0.0.0`), nor a shared Service VIP for a targeted
+role. Profile 2 uses the worker's configured UDS path or bound `host:port`; Profile 3 derives from the
+Kubernetes **downward API** - the stable per-pod FQDN `$(HOSTNAME).<service>.$(POD_NAMESPACE).svc` (+ container
+port) for targeted roles, or the Service DNS name for a load-balanced entrypoint. **Prefer stable DNS names /
+socket paths over raw IPs** (pod IPs churn on reschedule), so the advertised URI doubles as the durable identity
+and self-heals across restarts; consumers MUST NOT cache IPs but re-resolve through the directory (kept fresh by
+the registration/heartbeat loop). Exact wiring (downward-API keys, listener config) is owned by ADR-0001.
```
MikeFalcon77: **major — the framework's own default violates this MUST, silently**

`libs/toolkit/src/bootstrap/oop.rs:714-722` rewrites unspecified bind addresses to `127.0.0.1` / `[::1]` and registers that. In multi-host Profile 2 and in Profile 3 you get a registered-but-unreachable instance with no error.

A "recommendation" is not enough for a MUST that the default breaks. Make it a requirement on the toolkit: in OoP multi-host and k8s, refuse to start when `advertise_uri` is unset or loopback.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+  RPC request message is already additively extensible, but the Rust trait signature is not unless the argument
+  is a struct. This is the enumeration form the broker dispatcher uses to build/refresh a partition->instance
+  map, and that fan-out/broadcast consumers use to reach every match. Set-based / inequality **expression**
+  selectors (`in`, `notin`, `!=`) are out of scope for Layer 1 (equality-AND only); **health filtering stays
+  caller-side** (health contract below), so the selector carries no `only_serving`-style flag.
+* **"pick one from the matched set"** is an **application-owned selector** over an already-targeted label set
+  (first / consistent-hash / caller policy) - **not** toolkit load balancing. The toolkit's only load-balancing
+  primitive is round-robin over name/entrypoint resolution (§3-§4); `resolve_by_labels` merely returns the set,
+  and any pick-one over it (including consistent-hash for sharding) is domain policy **owned by the consuming
+  gear** (consistent with §2), not a directory-provided balancer.
```
MikeFalcon77, line 339: **major — correct boundary, expensive delivery plan**

Every consumer now writes membership polling, a refresh trigger, and a hash ring. The broker dispatcher and the cluster coordinator — the two named consumers — will write it twice, differently.

Envoy and gRPC keep subset selection in app-supplied metadata but keep `ring_hash`/`maglev` in the client library, precisely so this isn't duplicated. Ship a small `TopologyView` helper (label-keyed set + refresh + consistent hash) in the toolkit. It doesn't move policy into the framework; it stops policy from being reimplemented per gear.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+* **Empty-set semantics**: zero matches is `Ok(empty)` (or `Ok(None)` for the single form), kept distinct from a
+  directory-backend error - preserving the not-found vs. failure contract of ADR-0005, so an empty result reads
+  as "not ready yet," not "outage."
+* **Health contract (decision):** every lookup annotates each returned instance with its readiness state
+  (`InstanceState`: `Registered` / `Ready` / `Healthy` / `Quarantined` / `Draining`; only `Ready` and `Healthy`
+  are **serving** - `Registered` (not yet ready), `Quarantined` (failing health), and `Draining` (shutting down)
+  all count as **not-ready** and are excluded from RR's *preferred* set). This lets a caller always distinguish
+  **no match** (`Ok(empty)` / `Ok(None)`) from **matched-but-not-ready** (a returned instance that is not
+  serving). The three modes share these error semantics and differ only in health *filtering*:
```
MikeFalcon77: **major — `InstanceState` collides with the cluster gear's, with opposite meaning**

`gears/system/cluster/docs/ADR/008-service-discovery-state-is-intent-not-health.md` defines `InstanceState` as *module-declared intent, not a health observation*, with two variants. This ADR's `InstanceState` is heartbeat-derived health with five. A gear consuming both discovery paths gets two same-named enums meaning opposite things.

On top of that, `entrypoint` is a third axis in the same space — it is serving intent, which is exactly what cluster's `Enabled`/`Disabled` expresses. Pick one vocabulary across the platform, or rename here and say why the split exists.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+
+* **Additive contract change** (§2): the registration payload / proto gain a `labels` map and an `entrypoint`
+  **`optional bool`**, and `DirectoryClient` gains `resolve_instance` / `resolve_by_labels`. Name-based
+  resolution and the edge are narrowed to eligible entrypoints (`entrypoint` *unset* or `true`); a single-role
+  gear leaves the field *unset* and resolves over its whole fleet (§3). The field is wire-optional so **absence
+  maps to the safe default (eligible)** - an omitted field never silently drops an instance from name
+  resolution. Because the project is greenfield this three-state carries **no** back-compat / mixed-version
+  obligation; it exists solely for that safe-omission property.
+* **Consumer responsibilities**: gears own selection policy, poll `list_instances` to track membership (no
+  watch/push API exists), and MUST key targeting on **stable labels**, not `instance_id` (§2).
```
MikeFalcon77: **blocker — no correction path for a stale ownership map**

Poll-only plus app-owned shard maps, with no protocol for a stale map, means silent misrouting to a non-owner. That is the one failure mode partition-ownership systems always close: Kafka returns `NOT_LEADER_OR_FOLLOWER` and the client refreshes metadata; Akka/Orleans fence with a coordinator. The ADR names Kafka and Cassandra in its opening paragraph but adopts only the client-side map, not the correction path.

Add a required contract: a targeted call to an instance that no longer owns the requested shard MUST return a typed "not the owner" response that forces a refresh (optionally carrying the current owner). Without it, "selection ownership stays in the gear" also means "correctness is unspecified."

Note that `ServiceDiscoveryBackend::watch` already exists in `cluster-sdk`.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+  (poll -> reconcile), so `labels` / `entrypoint` become reusable there once carried on `InstanceInfo`.
+* Neutral, because it requires an additive change to a public contract (proto + SDK), reconciled with REST
+  contract codegen.
+* Bad, because it redefines name-based resolution semantics (entrypoint filtering) and adds consumer-side
+  complexity (membership polling, stable-key targeting).
+
+### Option C: Kubernetes-native addressing (headless Service / StatefulSet)
+
+* Good, because it leans on battle-tested platform primitives for stable per-pod identity.
+* Bad, because it only works in k8s (Profile 3) - Profile 1/2 and non-k8s deployments get nothing.
```
MikeFalcon77, line 525: **minor — this con is overstated, and it hides Option B's real driver**

§5 (L247-257) requires StatefulSet + headless Service for targeted roles in Profile 3 anyway — so in k8s you pay for Option C *and* Option B, and the directory labels duplicate k8s labels there.

The honest framing: Option B's real value is Profile 2 / non-k8s on-prem, where nothing equivalent exists. Put that in the drivers; it makes the decision stronger, not weaker.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+| Target a specific instance / label                    | 1     | No                  | directory SDK (proto + trait), `GearManager` selection  |
+| Entrypoint name resolution + edge exposure            | 1     | No                  | directory SDK + `GearManager` + `api-gateway` edge      |
+| Ownership / sharding / consistent-hash policy         | app   | No                  | consuming gear (broker dispatcher, cluster coordinator) |
+| Instance targeting via generated typed client         | 2     | Yes                 | key-aware `EndpointResolver` / `CallOptions` (codegen)  |
+| Per-instance edge routing                             | -     | No                  | out of scope (edge stays single-endpoint per gear/path) |
+
+### Relationship to Other ADRs
+
+* **ADR-0001 (Deployment Profiles)**: instance reachability differs per profile; Option B's k8s prerequisite
+  (headless Service / StatefulSet for targeted roles) is a Profile-3 concern.
```
MikeFalcon77: **major — these need amendments, not cross-references**

Per L252-257 this changes Profile 3's discovery mechanism and its "one gear = one pod + one Service" shape, so ADR-0001 and PRD `cpt-cf-fr-k8s-native` need edits.

ADR-0004 (Helm chart organization) too: the only chart in the repo today is `Deployment` + `Service` (`gears/mini-chat/deploy/helm/mini-chat/templates/`), with no StatefulSet or headless-Service shape and no per-role chart layout.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+  with `dispatcher` / `ingest` / `delivery` role deployments; `dispatcher` is the entrypoint, and its
+  routing to a specific `ingest` / `delivery` instance is realized via this decision's labels +
+  targeted resolution.
+* **Cluster gear (`GEARS.md`, Cluster Plane)** (consuming design): its service-discovery primitive - *instance
+  registration, serving intent, metadata filtering* - maps directly onto this decision. **Serving intent** is
+  the `entrypoint` flag (§2-§3); **metadata filtering** is `resolve_by_labels` over instance metadata, whose
+  selectable subset is the `labels` map (§2, §6). Its **topology watch / change-notification** feature is a
+  **cluster-gear-scoped** capability layered *above* the directory: this ADR's Layer 1 stays **poll-based**
+  (`list_instances`; no watch/push API - see Consequences), and the cluster gear builds watch/topology
+  semantics on top of that polling.
```
MikeFalcon77: **major — dependency direction is inverted**

The cluster gear is not a consumer waiting on this. It is a shipped provider of the same three capabilities (instance registration, serving intent, metadata filtering), plus the watch this ADR defers.

"the cluster gear builds watch/topology semantics on top of that polling" describes work that is already done, on a different substrate — see the comment on L36-37.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,499 @@
...
+clients) is a change inside that codegen and is deferred until it lands.
+
+### Relationship to the edge / gateway ADRs
+
+The built-in edge - specified by [`cpt-cf-adr-gateway-abstraction`](0003-cpt-cf-adr-gateway-abstraction.md)
+(ADR-0003) and [`cpt-cf-adr-edge-architecture`](0007-cpt-cf-adr-edge-architecture.md) (ADR-0007) - supplies the
+pieces this ADR's edge story depends on: the `libs/toolkit-gateway` crate (`GatewayProvider`, `ProxyRegistry`,
+`ToolKitGatewayProvider`), the `api-gateway` embedded reverse-proxy with its directory-sync loop (poll ->
+reconcile -> register/prune), the `ListAllInstances` RPC and the `InstanceInfo.openapi_spec` field, and the
+`.exposed()` / `x-cf-api-visibility` edge-visibility split (routes internal-by-default). This ADR is
```
MikeFalcon77: **minor — none of these prerequisites exist yet**

The edge-eligible predicate in §3 depends on all of them. In `libs/system-sdks/sdks/directory/proto/v1/directory.proto`: no `ListAllInstances`, and `openapi_spec` is on `RegisterInstanceRequest:84` only, not on `InstanceInfo:60-66`. No `.exposed()` or `x-cf-api-visibility` anywhere in the tree.

The ADR says Layer 1 "must merge after the edge implementation" — restate that as an ordering dependency in Consequences, so it isn't read as available today.

MikeFalcon77 (commented):
.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added. There's now a "Relationship to cluster ServiceDiscoveryV1" section plus a concept-mapping table and a Boundary paragraph,

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Reframed in "Relationship to the event-broker's current primitive"

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Fixed by taking your suggestion further: Option B (per-instance entrypoint) is rejected, and the chosen Option C makes the front-door decision structural via role-qualified names declared at the manifest level. There is no flag at all - a mis-wired instance is simply absent from a role (surfaces as not-ready), so a forgotten field cannot misroute. The tri-state-bool discussion is gone.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Dropped

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Done

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: These exist here: https://github.com/constructorfabric/gears-rust/pull/4403

See "Ordering dependency on the edge work"

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Corrected

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added amendments to relevant docs

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Adjusted wording

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added as a required contract in "Stale-ownership correction path"

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Not a reversal - basic k8s resolution still uses DNS and the directory stays optional for it; only the new specific-instance/shard targeting (which DNS can't do behind a VIP) goes through the directory. 

The multi-Service shape change is real and is now written as explicit amendments (PRD, DESIGN, ADR-0001, and ADR-0004)

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Good catch - made it a hard toolkit requirement, not a recommendation

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Agreed, added to the design for implementation

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Cluster service discovery will be removed

docs/arch/toolkit-oop/ADR/0001-cpt-cf-adr-deployment-profiles.md
```diff
@@ -54,6 +54,19 @@ Each gear runs as an independent Kubernetes pod. Service discovery uses K8s DNS
...
 public API ingress. SA tokens provide platform-plane identity; mTLS+SPIFFE (cert-manager) is the next phase.
 
+> **Amended by [ADR-0009 (Instance-Addressable Discovery)](0009-cpt-cf-adr-instance-addressable-discovery.md).**
+> Two refinements to the profile shapes above:
+> - **Profile 3:** a gear that runs in differentiated **roles** or is **sharded** maps to *multiple*
+>   role-qualified directory names — i.e. *multiple* pod sets / Services, not "one gear = one pod + one Service".
+>   Basic **name** resolution via k8s DNS is unchanged and DirectoryService remains **not required** for it;
+>   **instance / shard targeting** additionally requires each targeted instance to advertise a
+>   **per-instance-addressable** endpoint (not a shared Service VIP). The workload kind that provides it
+>   (`StatefulSet` + headless Service, a self-registering `Deployment`, …) is the gear developer's choice.
```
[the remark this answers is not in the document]
mattgarmon: Reworded

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,674 @@
...
+*different* axis and they compose:
+
+**Roles -> distinct directory names (manifest-declared).** When a gear runs in differentiated **roles**
+(front-door vs internal, or genuinely different contracts), each role registers under its **own directory
+name**. The gear stays **one gear** (one binary, one `deps`, mode-selected wiring per event-broker ADR-0007) -
+**gear identity is not the directory registration name** - and the boot mode selects which role-qualified name
+that instance registers under. Because internal roles are *not the bare name*, a bare-name resolve reaches
+**only** the front door - not by a filter that could be misconfigured, but because internal roles were never
+registered under that name. **A forgotten flag cannot misroute, because there is no flag**: a mis-wired
+instance is simply absent from a role (surfacing as "not ready", §6) rather than silently cross-routed.
```
MikeFalcon77, line 130: **"manifest-declared" has no mechanism yet, and the structural claim rests entirely on it.**

The two halves of this paragraph pull apart. `#[toolkit::gear(name = "...")]` takes a **string literal**, validated at macro expansion (`libs/toolkit-macros/src/lib.rs:233-243`), so one binary cannot declare three names that way. Meanwhile the name that actually reaches the directory is `OopRunOptions.gear_name`, a runtime `String` supplied by `main.rs` (`libs/toolkit/src/bootstrap/oop.rs:71`) — and §1 says "the boot mode selects which role-qualified name that instance registers under", which is runtime config.

If the name comes from runtime config, then "a forgotten flag cannot misroute, because there is no flag" does not hold: a mis-set mode registers an `ingest` instance under `event-broker` and a front-door caller round-robins straight into it. That is the same fail-open the `entrypoint` field was rejected for, moved from a boolean to a string.

The fix keeps the structural property: the **manifest declares the closed set** of role-qualified names, the boot mode maps into that set, and bootstrap **refuses to start** on a mode whose name is not in it. Worth specifying, since it is the mechanism the whole decision rests on.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,674 @@
...
+Worked example - `event-broker` (one binary, existing `DeploymentMode`), mapping mode -> registered name:
+
+| Boot mode           | Registers as            | Visibility                    |
+|---------------------|-------------------------|-------------------------------|
+| `cluster_dispatcher`| `event-broker`          | front door (bare name + edge) |
+| `cluster_ingest`    | `event-broker-ingest`   | internal, `shard` label       |
+| `cluster_delivery`  | `event-broker-delivery` | internal                      |
+
+```text
+directory:  event-broker         -> [dispatcher-a, dispatcher-b]
```
MikeFalcon77, line 141: **No readiness or `deps` story for role-qualified names.**

Splitting one gear into three directory names splits its dependency graph too, and nothing here covers it:

- A consumer writing `deps = [event-broker]` now gates only on the **dispatcher** name. Nothing gates on `event-broker-ingest` existing.
- The dispatcher itself depends on `event-broker-ingest` / `-delivery` — a gear depending on its own other role-names. The `deps` topo-sort has never seen a self-dependency of that shape.

Say what a role-split gear declares, and what a consumer's `deps` entry gates on.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,674 @@
...
+  dispatcher, cluster coordinator, …) does not re-implement the same membership plumbing, the toolkit ships a
+  small `TopologyView` helper over `resolve_by_labels`: a **label-keyed instance set** with **refresh** (poll
+  now; watch-driven once the change-notification follow-up lands, §7) and a **consistent-hash ring**
+  (`ring_hash` / `maglev`). It stays **mechanism**: the gear still supplies the hash key and owns the selection
+  *policy* and the §6 stale-owner correction; `TopologyView` only removes the poll/refresh/ring boilerplate. The
+  concrete API and hash-algorithm choice are implementation details deferred to the implementing change. This is
+  a Layer 1 deliverable (it depends only on `resolve_by_labels`, not on REST contract codegen).
+* **Empty-set semantics**: zero matches is `Ok(empty)`, kept distinct from a
+  directory-backend error - preserving the not-found vs. failure contract of ADR-0005, so an empty result reads
+  as "not ready yet," not "outage."
```
MikeFalcon77, line 350: **A permanently mis-wired role is indistinguishable from a cold start — forever.**

If no instance registers as `event-broker-ingest` (wrong mode, missing workload, typo in the name), `resolve_by_labels` returns `Ok(empty)`, which §6 says reads as "not ready yet," not "outage." A config error that will never resolve therefore looks exactly like a fleet still coming up, indefinitely.

This is better than the old fail-open, but it is the same gap §6 just closed for ownership: a stale target self-identifies, while an absent name does not. Something needs to eventually distinguish "no instances of a name a consumer declared a dependency on" from "not ready yet" — a readiness verdict after a grace window, as PR #4425 §4.7 proposes for its own permanent-error case.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,674 @@
...
+| `DiscoveryFilter` (AND-conjoined `MetaMatch::{Equals,OneOf}`) | `resolve_by_labels` `LabelSelector` (equality-AND; `OneOf`/expressions are the deferred `matchExpressions`, §6) |
+| `InstanceState::{Enabled, Disabled}` (serving intent)    | readiness state on results (`Registered`/`Ready`/`Healthy`/`Quarantined`/`Draining`, §6) - `Draining` == cluster `Disabled` |
+| `StateFilter::{Enabled, Disabled, Any}` (server-side serving-state filter) | **caller-side** health filtering (§6) - `resolve_by_labels` returns the full set annotated with readiness state; there is no server-side `only_serving` flag |
+| `ServiceHandle::set_state` (runtime serving-intent flip / drain) | no dedicated intent-flip RPC; an instance signals `Draining` through **readiness** (ADR-0005), which name round-robin de-prefers (§6) - the deliberate "intent, not health" split (cluster ADR-008) is folded into the readiness axis |
+| `ServiceHandle::update_metadata` (runtime metadata mutation) | no dedicated update RPC; label changes ride **re-registration** (labels MUST survive re-registration, §2 / Implementation notes) |
```
MikeFalcon77, line 563: **These two rows drop real capability, and one of them re-conflates what cluster ADR-008 separated.**

`set_state(Disabled)` is **declared intent** — cluster ADR-008 exists specifically to say intent is not a health observation. Routing it through readiness makes an instance that wants out of rotation for domain reasons fail its readiness probe, which also pulls it from the edge, flips `/readyz`, and in k8s pulls it from its Service (or restarts it, if wired to liveness). That is not an equivalent signal, and the mapping table presents it as one.

`update_metadata` via re-registration has an unspecified race: the registration/heartbeat self-heal loop re-registers periodically from its own recorded state, so a label update and a self-heal re-register can interleave. Which one wins is not stated, and "labels MUST survive re-registration" does not answer it.

Neither is necessarily wrong, but both are behaviour changes for current `ServiceDiscoveryV1` users and should be listed as such rather than as neutral mappings.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
```diff
@@ -0,0 +1,674 @@
...
+cluster gear's own decomposition as a separate, coordinated change, not by this ADR.** There is **one**
+service-discovery layer - the directory - not a directory layer plus a competing cluster layer. The prior
+concern that this ADR added a "second, weaker discovery layer next to a merged one" is resolved by
+*consolidation*: the cluster in-SDK discovery primitive
+(`gears/system/cluster/cluster-sdk/src/discovery/`, its default/standalone/Postgres backends, conformance
+suite, GTS specs, and wiring) will be **retired** by that decomposition work, after which the cluster gear
+keeps only its other three backend traits (cache, distributed lock, leader election). The dependency direction
+is therefore **not** inverted - cluster is not a shipped provider this ADR races; its discovery capability is
+the thing being consolidated here. The event-broker `domain/cluster.rs` `ServiceDiscoveryV1` handle and every
+cluster-side user are retired by that decomposition; until the directory replacement (Layer 1) lands the
```
MikeFalcon77, line 551: **This is now in direct conflict with open PR #4425, and neither PR has moved on it.**

[PR #4425](https://github.com/constructorfabric/gears-rust/pull/4425) is a design for running `cluster` as a deployable OoP gear. It keeps `ServiceDiscoveryV1` and makes it **remote**: §6.7 specifies a service-discovery contract, §12 sketches a `RemoteServiceDiscoveryBackend`, and its blocking decision 8c is specifically about `ServiceHandle` semantics across a gear restart. This section says that primitive is retired.

One of the two has to change, and the retirement here is scoped to "the cluster gear's own decomposition as a separate, coordinated change" — work that is not written, owned, or tracked anywhere I can find.

Given this ADR is now `accepted`, please link the tracking issue for the retirement and for the edge/directory prerequisites, and reconcile with #4425 before either lands. Otherwise "accepted" reads as "available", while both its prerequisites and its cleanup are unstarted.

MikeFalcon77 (commented):
.

MikeFalcon77:
merged with comments. At least two concerns should be addressed at hte next PR:
1. "manifest-declared" issue
2. Do something with 4425 conflict (service discovery)

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Implemented as specified in #4522: `#[toolkit::gear(roles = [...])]` declares the closed set of role-qualified names; the bare `name` must be a member (compile error otherwise), the set is emitted as a ROLES const and self-registered into a `DeclaredRoles` inventory, and OoP bootstrap auto-discovers and refuses to start if the resolved `gear_name` isn't in the set (`validate_role_name`). No per-instance flag, no main.rs wiring to forget - the fail-open you flagged is structurally impossible

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Added in ADR in #4522: `deps` is stated in terms of names, not gear identity. A consumer gates on the name it calls (front door, or an internal role-name it declares explicitly); a role-split gear declares its sibling role-names as ordinary directory-name edges - distinct names, so it's not a topo-sort self-dependency and introduces no cycle. Readiness gating is per-name.

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Removed this section from the ADR in #4426, no current users of cluster `ServiceDiscoveryV1`, and it was removed

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: Agreed - addressed in ADR in  #4426:  "Permanent-unresolvable verdict": for a name a consumer declared a dependency on, readiness emits a distinct unresolvable-dependency verdict (failed /readyz naming the name) once a configurable grace window elapses with zero instances, while ad-hoc resolve_by_labels probes stay a benign `Ok(empty)`

docs/arch/toolkit-oop/ADR/0009-cpt-cf-adr-instance-addressable-discovery.md
mattgarmon: #4425 has been updated with removal of `ServiceDiscoveryV1`

PR #4431, opened 2026-08-06
ci(release): pin publish to the release PR merge commit

MikeFalcon77:
- pin the checkout to `github.event.pull_request.merge_commit_sha`, falling back to `github.sha` for the workflow_dispatch path
- verify the merge commit's tree matches the release PR head's tree, i.e. the merge was trivial and PR CI actually covered the tree being published. This is the workflow-level equivalent of "require branches to be up to date before merging", which main does not currently enforce
- report release PR check status as advisory warnings plus a step summary. Merging a release PR is the maintainer's release decision, so this never blocks the publish and never waits on unfinished checks
- drop `make test-no-macros` and the nextest install from the release job: re-testing could only override that decision, never inform it, and `cargo publish` still compiles every crate. Saves ~25 min per release

- use one constant concurrency group so `release-pr` (refs/heads/main) and `release` (refs/pull/N/merge) can no longer run concurrently on the same merge, which let release-plz open the next release PR mid-publish

- add `!cancelled()`, so a transient failure in cache housekeeping cannot block an irreversible crates.io publish

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release reliability by preventing overlapping release runs.
  * Added safeguards to ensure releases match the reviewed release changes.
  * Releases now proceed when cache cleanup is skipped or fails, unless the workflow is cancelled.

* **Chores**
  * Release check results, including pending or failed checks, are now recorded for visibility.
  * Removed workspace test execution from the release job.

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -22,8 +22,11 @@ on:
 permissions:
   contents: read
 
+# One constant group for the whole workflow. `release-pr` (push to main) and
+# `release` (release PR merged) mutate the same state — tags, the release
+# branch, crates.io — and both fire on the same merge.
 concurrency:
-  group: release-plz-${{ github.workflow }}-${{ github.ref }}
+  group: release-plz-${{ github.repository }}
   cancel-in-progress: false
 
 jobs:
@@ -127,9 +130,13 @@ jobs:
     # IMPORTANT: don't publish on every push. Publish only:
     # - manually via workflow_dispatch (mode=release), or
     # - when a release PR (labeled `release-plz`) is merged into main.
+    # The `!cancelled()` below is what matters — it replaces the implicit
+    # `success()` on `needs`, so a failed or skipped cache cleanup no longer
+    # blocks the publish.
     if: >-
       ${{
-        github.repository == 'constructorfabric/gears-rust'
+        !cancelled()
+        && github.repository == 'constructorfabric/gears-rust'
         && (
           (github.event_name == 'workflow_dispatch' && github.event.inputs.mode == 'release')
           || (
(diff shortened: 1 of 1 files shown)


.github/workflows/release-plz.yml
```diff
@@ -22,8 +22,11 @@ on:
 permissions:
   contents: read
 
+# One constant group for the whole workflow. `release-pr` (push to main) and
+# `release` (release PR merged) mutate the same state — tags, the release
+# branch, crates.io — and both fire on the same merge.
 concurrency:
-  group: release-plz-${{ github.workflow }}-${{ github.ref }}
+  group: release-plz-${{ github.repository }}
```
capybutler, line 29: **Take the publish run back out of the shared concurrency queue.**

This collapses every trigger into one repository-wide group. GitHub allows at most one *pending* run per group, and cancels the previously pending run whenever a newer run joins. `cancel-in-progress: false` only protects the run that is already executing, not the one waiting behind it.

Before this change the publish run lived in its own group (`refs/pull/N/merge`) and never queued. Now it competes with every push to `main`.

Failure scenario, using this repository's own run history:

1. A release PR merge creates the `push` run and the `pull_request` run in the same second (observed pair at `2026-08-06T07:44:34Z`).
2. The short `release-pr` run wins the start race, so the publish run goes pending.
3. The next merge to `main` lands and silently cancels the pending publish run.

Both windows are wide. Merges to `main` are 29-54 seconds apart in recent history (`2026-08-05T18:50:40Z` / `18:51:09Z`, and `2026-08-06T06:50:44Z` / `06:51:38Z`). Publish runs take 46-54 minutes (`2026-08-06T07:44:34Z` to `08:30:22Z`, and `2026-08-05T18:51:09Z` to `19:45:07Z`). The mechanic already fires on push runs today: the run created at `2026-08-05T18:51:09Z` was cancelled one second later.

The result is no crates published, no tags, no GitHub release, and a run marked "cancelled" with no failure notification.

The `!cancelled()` added on line 138 does not help here. A run cancelled at the concurrency layer never starts its jobs, so no job-level `if` is evaluated.

Two workable shapes:

- Keep the constant group, but move `release-pr` off the `push` trigger onto `schedule` plus `workflow_dispatch`. Publish runs then become the only frequent members of the group.
- Restore per-trigger groups, and instead have the `release-pr` job poll `gh run list --workflow=release-plz.yml --status=in_progress` and exit early while a publish is live.

.github/workflows/release-plz.yml
```diff
@@ -188,20 +250,72 @@ jobs:
...
-      - name: cargo test (workspace)
-        run: make test-no-macros
-
-      - name: Clean build artifacts
-        run: rm -rf target
+      # No test run here on purpose. Merging a release PR is the maintainer's
+      # release decision, so re-testing could only override it, never inform it.
+      # The tree being published is the one CI validated on the release PR (see
+      # the tree check above); `cargo publish` still compiles every crate, so a
+      # crate that does not build cannot be published.
```
capybutler, line 257: **Restore a test gate on the published tree, or drop this justification.**

"The tree being published is the one CI validated on the release PR" does not hold for this repository, for two independent reasons.

**1. CI does not run on most release PRs.** `ci.yml` sets `paths-ignore: ['**/*.md', 'docs/**']` (`.github/workflows/ci.yml` lines 3-9). Seven of the last ten release-plz PRs changed `CHANGELOG.md` and nothing else, so CI never started on them.

Verified on PR #4370, head `3c7f0e27`:

- `gh run list --commit 3c7f0e27` lists no CI run.
- Its 14 check runs contain no test, clippy, or build job.
- The publish run's own log shows `make test-no-macros` executing 9112 tests. That was the only test execution against that tree.

**2. The integrated tip of `main` is never tested on its own.** `ci.yml` has no `push` trigger, so PR CI only ever validates `refs/pull/N/merge`. With merges landing under a minute apart, a semantic conflict between two PRs is caught by neither PR's CI. The step deleted here ran against the merge commit, which made it the only integration test of `main` anywhere in the pipeline.

`cargo publish` compiling every crate is a real check, but it builds lib and bin targets only. It does not build test targets and it runs no tests.

Publishing to crates.io is irreversible. Yank hides a version, it does not remove it. So this removes the last gate in front of an unrecoverable action.

Two ways to close it:

- Keep `make test-no-macros` in the release job.
- Add `push: branches: [main]` to `ci.yml` and make the publish depend on that run's result. This also fixes reason 2 for every other consumer of `main`.

The tree check added above does not substitute for either. Its reasoning is sound: if `merge_tree == head_tree` then `main` at merge time was an ancestor of the PR head, so the `refs/pull/N/merge` ref that CI ran against had the same tree. But it proves equality with a tree that, on most release PRs, nothing ever tested.

.github/workflows/release-plz.yml
```diff
@@ -141,13 +148,68 @@ jobs:
...
+          echo "merge commit $MERGE_SHA  tree=$merge_tree"
+          echo "PR head      $HEAD_SHA  tree=$head_tree"
+
+          if [ "$merge_tree" != "$head_tree" ]; then
+            echo "::error::main advanced after the release PR was last updated, so the merge is not trivial." \
+                 "The tree about to be published was never tested as a whole. Re-run release-plz to refresh" \
+                 "the release PR against current main, or publish via workflow_dispatch to override."
+            git --no-pager diff --stat "$HEAD_SHA" "$MERGE_SHA" | tail -20
+            exit 1
+          fi
```
capybutler, line 209: **Decide what happens when this check fails, because it will fail on roughly a third of releases and the printed remedy cannot be followed.**

Replaying `merge_tree != head_tree` across the last 20 release-PR merges, 6 would have aborted the publish: `9d310b0d`, `3f7c67ab`, `6f979e0c`, `70030c02`, `583748f9`, `8bd291ec`.

The check is correct to flag them. CI genuinely did not cover the merged tree. The problem is what happens next.

This runs after the maintainer has already merged. At that point `main` carries the version bumps and the changelog, and crates.io has nothing. The error text tells the operator to "Re-run release-plz to refresh the release PR against current main". That PR is closed and merged, and its version bumps are already on `main`, so a fresh `release-pr` run has nothing left to bump and will not reproduce it. The instruction cannot be followed.

The only remaining path is the `workflow_dispatch` override. That path skips this check (line 176) and, with the test step removed, runs no tests either. So the failure mode of this gate is to funnel one release in three onto the least-verified path available.

Two options:

- Make it advisory: a warning plus a step summary entry, matching the stance this PR takes on release-PR check status.
- Keep it blocking, and fix the message to describe the real recovery: re-run this workflow via `workflow_dispatch` with `mode=release`, after confirming `main` is green.

Either way the frequency is worth reducing at the source. All six failures are cases where `main` advanced after release-plz last refreshed the release branch. The release branch is only rebased by `release-pr` runs, and those already get cancelled when pushes to `main` bunch up, which is exactly when `main` advances fastest.

PR #4434, opened 2026-08-06
feat(service-principal): add service-accounts into AM and keycloak plugin

### LINKED ISSUE #4322: CORE - Service API Account Manager
**Description**: Machine-identity contract for services (SDK-only contract).
**Is Plugin**: NO
**Has Plugins**: NO
**Has Extension Points**: YES (principal identity model)

> Source: milestones doc §2 (vhp-core); VZ Q2 (inf).

fdlockgraf:
## Description

Closes: https://github.com/constructorfabric/gears-rust/issues/4322

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- [x] Unit tests pass (50 tests across both crates, incl. sanitizer boundary tests)
- [x] Integration tests pass (e2e router tests incl. 400/409/503 error paths)
- [x] New tests added

## Documentation

- [x] Service Principal PRD aligned with the implementation
- [x] Service Principal DESIGN aligned with the implementation (incl. decision-rationale appendix)

## Checklist

- [x] Self-review completed
- [x] Unrelated `.cf-studio/whatsnew.toml` and `.change-impact/` changes excluded


gears/system/service-principal/docs/PRD.md
```diff
@@ -0,0 +1,848 @@
...
+
+#### Creation Collision Safety
+
+- [ ] `p1` - **ID**: `cpt-cf-service-principal-fr-create-collision`
+
+The system **MUST** reject a creation request when its provider-side identifier is already occupied and **MUST NOT** resume, reveal, or modify the existing principal as part of that request.
+
+- **Rationale**: Reusing an existing identity could disclose credentials or attach privileges to the wrong workload.
+- **Actors**: `cpt-cf-service-principal-actor-provider-adapter`
+
```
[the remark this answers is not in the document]
fdlockgraf: I think this belong to design doc.

gears/system/service-principal/docs/PRD.md
```diff
@@ -0,0 +1,848 @@
...
+- **Physical safety**: Not applicable because this is an information-system control plane with no direct physical actuation.
+- **End-user accessibility and internationalization**: Not applicable to the initial server/API-only capability; future user interfaces must define their own requirements.
+- **Gear-specific disaster recovery**: Not applicable beyond platform and provider recovery obligations.
+- **Personal-data privacy**: No gear-specific privacy regime applies because the capability intentionally processes machine identities rather than human profiles. This exclusion ceases to apply in deployments where tenant or account identifiers can identify natural persons.
+- **Industry-specific regulation**: No regulation is introduced by this provider-neutral capability. Deployment-specific regulatory controls remain mandatory where the platform, provider, tenant, or workload is subject to them.
+- **Data retention and residency**: The identity provider owns principal-record retention and residency, while platform observability owns telemetry retention and residency. The Service Principal capability owns neither a principal datastore nor a telemetry archive.
```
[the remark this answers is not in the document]
fdlockgraf: I don't think we store in this gear anything sensitive

gears/system/service-principal/docs/DESIGN.md
```diff
@@ -0,0 +1,900 @@
...
+| `ProviderUnavailable` | No `ServicePrincipalClientV1` registered in `ClientHub` | `503` | Service-unavailable |
+| `Upstream { detail }` | `ServicePrincipalFailure::CleanFailure` or `authz::map_enforcer_err`'s `EvaluationFailed` | `503` | Service-unavailable |
+| `Ambiguous { detail }` | `ServicePrincipalFailure::Ambiguous` | `409` | `AMBIGUOUS_OUTCOME` |
+
+The taxonomy on both sides of the boundary (`ServicePrincipalFailure` and `DomainError`) is closed:
+neither type is `#[non_exhaustive]`, so every match is compile-checked to cover every variant.
+`Ambiguous` carries `409`, a status distinct from the `503` used for `Upstream` and
+`ProviderUnavailable` (rationale: [Appendix A](#appendix-a-decision-rationale)).
+Provider-specific diagnostic detail stays in the `detail` string surfaced by canonical error
+rendering; no adapter-internal state crosses the wire.
```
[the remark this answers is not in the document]
fdlockgraf: Fixed in a7608411: adapter-supplied `detail` is now sanitized at the `From<ServicePrincipalFailure> for DomainError` boundary (ASCII-graphic-only, whitespace collapsed, 200-char cap with truncation marker) before it can reach the RFC-9457 `Problem` body; server-side logs carry the sanitized text plus original length, never the raw adapter string. The SPI doc contract now requires caller-safe operator text in `detail`, and DESIGN §3.3/§4.1/§4.5 were reconciled to state one consistent behavior. Covered by sanitizer unit tests and e2e router tests asserting the 409 wire detail arrives cleaned.

gears/system/service-principal/docs/PRD.md
```diff
@@ -0,0 +1,901 @@
...
+
+### 2.2 System Actors
+
+#### Machine Workload
+
+**ID**: `cpt-cf-service-principal-actor-machine-workload`
+
+- **Role**: Uses issued credentials to request access tokens through the OAuth 2.0 `client_credentials` grant.
+- **Needs**: A stable tenant-scoped identity and valid credentials, with superseded or revoked credentials
+  becoming unusable.
```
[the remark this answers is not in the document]
fdlockgraf: Disposition (close-with-rationale): invalidation of superseded/revoked credentials is authoritatively owned by the provider adapter behind `ServicePrincipalClientV1`, and no adapter ships in this repository, so the gear cannot enforce or verify it in-repo today. The PRD records adapter conformance verification as an explicit Open Question with an owner, and DESIGN §4.10 states the limitation honestly rather than claiming a guarantee. Tightening the SPI so failure-to-invalidate must surface as `Ambiguous`/`CleanFailure` is a contract change we would prefer to land together with the first real adapter (where it can be conformance-tested) rather than as an untestable doc-comment promise here.

gears/system/service-principal/docs/PRD.md
```diff
@@ -0,0 +1,901 @@
...
+- **Rationale**: Inventory access must not grant credential access.
+- **Actors**: `cpt-cf-service-principal-actor-tenant-automation-admin`
+
+#### Ownership-Checked Addressing
+
+- [ ] `p1` - **ID**: `cpt-cf-service-principal-fr-ownership-addressing`
+
+The system **MUST** treat a `client_id` as addressable, for rotate-secret and revoke, only when it resolves
+within the explicit target tenant. An address that does not resolve within that tenant **MUST** return not-found
+without disclosing whether the id exists under a different tenant.
```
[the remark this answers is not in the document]
fdlockgraf: Disposition: in the implemented behavior there is no information-disclosure conflict — the PDP authorization gate runs before any provider call, so a caller for tenant A can never learn anything about tenant B's principals (cross-tenant requests end at 403, covered by the BOLA test), and the idempotent-revoke rule (provider `NotFound` treated as success) only applies after that gate passes. The two PRD requirements describe the authz layer and the provider layer respectively; we agree the wording can be tightened to make that layering explicit and will fold it into the next PRD touch-up rather than hold this PR on it.

gears/system/service-principal/service-principal-sdk/src/models.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Request to create a confidential `client_credentials` service client.
+#[derive(Debug, Clone)]
+pub struct CreateServicePrincipalRequest {
+    /// Tenant that owns the principal; lands in the token's `tenant_id`.
+    pub tenant_id: TenantId,
+    /// Short caller-chosen suffix (lowercase alnum + '-', max 40 chars); the
+    /// adapter builds the final client id as `svc-<tenant_id>-<name>`.
+    pub name: String,
+    /// Client scopes to attach; validated against the adapter allowlist.
+    pub scopes: Vec<String>,
```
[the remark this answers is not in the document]
fdlockgraf: Disposition (close-with-rationale): correct that `svc-<tenant_id>-<name>` is an unenforced convention and `ServicePrincipalSummary` carries no caller-supplied correlation field. The documented recovery for an ambiguous create is caller-driven `revoke` + `create` (DESIGN Appendix A explains why 409 signals that rather than inviting a blind retry), which does not depend on correlating the half-created principal. Adding a first-class correlation/idempotency field is a deliberate SPI extension we've kept out of this port (the PRD tracks it as an Open Question); it changes the trait and every future adapter, so it should land as its own reviewed change.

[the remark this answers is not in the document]
fdlockgraf: Retracting my earlier rationale on this thread — you were right and I was wrong. I claimed the documented recovery does not depend on correlating the half-created principal. It does: `revoke`/`rotate_secret` are addressed by `client_id`, `ServicePrincipalSummary` carried only `{client_id, enabled, scopes}`, and the `svc-<tenant_id>-<name>` derivation was a descriptive doc comment, never a contract. An adapter assigning opaque client ids satisfied the trait while making the recovery unachievable, exactly as you described.

Fixed in d22843da by defining the correlation contract in the SDK rather than deferring it:

- `ServicePrincipalSummary` (and the REST listing entry) now carry the caller-supplied `name`, with a normative obligation that adapters report it verbatim. That is the name-to-client_id bridge, and it works for opaque-client-id adapters.
- Uniqueness is pinned so the key is unambiguous: `create` MUST reject a name already live in the tenant, and MUST do so atomically, so `(tenant_id, name)` identifies at most one live principal. `fr-create-collision` now binds on the requested name instead of the provider's internal target identifier — previously it never actually constrained names.
- The `svc-<tenant_id>-<name>` format is demoted to an explicitly non-normative convention that callers must not parse, so the unenforced convention you flagged is no longer load-bearing anywhere.
- DESIGN 4.2 states recovery as three concrete steps (list, match on submitted name, then `rotate_secret` on that client_id or `revoke` + fresh `create`), notes absence means the create did not land so a plain retry is safe, and requires a caller that somehow sees multiple matches to escalate rather than act on a guess.
- Regression test `list_reports_caller_supplied_name_so_opaque_client_ids_stay_correlatable` drives an adapter assigning bare-UUID client ids (asserted to contain neither tenant nor name) through the real router and proves a caller reaches the principal from nothing but the submitted name; a second test pins rejection of an already-live name.

The PRD risk row no longer claims recovery is unconditionally reconcilable — it is conditional on the provider upholding the uniqueness obligation, and the residual gaps (no cross-adapter conformance harness, no automatic idempotency-key retry) are named explicitly.

gears/system/service-principal/docs/PRD.md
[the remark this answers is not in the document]
fdlockgraf: Following up on my earlier reply here — I said the wording tighten-up would come in the next PRD touch-up, and it has landed in be6e92d4.

Both requirements now state the layering instead of leaving it implied:

- `fr-ownership-addressing` no longer says an unresolved address must "return not-found" (too literal for revoke). It now says such an address MUST NOT act on any principal and MUST NOT disclose whether the id exists under another tenant, then spells out that authorization for the explicit target tenant is evaluated before any address lookup, and that past that gate rotate-secret reports not-found while revoke reports the same indistinguishable success it reports for an already-absent principal. In neither case can a caller separate never-existed from already-removed from exists-under-another-tenant.
- `fr-revoke-idempotent` now cross-references that obligation explicitly: the indistinguishability is what stops revoke from becoming a probe for principals owned elsewhere.

So the two rules are complementary rather than competing, and the pair reads coherently from either side. Implemented behavior is unchanged — only the specification was ambiguous.

gears/system/service-principal/service-principal/src/domain/error.rs
```diff
@@ -0,0 +1,297 @@
...
+    let collapsed = printable.split_whitespace().collect::<Vec<_>>().join(" ");
+    if collapsed.len() <= MAX_PROVIDER_DETAIL_LEN {
+        return collapsed;
+    }
+    // The string is now ASCII-only (filtered above), so every byte index is
+    // also a char boundary — slicing at `MAX_PROVIDER_DETAIL_LEN` is safe.
+    let mut truncated = collapsed[..MAX_PROVIDER_DETAIL_LEN].to_owned();
+    truncated.push_str(TRUNCATION_MARKER);
+    truncated
+}
```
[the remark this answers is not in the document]
fdlockgraf: Agreed, and fixed in ae807ea0 by taking your prescription rather than patching the filter. You were right that this was redaction theater: `secret=abc123` is pure ASCII graphic text, so it passed the filter untouched and reached both the response body and the log record.

`sanitize_provider_detail`, `MAX_PROVIDER_DETAIL_LEN` and `TRUNCATION_MARKER` are deleted. Adapter text is now discarded at the `From<ServicePrincipalFailure> for DomainError` boundary and replaced with fixed gear-owned messages — `InvalidInput`, `CleanFailure` and `Ambiguous` each map to a constant, and the adapter's `field` attribution is replaced by a neutral `request` placeholder rather than forwarded. Every arm now destructures with `{ .. }`, so no binding from the adapter's failure can survive into the constructed error; the property holds by construction, not by discipline.

Log records carry only the failure variant, the length of the discarded text, and whether a field was reported. The SDK docs now state that the gear treats `detail`/`field` as untrusted, never relays or logs their content, and that adapters must log their own diagnostics in-process where they hold the context to judge what is safe.

Tests assert the property on both channels: router tests drive `secret=abc123` and an oversized ANSI/host-bearing variant through the 400, 409 and 503 paths and scan the whole rendered problem envelope (not just `detail`), and `tracing-test` log-capture tests assert the emitted record contains no adapter text while pinning the metadata fields.

gears/system/service-principal/service-principal/src/domain/error.rs
```diff
@@ -0,0 +1,297 @@
...
+        .collect();
+    let collapsed = printable.split_whitespace().collect::<Vec<_>>().join(" ");
+    if collapsed.len() <= MAX_PROVIDER_DETAIL_LEN {
+        return collapsed;
+    }
+    // The string is now ASCII-only (filtered above), so every byte index is
+    // also a char boundary — slicing at `MAX_PROVIDER_DETAIL_LEN` is safe.
+    let mut truncated = collapsed[..MAX_PROVIDER_DETAIL_LEN].to_owned();
+    truncated.push_str(TRUNCATION_MARKER);
+    truncated
```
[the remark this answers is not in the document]
fdlockgraf: Correct, and this one is retired rather than patched: 200 bytes plus the 14-byte marker did exceed the documented maximum. Rather than reserving marker space, ae807ea0 removes the truncation path entirely along with `sanitize_provider_detail` and `MAX_PROVIDER_DETAIL_LEN` — adapter text is no longer placed on the wire at all, so there is nothing left to cap. See the reply on the sibling thread for the full change.

gears/system/service-principal/service-principal/src/api/rest/routes_tests.rs
```diff
@@ -0,0 +1,686 @@
...
+    ) -> Result<(), ServicePrincipalFailure> {
+        Ok(())
+    }
+    async fn list(
+        &self,
+        _ctx: &SecurityContext,
+        _tenant_id: TenantId,
+    ) -> Result<Vec<ServicePrincipalSummary>, ServicePrincipalFailure> {
+        Ok(self.created.lock().expect("mock lock").clone())
+    }
```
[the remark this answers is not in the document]
fdlockgraf: Good catch, and fixed in ae807ea0 — you are right that the double could have masked a tenant-isolation defect in the REST flow.

`OpaqueIdSp` now stores `Mutex<Vec<(TenantId, ServicePrincipalSummary)>>`. `create` rejects a duplicate only when the owner tenant and the name both match `req.tenant_id`, so uniqueness is per-tenant as the contract states rather than global, and `list` filters by the addressed `tenant_id`.

Added `same_name_in_two_tenants_succeeds_and_each_listing_stays_tenant_scoped`: it creates the same name under two tenants against one shared adapter, asserts both succeed with distinct client ids, and asserts each tenant's listing returns exactly one entry carrying that tenant's own client id. Against the previous global double this fails at the second create. The tests you flagged at 482-531 were checked too — they operate within a single tenant, so per-tenant uniqueness is exactly what they need, and they still pass unchanged.

fdlockgraf:
hey @aviator5 ! I've checked the codebase for the overlap of s2s and service principal.

From my research, they do not overlap.  In `authn-resolver-sdk/src/api.rs:69` we have 
```
async fn exchange_client_credentials(&self, request: &ClientCredentialsRequest)
      -> Result<AuthenticationResult, AuthNResolverError>;
```
that is used by s2s to communicate. 
on other hand, service principal is responsible for creation of the credentials 
```
 async fn create(&self, ctx: &SecurityContext, req: &CreateServicePrincipalRequest)                                                      
        -> Result<ServicePrincipalCredentials, ServicePrincipalFailure>;            
```

that they later can use
```
authn.exchange_client_credentials(&ClientCredentialsRequest {
          client_id:     creds.client_id,       // ← from create()
          client_secret: creds.client_secret,   // ← from create()
          scopes:        vec!["files.write".into()],
      }).await?;
```

they both use `SecurityContext` but differently ( service principal will accept it because creating one happens inside a tenant ) and the token exchange for s2s will produce one.

diffora:
> hey @aviator5 ! I've checked the codebase for the overlap of s2s and service principal.
> 
> From my research, they do not overlap. In `authn-resolver-sdk/src/api.rs:69` we have
> 
> ```
> async fn exchange_client_credentials(&self, request: &ClientCredentialsRequest)
>       -> Result<AuthenticationResult, AuthNResolverError>;
> ```
> 
> that is used by s2s to communicate. on other hand, service principal is responsible for creation of the credentials
> 
> ```
>  async fn create(&self, ctx: &SecurityContext, req: &CreateServicePrincipalRequest)                                                      
>         -> Result<ServicePrincipalCredentials, ServicePrincipalFailure>;            
> ```
> 
> that they later can use
> 
> ```
> authn.exchange_client_credentials(&ClientCredentialsRequest {
>           client_id:     creds.client_id,       // ← from create()
>           client_secret: creds.client_secret,   // ← from create()
>           scopes:        vec!["files.write".into()],
>       }).await?;
> ```
> 
> they both use `SecurityContext` but differently ( service principal will accept it because creating one happens inside a tenant ) and the token exchange for s2s will produce one.

Exactly - for change the credentials we need to create them.

aviator5:
Agreed with @diffora, @fdlockgraf, and @striped-zebra-dev to consider an option to add service accounts provisioning in the already existing account-management gear.

diffora:
Went through this end to end against merge-base `462aae290` (96 files, +31359/-133), reading the
code and the tests rather than the diff summary. Overall: the ambiguity/idempotency model in the
Keycloak facade is the strongest part of the PR — three attacks I ran at it (an orphaned client
invisible to `list`, `rotate` handing out a secret for a half-bound principal, a name collision
resuming someone else's principal) are all already closed, and I've recorded those as verified
rather than leaving them as open questions. Two things need resolving before merge: a naming and
classification contract that is asserted but unenforced, and the complete absence of CI evidence.

## Ask: rename `service-principals` → `service-accounts`

"Principal" is not this platform's vocabulary. The subject axis in the toolkit is
`subject_type ∈ {"user", "service"}` (`libs/toolkit-security/src/context.rs:26`) and the only
registered subject type anywhere is `gts.cf.core.security.subject_user.v1~`
(`oidc-authn-plugin/src/config.rs:1125`, `config/e2e-local.yaml:271`); the word "principal"
appears nowhere in the codebase before this PR. `service_account` is also what the implementation
actually manipulates — `/service-account-user`, `bind_sa_attributes`, `SA_TENANT_ID_ATTR`,
`SA_USER_TYPE_ATTR`.

The bigger win is that it dissolves H1/H2 below: renaming the subject type to
`gts.cf.core.security.subject_service.v1~` and the managed resource to
`gts.cf.core.am.service_account.v1~` turns the substring validator into an equality check against a
named constant and leaves the collision test with nothing to protect (the two names then share no
substring). `USER_RESOURCE_TYPE` is untouched — nothing existing gets renamed.

Cost asymmetry: nearly free today (nothing seeded, no role definitions reference the permission ids,
no Keycloak SA user carries the `user_type` value), and after merge it becomes a wire break plus a
vendor-side attribute migration that interacts badly with M3.

Wire surface that must move together — and note each item lands twice, since
`docs/account-management-v1.yaml` is a hand-maintained parallel document with its own conventions
(camelCase operation ids, lowercase-plural tags; pre-existing across every AM resource, e.g.
`listTenantUsers` vs `account_management.list_tenant_users`):

- paths `/tenants/{tenant_id}/service-principals` → `…/service-accounts` (the `rotate-secret`
  suffix is unchanged);
- operation ids `{create,list,rotate,revoke}_tenant_service_principal[s|_secret]` →
  `…_tenant_service_account[s|_secret]`;
- response field `service_principals` → `service_accounts` (yaml:1638,1643);
- **the four DTO type names — they are the OpenAPI schema component names.** The `Dto` suffix is
  stripped but the stem is the Rust type verbatim, so `ServicePrincipalListDto` /
  `ServicePrincipalCreateRequestDto` / `ServicePrincipalCredentialsDto` / `ServicePrincipalDto`
  surface as `#/components/schemas/ServicePrincipalList`, `…CreateRequest`, `…Credentials`
  (yaml:435,469,488,539) and become class names in generated clients;
- tags (`service-principals` in the yaml, `"Tenant Service Principals"` in the code);
- both GTS ids and the four permission instance ids;
- the Problem envelope's `resource_type` value (follows the GTS id, but clients switch on it).

Unchanged: the action vocabulary (`create` / `list` / `rotate_secret` / `revoke`),
`AMBIGUOUS_OUTCOME`, and the `idp_unsupported_operation` / `idp_unavailable` codes.

Non-breaking but should follow so nothing reads half-renamed: the SDK module and its seven
`Idp*ServicePrincipal*` types plus the four trait methods, `domain/service_principal/`,
`ServicePrincipalService`, `pep::SERVICE_PRINCIPAL`, `ServicePrincipalFailureExt`, the three
`DomainError::ServicePrincipal*` variants, `UnsupportedResource::ServicePrincipal`, the
`ServicePrincipalResource` marker, the plugin's `service_principal_facade.rs` / `sp_impl.rs`, the
`[service_principal]` config key, and the metrics — `sp_op_duration` plus its `sp_create` /
`sp_rotate_secret` / `sp_revoke` / `sp_list` / `sp_purge` label values
(`domain/ports/metrics.rs:114-118,185-189`), free now and a dashboard migration later.

Please **don't** rename `SP_CLIENT_ID_PREFIX = "svc"` or the `svc-<tenant_id>-<name>` shape — the
contract declares `client_id` opaque, it becomes Keycloak object names, and `svc` reads fine for a
service account.

Budget for the traceability anchors: `cpt-cf-account-management-*-service-principals-*` is 154
`@cpt-begin`/`@cpt-end` occurrences across 18 distinct ids, plus `feature-service-principals.md`
and its PRD / DESIGN / DECOMPOSITION entries. Optional in principle, but spec-check pins the code
anchors to the document, so a half-rename is worse than either end state.

## High

**H1 — `cf.core.security.subject_service_principal.v1~` is not a registered GTS type.**
`plugins/keycloak-idp-plugin/src/config.rs:457`, `src/domain/system_actor.rs:53`. The value is
written into every principal's Keycloak `user_type` attribute and onto the plugin's own system
context, so it reaches issued tokens. Register it where `subject_user.v1~` lives, or point at the
component that owns the namespace.

**H2 — the `contains("subject_service")` rule has no consumer in this repository.** `config.rs:486`
rejects a `subject_type` lacking the substring, citing the "RBAC principal-resolver classification
pattern"; `system_actor.rs:47-52` states the resolver "rejects an unknown `subject_type`
fail-closed (Internal/500)" and names `classify_subject_type`. Neither exists here — `subject_type`
is an `Option<String>` (`authz-resolver-sdk/src/models.rs:93`) forwarded by `pep/enforcer.rs:318`,
and both in-repo authz plugins decide purely on tenant hierarchy. The SDK test
`gts.rs::service_principal_type_is_not_a_subject_type` rests on the same assumption. **Q:** is the
classifier an external/closed component? If so please cite it in both docstrings; if not, the
validator should be an equality check and the collision test should go.

**H3 — no CI evidence.** `gh pr checks 4434` returns DCO + CodeRabbit only; code-ranker reports
"Built on a fork". Compile, clippy, dylint, unit, integration and e2e never ran on this branch.

## Medium

**M1 — the PEP resource id carries no information.** `domain/service_principal/service.rs::authorize`
passes `tenant_id` as `OWNER_TENANT_ID`, as `RESOURCE_ID`, and as the `access_scope_with` resource
argument. Authorization is tenant-granular; the `client_id` is never an authz input and its
ownership is delegated entirely to the adapter. Both in-tree adapters do enforce it
(`lookup_owned_client` checks the id prefix *and* the `attributes.tenant_id` marker), but a
third-party adapter that forgets has no platform-side backstop. This matches `pep::USER` exactly,
so it is an inherited convention rather than a regression — the issue is that the new module's
docstring presents it as an object-level check ("step 2 already is one"), which covers the tenant
but not the principal address.

**M2 — nothing proves the four per-verb permissions gate anything.** `tests/common/mod.rs:105` —
`PermitWithSubtreeResolver::evaluate` never inspects `action`, so it permits every verb for every
caller, and all of `api_service_principals_test.rs` runs under it. The independent-permissions DoD
is the entire justification for four actions instead of a read/write triad and has no probe. One
action-aware fake PDP plus four scenarios closes it.

**M3 — principal health is keyed on the currently-configured `subject_type`.**
`service_principal_facade.rs::sa_user` sets `attrs_bound` only when `user_type` equals
`cfg.subject_type`, and `rotate_secret_impl` refuses on `false`. Changing that config value
retroactively marks every existing principal corrupt — `rotate` breaks fleet-wide while `list` and
`revoke` keep working, and the documented recovery is re-issuing every credential. Given H1 that
value is likely to change. Suggest checking `user_type` for presence, or carrying an
accepted-legacy-values list.

**M4 — `Location` is concatenated from an opaque provider identifier without encoding.**
`api/rest/handlers/service_principals.rs`. The contract declares `client_id` opaque and
adapter-chosen, and AM validates only the name's length. A `client_id` containing `/`, `?`, `#`, a
space or a control character yields a wrong URL or an invalid header value. `encode_component`
already exists in the plugin.

**M5 — the Keycloak plugin has no e2e coverage.** `test_service_principals.py` states in its own
docstring that the seam under test is AM REST → `IdpPluginClient` → **static-idp-plugin**; ~9k
lines of adapter are exercised only against fake transports.

**M6 — module cycle** `service_principal_facade.rs:30` ↔ `tenant_facade.rs:327` (code-ranker 🔴).
Moving `PROVISIONING_TENANT_ID_ATTR` + `ambiguous_from_kc_error` to a neutral module — and, if the
reverse edge should go too, having `tenant_facade` depend on a small purge-hook trait — cuts it.

## Low

- **L1** `ServicePrincipalNotFound` carries `resource: String::new()` in the shared mapping
  (`domain/idp/mod.rs`); only `rotate_secret` pre-empts the arm, so every other path emits an
  AIP-193 envelope with an empty `resource`.
- **L2** `list` maps a provider `Ambiguous` to `409 AMBIGUOUS_OUTCOME`, telling the caller to
  reconcile a read. The docstring says list has no ambiguous outcome; nothing enforces that.
- **L3** the deprovision credential purge is fail-open — `tenant_facade.rs:660` guards it with
  `if let Some(...)` over a `OnceLock`. It *is* wired (`module.rs:185`), but a wiring regression
  would silently leave live machine credentials after a tenant is deprovisioned, with no error and
  no test asserting the hook is installed.
- **L4** `list_owned_reps` / `enforce_quota` issue `GET /clients?clientId=<prefix>&search=true`
  with no `first`/`max`, so both the quota gate and the reconciliation listing are bounded by the
  server default. Not reachable at `per_tenant_quota = 10`; a marker for whoever raises it.
- **L5** `encode_component` is `form_urlencoded` (space → `+`), query encoding applied to a path
  segment. The docstring justifies it as "inputs are `[a-z0-9-]`", but `admin_base()` feeds it
  `cfg.realm`, which is free-form operator config.
- **L6** the contract offers a recovery this adapter never provides:
  `provision_service_principal`'s docs allow resolving an ambiguous create by "rotate … without
  deleting", but an ambiguous create failed at or before the SA-attribute bind, so `attrs_bound` is
  false and rotate always refuses. The route description says so; the SDK contract does not.
- **L7** `scope_allowlist` defaults to empty, so any non-empty `scopes` request is a 400 until the
  section is configured. Fail-closed is right, but it deserves the note `realm` got.
- **L8** the e2e cross-tenant assertion is `assert r.status_code in (403, 404)`
  (`test_service_principals.py:278,289`). The contract is specifically 404 — "invisible rather than
  merely forbidden, which keeps tenant topology from leaking through a 403" — and a disjunctive
  assertion cannot fail if that regresses.

## Verified clean

Recording these so they don't get re-litigated: the ownership marker is written in the client
`POST` body rather than as a follow-up call, so a create that dies later still leaves a client that
`list` returns and `revoke` deletes — no orphan can permanently squat a name; KC 409 → permanent
`SpInvalidInput` regardless of owner, with uniqueness enforced by Keycloak's own `clientId`
constraint, so the SDK's atomicity obligation is genuinely met, not merely asserted; `rotate`
checks `attrs_bound` before the secret POST; all validation and the quota check precede the first
mutation; `purge_tenant_principals` is wired into deprovision, swallows per-client 404 and
deliberately avoids the success-equivalent error variant; every new `DomainError` variant is
covered in the single `From<DomainError> for CanonicalError` ladder and no adapter text reaches a
tracing macro on this path; `no-store` on both credential responses with no secret in any listing;
`am_permission_catalog_matches_enforced_action_vocabulary` is a genuinely good addition that closes
the enforced-but-undeclared direction for all five AM resource types; and the placement of
`idp_service_principal.rs` beside `idp.rs` correctly mirrors the existing `idp_user.rs` split —
please don't fold it into `idp.rs`.

## On the authorization resource

Given M1, the minimal honest fix is to stop passing `RESOURCE_ID` (pass `None`, as
`TenantService::create` / `list_children` already do) and drop it from the `ResourceType`
declaration — unless it is genuinely load-bearing for the subtree clamp resolving through the
`tenants` entity's `resource_col = "id"`, in which case it is misnamed rather than redundant and
the docstring should say "the tenant address the clamp resolves through". If object-level policy is
ever wanted, the key already exists and already flows: `IdpServicePrincipalCredentials::subject_id`
is a `Uuid`, which `client_id` can never be — though making it the addressing key is a contract
change for a follow-up. Independently of either: the "a `client_id` belongs to the addressed
tenant" guarantee is currently a docstring sentence; promoting it to a shared SDK conformance test
would make a third adapter honour it too.

diffora:
## HIGH

### X-1 — Red `Validate CFS Artifacts`: a semantic merge conflict with current `main`, invisible on the branch head

`gears/system/account-management/account-management/src/infra/storage/migrations/m0005_tenant_metadata_indexes.rs`

- merge-base has a `@cpt-begin` with no `@cpt-end`
- this PR adds an `@cpt-end` before the `_ =>` arm (line 65)
- `upstream/main` independently added its own, trailing the closing brace (line 67)
- git merges both cleanly → one begin, two ends → `marker-end-no-begin`

Local `cfs validate` on the head passes (`Errors: 0`); `actions/checkout` on a `pull_request` event checks out the *merge* commit. Every other branch that ran CFS on 08-17 passed; only this one fails, twice.

```console
$ git merge-tree ca4736a6d HEAD upstream/main
@@ -65,7 +65,7 @@
             // @cpt-end:cpt-cf-account-management-dbtable-tenant-metadata:p2:inst-dbtable-tenant-metadata-index-schema
             _ => {
                 return Err(DbErr::Custom(MYSQL_NOT_SUPPORTED.to_owned()));
-            }
+            } // @cpt-end:cpt-cf-account-management-dbtable-tenant-metadata:p2:inst-dbtable-tenant-metadata-index-schema
```

> **Fix:** rebase onto `upstream/main` and drop the marker line this PR adds — `main` already carries one. Validate the merge, not the head.

### Z6-1 — `POST /clients` is replayed on a KC 5xx, and the resulting 409 becomes a permanent reject instead of `Ambiguous`

1. `infra/kc_http.rs:274-283` — retries whenever `is_retryable_status(status) && attempt < max_retries`
2. `domain/error.rs:340-342` — `is_retryable_status(s) = s >= 500 || s == 429`
3. `infra/kc_http.rs:402-407` — `retry_transport_errors` (false for POST) gates **only** connect/timeout; its own doc: *"Retryable HTTP statuses (5xx/429) are unaffected"*
4. `config.rs:128` — `max_retries` defaults to 3 → live by default
5. `domain/service_account_facade.rs:542-545` — 409 → `SaInvalidInput{field:"name"}`, commented *"the id was taken before this request"*

That premise is false when attempt 1 landed the client and its response was a 502/503/504: attempt 2 gets a 409 because *this call* took the name. The caller receives a permanent 400 with no `AMBIGUOUS_OUTCOME` and no reason to reconcile; Keycloak keeps an orphaned confidential client holding a secret nobody has. Violates `dod-service-accounts-ambiguous-outcome`.

Sibling: `tenant_facade.rs:1406-1481` (`ensure_realm`) solves the identical duplicate-POST 409 on `POST /admin/realms` by re-probing — hardened after the same incident. The SA path is the only one that doesn't.

> **Fix (preferred):** a second flag on `execute_with_retry` so a non-idempotent verb doesn't replay on a retryable status either. Leave `with_reactive_401`'s 401 re-mint alone — a 401 proves rejection before any mutation.
> **Fix (alternative):** surface whether a replay occurred and map 409-after-replay to `AmbiguousCreated{stage: KcClientCreate}`.
> **Not the fix:** copying `ensure_realm`'s adopt-on-409 arm — the contract forbids resuming an existing account.
> **Probe:** wiremock `503` then `409` on `POST /clients`, assert `Ambiguous`. The current 409 test pins only the single-attempt case and stays green either way.

## MEDIUM

### Z3-1 — Unmapped failure category answers 500 where the DoD requires 503, and the `@cpt` markers claim compliance

`domain/idp/mod.rs:680-700` — `other =>` returns `DomainError::Internal`, wrapped in `@cpt-…:inst-algo-sa-text-discard-unknown-category`. `feature-service-accounts.md:261` and `dod-service-accounts-no-adapter-text` both say *"RETURN the conservative upstream-unavailable answer"*. The code's argument (500 is honest, since AM can't vouch for 503's "nothing retained, retry is safe") is better than the DoD's — but spec-check currently reads the obligation as paid. The loud `error!` half is implemented.

> **Fix:** amend step 5 + the DoD sentence to "the conservative *internal* answer" with the reasoning (keeps the markers truthful), or change the arm to `IdpUnavailable`.

### Z8-1 — The four `LazyIdpProvider` service-account forwards ship with no test

`src/infra/idp/lazy.rs` +105; `git diff` over `src/infra/idp/` shows that file only. All five existing tests drive `provision_tenant`. Nothing fails if a forward is dropped again, or if a future trait method with a declining default is added and not forwarded. The "no mock could surface this" claim holds for the Keycloak unmanaged-attributes defect, not for this seam — `ClientHub` and `TypesRegistryClient` are already faked at `lazy_tests.rs:71`.

> **Fix:** add a call recorder to the existing fake plugin and assert all ten trait methods reach it — one exhaustive test, near-zero fixture cost.

## LOW

### Z1-1 — `IdpServiceAccountFailure::Display` prints the untrusted `detail`

`account-management-sdk/src/idp_service_account.rs:328-332`. AM is clean today (every arm binds `{ .. }`; grepped the crate for formatting sites), but one future `error = %failure` defeats a MUST silently.

> **Fix:** drop `detail` from `Display` — the label is all AM wants — or add a guard like the workspace's `secrecy` MUST-wrap rule.

### Z2-1 — `list`'s `Ambiguous` → 503 remap is correct but undocumented in the spec

`domain/service_account/service.rs:351-353`. Algorithm step 7 says `Ambiguous` → `aborted` unconditionally, and the list flow invokes it with no carve-out; the justification lives in a comment.

> **Fix:** add a step 7a ("a non-retaining read reports uncertainty as the clean-failure category") or amend the list flow's error scenarios.

### Z3-2 — `SA_UNSUPPORTED_MESSAGE` never reaches the wire; two texts for one category

`domain/idp/mod.rs:540` sets it, `infra/sdk_error_mapping.rs:456` discards it for its own `UNSUPPORTED_DETAIL`. Only reader is `service_tests.rs:459`, pinning a string no client can observe. Also `redact_provider_detail` now digests AM's own constant.

> **Fix:** forward the domain `detail` on that arm, or delete the constant. One text, one owner.

### Z5-1 — `RESOURCE_ID` == `OWNER_TENANT_ID`, so no PEP property distinguishes one account from another

`domain/service_account/service.rs:146-149`. Documented seam, not a silent gap: the reason is given, and the module states that below the tenant clamp nothing AM-side constrains the `client_id`, so a third-party adapter that omits the ownership check has no platform-side backstop.

> **Fix:** none in this PR — the obligation is on the trait and both in-tree adapters enforce it. Open item is the absent cross-adapter conformance harness.

### Z6-2 — `kc_provisioning.rs` documents a reconciliation guarantee the SA facade doesn't implement

`domain/kc_provisioning.rs:20-31` claims the shared marker makes a duplicate-POST 409 reconcile to idempotent success. True for realms; `post_client`'s 409 arm never reads `PROVISIONING_TENANT_ID_ATTR`.

> **Fix:** qualify the sentence to name the realm surface, or point it at Z6-1's fix.

### Z7-1 — Two path-segment encoders in one crate

`tenant_facade.rs` uses `urlencoding::encode`; `service_account_facade.rs` uses `encode_component` over `percent_encoding::NON_ALPHANUMERIC` with a docstring warning against form-style encoders. Both correct today; the rule lives only in one file's prose.

> **Fix:** move `encode_component` into `kc_provisioning` and drop `urlencoding` from the crate.

### Z9-1 — `StoredAccount.secret` retains a plaintext credential with no reader

`static-idp-plugin/src/domain/service.rs:71-87` — the struct doc says "No secret is retained", then declares `pub secret: String`. Its only reader, `credentials_for` (`service.rs:318`), is called from the two sites that already hold the freshly minted value.

> **Fix:** pass the minted secret into `credentials_for` and drop the field, or amend the doc.

### Z9-2 — `generation` is written and never read; module doc describes the superseded scheme

`service.rs:396` bumps it, nothing reads it. `service.rs:13` still claims the secret "carries a per-account generation counter" — false since minting went random.

> **Fix:** delete the field, correct the header sentence.

### Z10-1 — Guard order (tenant guard before input caps) is unpinned

Both halves are tested separately; no test sends an over-cap name to an invisible tenant. Swapping the two lines in `create` keeps the suite green while leaking tenant existence through the error shape.

> **Fix:** one test with both violations at once — out-of-subtree tenant + over-cap name → `NotFound`.

### X-3 — `service_account.subject_type` is settable but has exactly one legal value

`config.rs:499` declares it; `validate` (`config.rs:545`) rejects anything else; the default already supplies it.

> **Fix:** drop the field and read the constant, or document the future subject arm it exists for.

diffora:
There are no service-account methods in `AccountManagementClient` — you can only reach this over REST. Other gears resolve `dyn AccountManagementClient` through ClientHub, so right now they can't create or revoke a machine identity without doing an HTTP call into a sibling gear. Can you add them?

diffora (approved):

fdlockgraf:
@aviator5 could you please also take a look?

aviator5:
## Findings

### Critical

None found.

### Important

- **Deprovision saga aborts on tenant-group 404, leaking the OpenBao realm-admin secret + orphaning the realm** - `gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/tenant_facade.rs:785-798`
  - **What:** `deprovision_after_decode` performs the tenant-group `DELETE` first (step 2, lines 767-798) and maps vendor `404|410` to `Err(PluginError::DeprovisionNotFound)`. Only afterwards (steps 3-5, lines 800-881+) does it run the parent-emptiness probe, the realm `DELETE` (which correctly treats `404|410` as success at line 867), and the `cs_writer.delete(&secret_ref)` for the OpenBao-backed realm-admin secret. `deprovision_tenant_saga` (lines 687-704) catches `DeprovisionNotFound` and translates it to an `already_absent` audit + `IdpDeprovisionFailure::NotFound`, which AM treats as success-equivalent — so no retry is triggered.
  - **Why:** If a first deprovision attempt deletes the tenant group successfully but then fails (5xx/transport/timeout) before completing the realm DELETE and the OpenBao secret delete, every retry will hit a `404` on the group DELETE (step 2), abort the saga as `DeprovisionNotFound`, and be acknowledged as success. The per-realm admin `client_secret` then lingers in OpenBao indefinitely and the orphaned KC realm is never torn down — a credential-lifecycle leak with no operator signal beyond the `already_absent` audit line. The hazard is structural in the step ordering, not dependent on a rare KC misfeature.
  - **Fix:** Make the tenant-group `DELETE` tolerate `404|410` as success (mirror the realm `DELETE` arm at line 867: `200..=299 | 404 | 410 => Ok(())`) and let the saga continue to the Created-mode cleanup steps. Only surface `DeprovisionNotFound` if the *realm* probe shows the realm is already gone (in which case the secret delete should still be attempted). Alternatively, reorder so the OpenBao secret delete runs before/with the realm DELETE and is idempotent on absent refs.
  - **Confidence:** 78/100

- **`read_realm_admin_client_secret` non-2xx body emitted without `redact_secrets`** - `gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/tenant_facade.rs:1924-1932`
  - **What:** The non-2xx arm of `read_realm_admin_client_secret` builds an `AmbiguousCreated { detail: format!(... -> {status}: {}", truncate_2kb(&body_text)) }` without running the body through `redact_secrets`. The sibling helper `service_account_facade::read_secret` (lines 758, 771) and every `body_first_2kb` emitter in `user_facade.rs` consistently pair `redact_secrets` with `truncate_2kb`. The endpoint in question returns the plaintext `client_secret` on 2xx, so a misconfigured KC or an upstream proxy that echoes fragments of the rejected credential/Authorization header in the error body can leak into the persisted `detail` field.
  - **Why:** Violates the DESIGN "Security and Data Protection" defence-in-depth mandate that every `body_first_2kb` emitter run `redact_secrets`. The asymmetry with the SA/user facades is a real redaction gap on a credential-bearing endpoint's error path.
  - **Fix:** Wrap with `redact_secrets(&truncate_2kb(&body_text))` at line 1929 (and apply the same to the JSON-decode branch at line 1937 for consistency, though that branch carries no body).
  - **Confidence:** 75/100

## Test Gaps

- **`contract_am_sdk.rs` is a stub placeholder, not a contract test** - `gears/system/account-management/plugins/keycloak-idp-plugin/tests/contract_am_sdk.rs:32-37`
  - The file's single `#[test] fn placeholder_until_am_sdk_conformance_harness_lands()` has a comment-only body and the doc-string explicitly defers wiring to "a follow-up ticket". A reviewer scanning the test list sees a `contract_am_sdk.rs` and assumes SDK-contract behaviour is asserted; it is not. Rename to `*_placeholder.rs` and mark `#[ignore]` so it does not inflate the test-count metric, or add at least one in-process `Arc<dyn IdpPluginClient>` dispatch test pinning the four SA methods through the trait object.

- **E2E covers 400/404/405 but NOT the 409 / 503 the PR description claims** - `testing/e2e/gears/account_management/test_service_accounts.py`
  - Status codes asserted across the suite: 201, 200, 204, 400 (duplicate name), 404 (rotate-absent, cross-tenant), 405 (no-GET). Zero hits for `503` and `409`. The in-process `api_service_accounts_test.rs` also does not exercise `Ambiguous` (→409) or `CleanFailure` (→503) end-to-end: its `FakeIdpPlugin` only returns `InvalidInput` for duplicate names. These are the two failure paths most likely to be regressed by future error-mapping refactors and they carry the security-sensitive wire-shape (`Cache-Control`, no adapter detail leak). Add a `FakeIdpPlugin` failure-injection hook for `Ambiguous`/`CleanFailure` (the in-process fake already supports it via `FakeServiceAccountOutcome` in `service_tests.rs`) and add two REST cases asserting the 409 body and 503 envelope.

- **`token_cache_tests.rs` is fully sequential; the `Arc::ptr_eq` race guard is untested** - `gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/kc/token_cache_tests.rs`
  - All 11 tests are `#[test]` (not `#[tokio::test]`). `TokenCache::get` advertises an `Arc::ptr_eq`-guarded removal invariant ("a concurrent fresh insert wins the race") that has no test, and `InflightLocks::prune_inactive` similarly has only sequential coverage. The factory-level single-flight invariant IS concurrently exercised in `factory_tests.rs:614` (`concurrent_acquire_on_cold_cache_hits_token_endpoint_once`), so practical behaviour is pinned one layer up; the cache-level invariant specifically is not. Add a `#[tokio::test]` that spawns N tasks mixing `insert(ttl=10ms)` + `invalidate` + `get` in a tight loop and asserts no panic/deadlock + post-loop `cache.len()` consistency.

- **`rotate_secret` Ambiguous outcome not pinned at the service-test level** - `gears/system/account-management/account-management/src/domain/service_account/service_tests.rs:624-639`
  - The only rotate-failure test pins `NotFound`. The Ambiguous/CleanFailure/InvalidInput outcomes for rotate are not asserted; `set_rotate_outcome` is only ever called with `NotFound`. Because `rotate_secret` passes `client_id` (not `name`) as the `resource` to `into_domain_error`, a future refactor that accidentally routes rotate through the list-mapper (`into_list_domain_error`, which collapses Ambiguous→`IdpUnavailable`) would pass every existing test. Add `rotate_secret_maps_ambiguous_to_its_own_variant` mirroring `create_maps_ambiguous_to_its_own_variant_steering_reconciliation`.

## Notes

- **`update_user` default-impl gap in keycloak plugin** (confidence 70, below reporting bar but worth surfacing): `KeycloakIdpPlugin` implements 9 of 10 `IdpPluginClient` methods; `update_user` inherits the SDK default that returns `UnsupportedOperation` → HTTP 501. This is explicitly documented as intentional in `plugins/keycloak-idp-plugin/docs/DESIGN.md:62,596` (deferred, needs JSON Merge Patch + duplicate/password-policy classification). The SDK default returns a typed error rather than silently no-op'ing, so the "MUST NOT silently no-op" contract half is satisfied. Confirm the REST layer surfaces 501 (not 500) on this path — `sdk_error_mapping_tests.rs` pins 501 + `User` resource_type, so the wire shape is correct.

- **`revoke` folds adapter `NotFound` (including foreign-tenant coupling rejection) into `outcome=ok` with no `am.idp` failure log** (confidence 62, below bar): `service.rs:475-490` matches `Ok(()) | Err(NotFound { .. })` into a single success arm and emits one `service_account_revoked / outcome=ok` audit record. Per the contract, `NotFound` covers both "account genuinely absent" and "client_id not owned by this tenant" (adapter coupling check). Both produce an identical 204 + audit line, with no `am.idp` category record (the folded arm skips `log_service_account_failure`). An auditor attributing "who revoked account X" cannot distinguish a blocked cross-tenant attempt from a real destroy. Consider calling `log_service_account_failure(&NotFound{..})` inside the folded arm and/or adding an `outcome = "ok_removed"` vs `outcome = "ok_already_absent"` discriminator. Idempotent 204 contract stays intact.

- **Widespread `truncate_2kb`-without-`redact_secrets` in `tenant_facade.rs`** (confidence 70, below bar): lines 1528, 1625, 1674, 1843, 1885, 1979, 2055, 2069, 2195, 2203, 2282, 2310, 2320, 2344, 2395, 2412, 2434, 2474, 2535 build `body_first_2kb` / `detail` fields from `truncate_2kb(&body)` without the redaction wrapper that `service_account_facade.rs`, `user_facade.rs`, and `kc/factory.rs` apply consistently. These endpoints don't normally carry secrets, but a misconfigured KC error body could echo form fragments, bearer tokens, or sensitive `Location` URLs. Notable: the `bind_admin_user_to_tenant` 401 arms (2055, 2195) and the realm-create body (1625). Defence-in-depth cleanup; bulk-apply `redact_secrets(&truncate_2kb(&body))` at these sites.

- **`CredStoreWriter::put` copies the secret into a non-zeroized `Vec<u8>`** (confidence 65, below bar): `credstore/writer.rs:54-56` — `let bytes = value.as_bytes().to_vec();` holds the plaintext across the `create().await` and is dropped normally (no zeroization) on the success leg. Consider `zeroize::Zeroizing<Vec<u8>>`. The `SecretValue` newtype is documented as non-`Clone`; the intermediate copy is the gap.

- **No response-body size cap in `kc_http.rs`** (confidence 60, below bar): `kc_http.rs:296,309-324` calls `resp.bytes().await` (unbounded) before `truncate_2kb`. A misconfigured or compromised KC / upstream proxy returning a large HTML error can drive the worker OOM. Use `resp.body().limit(N)` or a capped reader, aborting above ~1 MiB.

- **Secrets returned as bare `String` from secret-read helpers** (confidence 60, below bar): `service_account_facade.rs:738-783` (`read_secret`) and `tenant_facade.rs:1903-1939` (`read_realm_admin_client_secret`) both decode `CredentialRepresentation.value` into a plain `String` that callers immediately wrap into `SecretString`/`SecretValue`. The intermediate `String` (and the `serde_json` buffer) lives on the stack/heap without zeroization. Consider having these helpers return `SecretString` directly and parsing into a `Zeroizing<String>`.

- **`into_domain_error` SA "unmapped variant → Internal" arm has no test** (confidence 70, below bar): `idp/mod.rs:704-712` is the `other => DomainError::Internal` arm for the `#[non_exhaustive]` `IdpServiceAccountFailure`. The provision-side analogue IS tested for redaction (`idp_tests.rs:49,84`); the SA side has no equivalent. The arm carries a long comment about not leaking adapter text — the property it enforces is security-sensitive and untested. Extracting the helper into a `pub(crate) fn` and testing it directly would close the gap.

## Assessment
**Ready to merge?** Yes, with minor fixes
**Reasoning:** The architecture, trait wiring, tenant-isolation, and credential-redaction discipline are sound and well-tested on the happy paths and most failure paths. The one materially concerning correctness issue is the deprovision saga's step ordering (tenant-group DELETE before realm/secret cleanup, with the group's 404 aborting the saga and being acknowledged as success), which can leak the OpenBao realm-admin secret and orphan the realm on retry after a partial failure; that should be fixed before or shortly after merge. The `read_realm_admin_client_secret` redaction gap is a quick fix. The 409/503 e2e coverage gap and the stub `contract_am_sdk.rs` are worth closing but are not blocking.

aviator5 (approved):

aviator5:
@fdlockgraf please check https://github.com/constructorfabric/gears-rust/pull/4434#issuecomment-5342621925.
I think PR is ready to merge and all those findings could be fixed in the follow-up PR.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/kc/token_cache.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// *active-token* realms — a realm that was fetched once, expired,
+    /// and never re-accessed has its entry physically removed without
+    /// waiting for a re-read. Called from the factory's slow-path
+    /// `acquire_client` to amortise the sweep across cold-cache hits.
+    #[must_use = "the eviction count is a useful gauge; if you don't need it, prefix with `_ =`"]
+    pub fn prune_expired(&self) -> usize {
+        let now = Instant::now();
+        let before = self.inner.len();
+        self.inner.retain(|_, v| now < v.expires_at);
+        before - self.inner.len()
```
MikeFalcon77: `before - self.inner.len()` can underflow.

`DashMap::retain` is not atomic with respect to the two `len()` calls around it. A concurrent `TokenCache::insert` — which `acquire_client` does on every slow-path miss, right before it calls this method — can leave the map larger after the sweep than `before`. The subtraction then panics in debug and wraps to a huge `usize` in release, which is also what gets returned as the "evicted" gauge.

`before.saturating_sub(self.inner.len())`

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/kc/token_cache.rs
```diff
@@ -0,0 +1,171 @@
...
+    #[must_use = "the eviction count is a useful gauge; if you don't need it, prefix with `_ =`"]
+    pub fn prune_inactive(&self, token_cache: &TokenCache) -> usize {
+        let before = self.locks.len();
+        // `TokenCache::get` does read-time eviction, so its `is_some`
+        // answer reflects the cache's POST-prune state. Calling it
+        // here without a paired `prune_expired` is fine but less
+        // efficient — the typical caller invokes both back-to-back.
+        self.locks
+            .retain(|key, _| token_cache.get(&key.0, &key.1).is_some());
+        before - self.locks.len()
```
MikeFalcon77: Same underflow as `TokenCache::prune_expired`: `lock_for` inserts into `self.locks` from other tasks while `retain` is running, so `self.locks.len()` can exceed `before`. Use `saturating_sub`.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/kc/factory.rs
```diff
@@ -0,0 +1,565 @@
...
+            tracing::warn!(
+                target: "keycloak_idp_plugin.factory",
+                realm,
+                client_id,
+                kc_expires_in_secs = body.expires_in,
+                safety_ms = self.cfg.admin_token_lifetime_safety_ms,
+                "kc token expires_in <= safety margin; clamping cache TTL to MIN_TOKEN_TTL"
+            );
+        }
+        let expires_at = Instant::now() + effective_lifetime;
```
MikeFalcon77: `Instant::now() + effective_lifetime` panics with "overflow when adding duration to instant".

`expires_in` is taken verbatim from the Keycloak token response and is only floored (`.max(MIN_TOKEN_TTL)`), never capped. A bogus or hostile value — `u64::MAX`, or anything past a few hundred years — builds a valid `Duration` and then panics on this line, inside `acquire_client`, while the per-key inflight mutex is held.

Either clamp `body.expires_in` when the response is parsed, or use `Instant::now().checked_add(effective_lifetime)` with a bounded fallback.

gears/system/account-management/account-management-sdk/src/idp_service_account.rs
```diff
@@ -0,0 +1,343 @@
...
+/// Plugin authors therefore do not redact `detail`; they log their own
+/// diagnostics in-process instead.
+///
+/// `#[non_exhaustive]` for the same reason as the sibling `IdP` failure
+/// enums: the SDK can classify a further category (say, a provider-side
+/// quota refusal) in a minor release. AM's boundary maps an unknown
+/// variant conservatively onto an internal error with a loud `error!`,
+/// so a new variant surfaces as a mapping gap in operator logs rather
+/// than as a silently mis-typed response.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
MikeFalcon77: `Display` is hand-written to emit only the category label because `detail` may quote a freshly minted credential — but the derived `Debug` still prints `detail` in full.

That undoes the guarantee on every `{:?}` path: a `?failure` tracing field, a `dbg`-style format, or `.unwrap()` / `.expect()` on any `Result<_, IdpServiceAccountFailure>` (both require `E: Debug`). This is a public SDK type, so third-party adapters and consumers hit it too, not just the AM boundary that currently discards the text.

Hand-write `Debug` the same way `Display` is written, or hold `detail` in a `SecretString`.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/service_account_facade.rs
```diff
@@ -0,0 +1,1139 @@
...
+    /// 404/410 on POST map to `SaNotFound` (lookup→rotate deletion race).
+    /// Non-2xx on GET maps to `AmbiguousCreated` (post-mutation read).
+    async fn read_secret(
+        &self,
+        bootstrap: &KcAdminClient,
+        client_id: &str,
+        client_uuid: Uuid,
+        method: &'static str,
+        stage: AmbiguousStage,
+    ) -> Result<String, PluginError> {
```
MikeFalcon77: The freshly minted client secret leaves this function as a plain `String` and stays one until `IdpServiceAccountCredentials::new` wraps it, several frames up in `create_impl` (line 367) and `rotate_secret_impl` (line 958). In between it is an ordinary heap allocation: no zeroize on drop, and printable by any `{:?}`.

Return `SecretString` and declare `SecretRep.value` as `SecretString` — `secrecy` implements `Deserialize` for it, so `serde_json::from_str` at line 775 keeps working and the plaintext is never reachable outside the wrapper.

`TokenResponse.access_token` in `domain/kc/factory.rs` has the same shape.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/tenant_facade.rs
```diff
@@ -0,0 +1,2560 @@
...
+    ///   be resolved from `OpenBao` (Adopted / Created modes).
+    /// * [`PluginError::AmbiguousCreated`] with stage [`AmbiguousStage::Timeout`]
+    ///   when the saga exceeds `provision_timeout_ms`.
+    pub async fn provision_tenant_inner(
+        &self,
+        ctx: &SecurityContext,
+        req: &IdpProvisionTenantRequest,
+    ) -> Result<TenantIdpMetadataV1, PluginError> {
+        let timeout = std::time::Duration::from_millis(self.cfg.provision_timeout_ms);
+        match tokio::time::timeout(timeout, self.provision_tenant_saga(ctx, req)).await {
```
MikeFalcon77, line 369: When this timeout fires, `provision_tenant_saga` is dropped mid-flight, and so is every future nested inside it — `with_reactive_401`, the transport retry loop, whatever KC call was in progress. The code knows that: it is why the arm below returns `AmbiguousCreated`.

But neither saga carries the cancel-safety annotation the guidelines require on every async fn that can run inside `timeout` / `select!`, so the next person who wraps one of these — or one of the `with_reactive_401` closures — in another timeout has nothing to read first. Add a `// NOT cancel-safe:` line with the reasoning on `provision_tenant_saga` and on `deprovision_tenant_saga` (line 523).

gears/system/account-management/account-management-sdk/src/client.rs
```diff
@@ -386,6 +387,67 @@ pub trait AccountManagementClient: Send + Sync + 'static {
...
+
+    /// Generate and return a new secret for the tenant-owned `client_id`.
+    /// The old secret is invalid after a successful return. An absent or
+    /// foreign account returns `NotFound`; an uncertain mutation returns
+    /// `Aborted` and must not be blindly retried.
+    async fn rotate_service_account_secret(
+        &self,
+        ctx: &SecurityContext,
+        tenant_id: Uuid,
+        client_id: String,
```
MikeFalcon77: `client_id: String` forces every caller to allocate for a value that is only read. `AccountManagementClientImpl::rotate_service_account_secret` takes the `String` and immediately does `.rotate_secret(ctx, tenant_id, &client_id)`, and the REST handler already holds a `String` it has to clone or move away.

`&str` here. No other parameter in this trait takes an owned `String`. Same on `revoke_service_account` (line 447).

gears/system/account-management/account-management/src/domain/tenant/resolve.rs
```diff
@@ -0,0 +1,112 @@
...
+        Ok(schema) => schema.type_id,
+        Err(err) => {
+            tracing::warn!(
+                target: "am.tenant.resolve",
+                resource_type = TENANT_RESOURCE_TYPE,
+                tenant_type_uuid = %tenant.tenant_type_uuid,
+                error = %err,
+                "tenant_type uuid -> chained-id resolution failed; surfacing ServiceUnavailable"
+            );
+            return Err(DomainError::service_unavailable(format!(
```
MikeFalcon77: `{err}` is the raw types-registry client error, and this detail reaches the caller verbatim: `DomainError::ServiceUnavailable` is mapped in `infra/sdk_error_mapping.rs` with `.with_detail(detail)`, no curation. Compare the `IdpUnavailable` arm directly above it, which digests the upstream text precisely to avoid this.

Whatever the registry client's `Display` carries — host, URL, transport internals — ends up in the public 503 body. The `error = %err` field on the `warn!` above already records the full text for operators, so the envelope detail can stay fixed.

Behaviour carried over from `UserService`, but the extraction now puts it on the service-account surface too.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/user_facade.rs
```diff
@@ -0,0 +1,1986 @@
...
+    ///     so this is structurally invalid.
+    fn decode_cursor(raw: &str) -> Result<ListUsersCursor, PluginError> {
+        let invalid = || PluginError::UserOpRejected {
+            detail: "invalid cursor".into(),
+        };
+        let cursor = CursorV1::decode(raw).map_err(|_| invalid())?;
+        if cursor.o != CURSOR_ORDER_DIR || cursor.s != CURSOR_ORDER_TOKENS {
+            return Err(invalid());
+        }
+        if cursor.k.len() != 2 {
```
MikeFalcon77: Length check followed by `cursor.k[0]` / `cursor.k[1]` — a slice pattern says the same thing without the indexing, and keeps the shape check and the reads in one place:

```rust
let [created, id] = cursor.k.as_slice() else { return Err(invalid()) };
let last_created_at: i64 = created.parse().map_err(|_| invalid())?;
let last_user_id: Uuid = id.parse().map_err(|_| invalid())?;
```

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/service_account_facade.rs
```diff
@@ -0,0 +1,1139 @@
...
+            "{}/clients?clientId={}",
+            self.admin_base(),
+            encode_component(client_id),
+        );
+        let (status, body) = bootstrap
+            .raw_request("GET", &url, None, "/admin/realms/{realm}/clients")
+            .await?;
+        let mut reps: Vec<SaClientRep> =
+            kc_json("GET", "/admin/realms/{realm}/clients", status, &body)?;
+        Ok(if reps.is_empty() {
```
MikeFalcon77: `reps.remove(0)` shifts the rest of the vector and needs the `mut` binding plus the `is_empty` branch. `Ok(reps.into_iter().next())` is the whole thing, and `reps` can drop the `mut`.

gears/system/account-management/plugins/keycloak-idp-plugin/src/config.rs
```diff
@@ -0,0 +1,545 @@
...
+    /// requestable", which is a legitimate configuration and so cannot be
+    /// an init-time error. It is reported as a startup log line at the
+    /// `module.rs` init site instead, because the failure it produces
+    /// otherwise — every scoped create answering 400 — reads as a caller
+    /// bug rather than as missing configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if `realm` is blank.
+    pub fn validate(&self) -> Result<(), String> {
```
MikeFalcon77: `Result<(), String>` on a public validation method. The crate already has a typed error for exactly this — `PluginError::Config { detail }` — and the only caller (`module.rs`) re-wraps the `String` into `anyhow!` anyway. Returning `PluginError::Config` keeps the error typed and matches every other fallible path in the crate.

gears/system/account-management/plugins/keycloak-idp-plugin/src/domain/kc/token_cache.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Return the cached token for `(realm, client_id)` if present **and not yet expired**.
+    /// Returns `None` on miss OR on expired entry (caller should re-acquire).
+    ///
+    /// An expired entry is physically removed from the map (read-time
+    /// eviction), keeping the cache bounded under realm churn without an
+    /// explicit `invalidate`. The removal is guarded by `Arc::ptr_eq` so a
+    /// concurrent `insert` of a fresh token wins the race.
+    #[must_use]
+    pub fn get(&self, realm: &str, client_id: &str) -> Option<Arc<CachedToken>> {
+        let key = (realm.to_owned(), client_id.to_owned());
```
MikeFalcon77: Two `String` allocations per lookup, on the hottest path in the plugin — `acquire_client` calls `get` before every Keycloak request, and the warm-cache hit is supposed to be the cheap case. `invalidate` (line 82) and `InflightLocks::lock_for` / `forget` do the same.

`DashMap<(String, String), _>` has no borrowed query form, so the fix is in the key type: a single owned key the caller can build once and reuse, or a newtype whose `Borrow` target is a `&str`-shaped view.

MikeFalcon77 (approved):

PR #4439, opened 2026-08-06
Implement event-broker dispatcher routing, REST handlers, and standalone single-process service (gears-rust#4345 gears-rust#4346 gears-rust#4347)

### LINKED ISSUE #4345: [EB][core] Implement dispatcher routing (ingest topic routing, delivery cache-based routing, failover)
Part of #4331.

## Problem Statement

The dispatcher's routing/proxy logic is fully designed but has no implementation
and no ticket owns it. `docs/DESIGN.md:2224` (§4.1 Deployment Modes) shows that
in cluster mode, a process only has the service(s) for its role active
(`cluster_ingest` has no `DeliveryService`, `cluster_delivery` has no
`IngestService`, etc.) - so a request landing on the "wrong" process for its
role has nowhere to go without a router in front. `DESIGN.md:579-612`'s module
tree names the unimplemented files: `infra/dispatcher/proxy.rs` (routes to
ingest) and `router.rs` (routes to delivery).

**Not a design gap** - `DESIGN.md:2581` is explicit that this content is
"documented inline in this DESIGN.md and will be extracted into standalone
canonical ADR files," i.e. the design work is done; only the file-split and
the implementation are pending.

**Scoped down from the original ask**: `DESIGN.md:1216-1391` (§3.2 Dispatcher
Routing) specifies a full per-instance selection algorithm (ingest:
topic-pattern matching across hetero/sharded modes; delivery: cache-based
consumer-group ownership with CAS placement and SD-gated failover
invalidation). That algorithm turned out to need its own design pass - the
`cluster-sdk` that actually shipped has no `role`/`shards` discovery
dimension and no clean place to encode a per-instance pattern set, so
DESIGN.md's prose doesn't port over mechanically. Rather than block this
ticket on that redesign, it's split out to **#4438**, and this ticket is
narrowed to what a pass-through dispatcher needs: real endpoint
classification, real `ServiceDiscoveryV1`-backed resolve-and-forward, and
self-registration - which is also exactly what `DESIGN.md:1391`'s "a stubbed/
pass-through dispatcher is acceptable to unblock 03/04" already called
sufficient for now.

**Out of scope:** in standalone mode the dispatcher is not needed at all
(`DESIGN.md:1391`) - 04 (standalone runtime) does not depend on this ticket.

## Design / Proposed Solution(s)

Implemented in `openspec/changes/eb-dispatcher-routing/` (design.md D1-D13):

- All 19 REST endpoints classified Ingest/Delivery/Shared
  (`api/rest/routes/mod.rs`), each forwarding through a shared `forward()`
  seam.
- `forward()` resolves instances via a real `ServiceDiscoveryV1` (standalone
  or postgres cluster-cache provider, no mock), picks uniformly at random
  among them, and proxies via a minimal Pingora HTTP/1 connector
  (`pingora_core::connectors::http::v1::Connector`) - not the full
  `pingora-proxy` session engine.
- Ingest/delivery self-registration on boot, with a config-based
  advertise-address resolution seam (fails startup, not lazily, on a
  wildcard bind with no `advertise_addr` configured).
- Idle-timeout (not absolute-timeout) on proxied streaming connections -
  reset per byte read, not per connection.
- Both "no instance registered" and "resolved instance unreachable" map to
  canonical `503 Service Unavailable`, distinguished by `detail` (no `502`
  equivalent in the platform's canonical error taxonomy).

## Definition of Done

- [x] All 19 endpoints route correctly per the DESIGN.md Ingest/Delivery/
      Shared routing table
- [x] Real `ServiceDiscoveryV1`-backed resolve-and-forward (no per-instance
      selection algorithm - uniform random pick, `TODO(#4438)`)
- [x] Ingest/delivery self-registration, advertise-address resolution,
      idle-timeout on streaming connections, canonical `503` error semantics
- [x] A pass-through stub is the shipped behavior for cluster mode; standalone
      mode constructs no dispatcher at all (ADR-0007, #4343), so there's
      nothing separate to stub there

**Descoped to #4438**: ingest topic-pattern routing, delivery cache-based
routing with CAS placement and SD-gated failover invalidation. Correct today
only with exactly one instance registered per role (documented limitation on
`DeliveryRouter`, `eb-dispatcher-routing` design.md D11).

## Validation

- `cargo test -p cf-gears-event-broker` - end-to-end tests against real mock
  instances (real axum servers registered via real `ServiceDiscoveryV1`, not
  hand-rolled fakes): resolve-and-forward, random selection across multiple
  registered instances, both `503` failure modes distinguished by `detail`,
  idle-timeout behavior, self-registration, advertise-address resolution.
- `openspec validate --change eb-dispatcher-routing --strict` passes.

## References

- `gears/system/event-broker/docs/DESIGN.md:1216-1391` (§3.2 Dispatcher
  Routing, Dispatcher Routing Table)
- `gears/system/event-broker/docs/DESIGN.md:2224` (§4.1 Deployment Modes)
- `gears/system/event-broker/docs/DESIGN.md:579-612` (module tree:
  `infra/dispatcher/proxy.rs`, `router.rs`)
- `gears/system/event-broker/docs/DESIGN.md:1391`, `2581`
- `openspec/changes/eb-dispatcher-routing/` (proposal.md, design.md,
  specs/event-broker-dispatcher-forwarding/spec.md)
- #4438 (successor - dispatcher instance-selection algorithm)

### LINKED ISSUE #4346: [EB][core] Implement the event-broker REST API (ingest + delivery + dispatcher) with unit/local test coverage
Part of #4331.

## Problem Statement

`docs/DESIGN.md:1021-1043` documents 19 REST endpoints across
Ingest/Delivery/Shared, backed by a full OpenAPI spec (`docs/openapi.yaml`).
None of it is implemented: no web-framework dependency (axum/actix/warp/hyper/
tower-http) exists anywhere under `gears/system/event-broker/` (verified across
all `Cargo.toml` files), and none of the documented module paths
(`api/http/ingest.rs`, `api/http/delivery.rs`, `domain/ingest.rs`,
`domain/delivery.rs`) exist in `event-broker-sdk/src/`. What exists instead is
`api.rs`'s `EventBrokerApi`/`EventBrokerBackend` traits (contracts only) and
SDK-side producer/consumer/dlq/mock code that talks to a `MockBroker` test
double, not a real service.

**Out of scope:** the storage backend behind these handlers (05/06/07), the
deployment/process topology and module skeleton this runs inside (01), and the
dispatcher's routing/proxy algorithm (02) - this ticket is the HTTP handler
layer itself, which can be built and tested against the mock backend first.

## Design / Proposed Solution(s)

*(for discussion)*

- Web framework is already decided, not open: `DESIGN.md:108` (tech
  dependency table) and §1.3 (Architecture Layers) both specify **Axum** for
  the Transport layer. Implement against Axum directly rather than
  re-evaluating framework choice.
- Implement ingest and delivery handlers as thin adapters over the existing
  `EventBrokerApi`/`EventBrokerBackend` traits, so the handler layer doesn't
  need to wait on backend implementations (05/06/07) - it can run against
  `MockBroker` for its own tests.
- Sequence: implement against the mock backend first (unblocks 04's smoke/e2e
  harness), then swap in real backends as 05/06/07 land.
- Dispatcher routing (proxying a request that lands on the wrong process for
  its role) is owned by 02, not this ticket - 02 provides a pass-through stub
  for standalone/single-process testing, so this ticket doesn't need to
  resolve dispatcher scope itself.

## Definition of Done

- [ ] All 19 endpoints from `DESIGN.md:1021-1043` implemented, matching
      `docs/openapi.yaml`
- [ ] Handlers implemented as adapters over `EventBrokerApi`/`EventBrokerBackend`,
      not backend-specific
- [ ] Unit/local test coverage for every endpoint (happy path + documented
      error codes from each feature doc's Hard-Error Catalog)
- [ ] OpenAPI spec kept in sync with the implementation (no drift between
      `openapi.yaml` and actual routes)

## Validation

- Per-endpoint unit tests hitting the handler layer directly (framework test
  client), asserting status codes and response shapes against
  `docs/openapi.yaml` and the relevant feature doc's Hard-Error Catalog (e.g.
  `docs/features/0001-idempotent-producers.md` §7).
- Contract check: existing SDK tests under
  `event-broker-sdk/tests/{producer,consumer,sdk}` continue to pass unchanged
  when pointed at the new handler layer instead of `MockBroker` directly
  (proves the handler layer is a faithful adapter).

## References

- `gears/system/event-broker/docs/DESIGN.md:1021-1043, 579-612`
- `gears/system/event-broker/docs/openapi.yaml`
- `gears/system/event-broker/event-broker-sdk/src/api.rs`
- Existing SDK test structure: `event-broker-sdk/tests/{producer,consumer,sdk}`
- Related: 01-service-topology-deployment-design.md,
  02-dispatcher-routing-implementation.md

### LINKED ISSUE #4347: [EB][core] Build a runnable standalone single-process service with a real in-memory backend, plus smoke/e2e tests
Part of #4331.

## Problem Statement

`DESIGN.md:601-602` plans a `builtin/` module directory for "Built-in backend
(in-memory for dev/test)" - this directory does not exist in source. The only
in-memory implementation today is `event-broker-sdk/src/mock/backend.rs`, an
`EventBrokerBackend` impl built specifically for SDK unit tests (its own doc
comment flags `truncate`/`segments` as `TODO(M5b-flattening)` stubs, see 11) -
it is not a runnable service. There is no standalone-mode binary anywhere, and
no smoke/e2e harness for the event broker exists under the repo-root `testing/`
directory (grepped for "event broker"/"evbk" - zero hits).

**Out of scope:** the Postgres/Kafka backends (06/07) and backend routing (05) -
standalone mode is explicitly the simplest deployment shape (`PRD.md:677`:
"Standalone deployment (single process, in-process cluster provider, in-memory
backend) functions end-to-end") and is the MVP acceptance-criteria starting
point. The dispatcher (02) is not needed in standalone mode at all
(`DESIGN.md:1391`), so this ticket has no dispatcher dependency.

## Design / Proposed Solution(s)

*(for discussion)*

- Build a real `builtin` in-memory backend (distinct from the test-only
  `mock/backend.rs`) implementing the full `EventBrokerBackend` contract
  without the M5b-flattening TODOs.
- Wire a standalone-mode binary that runs ingest+delivery in-process (per 01's
  topology decision and module skeleton) against this backend - no dispatcher
  hop needed per `DESIGN.md:1391`.
- Add a smoke/e2e test harness under `testing/` (or `event-broker-sdk/tests/`)
  that boots the standalone binary and exercises the REST API (03) end-to-end.
- Sequencing note: this depends on 01 (topology/skeleton) and 03 (REST API)
  landing far enough to have something to run - likely proceeds in parallel
  with 03, converging once handlers exist.

## Definition of Done

- [ ] A real `builtin` in-memory `EventBrokerBackend` implementation exists,
      with no M5b-flattening stubs
- [ ] A standalone-mode entry point/binary boots ingest+delivery in-process
      against it (no dispatcher hop, per `DESIGN.md:1391`)
- [ ] A smoke/e2e test suite exists and runs the standalone binary through
      publish -> consume flows
- [ ] MVP acceptance criterion from `PRD.md:677` ("Standalone deployment ...
      functions end-to-end") is demonstrably met

## Validation

- E2E smoke test: publish an event via the REST API, consume it back, assert
  content/ordering - run against the real standalone binary, not the SDK-level
  mock.
- Compare against the scenario style already used elsewhere in the module
  (`gears/system/event-broker/scenarios/flows/1.01-flow-publish-subscribe-consume.md`)
  - reuse that flow as the smoke test's script if reasonable.

## References

- `gears/system/event-broker/docs/DESIGN.md:601-602`
- `gears/system/event-broker/event-broker-sdk/src/mock/backend.rs`
- `gears/system/event-broker/docs/PRD.md:677`
- `gears/system/event-broker/scenarios/flows/1.01-flow-publish-subscribe-consume.md`
- repo-root `testing/` (current structure, no event-broker content)
- Related: 01-service-topology-deployment-design.md,
  03-rest-api-implementation.md

striped-zebra-dev:
Wire ingest/delivery/dispatcher routing across all 19 REST endpoints with a
real ServiceDiscoveryV1-backed forward seam, and implement the handlers
themselves (ingest, producers, topics, event-types, consumer-groups,
subscriptions) with a full test harness and GTS-id-validated fixtures.
REST DTOs are consolidated into per-domain dto.rs files (ingest/delivery),
mirroring the house api/rest/dto.rs convention split across the two handler
groups this crate has.

Build the real, runnable standalone single-process deployment on top of
that: SpecificationManager's SQLite-backed startup cache (bulk-loaded from
types-registry), a Storage facade (SQLite for durable state, ClusterCacheV1
for the ephemeral subscription namespace), a real SQLite EventBrokerBackend
replacing the in-memory stub, a transactional ingest outbox
(toolkit-db's Outbox/LeasedMessageHandler) for atomic publish with
out-of-transaction backend persist, and a standalone [[bin]] with a
verified config (config/event-broker-standalone.yaml) that actually boots
and serves real REST traffic end to end.

Also fixes two real gaps found only by booting the binary for real: (1)
event-broker-sdk's TopicV1/EventTypeV1 GTS schemas were id-only, rejecting
every other real topic/event-type field on registration; and (2)
SpecificationManager's startup bulk-load ran too early (module init()),
before types-registry's own post_init commits config-seeded entities to
queryable storage - moved to the module's serve() lifecycle entry, the same
ordering fix mini-chat already uses for its own OAGW upstream registration.

Closes https://github.com/constructorfabric/gears-rust/issues/4345
Closes https://github.com/constructorfabric/gears-rust/issues/4346
Closes https://github.com/constructorfabric/gears-rust/issues/4347

## Summary by CodeRabbit

* **New Features**
  * Added standalone Event Broker operation with SQLite-backed persistence.
  * Added APIs for publishing events, registering producers, managing topics, event types, consumer groups, and subscriptions.
  * Added multipart and SSE event streaming with heartbeats, seeking, rebalancing, and prompt delivery wake-ups.
  * Added pagination, filtering, authorization, validation, and standardized API error responses.
  * Added dispatcher forwarding for ingest and delivery requests.
  * Added ISO 8601 duration support and updated event schemas.

* **Tests**
  * Added comprehensive integration and end-to-end coverage for publishing, consuming, persistence, streaming, and rebalancing.


gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            });
+        }
+        self.repo.delete_subscription(subscription_id).await
+    }
+
+    async fn list_subscriptions(
+        &self,
+        _ctx: &SecurityContext,
+    ) -> Result<Vec<Subscription>, DomainError> {
+        self.repo.list_subscriptions().await
```
MikeFalcon77: list_subscriptions takes ctx but never uses it (it's prefixed _ctx) and returns every subscription from every tenant with no filtering. Filter the results by the caller's tenant before returning, the same way create_consumer_group below derives tenant_id from ctx.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+    }
+
+    async fn get_subscription(
+        &self,
+        _ctx: &SecurityContext,
+        subscription_id: Uuid,
+    ) -> Result<Subscription, DomainError> {
```
MikeFalcon77, line 561: get_subscription loads a subscription by UUID only and ignores ctx, so any authenticated caller can read another tenant's subscription details (interests, assignments, expiry) if they know or guess the id. Check the subscription's tenant against ctx before returning it.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            .ok_or_else(|| DomainError::NotFound {
+                code: "SubscriptionNotFound",
+                message: format!("subscription '{subscription_id}' does not exist"),
+                resource: subscription_id.to_string(),
+            })
+    }
+
+    async fn stream(
+        &self,
+        _ctx: &SecurityContext,
```
MikeFalcon77: stream opens the live event stream for whatever subscription_id is passed in without checking it against ctx, so any authenticated caller can read another tenant's live event payloads through GET /v1/events:stream or :sse. Verify the subscription belongs to the caller's tenant before marking it streaming.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            .await?
+            .is_none()
+        {
+            return Err(DomainError::NotFound {
+                code: "SubscriptionNotFound",
+                message: format!("subscription '{subscription_id}' does not exist"),
+                resource: subscription_id.to_string(),
+            });
+        }
+        self.repo.delete_subscription(subscription_id).await
```
MikeFalcon77: leave deletes a subscription by id without checking ctx against the subscription's owner, so any authenticated caller can terminate another tenant's subscription. Add an ownership/tenant check before calling delete_subscription.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+        }))
+    }
+
+    async fn seek(
+        &self,
+        _ctx: &SecurityContext,
+        subscription_id: Uuid,
+        targets: Vec<SeekTarget>,
+    ) -> Result<Vec<SeekPosition>, DomainError> {
+        let subscription = self
```
MikeFalcon77, line 679: seek resolves and persists new cursor positions for subscription_id without checking ctx, so a caller can move another tenant's consumption cursor. Verify the subscription's owner against ctx before writing the new cursor.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+                code: "ConsumerGroupNotFound",
+                message: format!("consumer group '{id}' is not registered"),
+                resource: id.to_owned(),
+            })
+    }
+
+    async fn list_consumer_groups(
+        &self,
+    ) -> Result<Vec<crate::domain::model::ConsumerGroup>, DomainError> {
+        self.repo.list_consumer_groups().await
```
MikeFalcon77: list_consumer_groups doesn't even take a SecurityContext parameter, so it returns every consumer group across every tenant, including tenant_id and owner_principal_id, to any caller. Add a ctx parameter and filter to the caller's tenant.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            created_at: Utc::now(),
+        };
+        self.repo.create_consumer_group(group).await
+    }
+
+    async fn get_consumer_group(
+        &self,
+        _ctx: &SecurityContext,
+        id: &str,
+    ) -> Result<crate::domain::model::ConsumerGroup, DomainError> {
```
MikeFalcon77, line 812: get_consumer_group ignores ctx and returns any consumer group's tenant_id and owner_principal_id to any authenticated caller who knows or guesses its id. Check the group's tenant/owner against ctx before returning it.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+    ) -> Result<Vec<crate::domain::model::ConsumerGroup>, DomainError> {
+        self.repo.list_consumer_groups().await
+    }
+
+    async fn delete_consumer_group(
+        &self,
+        _ctx: &SecurityContext,
+        id: &str,
+    ) -> Result<(), DomainError> {
+        if self.repo.find_consumer_group(id).await?.is_none() {
```
MikeFalcon77: delete_consumer_group ignores ctx, so any authenticated caller can delete another tenant's consumer group by id. Verify tenant_id/owner_principal_id against ctx before deleting, the same way reset_producer in ingest.rs checks ownership.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+                resource: subscription_id.to_string(),
+            });
+        }
+
+        let (tx, rx) = mpsc::channel(16);
+        let repo = std::sync::Arc::clone(&self.repo);
+        let heartbeat_interval = self.heartbeat_interval;
+        let assigned = subscription.assigned.clone();
+        let consumer_group = subscription.consumer_group.clone();
+        tokio::spawn(async move {
```
MikeFalcon77: The per-stream delivery task is spawned with tokio::spawn and has no cancellation token or shutdown signal. It only stops when the receiver drops the channel, so on gear shutdown it gets severed instead of exiting cleanly, possibly mid cursor write. Thread a cancellation token into the loop and select on it alongside the send/sleep points.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            loop {
+                let mut delivered_any = false;
+                for assignment in &assigned {
+                    let Some(topic) = repo.find_topic(&assignment.topic).await.ok().flatten()
+                    else {
+                        continue;
+                    };
+                    let since = cursors[&(assignment.topic.clone(), assignment.partition)];
+                    let Ok(events) = repo
+                        .query(&topic, assignment.partition, since, i32::MAX)
```
MikeFalcon77: Every 50ms tick, for every open stream, this re-queries the partition with limit = i32::MAX, so a subscription with a large backlog or a hot partition triggers a full scan of that partition on every single tick, and the cost multiplies with the number of open streams. Cap the per-poll query to a small batch size.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+            let mut cursors: std::collections::HashMap<(GtsInstanceId, i32), i64> = assigned
+                .iter()
+                .map(|a| ((a.topic.clone(), a.partition), a.offset))
+                .collect();
+            let mut last_frame_at = tokio::time::Instant::now();
+
+            loop {
+                let mut delivered_any = false;
+                for assignment in &assigned {
+                    let Some(topic) = repo.find_topic(&assignment.topic).await.ok().flatten()
```
MikeFalcon77: repo.find_topic(...).ok().flatten() treats a real repository error the same as topic-not-found and just continues the loop with no logging. A persistent storage failure will look like a stream that's alive but only emits heartbeats. Log the error on the Err arm instead of discarding it, like the cursor-persist failure a few lines below already does.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -47,24 +178,568 @@ pub trait DeliveryService: Send + Sync {
...
+                    else {
+                        continue;
+                    };
+                    let since = cursors[&(assignment.topic.clone(), assignment.partition)];
+                    let Ok(events) = repo
```
MikeFalcon77: repo.query(...) failures here are swallowed by the let-else with no logging, masking a real backend error as 'nothing to deliver this tick' inside an unsupervised background task. Log the error before continuing.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
```diff
@@ -0,0 +1,212 @@
...
+/// JSON.
+pub async fn reset_producer(
+    Extension(ctx): Extension<SecurityContext>,
+    Extension(state): Extension<HandlerState>,
+    Path(raw_id): Path<String>,
+    request: Request,
+) -> ApiResult<StatusCode> {
+    let producer_id = parse_action_suffixed_id(&raw_id, "reset", "producer")?;
+
+    let body = axum::body::to_bytes(request.into_body(), usize::MAX)
```
MikeFalcon77: axum::body::to_bytes(request.into_body(), usize::MAX) puts no cap on the request body size before parsing. Pass a concrete size limit instead of usize::MAX so a large body can't be forced into memory.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/consumer_groups.rs
```diff
@@ -0,0 +1,168 @@
...
+/// non-empty body isn't valid JSON matching `CreateConsumerGroupRequest`.
+pub async fn create_consumer_group(
+    Extension(ctx): Extension<SecurityContext>,
+    Extension(state): Extension<HandlerState>,
+    uri: Uri,
+    request: Request,
+) -> ApiResult<impl IntoResponse> {
+    // [question]: don't we have named groups? How does the user specify named group? check in DESIGN.md; report
+
+    let body = axum::body::to_bytes(request.into_body(), usize::MAX)
```
MikeFalcon77: Same pattern as reset_producer: axum::body::to_bytes(..., usize::MAX) has no size cap, so a client can force an arbitrarily large body into memory. Bound it with an explicit limit.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub partition: Option<i32>,
+}
+
+impl From<ResetProducerRequest> for ProducerResetScope {
+    fn from(req: ResetProducerRequest) -> Self {
+        match (req.topic, req.partition) {
+            (Some(topic), Some(partition)) => {
+                ProducerResetScope::TopicPartition { topic, partition }
+            }
+            _ => ProducerResetScope::All,
```
MikeFalcon77: When a client sends only topic or only partition (not both), this silently falls back to ProducerResetScope::All instead of rejecting the request, so a partial/malformed request ends up wiping all producer state instead of failing with a validation error. Reject the case where exactly one of topic/partition is set.

gears/system/event-broker/event-broker/src/infra/specification.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    async fn register_event_type(&self, spec: EventType) -> Result<EventType, DomainError> {
+        crate::domain::specification::validate_allowed_subject_types(&spec.allowed_subject_types)?;
+        register_one(&self.client, &spec).await?;
+        Ok(spec)
+    }
+
+    async fn get_topic(&self, id: &GtsInstanceId) -> Option<Topic> {
+        let instance = self.client.get_instance(id.as_ref()).await.ok()?;
+        serde_json::from_value(instance.object).ok()
```
MikeFalcon77: get_topic collapses both a real types-registry error and a schema-mismatch deserialization failure into None via .ok(), and callers turn that into a client-facing 404 TopicNotFound. A real backend outage or a stored-record schema drift is indistinguishable from a genuine not-found, and nothing is logged either way. Return a Result so a backend failure surfaces as 500 instead of a false 404, or at least log before returning None.

gears/system/event-broker/event-broker/src/infra/specification.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    async fn get_event_type(&self, id: &GtsInstanceId) -> Option<EventType> {
+        let instance = self.client.get_instance(id.as_ref()).await.ok()?;
+        serde_json::from_value(instance.object).ok()
```
MikeFalcon77: Same issue as get_topic: get_event_type turns any types-registry error or deserialization failure into None with no logging, so publish_event reports a false EventTypeNotFound on backend failures.

gears/system/event-broker/event-broker/src/infra/specification.rs
```diff
@@ -0,0 +1,307 @@
...
+
+async fn list_all<T: serde::de::DeserializeOwned>(
+    client: &Arc<dyn TypesRegistryClient>,
+    type_id: &str,
+) -> Vec<T> {
+    let query = InstanceQuery::new().with_pattern(format!("{type_id}*"));
+    client
+        .list_instances(query)
+        .await
+        .unwrap_or_default()
```
MikeFalcon77: list_all uses .unwrap_or_default() on client.list_instances(...), so a types-registry failure silently returns an empty list instead of an error. GET /v1/topics or /v1/event-types would report zero results during a backend outage instead of failing loudly. Propagate the error instead of defaulting to empty.

gears/system/event-broker/event-broker/src/infra/dispatcher/forward.rs
```diff
@@ -0,0 +1,121 @@
...
+    req: Request,
+) -> Result<Response, CanonicalError> {
+    let name = role.service_name();
+
+    let instances = state
+        .cluster
+        .service_discovery
+        .discover(name, DiscoveryFilter::default())
+        .await
+        .map_err(|_| {
```
MikeFalcon77: The service_discovery.discover error is discarded via map_err(|_| ...) with no logging, so a real discovery failure (timeout, auth, backend outage) looks identical in logs and to the client to the case where no instance is registered. Log the original error before mapping to service_unavailable.

gears/system/event-broker/event-broker/src/infra/dispatcher/forward.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    // D8: uniformly random pick among returned instances - a placeholder for
+    // real load-balancing / consumer-group-ownership-aware routing, which
+    // lands with #4438. TODO(#4438): replace with the real algorithm.
+    let picked = &instances[rand::random_range(0..instances.len())];
+
+    proxy_client::proxy(&picked.address, &state.connector, req, state.idle_timeout)
+        .await
+        .map_err(|_| {
```
MikeFalcon77: proxy_client::proxy's error is discarded here too, with no logging, so operators get no information about why the resolved instance was unreachable. Log the error before converting to service_unavailable.

gears/system/event-broker/event-broker/src/infra/dispatcher/proxy_client.rs
```diff
@@ -0,0 +1,146 @@
...
+    addr: &str,
+    connector: &Connector,
+    req: Request,
+    idle_timeout: Duration,
+) -> Result<Response, ProxyError> {
+    let host_port = addr
+        .strip_prefix("http://")
+        .or_else(|| addr.strip_prefix("https://"))
+        .unwrap_or(addr);
+    let peer_addr: std::net::SocketAddr = host_port.parse().map_err(|_| ProxyError)?;
```
MikeFalcon77: ProxyError is a unit struct with no source, and every fallible step in proxy() (peer parse, get_http_session, header build, write, read, response parsing) is mapped with map_err(|_| ProxyError), so the real cause is discarded and nothing is logged anywhere in this function. When forwarding breaks in production this is undiagnosable beyond 'unreachable'. Give ProxyError a source per failure site, or at least log the underlying error before discarding it.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Maps one domain `Frame` to a multipart/mixed part (`stream_events`'s wire
+/// shape) - a `--boundary`/`Content-Type`/JSON-body/`\r\n` part, matching
+/// `event-broker-consumption-frames`.
+// `Body::from_stream` requires `Stream<Item = Result<_, E>>` - the `Result`
+// wrapper is a call-site contract, not a spurious `Ok`.
+#[allow(clippy::unnecessary_wraps)]
+fn frame_to_multipart_part(frame: Frame) -> Result<Bytes, Infallible> {
+    let dto = FrameDto::from(frame);
+    let json = serde_json::to_vec(&dto).unwrap_or_default();
```
MikeFalcon77: serde_json::to_vec(&dto).unwrap_or_default() silently emits an empty JSON body inside the multipart part on a serialization failure, corrupting the stream with no log and no client-visible signal. frame_to_sse_event further down at least falls back to a heartbeat event for the same failure - do the same here and log the error.

gears/system/event-broker/event-broker/src/domain/error.rs
```diff
@@ -1,41 +1,119 @@
...
-    /// `SpecificationManager::validate_event_data`, `IngestService::publish_event`).
-    /// Distinct from `Internal` so REST error mapping (#4346) can lift it to
-    /// `400`/`422` instead of `500`.
-    #[error("validation failed: {0}")]
-    Validation(String),
+    /// Client-supplied input failed validation - `api/rest/error.rs` lifts
+    /// this to `400`. `code` is the `docs/openapi.yaml` machine code (e.g.
+    /// `BadTypePattern`, `InvalidMode`, `InvalidPartition`).
+    #[error("{code}: {message}")]
+    Validation { code: &'static str, message: String },
```
MikeFalcon77, line 20: Validation/Forbidden/NotFound/Conflict/RateLimited all carry their machine code as a bare &'static str instead of a closed enum, so a typo'd code at any call site compiles fine and just falls through to a generic mapping at the REST boundary. A closed error code enum would catch that at compile time.

gears/system/event-broker/event-broker/src/api/rest/error.rs
```diff
@@ -0,0 +1,285 @@
...
+        "SubscriptionNotFound" => SubscriptionResourceError::not_found(message)
+            .with_resource(resource)
+            .create(),
+        "ConsumerGroupNotFound" => ConsumerGroupResourceError::not_found(message)
+            .with_resource(resource)
+            .create(),
+        "ProducerNotFound" => ProducerResourceError::not_found(message)
+            .with_resource(resource)
+            .create(),
+        _ => EventBrokerResourceError::not_found(message)
```
MikeFalcon77, line 52: not_found/forbidden/conflict dispatch on DomainError's string code via a match with a wildcard fallback, so a mistyped or newly added code silently maps to the generic EventBrokerResourceError instead of failing to compile or being caught by a test. This is the flip side of domain/error.rs carrying code as a plain string.

gears/system/event-broker/event-broker/src/domain/specification.rs
```diff
@@ -11,11 +12,71 @@ use crate::domain::model::{EventType, Topic};
...
+///
+/// # Errors
+/// Returns `DomainError::Internal` if `event_type.data_schema` itself is not
+/// a valid JSON Schema, or `DomainError::Validation { code: "SchemaViolation",
+/// .. }` if `data` fails validation against it.
+pub fn validate_against_schema(
+    event_type: &EventType,
+    data: &JsonValue,
+) -> Result<(), DomainError> {
+    let validator = jsonschema::validator_for(&event_type.data_schema).map_err(|err| {
```
MikeFalcon77, line 73: validate_against_schema calls jsonschema::validator_for on every call, and this runs on the hot publish_event/publish_batch path, so the schema gets recompiled from scratch for every published event instead of once per event type. Cache the compiled validator per event_type.id and reuse it.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/events.rs
```diff
@@ -0,0 +1,163 @@
...
+use crate::api::rest::state::HandlerState;
+use crate::domain::error::DomainError;
+use crate::domain::model::{Event, Meta};
+
+/// Publish-time transport-metadata block (`event.v1.schema.json`'s `meta`) -
+/// producer-protocol fields for chained/monotonic modes. Omit entirely for
+/// stateless publish.
+#[derive(Debug, Deserialize, Serialize, utoipa::ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MetaDto {
```
MikeFalcon77: MetaDto and the other request/response DTOs in this file are defined directly in the handler module instead of a shared api/rest/dto.rs, unlike other gears in this repo (chat-engine, oagw, mini-chat, account-management, resource-group, usage-collector) which keep DTOs in dto.rs and handlers with just handler functions plus From/TryFrom glue. Move the DTOs out to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::api::rest::state::HandlerState;
+use crate::domain::error::DomainError;
+use crate::domain::ingest::{
+    ProducerCursors, ProducerMode, ProducerRegistrationInput, ProducerResetScope,
+};
+
+#[derive(Debug)]
+#[toolkit_macros::api_dto(request)]
+#[serde(deny_unknown_fields)]
+pub struct RegisterProducerRequest {
```
MikeFalcon77: RegisterProducerRequest/Response and the cursor DTOs are declared here in the handler module rather than a shared api/rest/dto.rs, unlike the rest of the repo's gears. Move them to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/topics.rs
```diff
@@ -0,0 +1,108 @@
...
+use toolkit_odata::ast::Value as ODataValue;
+
+use crate::api::rest::pagination::{ListQuery, eval_filter, paginate};
+use crate::api::rest::state::HandlerState;
+use crate::domain::error::DomainError;
+use crate::domain::model::{Topic, TopicSegmentManifest};
+
+#[derive(Debug)]
+#[toolkit_macros::api_dto(response)]
+pub struct TopicDto {
```
MikeFalcon77: TopicDto and TopicSegmentsResponse are declared in the handler module rather than a shared api/rest/dto.rs. Move them to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/event_types.rs
```diff
@@ -0,0 +1,55 @@
...
+use toolkit::api::canonical_prelude::*;
+use toolkit_odata::ast::Value as ODataValue;
+
+use crate::api::rest::pagination::{ListQuery, eval_filter, paginate};
+use crate::api::rest::state::HandlerState;
+use crate::domain::model::EventType;
+
+#[derive(Debug)]
+#[toolkit_macros::api_dto(response)]
+pub struct EventTypeDto {
```
MikeFalcon77: EventTypeDto is declared in the handler module rather than a shared api/rest/dto.rs. Move it to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/consumer_groups.rs
```diff
@@ -0,0 +1,168 @@
...
+            ConsumerGroupKind::Anonymous => ConsumerGroupKindDto::Anonymous,
+            ConsumerGroupKind::Named => ConsumerGroupKindDto::Named,
+        }
+    }
+}
+
+#[derive(Debug)]
+#[toolkit_macros::api_dto(response)]
+// [todo]: what about TTL for the consumer group? check the DESIGN.md
+pub struct ConsumerGroupDto {
```
MikeFalcon77: ConsumerGroupKindDto, ConsumerGroupDto and CreateConsumerGroupRequest are declared in the handler module rather than a shared api/rest/dto.rs. Move them to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/subscriptions.rs
```diff
@@ -0,0 +1,370 @@
...
+        Self {
+            topic: a.topic.into_string(),
+            partition: a.partition,
+        }
+    }
+}
+
+#[derive(Debug)]
+#[toolkit_macros::api_dto(response)]
+pub struct SubscriptionDto {
```
MikeFalcon77: SubscriptionDto and its request/response companions are declared in the handler module rather than a shared api/rest/dto.rs. Move them to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming.rs
```diff
@@ -0,0 +1,244 @@
...
+    offset: i64,
+    last_examined: i64,
+}
+
+/// Wire shape for one frame (`event-broker-consumption-frames`): a JSON
+/// object with a top-level `kind` discriminant.
+#[derive(Debug)]
+#[toolkit_macros::api_dto(response)]
+#[serde(tag = "kind")]
+pub(crate) enum FrameDto {
```
MikeFalcon77: FrameDto, EventPayloadDto and PositionDto are declared in the handler module rather than a shared api/rest/dto.rs. Move them to api/rest/dto.rs.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers_tests.rs
```diff
@@ -0,0 +1,291 @@
...
+        .api_v1()
+        .post_producers()
+        .with_body(Json(&json!({ "mode": "chained", "client_agent": "a" })))
+        .send()
+        .await;
+    let id = register.json()["id"].as_str().unwrap().to_owned();
+
+    let resp = harness.api_v1().post_producer_reset(&id).send().await;
+
+    resp.assert_status(200);
```
MikeFalcon77, line 281: This test only checks the HTTP status and empty body. It never verifies that the producer's cursor/sequence state was actually cleared by the reset, for example via get_producer_cursors or a previously-rejected sequence now succeeding. Assert the actual reset side effect, not just the response shape.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers_tests.rs
```diff
@@ -0,0 +1,291 @@
...
+    let resp = harness
+        .api_v1()
+        .post_producer_reset(&id)
+        .with_body(Json(
+            &json!({ "topic": "gts.example.topic.v1~t1", "partition": 0 }),
+        ))
+        .send()
+        .await;
+
+    resp.assert_status(200);
```
MikeFalcon77, line 366: This test asserts only status and empty body and never confirms the reset was scoped to the given topic/partition instead of clearing everything, so it can't tell a correct scoped reset from an accidental full reset. Register cursors on two different topic/partition pairs, reset one, and assert the other is unchanged.

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: thx fixed

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
striped-zebra-dev: Fixed - switched to the axum Bytes extractor, which gets axum's implicit 2MiB body limit for free.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/consumer_groups.rs
striped-zebra-dev: Same fix - Bytes extractor, same implicit 2MiB cap.

gears/system/event-broker/event-broker/src/infra/dispatcher/proxy_client.rs
```diff
@@ -0,0 +1,146 @@
...
+    let peer = HttpPeer::new(peer_addr, false, String::new());
```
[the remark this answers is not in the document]
striped-zebra-dev: Confirmed - fix designed (async DNS resolution instead of the IP-literal-only parse), not applied yet. Tracking separately for now.

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: Fixed - logs the error now instead of swallowing it into a silent skip.

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: Fixed - same poll loop, logs before continuing instead of swallowing.

gears/system/event-broker/event-broker/src/infra/specification.rs
striped-zebra-dev: Fixed - logs the error before returning None.

gears/system/event-broker/event-broker/src/infra/specification.rs
striped-zebra-dev: Fixed - same fix, logs before returning None.

gears/system/event-broker/event-broker/src/infra/specification.rs
striped-zebra-dev: Fixed - logs before returning an empty list.

gears/system/event-broker/event-broker/src/domain/repo.rs
```diff
@@ -30,23 +44,57 @@ pub trait EventRepo: Send + Sync {
...
+                  client-side; needs pagination pushdown once a real backend \
+                  replaces the in-memory store"
+    )]
+    async fn list_consumer_groups(&self) -> Result<Vec<ConsumerGroup>, DomainError>;
+    async fn delete_consumer_group(&self, id: &str) -> Result<(), DomainError>;
+
+    /// Whether any subscription currently has an active membership in this
+    /// group - guards `DELETE /v1/consumer-groups/{id}`
+    /// (`ConsumerGroupHasActiveMembers`).
+    async fn has_active_members(&self, id: &str) -> Result<bool, DomainError>;
```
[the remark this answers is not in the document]
striped-zebra-dev: Already fixed by the tenant-isolation change - list_consumer_groups now filters anonymous groups by tenant and named groups by :consume.

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: Valid, but future development - not fixing now.

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: Same - future development.

gears/system/event-broker/event-broker/src/domain/error.rs
striped-zebra-dev: Valid, but a real redesign (per-kind closed enums across ~24 codes and every call site that constructs one), not a quick fix - and not a live bug, since the wire status/message are still correct either way, only the resource-type attribution for an uncovered code could be off. Future development.

gears/system/event-broker/event-broker/src/api/rest/error.rs
striped-zebra-dev: Same call as the domain/error.rs finding this mirrors - future development, tracked together.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/events.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/topics.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/event_types.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/consumer_groups.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/subscriptions.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming.rs
striped-zebra-dev: Valid pattern elsewhere in the repo, but this needs a broader discussion on the right module structure for this gear before moving anything - not a per-file fix in this PR. Future development.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/events.rs
```diff
@@ -0,0 +1,163 @@
...
+            subject_type: req.subject_type,
+            occurred_at: req.occurred_at,
+            trace_parent: req.trace_parent,
+            data: req.data.unwrap_or(JsonValue::Null),
+            meta: req.meta.map(|m| Meta {
+                version: m.version,
+                producer_id: m.producer_id.unwrap_or_default(),
+                previous: m.previous.unwrap_or(0),
+                sequence: m.sequence.unwrap_or(0),
+            }),
```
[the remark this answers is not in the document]
striped-zebra-dev: Future development - out of scope for this PR (REST API wiring, not producer-protocol field enforcement).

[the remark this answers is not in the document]
striped-zebra-dev: We already have all the tickets.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers_tests.rs
striped-zebra-dev: Fixed - both tests now register a topic, publish a chained event to build cursor state, and assert the actual cursor via GET .../cursors before/after reset, plus that a previously-would-be-412 sequence now succeeds post-reset.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers_tests.rs
striped-zebra-dev: Fixed - now publishes to two different topics under the same producer and asserts the scoped reset clears only the targeted (topic, partition), leaving the other topic's cursor unchanged.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming.rs
striped-zebra-dev: outdated

gears/system/event-broker/event-broker/src/domain/delivery.rs
striped-zebra-dev: No longer applies - delivery is pull-based. There's no `tokio::spawn` in `domain/` or `api/`; `domain/streaming/session.rs` produces frames on demand, driven by the axum response body. A disconnect drops the session, and `StreamLease`/membership release on `Drop`. Streaming never writes cursors either - the only delivery-side `put_cursor` is in `seek`.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
striped-zebra-dev: Fixed. The scope decision moved into `TryFrom<ProducerResetSelector> for ProducerResetScope` (`domain/ingest.rs`) with four explicit arms and no wildcard - `(Some(topic), None)` and `(None, Some(partition))` are both rejected as `Validation { InvalidBody }`.

gears/system/event-broker/event-broker/src/infra/dispatcher/forward.rs
striped-zebra-dev: Fixed. `forward.rs` logs `warn!(error = %err, gear = name, "directory resolution failed")` before mapping, and the two cases now differ on the wire too - `no {name} instance registered` vs `service discovery unavailable for {name} instances`.

gears/system/event-broker/event-broker/src/infra/dispatcher/forward.rs
striped-zebra-dev: Fixed. `warn!(error = %err, gear = name, "forwarding to resolved instance failed")` runs before the `ProxyError` match.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming.rs
striped-zebra-dev: Fixed. Both `to_multipart_part` and `to_sse_event` (now in `delivery/encode.rs`) emit a `tracing::error!` on the serialization failure before falling back. The empty-part vs heartbeat fallback is kept deliberately - a malformed part is recoverable, a truncated stream isn't - but it's no longer silent.

gears/system/event-broker/event-broker/src/api/rest/error.rs
striped-zebra-dev: Fixed. `DomainError::{Validation,Forbidden,NotFound,Conflict,RateLimited}` now carry `code: ErrorCode` (a closed enum), and the `api/rest/error.rs` builders take it, so a mistyped code is a compile error. One residual: a newly *added* `ErrorCode` variant still falls into the `_ =>` arms of the `resource_type`/`precondition_reason`/`wire_override` const fns - making those exhaustive is a separate, smaller hardening.

gears/system/event-broker/event-broker/src/domain/specification.rs
striped-zebra-dev: Fixed. `TypesRegistrySpecificationManager` caches the compiled validator keyed by event-type id plus a fingerprint of `data_schema`, so a batch compiles once per (type, schema) rather than per event, and a re-registered type self-invalidates when the fingerprint changes. The free fn is split into `compile_schema` + `check_against_schema` so the error mapping stays shared.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/producers.rs
striped-zebra-dev: Done. Every ingest DTO now lives in `api/rest/handlers/ingest/dto.rs` and `producers.rs` declares no types of its own. Split `ingest`/`delivery` rather than one `api/rest/dto.rs` because this crate has two distinct handler groups.

gears/system/event-broker/event-broker/src/domain/ingest.rs
```diff
@@ -37,4 +106,412 @@ pub trait IngestService: Send + Sync {
...
+        ctx: &SecurityContext,
+        events: Vec<Event>,
+    ) -> Result<BatchResult, DomainError> {
+        const MAX_BATCH: usize = 100;
+        if events.len() > MAX_BATCH {
+            return Err(DomainError::BatchTooLarge {
+                count: events.len(),
+                max: MAX_BATCH,
+            });
+        }
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `BatchConfig` is now wired: `max_size` replaces the hardcoded `MAX_BATCH` in `publish_batch`, and `max_payload_bytes` is enforced (new `BatchPayloadTooLarge` -> 413). Threaded through `module.rs` -> `wiring::build_handler_state` -> `IngestServiceImpl`.

gears/system/event-broker/event-broker/src/infra/specification.rs
```diff
@@ -0,0 +1,307 @@
...
+#[async_trait]
+impl SpecificationManager for TypesRegistrySpecificationManager {
+    async fn register_topic(&self, spec: Topic) -> Result<Topic, DomainError> {
+        register_one(&self.client, &spec).await?;
+        Ok(spec)
+    }
+
+    async fn register_event_type(&self, spec: EventType) -> Result<EventType, DomainError> {
+        crate::domain::specification::validate_allowed_subject_types(&spec.allowed_subject_types)?;
+        register_one(&self.client, &spec).await?;
```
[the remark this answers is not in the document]
striped-zebra-dev: No longer applies. Event-broker doesn't register event types - `register_one` is gone. It reads them and caches a projection whose persisted field is `topic: GtsInstanceId`, matching both the SDK model and the committed schema (`required: [topic]`). No `topic_id` remains in the models or the projection.

gears/system/event-broker/event-broker/src/api/rest/routes/routes_tests.rs
```diff
@@ -0,0 +1,623 @@
...
+
+    assert!(
+        drained.is_ok(),
+        "the idle-timeout must close the connection well within 2s of the last heartbeat, \
+         not hang forever"
+    );
+    assert!(
+        heartbeat_bytes >= 1,
+        "expected at least one heartbeat byte to arrive before the connection closed"
+    );
```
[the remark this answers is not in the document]
striped-zebra-dev: Done - `assert_eq!(heartbeat_bytes, 3, ...)` now, so an idle-timeout that fired after the first heartbeat would fail the test.

gears/system/event-broker/event-broker/src/config.rs
```diff
@@ -89,6 +114,29 @@ impl Default for SubscriptionConfig {
...
+    pub heartbeat_interval_secs: u32,
+}
+
+impl Default for StreamingConfig {
+    fn default() -> Self {
+        Self {
+            heartbeat_interval_secs: 5,
+        }
+    }
+}
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed, though the mechanism differs from the suggestion: the value never reaches `tokio::time::interval`. It's a `HeartbeatSchedule` deadline consumed by `sleep_until`, so `0` doesn't panic - it makes the schedule permanently due and spins the session loop (the same failure the progress timer already documents). Rejected at deserialization via a `deserialize_with` on both `heartbeat_interval_secs` and `progress_min_interval_secs`, with config tests.

gears/system/event-broker/event-broker/src/infra/dispatcher/proxy_client.rs
```diff
@@ -0,0 +1,242 @@
...
+        .map_err(ProxyError::Session)?;
+    if !body_bytes.is_empty() {
+        session
+            .write_body(&body_bytes)
+            .await
+            .map_err(ProxyError::Session)?;
+    }
+    session.finish_body().await.map_err(ProxyError::Session)?;
+
+    session.read_response().await.map_err(ProxyError::Session)?;
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `proxy()` sets `peer.options.connection_timeout`/`total_connection_timeout` for the connect, and `session.read_timeout`/`write_timeout` for the header phase and each read. One gotcha worth noting: this connector builds the session with `new_with_options`, which does *not* copy the peer's read/write timeouts, so they have to be set on the session directly. Covered by an accept-but-never-respond test.

gears/system/event-broker/docs/DESIGN.md
```diff
@@ -736,6 +752,19 @@ Key methods:
...
+(`#[gts_type_schema(..., base = true)]`), discoverable via the process-wide `toolkit-gts` inventory
+and auto-seeded into `types-registry` at startup. `SpecificationManager` itself has a real,
+`types-registry`-backed production implementation (`infra/specification.rs`'s
+`TypesRegistrySpecificationManager`, resolving `TypesRegistryClient` via `ClientHub`) - it queries
+`register_instances`/`get_instance`/`list_instances` directly per call, with no caching,
+hash-based change detection, or indexed UUID lookup yet (all still aspirational, as described above).
+This supersedes the previously-stale `infra/type_provisioning.rs`/`#4346`/`#4347` pointer (that file
+was a dead, never-called `todo!()` stub and has been deleted - those tickets never actually tracked
+this work). `TopicRepo`/`EventRepo` (event/cursor storage, a separate concern from topic/event-type
+*metadata*) remain `InMemoryDomainRepo`-backed, pending `#4347`/`#4348`'s real storage-backend work.
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. Rewrote the storage-status paragraph: cursors are durable and SeaORM-backed (`Storage`'s `impl CursorRepo`), events go through the resolved `EventBrokerBackend`, and the old `TopicRepo`/`EventRepo` traits and `InMemoryDomainRepo` no longer exist. Also corrected the `repo.rs` trait listing to `SubscriptionRepo, CursorRepo, ConsumerGroupRepo`.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/streaming_tests.rs
```diff
@@ -0,0 +1,883 @@
...
+            .with_body(Json(&json!({
+                "partition_positions": parts.iter().map(|&p| json!({
+                    "topic": "gts.cf.core.events.topic.v1~x.eb.t1.topic.v1",
+                    "partition": p,
+                    "value": "earliest",
+                })).collect::<Vec<_>>()
+            })))
+            .send()
+            .await;
+    }
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. The loop now reads each subscription's live `assigned` set and seeks only those partitions, and chains `.assert_status(200)`, so a rejected seek fails the test rather than passing silently.

gears/system/event-broker/event-broker/src/api/rest/handlers/delivery/subscriptions.rs
```diff
@@ -0,0 +1,176 @@
...
+/// Returns the mapped `CanonicalError` for `DomainError::NotFound` if the
+/// subscription doesn't exist.
+pub async fn leave_subscription(
+    Extension(ctx): Extension<SecurityContext>,
+    Extension(state): Extension<HandlerState>,
+    Path(id): Path<Uuid>,
+) -> ApiResult<StatusCode> {
+    state.delivery.leave(&ctx, id).await?;
+    Ok(StatusCode::NO_CONTENT)
+}
```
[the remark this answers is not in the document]
striped-zebra-dev: Production doesn't return plain text here - the gear is wrapped in toolkit's `canonical_error_middleware`, which converts any untyped/`text/plain` error into an RFC 9457 `Problem`, so a malformed UUID comes back as `400 application/problem+json` (generic `about:blank`). Turning it into a typed `invalid_path_params` violation means adopting `toolkit::api::rest::extract::Path`, which is the house-wide item tracked as #4596.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/dto.rs
```diff
@@ -0,0 +1,314 @@
...
+impl From<ResetProducerRequest> for ProducerResetScope {
+    fn from(req: ResetProducerRequest) -> Self {
+        match (req.topic, req.partition) {
+            (Some(topic), Some(partition)) => {
+                ProducerResetScope::TopicPartition { topic, partition }
+            }
+            _ => ProducerResetScope::All,
+        }
+    }
+}
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `ResetProducerRequest -> ProducerResetSelector` is now a plain field copy; the scope decision lives in `TryFrom<ProducerResetSelector> for ProducerResetScope` with four explicit arms and no wildcard, so both half-specified bodies are `Validation { InvalidBody }`.

gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/events.rs
```diff
@@ -0,0 +1,67 @@
...
+    Query(query): Query<SyncWaitQuery>,
+    Json(req): Json<PublishEventRequest>,
+) -> ApiResult<StatusCode> {
+    state.ingest.publish_event(&ctx, req.try_into()?).await?;
+    Ok(if wants_sync_wait(&headers, &query) {
+        StatusCode::CREATED
+    } else {
+        StatusCode::ACCEPTED
+    })
+}
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - the handler never returns 201. `publish_event` returns `202 Accepted`, or `200` for an idempotent duplicate; `Prefer: wait` is an explicit `501 Not Implemented`, not a silent 201. openapi documents exactly that.

gears/system/event-broker/event-broker/src/domain/consumer_group_coordinator/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            for &p in &splits[i] {
+                entries.push(Assignment {
+                    topic: topic_id.clone(),
+                    partition: p as i32,
+                    offset: 0,
+                    last_examined: 0,
+                });
+            }
+        }
+    }
```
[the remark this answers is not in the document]
striped-zebra-dev: Handled. The conversion is `range_split(n_partitions.unsigned_abs(), ...)` - no `as u32` - and `partitions` is resolved through `domain/resolution.rs`, which rejects `< 1`, so a topic whose settings fail to resolve never enters the spec cache.

gears/system/event-broker/event-broker/src/domain/delivery.rs
```diff
@@ -51,24 +179,800 @@ pub trait DeliveryService: Send + Sync {
...
+                                offset: sequence,
+                            })
+                            .await
+                        {
+                            tracing::warn!(?err, "failed to persist session cursor after delivery");
+                        }
+                        delivered_any = true;
+                        last_frame_at = tokio::time::Instant::now();
+                    }
+                }
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `stream` compiles `InterestFilter::compile(&subscription.interests)` per open and applies it per event; non-matching events still advance `last_examined`/`undelivered` in the read set, so the cursor doesn't stall on filtered spans. The expression filter (`Interest.filter`) is carried but deliberately not evaluated yet - that engine is ADR-0005's.

gears/system/event-broker/event-broker/src/infra/partition_cache/budget.rs
```diff
@@ -0,0 +1,517 @@
...
+    /// `demand` must be the same slice, in the same order, that produced this
+    /// allocation - the pairing is positional, and a mismatched slice silently
+    /// returns a wrong total.
+    #[must_use]
+    pub fn committed_bytes(&self, demand: &[ExtentDemand]) -> usize {
+        self.grants
+            .iter()
+            .zip(demand.iter())
+            .map(|(grant, want)| want.bytes_for(grant.runway_events()))
+            .fold(0, usize::saturating_add)
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. `committed_bytes` now looks each demand's grant up by key via `runway_for(key, segment_from)` instead of `zip`, so order no longer matters and an unmatched demand contributes nothing - the length/reorder hazard is structurally gone, which is why the length-mismatch/reordered tests aren't needed.

gears/system/event-broker/event-broker/src/infra/partition_cache/runway.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+fn clamp_to_bounds(target: usize, policy: &RunwayPolicy, scanning: bool) -> usize {
+    let ceiling = if scanning {
+        policy.scanner_cap_events
+    } else {
+        usize::MAX
+    };
+    target.max(policy.floor_events).min(ceiling)
+}
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - `policy.scanner_cap_events.max(policy.floor_events)`, so a cap below the floor is raised to the floor. Added a `scanner_cap_events < floor_events` test.

gears/system/event-broker/plugins/sqlite-event-broker-plugin/src/backend.rs
```diff
@@ -0,0 +1,323 @@
...
+            .secure()
+            .scope_with(&AccessScope::allow_all())
+            .all(&conn)
+            .await
+            .map_err(|e| read_failed(e.to_string()))?
+            .into_iter()
+            .map(|r| r.partition)
+            .collect::<std::collections::BTreeSet<_>>()
+            .into_iter()
+            .collect();
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - reads `partition_state` (one row per `(topic, partition)`) instead of scanning the event log. One deliberate difference, pinned with a test: a `partition_state` row survives retention emptying its partition, so a pruned-empty partition stays listed as led - which is the more correct answer for a *leader* listing.

gears/system/event-broker/event-broker/tests/standalone_integration_tests.rs
```diff
@@ -0,0 +1,551 @@
...
+        if segments["end_sequence"] == json!(1) {
+            break;
+        }
+        tokio::time::sleep(Duration::from_millis(20)).await;
+    }
+    assert_eq!(
+        segments["end_sequence"],
+        json!(1),
+        "exactly one event must have drained from the outbox: {segments}"
+    );
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - the poll is now a 10s deadline loop, matching this suite's other waits (5s frame timeouts, 15s `wait_for_healthy`), instead of the ~1s (50x20ms) budget.

testing/e2e/suites/event_broker/test_stream_guardrails.py
```diff
@@ -0,0 +1,91 @@
...
+    whether the subscription exists, never 406).
+    """
+    sub_id = str(uuid.uuid4())
+    async with api() as client:
+        resp = await client.get(
+            f"/events:stream?subscription_id={sub_id}",
+            headers={"Accept": "*/*"},
+        )
+
+    assert resp.status_code != 406
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - tightened to `== 404` (a fresh-UUID subscription misses -> `NotFound` -> 404 deterministically; the 409 paths need an existing subscription), plus `status`/`type` body asserts, and dropped the `!= 406`/'or 409' hedge.

gears/system/event-broker/plugins/sqlite-event-broker-plugin/src/backend.rs
```diff
@@ -0,0 +1,323 @@
...
+                    }
+
+                    let mut new_last_chain_sequence = last_chain_sequence;
+                    let mut models = Vec::with_capacity(events.len());
+                    for event in &events {
+                        let sequence = next_sequence;
+                        next_sequence += 1;
+                        if let Some(chain_sequence) = event.meta.as_ref().and_then(|m| m.sequence) {
+                            new_last_chain_sequence = Some(chain_sequence);
+                        }
```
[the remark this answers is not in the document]
striped-zebra-dev: Real but latent, and deliberately deferred. The only production caller (`ingest_outbox`) always persists exactly one event, and the single-event path dedups correctly; no caller passes a chained multi-event batch. The batch-level dedup and the last-vs-max watermark belong to the storage-backend redesign, so we're not changing the sqlite backend for it now. Leaving this open as a tracker for that work.

gears/system/event-broker/event-broker/src/infra/workers/ingest_outbox.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        let Some(partition) = event.partition else {
+            error!(topic = %event.topic, "ingest outbox payload has no stamped partition");
+            return MessageResult::Reject("missing partition".to_owned());
+        };
+
+        let Some(topic) = self.spec_manager.get_topic(&event.topic).await else {
+            error!(topic = %event.topic, "ingest outbox payload references a topic that is no longer registered");
+            return MessageResult::Reject(format!("topic '{}' is not registered", event.topic));
+        };
```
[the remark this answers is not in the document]
striped-zebra-dev: Acknowledged and deliberately deferred. For now `Reject` (dead-letter) is the accepted behaviour, and a code comment at the call site records that a `None` conflates 'topic deregistered' with 'spec-cache read failed'. Distinguishing a transient read failure and retrying it belongs to the backend-integration design, which is out of scope for this PR. Leaving open as a tracker.

gears/system/event-broker/plugins/sqlite-event-broker-plugin/src/backend.rs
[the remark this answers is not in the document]
striped-zebra-dev: Resolving - this will be addressed in future development (the backend redesign / backend-integration work).

gears/system/event-broker/event-broker/src/infra/workers/ingest_outbox.rs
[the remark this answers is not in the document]
striped-zebra-dev: Resolving - this will be addressed in future development (the backend redesign / backend-integration work).

PR #4443, opened 2026-08-07
fix(toolkit): emit array responses inline instead of a `Vec` component

MikeFalcon77:
- utoipa's default `ToSchema::name()` strips generic arguments, so every `Vec<T>` resolves to the single component name `Vec`. Four list endpoints across gear-orchestrator, nodes-registry and oagw register a `Vec<T>` response into the same registry, so they all wrote to `components.schemas.Vec` and the last one won. The committed spec shows the damage: `Vec` holds oagw's RouteResponse array shape, all four endpoints `$ref` it, and `GearDto` is absent entirely — `Vec<T>::schemas()` forwards to `T::schemas()`, which pushes T's fields but never T itself.

- Top-level arrays are now emitted inline — `{type: array, items: {$ref: T}}` — with only the item type registered as a named component. That is the OpenAPI norm (cf. the OAI Petstore example), what utoipa's own `#[utoipa::path]` produces for a `Vec<T>` body, and what lets client generators emit `Vec<Pet>` rather than a wrapper type named after the container.

  - `ResponseSpec.schema_name: Option<String>` becomes `schema: Option<ResponseSchema>`, mirroring the `RequestBodySchema` enum already used on the request side. `ResponseSpec::schema_name()` keeps the read sites a one-word change. Adding any field would have broken the same 22 construction sites, all in-repo, so the enum costs nothing extra and rules out the illegal `(None, is_array: true)` state.
  - New `OperationBuilder::json_array_response_with_schema::<Item>()` on both response states.
  - `ensure_schema::<Vec<_>>()` now asserts: `Vec` is a std type, so `#[schema(as = ...)]` cannot disambiguate it and callers must use the array method. Guarded on that one name only — a real DTO could be called `Option` or `Map`.

- Also turns a schema-name collision into a hard error at registration instead of a warn-and-override, and fixes the one other collision that surfaced: users-info and account-management both ship a `UserDto`. The example's is aliased to `UsersInfoUserDto`, leaving the production component name (6 refs in the spec) untouched.

Spec regenerated: `Vec` removed, `GearDto` and `UsersInfoUserDto` added, no dangling refs. Seven endpoints' 200 responses change because the type they document was wrong, not because the wire format moved — the servers have always returned these shapes.

## Summary by CodeRabbit

- **Documentation**
  - Updated API documentation with explicit schemas for gears, nodes, routes, upstreams, and user information.
  - List endpoints now clearly describe their typed array responses.
  - User information endpoints now document dedicated user and pagination response formats.

- **Bug Fixes**
  - Improved API response definitions for more accurate schema validation and clearer integration expectations.
  - Preserved existing response behavior while resolving ambiguous generic collection descriptions.

### DIFF
--- a/docs/api/api.json
+++ b/docs/api/api.json
@@ -1252,6 +1252,62 @@
         ],
         "type": "object"
       },
+      "GearDto": {
+        "description": "Response DTO for a single registered gear",
+        "properties": {
+          "capabilities": {
+            "description": "Declared capabilities (e.g., \"rest\", \"grpc\", \"system\", \"db\")",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
+          "dependencies": {
+            "description": "Gear dependencies (other gear names)",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
+          "deployment_mode": {
+            "$ref": "#/components/schemas/DeploymentModeDto",
+            "description": "Whether the gear is compiled-in or out-of-process"
+          },
+          "instances": {
+            "description": "Running instances of this gear",
+            "items": {
+              "$ref": "#/components/schemas/GearInstanceDto"
+            },
+            "type": "array"
+          },
+          "name": {
+            "description": "Gear name",
+            "type": "string"
+          },
+          "plugins": {
+            "description": "Plugins provided by this gear (reserved for follow-up implementation)",
+            "items": {
+              "$ref": "#/components/schemas/PluginDto"
+            },
+            "type": "array"
+          },
+          "version": {
+            "description": "Gear version (if reported by a running instance)",
+            "type": [
+              "string",
+              "null"
+            ]
+          }
+        },
+        "required": [
+          "name",
+          "capabilities",
+          "dependencies",
+          "deployment_mode",
+          "instances"
+        ],
+        "type": "object"
+      },
       "GearInstanceDto": {
         "description": "Response DTO for a running gear instance",
         "properties": {
@@ -2884,6 +2940,24 @@
           "items": {
             "items": {
               "$ref": "#/components/schemas/UserDto"
+            },
+            "type": "array"
+          },
+          "page_info": {
+            "$ref": "#/components/schemas/PageInfo"
+          }
+        },
+        "required": [
+          "items",
+          "page_info"
+        ],
+        "type": "object"
+      },
+      "Page_UsersInfoUserDto": {
+        "properties": {
+          "items": {
+            "items": {
+              "$ref": "#/components/schemas/UsersInfoUserDto"
             },
             "type": "array"
           },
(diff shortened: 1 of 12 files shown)


MikeFalcon77:
## Why this PR reports 43 breaking changes

All reported breaking changes are corrections to **wrong documentation**. No server
behaviour changes: `json_response_with_schema::<T>()` only ever used `T` to build the
OpenAPI schema — handlers still return the same `Json(Vec<...>)` they always did. Not a
single line of response-serialization code is touched in this PR.

### Root cause

utoipa's default `ToSchema::name()` strips generic arguments:

```rust
type_name::<Vec<GearDto>>()  // "alloc::vec::Vec<...GearDto>"
  .split_once("<")           // "alloc::vec::Vec"
  .rsplit_once("::")         // "Vec"
```

So every `Vec<T>` registers under the single component name `Vec`. Four list endpoints in
the example server did exactly that, and the registry's old policy was warn-and-override —
the last gear to register won, and the other three endpoints silently `$ref`'d a schema
describing someone else's type.

On `main` today, `components.schemas.Vec` holds oagw's `RouteResponse` array shape, four
endpoints point at it, and `GearDto` is not in the spec at all (`Vec<T>::schemas()` forwards
to `T::schemas()`, which pushes `T`'s fields but never `T` itself).

### The report is the shape of the bug

Read what oasdiff says was "removed":

| Endpoint | Properties "removed" | What they actually are |
|---|---|---|
| `GET /gear-orchestrator/v1/gears` | `match`, `upstream_id`, `priority`, `enabled`, `tenant_id`, `cors`, `rate_limit`, `tags` | fields of **oagw's `RouteResponse`** |
| `GET /nodes-registry/v1/nodes` | same set | fields of **oagw's `RouteResponse`** |
| `GET /oagw/v1/upstreams` | `match`, `priority`, `upstream_id` | fields of **`RouteResponse`** — oagw collided with itself |
| `GET/POST/PATCH /users-info/v1/users*` | `username`, `first_name`, `last_name`; `email`/`display_name` stop being nullable | fields of **account-management's `UserDto`** |

The gears list was documented as returning routes. The nodes list was documented as
returning routes. The upstreams list was documented as returning routes. None of them ever
did. oasdiff has no way to know the old type was wrong, so it classifies the correction as
a breaking change.

### Second collision

`users-info` and `account-management` both define a `UserDto`, and both mount into the same
server. The published spec currently carries account-management's version (6 refs), so the
example's is aliased to `UsersInfoUserDto` via `#[schema(as = ...)]`. The production
`UserDto` component is byte-identical to `main`.

### What changed structurally

Top-level arrays are now emitted inline, referencing the item type:

```json
"schema": { "type": "array", "items": { "$ref": "#/components/schemas/GearDto" } }
```

instead of `$ref: "#/components/schemas/Vec"`. This is the OpenAPI norm (cf. the OAI
Petstore example) and what utoipa's own `#[utoipa::path]` produces for a `Vec<T>` body. It
also means client generators emit `Vec<GearDto>` / `[]GearDto` rather than a wrapper type
named after the container.

Net spec diff: `Vec` removed, `GearDto` and `UsersInfoUserDto` added, no dangling refs.

aviator5 (approved):

PR #4444, opened 2026-08-07
perf(resource-group,toolkit-db): fix every N+1 and redundant read, with the audit suite that guards them (VHP-2344)

ffedoroff:
Every N+1 and redundant-I/O finding of the `resource-group` DB-behaviour audit, and the test layer that keeps them fixed.

Lifted from `rg/triage-fixes` **by line, not by cherry-pick**: none of that branch's tenant scoping, authorization gates, error mapping or API changes come along.

> **Rescoped.** An earlier revision of this PR also carried transaction boundaries, retry, isolation and a metrics layer. Those moved to a follow-up on `perf/db-tx-isolation`, which is this branch plus 12 commits. This PR is now the query-count work alone.

---

## The write paths, and what they cost

**Closure maintenance was the expensive one.** `rebuild_subtree_closure` issued one `INSERT` per `ancestors × subtree` row: moving a 10 000-node subtree under a depth-10 parent was roughly 100 000 separate statements. Batching removed the statement count but not the rows — the process still built that many `ActiveModel`s, sent them, and held a `SERIALIZABLE` transaction open for the round-trip.

The rows are a function of rows the database already holds, so it computes them itself now: one `INSERT ... SELECT` over a cross product, and one set-based `DELETE` whose two predicates are the same subquery. `A × N` rows no longer cross the wire, and this gear's largest unbounded `IN (...)` list went with them. The rest are chunked against `max_bind_params_for`; the two that are not are bounded by `SMALLINT`'s domain and are named in the audit document as such, rather than left to be rediscovered.

`insert_ancestor_closure_rows` likewise — two statements to one, both of which sat inside the transaction a create holds open.

**Force delete** ran three statements per node. Memberships and closure rows have no ordering constraint among themselves, so each is one statement for the whole subtree; group rows do — a parent cannot precede its children — so they go depth level by depth level, deepest first. The count follows tree depth, not node count.

**Type junctions** inserted one row at a time, holding the enclosing transaction open for the length of the request.

## Reads and validation

**The move depth check** asked the database, per descendant, whether it is a descendant and how deep it sits. Both answers were already in the closure rows one query returns.

**`list_types`** resolved junction references per row, so a page of N types cost `2N+1` queries — the only finding on a read path.

**Type paths in `$filter`, and the removed-parent sweep** cost one `gts_type` SELECT per value. Both now collect, resolve once, and substitute in memory. The rejection message keeps its old ordering.

**Rows the database already returned.** `insert` discarded the model `secure_insert` hands back and re-read the row. `update` reported a row count, so the repository read the row straight back — and both callers threw that model away and read a *third* time to build the response; `update` returns `()` now and the callers assemble from what they wrote. A move read the new parent twice, once for the type check and once for the cross-tenant check; a delete read the group twice, discarding the second.

Measured with the audit recorder — move of a 15-node subtree: `resource_group` SELECTs **5 → 2**, `gts_type` **5 → 4**. Create of a root group: `resource_group` **2 → 1**, `gts_type` **2 → 1**, and the one that remains is SeaORM's non-RETURNING fallback, which PostgreSQL does not issue at all.

---

## New in `toolkit-db`

- `secure_insert_many` — batch insert keeping `secure_insert`'s per-row scope validation, chunked against the per-backend bind ceiling.
- `max_bind_params_for(runner)` — that ceiling, so callers building their own `IN (...)` chunk by the same number instead of discovering it as a driver error.
- `secure_insert_from_select` — the set-based counterpart. Rows produced by a `SELECT` inside the database cannot be scope-validated per row, so it is restricted to entities declaring **no** scope columns, and it builds the `INSERT` itself from `E`. An earlier revision took a prepared statement and checked `E` separately, which let a caller satisfy the guard and defeat it in one call.
- Jittered backoff on the transaction retry loop, and give-up logging that reports the error *class*: both `DbErr`'s `Display` and `SqlErr`'s payload embed conflicting values on PostgreSQL and MySQL, and this is shared infrastructure.

## What guards it

`toolkit-db::test_support` — a `QueryRecorder` behind a `test-support` feature, attached through SeaORM's metric callback, reading transaction membership from the task-local guard rather than parsing SQL. Operations that could scale with input run at two sizes and the **slope** is asserted; an absolute count rots on the next refactor.

`assert_closure_matches_parent_links` asserts the whole closure table equals the transitive closure of the `parent_id` links, depths included. It does not encode what a rebuild *should* write; it holds whatever wrote it to the hierarchy the group rows describe. Both it and the exact-row assertions were validated by breaking the depth arithmetic and confirming they fail.

That discipline found a coverage gap on the way: the cross-tenant check on `update_group`'s path had **no** test — deleting it left the whole suite green. Two tests now pin it, on both entry points.

## Risk

No schema change, no migration, no API change. Every rewrite replaces N statements with one that reads or writes the same rows. The scope check inside `secure_insert_many`, and the entity restriction on `secure_insert_from_select`, are the security-relevant lines.

**Two observable differences, and neither is in the contract.** `PUT /groups/{id}` builds its answer from the request instead of re-reading the row, so `metadata` comes back exactly as sent — and `PUT /types/{code}` does the same with `metadata_schema`. The columns are `JSONB`, and PostgreSQL normalizes on the way in — duplicate keys collapse, key order is the server's, numbers are canonical — while the echoed request is not (this workspace resolves `serde_json` with `preserve_order`, so key order survives to the response). A response that previously matched the following `GET` byte for byte is no longer guaranteed to. Group create and move are unaffected: both take the value from the row. Restoring the reads would restore the match, at the cost of the thing this branch is for.

**Formally semver-breaking for `cf-gears-resource-group`.** The crate publishes to crates.io with `semver_check = true`. `GroupRepositoryTrait::update` returns `()`, `TypeRepositoryTrait::update_type` returns its timestamp rather than the row, `find_by_code_with_id` returns the row rather than its id, and several methods are added without default implementations. Practically the risk is nil — the traits are infrastructural and have one implementor each — but release-plz will ask for a minor bump, and that is correct.

## Reviewed against `main` by four engines

codex, fable, opus and DeepSeek read the whole diff independently, with no access to each other's findings. What they surfaced is fixed in the last commit: the read path's unbounded `IN (...)` on `GET /groups/{id}/descendants`, list literals from the request body binding unchunked, a shared-budget calculation that overran whenever one of its two lists did, two writes still reading themselves back, two assertions claiming a fix this branch does not carry, and a static rule that passed because it read only the file that never had the defect. Dangling references to Postgres suites that live on the follow-up branch went too — including one from `toolkit-db`, which ships to crates.io — along with an `integration` feature gating nothing and four dev-dependencies no test names.

Three things I had told the reviewers turned out to be wrong and are corrected in that commit's message: the closure helpers' return types, whether this crate is published, and whether the remaining `IN (...)` lists were bounded by construction.

## Verification

- `cargo test -p cf-gears-resource-group -p cf-gears-toolkit-db` — **653 passed, 0 failed**. The reported ignored are documentation examples; no test function carries `#[ignore]`.
- `cargo fmt --all --check` and `clippy --all-targets -D warnings` clean.
- CDSL markers unchanged in every file touched.

- PostgreSQL smoke: **3/3** against testcontainers PostgreSQL (`--features integration`, `make test-rg-pg`), also wired into CI with `RG_PG_REQUIRE_DOCKER=1` — the `INSERT ... SELECT` cross product, the derived-table `DELETE` and the force-delete path now run against the production backend in this branch's own CI, not only after merge.

**Not verified locally:** `cfs` 1.5.10 fails locally at `validate-kits` on an out-of-date engine, so `Validate CFS Artifacts` is CI-only.


gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -916,101 +1084,117 @@ impl GroupRepositoryTrait for GroupRepository {
...
+        // The DELETE runs before the INSERT below and cannot disturb either of
+        // its sources: it only removes rows whose ancestor is *outside* the
+        // subtree, while the `st` source ranges over `ancestor_id = group_id`
+        // (inside by definition) and the `pa` source over
+        // `descendant_id = parent_id`, whose ancestors are all outside --
+        // guaranteed by the `is_descendant` check the service makes before
+        // calling this.
+        ClosureEntity::delete_many()
+            .filter(closure_entity::Column::DescendantId.in_subquery(subtree_query.clone()))
+            .filter(closure_entity::Column::AncestorId.not_in_subquery(subtree_query))
```
Artifizer: _🩺 Stability & Availability_ | _🟠 Major_ | _Needs changes_

**Avoid a same-target-table subquery in this `DELETE`; it breaks the MySQL backend.**

This deletes from `resource_group_closure` while both predicates use subqueries that also read `resource_group_closure`. MySQL rejects this shape with the classic target-table restriction (`You can't specify target table ... for update in FROM clause`). The generic helper documentation added in this PR even calls out that `DELETE` with a subquery naming the target table is PostgreSQL/SQLite-only.

Resource Group storage is otherwise backend-agnostic, so this makes `move_group` / `update_group` parent changes fail on MySQL. Please rewrite this as a MySQL-safe form, for example by materializing the subtree into a derived table / temp table per backend support, or fall back to the previous materialized-id delete path for MySQL while keeping the set-based path for backends that support it.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
```diff
@@ -40,6 +40,143 @@ fn system_scope() -> AccessScope {
...
+            .map(|r| r.parent_type_id)
+            .chain(membership_rows.iter().map(|r| r.membership_type_id))
+            .collect();
+
+        let code_by_id: std::collections::HashMap<i16, String> = if referenced_ids.is_empty() {
+            std::collections::HashMap::new()
+        } else {
+            let ids: Vec<i16> = referenced_ids.into_iter().collect();
+            GtsTypeEntity::find()
+                .filter(gts_type::Column::Id.is_in(ids))
```
Artifizer: _🩺 Stability & Availability_ | _🟠 Major_ | _Needs changes_

**Chunk the referenced type-id lookup; this new batch path can exceed backend bind limits.**

`referenced_ids` is built from every allowed-parent and allowed-membership junction row for the whole returned page. Those relationship lists are user/data controlled and can be much larger than the page size, but this query binds the whole set in one `IN (...)`. A sufficiently connected type registry will fail at the driver once the list exceeds the backend parameter limit.

Please chunk `ids` using `toolkit_db::secure::max_bind_params_for(db)` and merge the resulting `id -> schema_id` maps, matching the chunking pattern used elsewhere in this PR.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: Fixed in 8531f0e74 via the derived-table form: each predicate now wraps the subtree subquery in `from_subquery` (`st_desc` / `st_anc`), which MySQL materializes before deleting (lifting the ER 1093 restriction) while PostgreSQL/SQLite accept either form unchanged — so the delete stays set-based on every backend instead of falling back to materialized ids. The insert-from-select helper's rustdoc now names this workaround too. The audit suite and group service tests (which exercise `move_group`) stay green.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
ffedoroff: Fixed in 3453298ef: the id -> schema_id resolution now chunks `ids` with `toolkit_db::secure::max_bind_params_for(db)` and merges the maps, matching the chunking pattern used elsewhere in this branch. The `is_empty` special case became unnecessary (an empty list yields no chunks and no query), and `load_full_types_batch`'s rustdoc now spells out that the referenced set is data-controlled rather than page-controlled.

libs/toolkit-db/src/secure/db_ops.rs
```diff
@@ -202,6 +202,235 @@ where
...
+/// Nothing here can derive it -- the rows never enter this process -- so a
+/// caller that needs the figure (a metric, a budget check) has to be given
+/// it.
+///
+/// # Errors
+///
+/// - `ScopeError::Invalid` if `E` declares any scope column, or if the target
+///   column count does not match the source `SELECT`.
+/// - `ScopeError::Db` if the statement fails.
+pub async fn secure_insert_from_select<E, C>(
```
MikeFalcon77, line 391: **HIGH**

`secure_insert_from_select` takes no `AccessScope` and performs no scope check on any path.

The guard `insert_from_select_allowed::<E>()` only proves `E` declares no scope columns — but for exactly that class of entity the per-row helpers are *not* permissive: `#[secure(no_tenant, no_resource, no_owner, no_type)]` sets `IS_UNRESTRICTED = false`, so under a deny-all scope `validate_insert_scope` returns `Denied`, and under a constrained scope every filter's `resolve_property` yields `None`, so all constraints fail closed and it returns `Denied` too. The doc's claim that per-row validation "has nothing to examine and skipping it forfeits nothing" is therefore false: it forfeits the deny-all and constrained-scope rejections. No current call site is affected — the gear passes `AccessScope::allow_all()` — but this is a `secure_*` export in a crates.io framework crate.

Add an `&AccessScope` parameter and reject before building the statement whenever the per-row path would: `scope.is_deny_all()`, and any constrained scope for an entity whose `resolve_property` resolves nothing. Correct the `# Security` doc paragraph.

gears/system/resource-group/resource-group/src/domain/type_service.rs
```diff
@@ -313,9 +316,21 @@ impl<TR: TypeRepositoryTrait> TypeService<TR> {
...
                     .await?;
+
+                // Assembled, not read back (RG-08). `metadata_schema` and
+                // `updated_at` are what the write just set, `id` and
+                // `schema_id` are the keys it was addressed by, and
+                // `created_at` is immutable -- carried from the row this
+                // transaction read at the top. The re-read this replaces
+                // could only return these same five values.
+                let updated_model = gts_type::Model {
+                    metadata_schema: Some(stored_schema),
```
MikeFalcon77, line 330: **HIGH**

`PUT /types/{code}` assembles its response from the request instead of re-reading the row — a second instance of the wire change the PR body declares as the only one.

`metadata_schema: Some(stored_schema)` is the request-derived value, not the JSONB the database normalized on the way in, so the same duplicate-key collapsing, key reordering and numeric canonicalization the PR body documents for `PUT /groups/{id}` applies here too. The body states "One observable difference, and it is not in the contract" and names only the groups endpoint; a reviewer trusting that scope will not look at this path.

Either restore the post-write read on this path, or list `PUT /types/{code}` alongside `PUT /groups/{id}` in the PR body and in the OpenAPI contract as an observable change.

gears/system/resource-group/resource-group/tests/common/mod.rs
```diff
@@ -471,6 +553,38 @@ pub fn assert_no_surrogate_ids(json: &serde_json::Value) {
...
+/// The callback goes on before the connection is wrapped, unlike [`test_db`]:
+/// `DBProvider`/`Db` never hand out the inner `SeaORM` connection, so
+/// attaching afterwards would miss every statement.
+pub async fn test_db_with_recorder() -> (Arc<DBProvider<DbError>>, QueryRecorder) {
+    let opts = ConnectOpts {
+        max_conns: Some(1),
+        min_conns: Some(1),
+        ..Default::default()
+    };
+    let (db, recorder) = toolkit_db::test_support::connect_with_recorder("sqlite::memory:", opts)
```
MikeFalcon77, line 573: **HIGH**

Every new test runs against `sqlite::memory:`, so the branch's riskiest new SQL never executes on the production backend.

`secure_insert_from_select` writing `resource_group_closure` from a self-join cross product, `insert_ancestor_closure_rows`' `INSERT ... SELECT` with a bound `Expr::val(child_id)` in the select list, and `rebuild_subtree_closure`'s `DELETE ... IN (subquery) AND ancestor_id NOT IN (derived table)` are exactly the statements PostgreSQL's type system can reject where SQLite's dynamic typing accepts — and the PR states the PostgreSQL suites live on a different branch.

Add an env- or `#[cfg]`-gated PostgreSQL variant of `test_db_with_recorder` and run `group_move_under_deep_parent_rebuilds_every_depth`, `assert_closure_matches_parent_links` and the force-delete path against testcontainers PostgreSQL in CI on this branch.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
```diff
@@ -0,0 +1,1328 @@
...
+        let last = build_chain(&group_svc, &ctx, &t.code, tenant_id, depth).await;
+
+        rec.clear();
+        common::create_child_group(&group_svc, &ctx, &t.code, last, "extra", tenant_id).await;
+        count_in(&rec.stats(), QueryKind::Insert, "resource_group_closure")
+    }
+
+    let small = closure_inserts_for_new_child_at_depth(3).await;
+    let large = closure_inserts_for_new_child_at_depth(15).await;
+    assert_eq!(
```
MikeFalcon77, line 671: **HIGH**

The scale tests compare statement counts, which cannot distinguish the fix from the regression a refactor is most likely to reintroduce.

`count_in(&rec.stats(), QueryKind::Insert, "resource_group_closure")` returns 1 at depth 3 and depth 15 both for the `INSERT ... SELECT` this PR wrote and for a version that materializes the ancestor rows in Rust and writes them with one `secure_insert_many` — the O(N) wire traffic RG-06 exists to remove. `normalize_sql` deliberately collapses `(?), (?)` groups, so batch width is invisible to statement counting, and `total_params()` was added for precisely this but is never called by any scale test.

Assert on `rec.total_params()` (or the closure INSERT's `param_count`) alongside the statement count, so a Rust-materialized batch fails the depth-3-vs-15 comparison.

libs/toolkit-db/src/secure/db_ops.rs
```diff
@@ -202,6 +202,235 @@ where
...
+    };
+    let rows_per_statement = rows_per_insert::<E>(backend);
+
+    let mut rest = models;
+    while !rest.is_empty() {
+        let take = rows_per_statement.min(rest.len());
+        let chunk: Vec<E::ActiveModel> = rest.drain(..take).collect();
+        match DBRunnerInternal::as_seaorm(runner) {
+            SeaOrmRunner::Conn(db) => {
+                E::insert_many(chunk).exec(db).await?;
```
MikeFalcon77: **MEDIUM**

`secure_insert_many` writes via `E::insert_many(...)`, which diverges from `secure_insert`'s `am.insert(db)` in two ways the scope validation assumes away.

Verified against sea-orm 1.1.19: `Insert::add_many` (`src/query/insert.rs`) binds a typed NULL for any column that is `NotSet` in one model but `Set` in a sibling in the same chunk, and `ActiveModelBehavior::before_save` is invoked only from `ActiveModelTrait::insert` (`src/entity/active_model.rs:280`), never from `Insert::exec`. `validate_insert_scope` skips a `NotSet` scope column on the stated assumption that "the column is not being set in this insert" and a default applies — on this path that row can instead be written with an explicit NULL scope column.

Reject heterogeneous `Set`/`NotSet` column sets within a chunk with `ScopeError::Invalid`, and either route models through `before_save` or document both divergences on `secure_insert_many`.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
```diff
@@ -408,55 +523,155 @@ impl TypeRepositoryTrait for TypeRepository {
...
+        // layer -- and `resource_group` declares scope columns. The subquery
+        // below is hand-built and inherits no scope predicate of its own; it
+        // is safe because this method runs under the system scope it
+        // constructs itself above, where `build_scope_condition` is empty.
+        // Give this method a caller-supplied scope and the subquery has to be
+        // scoped with it.
+        let budget = toolkit_db::secure::max_bind_params_for(db);
+        let mut violating: Vec<rg_entity::Model> = Vec::new();
+        for type_chunk in parent_type_ids.chunks(budget.saturating_sub(1).max(1)) {
+            let parents_of_candidate_type = Query::select()
```
MikeFalcon77, line 603: **MEDIUM**

The hand-built subquery over `resource_group` — an entity declaring `tenant_col` and `resource_col` — carries no scope predicate.

Only the outer `find()` gets `.secure().scope_with(&scope)`; `Query::select().column(rg_entity::Column::Id).from(ResourceGroupEntity)` bypasses `build_scope_condition` entirely. The in-code comment concedes it ("Give this method a caller-supplied scope and the subquery has to be scoped with it"), so the guarantee rests on prose plus the `system_scope()` this method happens to construct today. The comment's stated reason for going raw is also wrong: `secure/select.rs` already provides `project_all`, `and_scope_for`, `scope_via_exists` and `find_also_related`.

Derive the subquery from a scoped `SecureSelect` — `ResourceGroupEntity::find().filter(...).secure().scope_with(&scope)` lowered to its `SelectStatement` — so the tenant predicate is attached regardless of the scope the method is given.

gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
```diff
@@ -187,25 +197,24 @@ impl MembershipRepositoryTrait for MembershipRepository {
...
+        // The group ids feeding the second query were exactly the first
+        // query's result, so the database can derive them: one statement with
+        // two bind parameters, not two statements the second of which bound
+        // one parameter per membership of the resource -- unbounded, and
+        // unchunked.
+        //
+        // `resource_group_membership` declares no scope columns, so this
+        // hand-built subquery skips no scope predicate: for that entity
+        // `build_scope_condition` is `Condition::all()` under any scope.
+        let member_group_ids = Query::select()
```
MikeFalcon77, line 223: **MEDIUM**

The comment justifying this raw subquery states an invariant that is false.

It claims that for `resource_group_membership` "`build_scope_condition` is `Condition::all()` under any scope". In `secure/cond.rs`, a scope that is neither unconstrained nor deny-all compiles each constraint through `build_constraint_condition`, which returns `None` as soon as `resolve_property` fails; for a `no_tenant, no_resource, no_owner, no_type` entity every constraint drops, `compiled.len() == 0`, and the function returns `deny_all()`. The raw subquery is therefore strictly more permissive than the secure equivalent, and the stated invariant would license the same substitution in a caller-scoped method.

Correct the comment, and obtain the subquery from `MembershipEntity::find().filter(...).secure().scope_with(&scope)` so the predicate set tracks the entity's scope declaration.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -749,41 +820,135 @@ impl GroupRepositoryTrait for GroupRepository {
...
+    ) -> Result<(), DomainError> {
+        if ids.is_empty() {
+            return Ok(());
+        }
         let scope = system_scope();
+        // Chunk against the backend's bind-parameter ceiling, the same one
+        // `secure_insert_many` respects: one `IN (...)` predicate binds one
+        // parameter per id, so an unchunked delete of a large subtree fails
+        // in the driver rather than in the query.
+        for chunk in ids.chunks(toolkit_db::secure::max_bind_params_for(db)) {
```
MikeFalcon77, line 888: **MEDIUM**

Bind-parameter budgeting is re-implemented in the gear at nine call sites instead of living in the secure layer.

`max_bind_params_for` exposes a raw backend budget that group_repo.rs (128, 338, 866, 893, 922) and type_repo.rs (90, 554, 595, 744) each chunk against by hand, including two different hand-rolled budget-splitting formulas (`.div_euclid(2).max(1)` and `budget.saturating_sub(type_chunk.len()).max(1)`). `secure_insert_many` chunks internally via `rows_per_insert` and never exposes the limit — so the crate already knows this is infrastructure's job, and the gear now knows about wire-protocol limits anyway.

Add chunked list-predicate support to the secure layer (e.g. `filter_in_chunked(col, values)` on `SecureSelect`/`SecureDeleteMany` splitting internally like `secure_insert_many`) and drop `max_bind_params_for` from the gear.

gears/system/resource-group/resource-group/src/domain/type_service.rs
```diff
@@ -19,6 +19,7 @@ use crate::domain::error::DomainError;
 use crate::domain::repo::TypeRepositoryTrait;
 #[allow(unused_imports)]
 use crate::domain::validation;
+use crate::infra::storage::entity::gts_type;
```
MikeFalcon77, line 22: **MEDIUM**

The domain service imports a persistence entity: `use crate::infra::storage::entity::gts_type;`.

It is used at line 329 to hand-construct a `gts_type::Model` literal, so the domain layer now depends on the physical column layout of the `gts_type` table and on knowing which columns the repository's UPDATE touched. Add a column to the write and this struct goes stale with no compile error; add a non-`Option` column to the entity and the domain breaks.

Have `TypeRepositoryTrait::update_type` return the updated row (or a domain-owned snapshot) built inside the repository from the values it wrote, and delete this import.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -1075,7 +1113,9 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
         new_parent_id: Option<Uuid>,
         rg_type: &resource_group_sdk::ResourceGroupType,
         profile: &QueryProfile,
-    ) -> Result<(), DomainError> {
+    ) -> Result<Option<crate::infra::storage::entity::resource_group::Model>, DomainError> {
```
MikeFalcon77, line 1130: **MEDIUM**

`move_group_internal_impl` returns `Result<Option<crate::infra::storage::entity::resource_group::Model>, DomainError>`, putting a SeaORM entity in a domain-layer signature.

Beyond the layering inversion, `Option<Model>` re-encodes information the caller already holds in `new_parent_id: Option<Uuid>` without tying them together, so the caller must trust an undocumented invariant that the `Option` is `Some` exactly when `new_parent_id` is `Some`. `update_group_inner` discards the value and keeps its own lookup, so only one of the two callers benefits.

Return a domain-owned struct carrying just the fields the caller consumes (`tenant_id`), or return `()` and have both callers share one explicit parent lookup.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -209,80 +223,132 @@ impl GroupRepository {
...
+        // These paths come straight out of the client's `$filter`, so their
+        // number is the caller's choice. `ODataLimits::max_filter_length`
+        // would bound it, but `validate_filter` is not called anywhere in the
+        // workspace, so today the only ceiling is the transport's URI limit.
+        // Chunked against the bind ceiling like every other client-fed list
+        // here -- `type_repo`'s `resolve_ids` and the removed-parent sweep
+        // already are, and this was the one that was not.
+        let scope = system_scope();
+        let mut ids: std::collections::HashMap<String, i16> = std::collections::HashMap::new();
+        for chunk in paths.chunks(toolkit_db::secure::max_bind_params_for(db)) {
```
MikeFalcon77, line 347: **MEDIUM**

The number of GTS type paths resolved from a client `$filter` is attacker-controlled and bounded only by HTTP URI length.

`resolve_type_filter_node` collects every path in the filter and chunks the lookups against the bind ceiling, and the adjacent comment states that `ODataLimits::max_filter_length` "would bound it, but `validate_filter` is not called anywhere in the workspace". Hand-rolled chunking is compensating for a missing check at the OData boundary, which is where the framework provides the control.

Call the toolkit-odata `validate_filter` / `ODataLimits` check on the parsed query in the `OperationBuilder` `.with_odata_filter()` path, and reject an oversized filter instead of chunking an unbounded list.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -1121,21 +1162,18 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
...
-                            max_subtree_depth = depth;
-                        }
-                    }
-                }
+                // One query for the whole subtree. The rows already carry
+                // the depth relative to the moved group, and every id they
+                // contain is a descendant by construction -- so the previous
+                // `is_descendant` + `get_relative_depth` pair per row was
+                // re-asking the database what it had just returned.
+                let max_subtree_depth = group_repo
```
MikeFalcon77, line 1184: **MEDIUM**

The depth-limit check materializes the entire subtree to compute a single scalar.

`get_descendant_ids_with_depth(conn, group_id).await?.into_iter().map(|(_id, depth)| depth).max().unwrap_or(0)` transfers one `(Uuid, i32)` row per descendant, server-sorted by `ORDER BY depth`, and allocates it in process inside the SERIALIZABLE move transaction — then throws every id away. On the 10 000-node subtree the PR uses as its own example, the move path is still linear in node count while the closure rebuild beside it is not.

Add a repo method issuing `SELECT MAX(depth) FROM resource_group_closure WHERE ancestor_id = ?` returning `i32`, and call it here.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -891,11 +903,21 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
...
+        // second one after the write, the first being inside `update` itself.
+        Ok(ResourceGroup {
+            id: group_id,
+            code: existing_type_path,
+            name: req.name.clone(),
+            hierarchy: GroupHierarchy {
+                parent_id: req.parent_id,
+                tenant_id: existing.tenant_id,
+            },
+            metadata: req.metadata.clone(),
```
MikeFalcon77, line 926: **MEDIUM**

`metadata: req.metadata.clone()` echoes the client's JSON rather than the JSONB the database persisted.

`PUT /groups/{id}` and a following `GET /groups/{id}` now return different bytes for the same resource, so a client that caches or diffs the two sees a spurious change. The create path at line 690 uses `model.metadata` from the insert, so the two write paths are inconsistent with each other. The PR body documents this, but the API contract does not.

Return the persisted model from an `UPDATE ... RETURNING`, or record the divergence in the OpenAPI contract for `PUT /groups/{id}` rather than only in the PR description.

gears/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -89,7 +96,7 @@ pub trait GroupRepositoryTrait: Send + Sync + 'static {
         gts_type_id: i16,
         name: &str,
         metadata: Option<&serde_json::Value>,
-    ) -> Result<rg_entity::Model, DomainError>;
+    ) -> Result<(), DomainError>;
```
MikeFalcon77: **MEDIUM**

`GroupRepositoryTrait::update` returns `Result<(), DomainError>`, discarding the `rows_affected` count `update_many` already produced.

A caller cannot distinguish "row updated" from "row did not exist". The doc comment tells a future caller to check for itself, but the signature hands it nothing to check with — the only route back to that information is the extra SELECT this refactor exists to remove. Both current callers are safe only because they read the row earlier in the same SERIALIZABLE transaction, a precondition the type does not express.

Return `Result<u64, DomainError>` (or `bool`) from the count already in hand, and have `update_group_inner`/`move_group_inner` map zero rows to `DomainError::group_not_found(id)`.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -749,41 +820,135 @@ impl GroupRepositoryTrait for GroupRepository {
...
+            [
+                closure_entity::Column::AncestorId,
+                closure_entity::Column::DescendantId,
+                closure_entity::Column::Depth,
+            ],
+            source,
+            db,
+        )
+        .await
+        .map_err(|e| DomainError::database(e.to_string()))?;
```
MikeFalcon77: **MEDIUM**

The `ScopeError` from `secure_insert_from_select` is flattened to `DomainError::database(e.to_string())`, which wraps the text in `DbErr::Custom` and destroys the typed error.

This method runs inside `transaction_with_retry(TxConfig::serializable(), DomainError::db_err, ...)`, and `is_retryable_contention` (contention.rs:86) matches only `DbErr::Exec | DbErr::Query`, returning `false` for everything else — so a PostgreSQL 40001 raised by this new `INSERT ... SELECT` is classified non-retryable and the jittered backoff this PR adds can never fire for it. The give-up log also degrades to `sql_err = "not_a_sql_error"`. The stringify-the-error idiom is pre-existing in this file (30 occurrences before and after), but it now covers the set-based statements most likely to serialization-conflict.

Match `ScopeError::Db(db_err) => DomainError::Database(db_err)` and stringify only the non-`Db` variants, here and at line 1219.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
```diff
@@ -266,12 +375,13 @@ impl TypeRepositoryTrait for TypeRepository {
...
+        // re-reading by code cost a second `gts_type` SELECT on every
+        // create (RG-08); `secure_insert` returns the persisted model with
+        // the generated id in it.
+        toolkit_db::secure::secure_insert::<GtsTypeEntity>(model, &scope, db)
             .await
-            .map_err(|e| DomainError::database(e.to_string()))?;
-
-        // Re-read to get the auto-generated ID
-        Self::find_model_by_code(db, schema_id).await
+            .map_err(|e| DomainError::database(e.to_string()))
```
MikeFalcon77, line 385: **MEDIUM**

`TypeRepository::insert` does not classify unique-constraint violations, unlike both sibling repositories.

`group_repo.rs:725` and `membership_repo.rs:120` both branch on `e.is_unique_violation()`; this one maps every `secure_insert` failure to `DomainError::database(e.to_string())`. `gts_type` has `UNIQUE(schema_id)` and `create_type`'s pre-check only closes the window under SERIALIZABLE, so a concurrent duplicate create returns HTTP 500 instead of 409 `TypeAlreadyExists`.

Mirror the siblings: `if e.is_unique_violation() { DomainError::type_already_exists(schema_id) } else { ... }`.

libs/toolkit-db/src/secure/db_ops.rs
```diff
@@ -202,6 +202,235 @@ where
...
+/// # Errors
+///
+/// - `ScopeError::Invalid` if a tenant-scoped model is missing `tenant_id`.
+/// - `ScopeError::Denied` if a model's values do not satisfy the scope.
+/// - `ScopeError::Db` if the database insert fails.
+pub async fn secure_insert_many<E>(
+    models: Vec<E::ActiveModel>,
+    scope: &AccessScope,
+    runner: &impl DBRunner,
+) -> Result<(), ScopeError>
```
MikeFalcon77, line 257: **MEDIUM**

`secure_insert_many` returns `Result<(), ScopeError>` while both siblings return data (`secure_insert` -> `E::Model`, `secure_insert_from_select` -> `u64`).

The doc warns that "A split batch is not atomic on its own", but with a `()` return a caller whose second chunk fails cannot learn how many rows the first chunk committed — the recovery information for the exact partial-failure arm the doc raises is discarded inside the function.

Return `Result<u64, ScopeError>` accumulating `rows_affected` across chunks, matching `secure_insert_from_select`.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -916,101 +1084,139 @@ impl GroupRepositoryTrait for GroupRepository {
...
+                    closure_entity::Column::DescendantId,
+                )))
+                .expr(
+                    Expr::col((ancestors, closure_entity::Column::Depth))
+                        .add(Expr::col((subtree, closure_entity::Column::Depth)))
+                        .add(1),
+                );
+            // @cpt-end:cpt-cf-resource-group-algo-entity-hier-closure-rebuild:p1:inst-closure-rebuild-4a1
+
+            toolkit_db::secure::secure_insert_from_select::<ClosureEntity, _>(
```
MikeFalcon77, line 1236: **MEDIUM**

The row count returned by `secure_insert_from_select` is discarded and neither the set-based `DELETE` nor the rebuild emits any tracing.

The rows now exist only inside the database, so the process has lost the one signal it had. The doc comment at lines 1120-1123 identifies a concrete silent-corruption mode for this exact statement pair — "this DELETE would quietly stop deleting -- no error, no rows, a closure table that keeps its stale ancestors" — and nothing at runtime would distinguish that from a healthy rebuild.

Bind the `rows_affected` from the insert-from-select and the `DeleteResult::rows_affected` from the delete, emit a `tracing::debug!` with `group_id`, `deleted` and `inserted`, and warn when a rebuild with a `new_parent_id` inserts zero rows.

libs/toolkit-db/src/secure/db.rs
```diff
@@ -783,3 +1057,158 @@ impl std::fmt::Debug for DbTx<'_> {
...
+    }
+
+    #[test]
+    fn delay_grows_with_attempt_number() {
+        // Structural growth check on the *ceiling* (deterministic), not on
+        // any individual jittered sample -- two samples from a later
+        // attempt aren't guaranteed to exceed an earlier one (jitter can
+        // draw close to zero), but the achievable ceiling must strictly
+        // grow attempt-over-attempt up to the cap.
+        let ceilings: Vec<Duration> = (2..=5).map(pre_jitter_ceiling).collect();
```
MikeFalcon77: **MEDIUM**

`delay_grows_with_attempt_number` never calls `retry_backoff_delay` — it asserts monotonicity of `pre_jitter_ceiling`, a helper defined inside the test module.

`pre_jitter_ceiling` re-implements the production formula from the same constants, so the test asserts a property of test code. The sibling `delay_is_bounded_by_the_pre_jitter_ceiling` only checks `delay <= ceiling`, which a `Duration::ZERO` return would also satisfy — so between them, no test fails if `retry_backoff_delay` stops growing, ignores `next_attempt`, or drops `ExponentialBackoff` entirely.

Sample `retry_backoff_delay(n)` many times per attempt and require the observed maximum to grow from attempt 2 to attempt 5, or expose the pre-jitter base from production code and assert on that.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
```diff
@@ -0,0 +1,1328 @@
...
+    let src = include_str!("../src/domain/group_service.rs");
+    let unretried = count_occurrences(src, UNRETRIED_SERIALIZABLE);
+    let retried = count_occurrences(src, RETRIED_SERIALIZABLE);
+    assert_eq!(
+        unretried, 0,
+        "negative control violated: group_service.rs should not bypass \
+         transaction_with_retry for its SERIALIZABLE writes"
+    );
+    assert!(
+        retried >= 3,
```
MikeFalcon77: **MEDIUM**

`static_rule_passes_group_service_uses_retry` asserts `retried >= 3` while its own failure message names four operations.

The message reads "expected create/update/move/delete_group to all use transaction_with_retry", so dropping the retry wrapper from any one of the four still satisfies the threshold. The `unretried` half of the same rule is pinned exactly (`unretried == 0`), and the `type_service.rs` rule is exact in both directions — this one tolerates precisely the regression it exists to catch.

Change to `assert_eq!(retried, 4, ...)` so losing a retry wrapper on any single write path fails the rule.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
```diff
@@ -0,0 +1,1328 @@
...
+        rec.dump()
+    );
+    // Exactly 1 resource_group SELECT, and it is not ours: SeaORM's
+    // non-RETURNING-fallback re-select after the insert (no
+    // `sqlite-use-returning-for-3_35`), which PostgreSQL does not issue at
+    // all. `create_group_inner` used to add a second, reading the row back to
+    // build the response it could assemble from the insert (RG-08).
+    let rg_selects = count_in(&rec.stats(), QueryKind::Select, "resource_group");
+    assert_eq!(
+        rg_selects,
```
MikeFalcon77, line 232: **MEDIUM**

`trace_create_root_group` pins `rg_selects == 1` to a SQLite driver artifact.

The test's own comment says the single SELECT "is not ours: SeaORM's non-RETURNING-fallback re-select ... which PostgreSQL does not issue at all". So the assertion fails on the production backend (0 SELECTs) or if `sqlite-use-returning-for-3_35` is enabled, while a genuine read-back regression on PostgreSQL lands at exactly 1 and passes. The number encodes the test environment, not the invariant.

Replace with `rec.redundant_reads_after_write().is_empty()`, which `trace_update_group` already uses, and do the same for `trace_create_type`'s `type_selects == 2`.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
```diff
@@ -0,0 +1,1328 @@
...
+        &move_stats_for_subtree_size(3).await,
+        QueryKind::Insert,
+        "resource_group_closure",
+    );
+    let large = count_in(
+        &move_stats_for_subtree_size(15).await,
+        QueryKind::Insert,
+        "resource_group_closure",
+    );
+    assert_eq!(
```
MikeFalcon77, line 724: **MEDIUM**

`scale_move_closure_inserts_do_not_grow_with_subtree_size` has the same statement-count blind spot for RG-04.

The A x N cross product could be rebuilt in process memory and written with a single `secure_insert_many`, and the `resource_group_closure` INSERT statement count would still be 1 at N=3 and N=15. The comment claims the test proves "the pairs never become Rust values", but statement count cannot observe how many values crossed the wire — parameter count can, and the recorder already records it.

Add a `total_params()` comparison alongside the statement-count equality here and in `scale_move_descendant_depth_selects_do_not_grow_with_subtree_size`.

gears/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -224,13 +265,18 @@ pub trait TypeRepositoryTrait: Send + Sync + 'static {
...
+    /// caller can assemble the new row from what it already holds. Reading
+    /// it back cost a second `gts_type` SELECT per update (RG-08).
     async fn update_type<C: DBRunner>(
         &self,
         db: &C,
         type_id: i16,
-        code: &str,
         metadata_schema: Option<&serde_json::Value>,
-    ) -> Result<gts_type::Model, DomainError>;
+    ) -> Result<time::OffsetDateTime, DomainError>;
```
MikeFalcon77, line 282: **MEDIUM**

`TypeRepositoryTrait::update_type` returns a bare `time::OffsetDateTime` whose meaning lives only in the doc comment.

The return type carries no indication of which column or row it belongs to, and it obliges every caller to reconstruct a `gts_type::Model` by hand — which `type_service.rs:329` now does, duplicating knowledge of which columns the UPDATE touched into the domain layer where it will silently rot when a column is added.

Return the updated row built inside the repository from the values it wrote, so the column set stays in one place.

libs/toolkit-db/src/secure/db_ops.rs
```diff
@@ -202,6 +202,235 @@ where
...
+    }
+
+    let backend = match DBRunnerInternal::as_seaorm(runner) {
+        SeaOrmRunner::Conn(db) => sea_orm::ConnectionTrait::get_database_backend(db),
+        SeaOrmRunner::Tx(tx) => sea_orm::ConnectionTrait::get_database_backend(tx),
+    };
+    let rows_per_statement = rows_per_insert::<E>(backend);
+
+    let mut rest = models;
+    while !rest.is_empty() {
```
MikeFalcon77, line 285: **MEDIUM**

The chunking loop emits no tracing when a batch is split or when a later chunk fails.

This is shared infrastructure whose documented failure mode is a partially committed batch. With no log line, an operator seeing half a batch present has no signal that the write was split at all, nor any record of how many statements ran before the failure.

Emit a `tracing::debug!` with `rows` and `statements` when `models.len() > rows_per_statement`, and a `tracing::warn!` on the error path carrying the number of chunks already executed.

gears/system/resource-group/resource-group/src/domain/type_service.rs
```diff
@@ -417,34 +432,45 @@ impl<TR: TypeRepositoryTrait> TypeService<TR> {
...
+                .await?;
+            // @cpt-end:cpt-cf-resource-group-algo-type-mgmt-check-hierarchy-safety:p1:inst-hier-check-2a
+
+            // @cpt-begin:cpt-cf-resource-group-algo-type-mgmt-check-hierarchy-safety:p1:inst-hier-check-2b
+            // Report the first removed parent, in request order, that has a
+            // violation -- the same answer the per-parent loop gave, since
+            // it iterated in that order and returned on its first hit.
+            if let Some(removed_parent) = removed_parents
+                .iter()
+                .find(|p| violations.iter().any(|(code, _, _)| code == *p))
```
MikeFalcon77: **LOW**

`removed_parents.iter().find(|p| violations.iter().any(...))` re-scans the whole violations vector per removed parent.

Neither list is bounded: `removed_parents` comes from the request body's `allowed_parent_types`, and `find_groups_violating_removed_parents` returns one triple per violating group. The quadratic string comparison runs inside the SERIALIZABLE `update_type` transaction, extending the window it is held open.

Fold `violations` into a `HashMap<&str, Vec<&str>>` keyed by parent code in one pass, then look each `removed_parents` entry up in that map.

libs/toolkit-db/src/secure/db_ops.rs
```diff
@@ -202,6 +202,235 @@ where
...
+) -> Result<u64, ScopeError>
+where
+    E: ScopableEntity + EntityTrait,
+    C: IntoIterator<Item = E::Column>,
+{
+    insert_from_select_allowed::<E>()?;
+
+    let mut insert = sea_orm::sea_query::Query::insert();
+    insert.into_table(E::default()).columns(columns);
+    insert.select_from(source).map_err(|_| {
```
MikeFalcon77, line 410: **LOW**

`map_err(|_| ...)` on `SelectStatement::select_from` discards the sea_query error carrying the actual column and value counts.

This is shared library code and the replacement is a `&'static str`, so a caller that builds a target column list of the wrong arity learns neither which side is wrong nor by how much, and the original cause is unavailable for logging.

Add a `ScopeError` variant carrying a `String` and format the sea_query error's counts into it instead of dropping the error.

gears/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -176,6 +203,20 @@ pub trait TypeRepositoryTrait: Send + Sync + 'static {
...
 
+    /// Same lookup as `find_by_code`, but also returns the surrogate
+    /// SMALLINT id, letting a caller that needs both avoid a separate
+    /// `resolve_id` call for the same code (RG-11).
+    /// The row and the type built from it, in one query.
+    ///
+    /// Returns the `gts_type` row itself, not just its surrogate id: the
+    /// update path needs `created_at` to assemble its answer without reading
+    /// the row back after the write, and this query already held it.
+    async fn find_by_code_with_id<C: DBRunner>(
```
MikeFalcon77: **LOW**

`find_by_code_with_id` returns `Option<(gts_type::Model, ResourceGroupType)>` — the whole persistence row, not the surrogate id its name promises.

The doc's first paragraph promises the id and contradicts itself two paragraphs later. A name that understates what it hands back invites callers to take the entire row when they wanted one field, which is how the persistence entity reached `type_service.rs`; it also leaves `find_by_code` a near-duplicate of this body in `type_repo.rs`, differing only in what it drops.

Rename to `find_by_code_with_model`, or narrow the return to `Option<(i16, ResourceGroupType)>` and give the update path a dedicated `find_model_by_code`; implement `find_by_code` by delegating.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -916,101 +1084,139 @@ impl GroupRepositoryTrait for GroupRepository {
...
-                }
-                // @cpt-end:cpt-cf-resource-group-algo-entity-hier-closure-rebuild:p1:inst-closure-rebuild-4a
-            }
-
-            for row in new_rows {
-                toolkit_db::secure::secure_insert::<ClosureEntity>(row, &scope, db)
-                    .await
-                    .map_err(|e| DomainError::database(e.to_string()))?;
-            }
+            let mut source = Query::select();
```
MikeFalcon77, line 1195: The `INSERT ... SELECT` source is a raw `Query::select()` with two `from_as(ClosureEntity, alias)` self-joins and hand-written predicates, never passing through the secure query builder.

The scope-column check inside `secure_insert_from_select` covers the insert target only; the join sides here carry no scope condition. It is correct today only because `resource_group_closure` declares no scope columns — an entity attribute, not an invariant of this call site — and the correctness is re-derived by comment rather than by the type system.

Build the source from scoped `SecureSelect` values so the join sides inherit the entity's scope condition automatically if it ever gains one.

libs/toolkit-db/src/secure/mod.rs
```diff
@@ -140,6 +140,8 @@ pub use runner::DBRunner;
 pub(crate) use runner::{DBRunnerInternal, SeaOrmRunner};
 
 // Primary database types (new secure API)
+#[cfg(feature = "test-support")]
+pub use db::in_transaction_for_testing;
```
MikeFalcon77, line 144: **LOW**

`in_transaction_for_testing` is re-exported from the production `secure` facade under `#[cfg(feature = "test-support")]`.

Cargo features are additive and unified across a build graph, so any workspace member enabling `toolkit-db/test-support` — the resource-group gear does, via dev-dependencies — surfaces a `_for_testing` name in the production `secure` namespace of every other gear in the same build.

Move the re-export into the `test_support` module that owns the rest of this surface and keep `secure`'s public API free of test-only accessors.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -678,11 +676,19 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
...
 
-            let sys = toolkit_security::AccessScope::allow_all();
-            group_repo
-                .find_by_id(tx, &sys, group_id)
-                .await?
-                .ok_or_else(|| DomainError::database("Insert succeeded but group not found"))
+            // The row the insert returned, not a re-read of it. Its type path
+            // is the code this request named -- `type_id` was resolved from
+            // it above -- so nothing here needs the database again (RG-08).
+            Ok(ResourceGroup {
```
MikeFalcon77, line 682: **LOW**

The `rg_entity::Model` -> `ResourceGroup` mapping is hand-written four times in the service (682, 740, 911, 1005), duplicating `GroupRepository::model_to_resource_group`.

The DTO assembly is one decision now made in five places; adding a field to `ResourceGroup` compiles only after all five are updated, and missing one yields a silently divergent response rather than a compile error — which is the shape of the metadata divergence this PR introduces.

Make `model_to_resource_group` `pub(crate)` (or add a `From` impl in the infra module) and call it from all four sites.

MikeFalcon77 (changes requested):

libs/toolkit-db/src/secure/db_ops.rs
ffedoroff: Fixed in 4b7ca18a4. You are right, and the doc was wrong on top of it: `no_*` is not `unrestricted`, so the per-row path does reject a deny-all or constrained scope for exactly this class of entity. The helper now takes `&AccessScope` and accepts only an unconstrained one — deliberately *stricter* than the per-row path (which also admits an empty constraint): an entity with no scope columns has nothing to satisfy a non-empty constraint with, and telling the two apart here buys nothing. An sqlite-backed test pins the rejection. Both call sites already passed the system scope, so the gear's behaviour is unchanged — this hardens the crates.io export.

gears/system/resource-group/resource-group/src/domain/type_service.rs
ffedoroff: Confirmed — the PR body now names both `PUT /groups/{id}` and `PUT /types/{code}`. We also verified the mechanism is real in this workspace: `serde_json` resolves with `preserve_order` transitively, so the echoed request keeps its key order while jsonb does not. We did not restore the post-write read: that would trade the RG-08 fix for byte-equality of a JSON rendering the API contract does not promise.

gears/system/resource-group/resource-group/tests/common/mod.rs
ffedoroff: Fixed in 4b7ca18a4, scoped deliberately: a `--features integration` PG smoke suite (testcontainers) with three tests covering exactly the SQL flagged here — the `INSERT ... SELECT` cross product, the derived-table `DELETE` via move/rebuild, and the force-delete path — wired into CI (`make test-rg-pg`) with `RG_PG_REQUIRE_DOCKER=1`, so an unreachable daemon fails the step instead of skipping into green. Verified locally: 3/3 against testcontainers PostgreSQL. The full concurrency suite (barrier-synchronized races, SSI behaviour) stays a separate PR per the audit report's follow-up plan — it needs infrastructure this branch shouldn't grow.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
ffedoroff: Fixed in 4b7ca18a4. Good catch — `total_params()` was added for precisely this and called by nothing. Both scale tests now assert it alongside the statement count; a Rust-materialized batch fails the N=3 vs N=15 comparison on parameters even though `normalize_sql` keeps its statement count at 1.

libs/toolkit-db/src/secure/db_ops.rs
ffedoroff: Both divergences verified against sea-orm 1.1.20 and documented on `secure_insert_many` in 4b7ca18a4. We did not reject heterogeneous `Set`/`NotSet` batches: that shape is a legal way to let rows fall back to a DB default, both current call sites build homogeneous batches, and the tenant column already has an explicit `NotSet` reject. The doc now warns about the typed-NULL behaviour and that `before_save` never runs on this path.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
ffedoroff: The comment was wrong and is rewritten in 4b7ca18a4 — you are right that `project_all`/`into_inner` exist and the stated reason was false. But we deliberately did not scope the subquery: this is an integrity sweep. Under a constrained scope a scoped subquery compiles to `WHERE false` (every constraint drops on a scope-column-free entity, and `compiled.len() == 0` is deny-all), so the sweep would report "no violations" to a caller whose scope hides them, and `update_type` would pass with real violations standing. The new comment states the system-scope contract and what to do if this method ever takes a caller's scope: rethink the check, not scope the subquery.

gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
ffedoroff: Correct on both counts, and the comment is rewritten in 4b7ca18a4 — a constrained scope on a `no_*` entity compiles to deny-all, not `Condition::all()`. We did not adopt the scoped-subquery fix though: this is the tenant-compatibility check, an integrity read that must see every row. Under a constrained scope the scoped subquery becomes `WHERE false`, the "existing tenants" set comes back empty, and a cross-tenant membership passes vacuously — the false comment would have become a real hole. The new comment states the system-scope contract explicitly.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: Agreed on direction, declined here: a `filter_in_chunked` builder method can't exist for SELECT — merging chunked result sets silently breaks `order_by`/`limit`, so it has to be an executor (`all_in_chunked(col, values, runner)`), and three of the nine sites combine *two* unrelated list predicates in one statement, which no single-predicate helper covers. It's a real toolkit-db API design task; tracked as a separate PR.

gears/system/resource-group/resource-group/src/domain/type_service.rs
ffedoroff: Fixed in the stacked PR #4447 (6bdf1b042): `update_type` now returns the updated row assembled inside the repository next to its own `col_expr` list, the service's hand-built model is gone, and this import is deleted.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Fixed in the stacked PR #4447 (6bdf1b042): `move_group_internal_impl` takes a `ParentSnapshot { tenant_id, gts_type_id }` its caller builds from a read it already performed, and returns a domain-owned `MoveOutcome` — no entity in the signature, and the duplicate parent read on the update-with-reparent path went with it. One correction for the record: the invariant was documented (the doc comment right above the signature), and both callers were safe — but the type not expressing it was a fair point, and now it does.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: The comment was wrong in the other direction and is rewritten in 4b7ca18a4: the list is not "bounded only by URI length" — the toolkit extractor enforces `MAX_FILTER_LEN = 8 KiB` and `MAX_NODES = 2000` before this code runs (`libs/toolkit/src/api/odata.rs`). The genuinely dead mechanism is toolkit-odata's parallel `ODataLimits::validate_filter`; unifying the two is a toolkit-odata change affecting every gear — separate PR. The chunking here stays as defense in depth.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Fixed in the stacked PR #4447 (6bdf1b042): `get_max_descendant_depth` issues `SELECT MAX(depth)` and the move path uses it; the materializing method stays for force-delete, which actually needs the ids.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Documented rather than reverted: the PR body now records this (and the same shape on `PUT /types/{code}`) as the branch's observable differences, including why they are real — this workspace resolves `serde_json` with `preserve_order`, so the echoed request keeps its key order while jsonb normalizes. Restoring the read would restore byte-equality at the cost of the re-read RG-08 removes, for a rendering the API contract does not promise. The create/update/move source inconsistency you point at is fair; it is called out in the PR body alongside.

gears/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Fixed in 4b7ca18a4: `update` returns the `rows_affected` it always had, and both callers map zero to `group_not_found` instead of trusting a doc comment.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: Fixed in 4b7ca18a4: `ScopeError::Db` maps to `DomainError::Database`, so the retry loop sees the typed `DbErr` and the give-up log gets a real `sql_err` class. The stacked PR #4447 also fixes the root cause for the remaining stringify sites — `is_retryable_contention` learns `DbErr::Custom` — but the typed pass-through is better and lands here.

gears/system/resource-group/resource-group/src/infra/storage/type_repo.rs
ffedoroff: Fixed in the stacked PR #4447: `TypeRepository::insert` branches on `is_unique_violation`, mirroring the siblings. Kept there because the whole error-classification family (unique + FK) lands together with the isolation work that makes it load-bearing.

libs/toolkit-db/src/secure/db_ops.rs
ffedoroff: We looked hard at this one and declined: `Result<u64, _>` only returns the count on success, where it equals `models.len()` — a number the caller already has. The scenario that motivates it (second chunk fails; how much did the first commit?) lives in the `Err` arm, which a success-typed count cannot reach. Making that recoverable needs a structured error carrying `committed_rows`, which is a real API design change for a crates.io crate — happy to do it as its own PR if the recovery story is worth it to you.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: Fixed in the stacked PR #4447 (6bdf1b042): the delete's `rows_affected` is captured alongside the insert's, a debug event carries `group_id`/`deleted`/`inserted`, and a reparent that inserted zero rows warns — matching the silent-corruption mode the doc comment describes. The insert count also feeds the `closure_rows_written` metric there.

libs/toolkit-db/src/secure/db.rs
ffedoroff: Fixed in 4b7ca18a4: the growth test now samples `retry_backoff_delay` itself — 200 draws at attempts 2 and 5, asserting the observed maxima grow. With pre-jitter ceilings at 10ms and 80ms the false-failure probability is <= (10/80)^200. `pre_jitter_ceiling` stays for the bound test only.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
ffedoroff: Fixed in 4b7ca18a4, but not to `== 4`: the file has **five** call sites — the four in this thread's list plus `create_group_unscoped`, which was invisible to both the old floor and the proposed count. It is `assert_eq!(retried, 5)` now, with all five named in the message. The stacked PR pins its own counts exactly too, including `delete_type`, which the old `>= 2` in the type-service rule also missed.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
ffedoroff: Diagnosis accepted; the proposed replacement doesn't survive contact with the current backend: `redundant_reads_after_write()` flags the pair "INSERT -> next statement on the same table is a SELECT", and SeaORM's non-RETURNING fallback on SQLite is exactly that pair — the swap fails `trace_create_root_group` on healthy code today (it only works in `trace_update_group` because updates don't re-select). The honest fix is a per-backend expectation, which only makes sense together with running the suite on PG — tracked with the PG follow-up rather than half-done here.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
ffedoroff: Fixed in 4b7ca18a4 together with the sibling thread: both scale tests assert `total_params()` alongside the statement count.

gears/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Fixed in the stacked PR #4447 (6bdf1b042): `update_type` returns the updated row, assembled inside the repository from the values it wrote — the column set the UPDATE touches lives in one place, and the returned timestamp that reached no observable field is gone.

libs/toolkit-db/src/secure/db_ops.rs
ffedoroff: Declined: with two-column junction rows the split threshold is ~15 000 rows in one statement — an event production will not see, and a debug line that never fires is not a signal. If the partial-commit story matters, it needs the structured error from the `secure_insert_many` return-type thread first; happy to bundle both in a toolkit-db PR.

gears/system/resource-group/resource-group/src/domain/type_service.rs
ffedoroff: Fixed in 4b7ca18a4: violations fold into a `HashMap` in one pass; order of selection and the error text are unchanged.

libs/toolkit-db/src/secure/db_ops.rs
ffedoroff: Agreed, deferred: `ScopeError` is a public enum without `#[non_exhaustive]`, so a String-carrying variant is a semver break for the crates.io crate. Separate PR together with adding `non_exhaustive`.

gears/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Fixed in 4b7ca18a4: renamed to `find_by_code_with_model`, and the doc no longer promises an id in one paragraph and a row in the next.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
ffedoroff: Declined, with the mechanics spelled out: both join sides and the target are the same scope-column-free closure entity, and if it ever gains a scope column, `insert_from_select_allowed` rejects the whole call at the guard — before any statement is built. The hypothetical fails loudly, not silently. A scoped source under a constrained scope would instead compile to `WHERE false`, which is precisely the silent-corruption mode the neighbouring doc comment warns about.

libs/toolkit-db/src/secure/mod.rs
ffedoroff: The stated harm doesn't occur in this workspace: the root manifest pins `resolver = "3"`, and dev-dependency features don't unify into production graphs — the name only appears in builds that include dev units, where a test-only accessor in `secure` is harmless. Happy to move it into `test_support` as hygiene if you feel strongly; declined as a defect.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Partially declined: of the four named sites only two have an `rg_entity::Model` in hand — the others assemble from the request and from a modified parent, so a shared helper covers half the sites at the cost of fabricating models for the rest. The real issue underneath (three different sources for `metadata`) is now documented in the PR body as the branch's observable difference.

Artifizer (approved):

MikeFalcon77 (approved):

PR #4447, opened 2026-08-08
perf(resource-group,toolkit-db): shorten the transaction, then stop paying for the isolation it held (VHP-2344)

ffedoroff:
Where #4444 removed statements, this removes the time a transaction holds them open, and stops paying for an isolation level the work does not need.

---

## The retry loop was dead code

Every repository in this gear maps its errors through `DomainError::database(e.to_string())`, which wraps into `DbErr::Custom`. `is_retryable_contention` matched only `DbErr::Exec`/`Query` and returned `false` for everything else — so a serialization failure raised inside *any* repository call arrived in a shape the classifier refused to recognize. `transaction_with_retry` never retried a single write path here (RG-15).

That also means the jittered backoff added in #4444 could not run, and neither could the give-up logging it feeds. The classifier now checks `DbErr::Custom` against the same message patterns, with five regression tests.

Worth naming for anyone adding metrics later: a retry counter added *before* this fix would have honestly reported zero retries, and zero retries looks like a healthy system.

## Transaction window

**GTS metadata validation moved out of the transaction (RG-09).** Resolving the chained schema is a cross-gear network round-trip, and it is then compiled; both ran with a `SERIALIZABLE` transaction open, and every retry paid for them again. Neither reads database state. The fix is structural: `create_group_inner` and `update_group_inner` no longer take a `TypesRegistryClient`, so the call cannot drift back inside.

**`delete_type` had no transaction at all (RG-02)** — resolve, count and delete on a bare connection, so a concurrent `create_group` of that type could land between the count and the delete.

## Isolation, chosen per operation

| Operation | Level | Why |
|---|---|---|
| create, move, parent-changing update, force delete | `SERIALIZABLE` | rewrite closure rows across a subtree |
| rename/metadata-only update | default | one row by primary key |
| non-force delete | default | one row, with `SELECT ... FOR UPDATE` on the target |
| create type | default | `UNIQUE(schema_id)` holds the invariant at any level |
| delete type | default | `ON DELETE RESTRICT` holds it |

`update_group` cannot know before `BEGIN` whether it is a move, so it takes a hint from a read it already does and escalates: the authoritative in-transaction read returns `NeedsSerializable` **before any write**, and the operation reruns at `SERIALIZABLE`. Overshooting is harmless; undershooting cannot commit.

These operations do not stop being correct under `SERIALIZABLE` — they simply do not require it. What changes is where contention surfaces: on PostgreSQL's default they wait on a row lock instead of aborting with `40001`. Retry stays in place either way.

**What actually serializes the non-force delete against a concurrent create is the foreign key, and the code says so.** A plain `SELECT` does not wait on `FOR UPDATE`; what blocks is the `FOR KEY SHARE` the RI trigger takes on the parent row when inserting a child, which conflicts with the `FOR UPDATE` the delete holds. Whichever arrives first, the other waits — the delete's lock makes the ordering decidable rather than doing the blocking. Every foreign key into `resource_group` is `ON DELETE RESTRICT`, so no orphan is reachable; this was review-challenged (one engine called it Critical) and survived four independent attempts to construct one.

**And the race the lowered level opens gets a real answer.** Under `SERIALIZABLE` on both sides, a create that lost to a concurrent delete of its parent surfaced as `40001`, retried, and re-read a clean `group_not_found`. With the delete at the backend default, SSI has no second serializable party — the foreign key answers instead, and unmapped that was a 500. `GroupRepository::insert` now classifies the FK violation, and only for `fk_resource_group_parent` (matched by constraint name): the table's other FK, `fk_rg_gts_type`, fails on a concurrent `delete_type` and must not be reported as a missing parent.

**Two constraint classifications came along**, but only where a constraint was already enforcing the invariant: a duplicate type code was a 500 until `SERIALIZABLE` retried into the in-transaction check, and a type still in use answered from `ON DELETE RESTRICT` with no mapping. `is_foreign_key_violation` joins `is_unique_violation` in `toolkit-db`, with tests that they are not confused for each other.

## Measurement (P0)

`resource-group` was the only system gear without `infra/metrics.rs`; nine others have one, so this wires into an existing platform rather than building one. Five instruments: operation duration by operation and outcome, subtree size, closure rows written (the database's own count — the rows never reach the process), isolation escalations, and metadata-validation duration.

Not injected through the constructor: it has ~150 call sites, nearly all tests. The service defaults to a no-op port and `gear.rs` installs the real one, so the domain never mentions OpenTelemetry.

**Not here:** the retry loop's own counters — attempts, serialization failures, exhausted budgets. Those live in `toolkit-db`, which carries no instruments at all today; adding the first is a platform decision that deserves its own review.

## Behaviour changes

Three, all intentional, none in the API contract:

1. **Isolation levels** as tabled above.
2. **Error precedence on create.** Metadata is validated before the type is resolved, so a request with *both* an unregistered type and invalid metadata reports the metadata error.
3. **A create losing the delete race answers 404**, not 500 (see above).

No schema change, no migration, no API change.

## What guards it

Four source-scan rules in Section 4 cover what is not observable as SQL: retry on every transaction in **both** service files, the external call kept out of the transaction, and the non-force delete's row lock — invisible on SQLite, which has no row locks. Each rule is scoped to the function bodies it is about, through a shared `fn_body` helper that panics rather than scanning nothing: an earlier revision scanned whole files, and one rule passed on `update_group`'s literals regardless of what `delete_group` did — verified by hard-coding the level back and watching the rule stay green. It fails now.

Where a test pins a security check or a defect class, it was validated by breaking the thing it guards and confirming it fails. That found a coverage gap: the cross-tenant check on `update_group`'s path had **no** test — deleting it left the whole suite green.

## Verification

- `cargo test -p cf-gears-resource-group -p cf-gears-toolkit-db` — **700 passed, 0 failed**.
- `cargo fmt --all --check` and `clippy --all-targets -D warnings` clean.
- CDSL markers unchanged in every file touched.

**Not verified locally:** tests run on SQLite, which is serializable regardless and omits `FOR UPDATE` entirely — so it cannot distinguish the lowered levels or show the row lock. The isolation-escalation path needs a concurrent writer; the PostgreSQL suites that exercise it are not in this branch (see the audit report's deviation section for where they live).

## Summary by CodeRabbit

* **New Features**
  * Added resource-group operation metrics for timing, subtree size, closure updates, and isolation escalations.
  * Added safer, scope-aware batch database operations and improved large-query handling.
* **Bug Fixes**
  * Improved tenant protection, hierarchy consistency, locking, retries, and database error classification.
  * Reduced redundant queries and improved bulk-processing performance.
* **Documentation**
  * Expanded PostgreSQL testing guidance and added a defect-to-control reference.
  * Clarified concurrency, retry, timeout, and query-cost expectations.
* **Tests**
  * Added coverage for isolation, hierarchy changes, query efficiency, metrics, and database behavior.


Artifizer:
It seems like this PR is redundant and not needed anymore

PR #4450, opened 2026-08-10
docs(usage-collector)!: PRD Phase 2 - billing integration

Corw1n-of-Amber:
## Summary

Phase 2 of the Usage Collector PRD — making the gear a correct feed for the BSS Rating gear and a viable baseline for S3 metering. Delivered as in-place edits to the existing PRD rather than a separate document, at the Phase-1 developer's request.

Agreed in person with @capybutler on 2026-08-10; opening the PR so the review happens with the CF team.

New **§5.9 Billing Integration** adds: the billing UsageType profile, interval usage windows, counter/gauge quantity semantics under windows, canonical metering units and unit binding, dimension keys over declared metadata, a `reason_code` on billing corrections, the unstripped billing field set on read, a pull-primary replay-safe billing feed, an operator/plugin retention floor, origin-flagged backfill, ingestion rate limiting, and per-scope reconciliation metadata. Also: two Phase-2 NFRs in §6.1, capability categories in §7, three use cases in §8, a Phase-2 acceptance-criteria block in §9, and two assumptions in §11.

## ⚠️ Breaking change to the record model — needs a replacement ADR

The emitter-supplied event timestamp **`created_at` is removed for every UsageType** (not just billing ones) and replaced by the interval `window_start` / `window_end`; a point event sets the bounds equal. Consequently the dedup identity and the derived-`id` input become `(tenant_id, gts_id, idempotency_key, window_start, window_end)`, with the UUIDv5 namespace constant unchanged.

This **would supersede ADR-0014**. A PRD cannot retire an ADR, so a replacement ADR is required alongside this change — it is not in this PR yet.

Semantically the change is lossless (an emitter that sent one timestamp now sends that instant as both bounds), but it is wire-breaking for every caller. It is affordable on the same grounds ADR-0013 used for its own breaking change: the gear is unreleased.

**One subtlety worth reviewer attention:** a point event does not survive naive interval-overlap logic. Overlap is `window_start < range_end AND window_end > range_start`, so a point at 11:00 queried over a range starting at 11:00 fails the second test and disappears from the period it belongs to — the month-close boundary, i.e. exactly where money is counted, and silently. §5.9 therefore specifies **two** selection cases and requires both, and §9 carries an explicit boundary acceptance criterion.

## Decisions behind this

| | Decision |
|---|---|
| Feed | **Pull-primary** over the existing Downstream Usage Reader Contract, with **per-UsageType subscription**; ordering is a monotonic acceptance sequence scoped per `(tenant_id, gts_id)`; corrections arrive as append-only feed events. No event bus in this phase. "Pull, not push" describes only the outbound edge — usage sources including the S3 agent still push inbound via ordinary ingestion. |
| Windows | Replace the timestamp rather than sit alongside it (see above). Late arrival is measured against a **gear-assigned** `accepted_at`, so a skewed emitter cannot hide its own lateness. |
| Retention | A **per-UsageType retention policy** honoured by the storage deployment, with a floor of `max(replay horizon, backfill window)` = 90 days for meters a charging consumer reads. No aggregate floor. The gear is explicitly **not** the system of record for long-term billing evidence, which keeps the §6.2 "not a financial-reporting source" exclusion honest. |
| Feed freshness | A **plugin readiness gate** (published ceiling ≤ 5 min p95), not a gear-level bound — the gear floor of "eventually consistent, no upper bound" is unchanged. |
| Replay capacity | A **recovery objective** (a consumer 24h behind catches up within 6h), from which `R_read ≥ 5 × R_subscribed` follows — anchored to the consumer's subscription, not the gear-wide envelope. |

## Deliberately not included

- No commercial identity on the record — subscription, SKU, payer and seller axes are resolved downstream.
- No new record identifier: ADR-0013 stands and the `corrects_id` reference is unchanged; only the derivation *input* moves.
- No event-bus ingestion and no push feed.
- No materialized rollup mandate, no gear-level retention or freshness enforcement.
- No billing-profile flag and no two-class record model: after review, the metering model is unconditional for every UsageType (see the review threads), which also removed the enable-on-existing migration question.
- Deferred unchanged: Individual Event Amendment, Audit Events, Multi-Region Replication.

## Review notes

- Requirements that constrain a UsageType are written against the **declaration**, not against the gear-local catalog, so they survive the move to the types registry.
- Phase-1 text touched only where the record-model change made it self-contradictory: the idempotency FR, the query-freshness floor, the deactivation use case, and §14.
- Field naming follows the document's existing `snake_case` convention throughout.
- Pagination touches [#4397](https://github.com/constructorfabric/gears-rust/pull/4397) conceptually (the billing feed is cursor-based); that PR does not touch `PRD.md`, so there is no textual conflict, but the cursor shape should be reconciled with it at DESIGN time.

## Test plan

- [ ] Replacement ADR for ADR-0014 authored and reviewed
- [ ] §9 Phase-2 acceptance criteria reviewed against §5.9, in particular the point-event boundary criterion
- [ ] Confirm which Phase-2 requirements land in this milestone vs later (priorities `p1`/`p2`/`p3` are set in the document)
- [ ] DESIGN and `usage-collector-v1.yaml` updated in a follow-up

## Summary by CodeRabbit

* **Documentation**
  * Expanded billing integration requirements for usage windows, acceptance timestamps, canonical units, quantity semantics, and correction reasons.
  * Added guidance for billing data access, retention, backfills, rate limits, reconciliation metadata, replay, and freshness objectives.
  * Clarified record identity, idempotency, deactivation, traceability, metadata filtering, and downstream ownership of commercial identity.
  * Updated SDK, REST, downstream integration, use-case, acceptance, assumptions, and traceability documentation.
  * Added requirements for consistent billing records and reliable handling of corrections and historical data.


gears/system/usage-collector/docs/PRD.md
```diff
@@ -435,6 +440,198 @@ The system **MUST** treat its persisted data as one of three classes:
...
+
+#### Dimension Key (via declared metadata keys)
+
+- [ ] `p2` - **ID**: `cpt-cf-usage-collector-fr-dimension-key`
+
+A billing-relevant UsageType declaration **MAY** name an ordered subset of its closed metadata keys as its `dimension_key` — the pricing-relevant dimensions (e.g., storage class, traffic scope, operation). Values follow the closed-shape metadata contract (declared keys, string values). The gear **MUST NOT** interpret dimension semantics. Emission of dimension values is the source's responsibility; until sources emit them the `dimension_key` is empty and dimensional rating is out of MVP.
+
+This **extends the grouping set** of `cpt-cf-usage-collector-fr-query-aggregation`, which today groups by time bucket, tenant, subject, and resource only: for billing-relevant UsageTypes, aggregation **MUST** additionally support grouping by the declared dimension keys, and raw and aggregation queries **MUST** support equality filtering on them. Grouping and filtering on metadata keys that are *not* declared as dimension keys remains unsupported, so the addressable grouping surface stays bounded and declared at registration rather than open-ended over arbitrary metadata.
```
capybutler: ## `fr-dimension-key` restates a capability the closed-shape metadata model already provides

§5.9 introduces `dimension_key` on the premise that grouping over metadata doesn't exist yet. That premise doesn't hold — grouping over declared metadata keys is already specified *and* already shipped.

**Already specified.** `domain-model.md:668` defines `group_by` as:

> Any combination of the **fixed fields** in §2.9 plus every key in the queried usage type's **`metadata_fields`** (resolved per request from `gts_id`).

with the bounding invariant at `:676` — admissibility computed per request from the queried type's `metadata_fields`, anything outside that union rejected before plugin dispatch. And this PRD already reaches the query surface itself, in `fr-record-metadata` (§5.1, line 224):

> Downstream consumers (billing, reporting, analytics) extract declared keys by name; **the Usage Collector's query surface addresses the same declared keys.**

**Already implemented.**

- `usage-collector-sdk/src/models.rs:938` — `AggregationDimension::Metadata(MetadataKey)`
- `usage-collector/src/api/rest/dto.rs:338` — wire form `{"metadata": "<key>"}` inside `group_by`
- `plugins/timescaledb-usage-collector-plugin/src/infra/storage/query/aggregate.rs:100` — compiles to `metadata ->> $N`, key bound rather than interpolated
- `plugins/timescaledb-usage-collector-plugin/tests/records_query_integration_pg.rs:538` — `pg_aggregate_group_by_metadata`, green against real Postgres

So line 553's "which today groups by time bucket, tenant, subject, and resource only" is inaccurate in both directions: metadata grouping exists, and *time-bucket* grouping does not — `AggregationDimension` carries no time variant.

### The stated rationale is already satisfied

> so the addressable grouping surface stays bounded and declared at registration rather than open-ended over arbitrary metadata

`metadata_fields` **is** declared at registration and **is** closed — that is the entire point of `fr-record-metadata`'s "no free-form remainder, no open-extras escape hatch, and no silently-preserved undeclared properties". The grouping surface is therefore already bounded and already declared. A second declared list doesn't add boundedness; it adds a list that has to be kept in sync with the first.

The rest of the FR's surface thins out similarly:

- `:573` requires billing reads to return "declared `dimension_key` values" — metadata is already returned verbatim on every read path, so this is a restatement rather than a new obligation.
- `:1142` requires validating that "every declared dimension key is a member of `metadata_fields`" — validation that exists only because the second list exists.
- "ordered subset" — `group_by` is already an ordered list supplied per request by the caller, so declaration-time ordering isn't load-bearing for the rating tuple.

### What `dimension_key` would genuinely add, and where it belongs

There is one real difference worth naming: `dimension_key ⊆ metadata_fields`, so it can mark a key groupable while `metadata_fields` still holds keys nobody should group on (a high-cardinality `request_id`). That's a legitimate concern — but it's a *per-key property*, not a parallel ordered list, and the cardinality risk it guards is already bounded by `MAX_AGGREGATION_BUCKETS` (100k, rejected as `AGGREGATION_RESULT_TOO_LARGE`) plus the mandatory bounded time window on the aggregate path.

If a pricing-relevance marker is wanted for the Rating gear, it reads better as an annotation on a declared metadata key than as a second list. I'd argue it isn't needed in this phase regardless — the FR itself says "until sources emit them the `dimension_key` is empty and dimensional rating is out of MVP".

### Suggested change

1. Drop `fr-dimension-key` as a distinct concept.
2. Amend `fr-record-metadata` (§5.1) to state the grouping/filtering contract explicitly: declared keys are groupable and equality-filterable on both read paths, bounded per request to the queried type's `metadata_fields`. This promotes `domain-model.md:668`/`:676` into the PRD where the normative statement belongs, and makes line 224's "the query surface addresses the same declared keys" concrete instead of implied.
3. Correct line 553's premise wherever the surrounding text survives, and drop the time-bucket claim — grouping by time period is specified nowhere and implemented nowhere, so asserting it as existing behavior hides a real gap.

### Related implementation note

The `domain-model.md:676` admissibility check is **not implemented**. `usage-collector/src/domain/service.rs:1883` already resolves the usage type pre-dispatch and has `usage_type.metadata_fields` in hand, but only calls `require_op_allowed_for_kind` — so today an arbitrary undeclared key in `group_by` is accepted and silently yields an empty-string bucket (`record_store.rs:1307` maps a NULL dimension to `""`). The `metadata.<key>` filter side-channel has the same gap. Whichever way this FR lands, that check is missing and worth a follow-up.

gears/system/usage-collector/docs/PRD.md
```diff
@@ -435,6 +440,198 @@ The system **MUST** treat its persisted data as one of three classes:
...
+
+Enabling the profile on a UsageType that already has persisted records is **not specified here**: those records predate the billing model and carry no bound unit, so admitting them to the billing feed would require a migration rule this phase does not define. Deployments **SHOULD** declare billing meters as billing-relevant from first registration.
+
+- **Depends on**: `cpt-cf-usage-collector-fr-usage-type-existence-and-semantics`
+- **Rationale**: One switch scopes every Phase-2 obligation to the UsageTypes that are actually billed, preserving the lightweight contract for the rest of platform telemetry and keeping the gear generic. The enable-on-existing case is deliberately left unspecified rather than forbidden: the UsageType catalog is expected to move to a separate types registry, so writing a migration rule against today's catalog would specify a surface that is on its way out. Declaring billing meters as billing from the start costs nothing while the gear is unreleased and no billing meter exists yet.
```
capybutler: ## Drop the billing profile flag — make the billing model the only model

This FR's own rationale contains the argument for removing it:

> Declaring billing meters as billing from the start costs nothing **while the gear is unreleased and no billing meter exists yet**.

That premise is correct — the gear is unreleased, there are no installations, and no billing meter exists. But it applies symmetrically: there is likewise no legacy telemetry corpus that needs protecting from the strict model. The flag preserves a "lightweight contract" for a population of UsageTypes that doesn't exist yet, and charges a permanent fork in the record model for it.

### What the switch forks

Per `:447`, **every** requirement in §5.9 is conditioned on the profile except `fr-usage-windows`: the live future-time bound, metering-unit binding, canonical units, quantity semantics and the gauge sampling interval, dimension keys, `reason_code` on corrections, the unstripped field set on read, feed inclusion, the retention floor, backfill, ingestion rate limiting, and the billing-capable plugin ceiling (`:740`).

That is two ingestion validation paths, two correction models, two read projections, two retention obligations, and two plugin-qualification regimes — in a gear that hasn't shipped. §9 inherits the same split: `:1281` conditions the entire Phase-2 acceptance block on the profile, and `:1283` exists solely as a regression guard describing the other branch.

### The migration hole exists only because the flag does

`:457` leaves the one operation that will actually be needed unspecified:

> Enabling the profile on a UsageType that already has persisted records is **not specified here**

This isn't a gap that closes in a later phase — it's structural. The day a telemetry type needs billing, its history carries no bound unit, no `reason_code` on its corrections, and never appeared on the feed. "Left unspecified rather than forbidden" is, in practice, forbidden with extra steps.

Delete the flag and there is nothing to migrate *into*: every type is unit-bound, reason-coded, and feed-shaped from its first record. `:740` simplifies the same way — "operators MUST NOT enable the billing profile on a deployment whose active plugin publishes no qualifying ceiling" becomes a plain plugin readiness gate with no per-type enablement to police.

### The honest cost

Most of what the flag gates is **semantic**, and costs nothing to universalize — these are correctness properties that shouldn't be optional in a metering gear: unit binding, quantity semantics + sampling interval, `reason_code`, the unstripped field set on read, record identity on reads, the future-time bound.

Two are **operational** and do carry real cost:

- **Feed inclusion.** `nfr-replay-throughput` (`:757`) rests explicitly on the billing subset being smaller than the whole — "the gear-wide envelope covers all telemetry, including high-volume non-billing sources; the billing-relevant subset is a planning assumption". If every type is billing-relevant, `R_billing` becomes the full ~36M/h envelope and `R_read ≥ 5 × R_billing` lands near 180M/h rather than 50M/h.
- **Retention floor.** 90 days of raw retention across all telemetry rather than meters only.

I don't think that rescues the flag, but it does locate where scoping belongs: those two are **capacity and deployment** concerns, not correctness ones. Scope them as such — a consumer-side feed filter and a per-type retention policy — rather than through a profile that also decides whether a record is well-formed.

### Suggested change

1. Drop `fr-billing-usage-type-profile`. Make §5.9's semantic requirements unconditional for every UsageType — exactly the shape `fr-usage-windows` already takes, so `:484` shows the document already accepts the pattern.
2. Remove the "billing-relevant" qualifier from the FRs it gates, and drop `:457` together with the unspecified-migration hole.
3. Re-scope feed inclusion and the retention floor as deployment/capacity concerns, and revisit the `R_read ≥ 5 × R_billing` planning assumption at `:756` against the wider base.
4. Collapse the §9 conditioning at `:1281` and drop the `:1283` regression guard, which exists only to describe the non-billing branch.

Breaking changes are affordable here for exactly the reason this PR already invokes for `created_at` → windows: the gear is unreleased. Better to absorb the whole break in one Phase-2 wire change than to ship a mode switch and find out at first invoice which types ended up on the wrong side of it.

gears/system/usage-collector/docs/PRD.md
```diff
@@ -435,6 +440,198 @@ The system **MUST** treat its persisted data as one of three classes:
...
+
+Two consequences **MUST** hold. First, two records covering different windows can no longer collide on one `id`, so an emitter may keep a stable per-meter idempotency key across windows — the window bounds already distinguish them. Second, offline `corrects_id` pre-computation now requires the target's exact window bounds (µs) instead of its `created_at`.
+
+- **Depends on**: `cpt-cf-usage-collector-fr-idempotency`
+- **Rationale**: ADR-0014 folded `created_at` into the derivation precisely so that two submissions differing only in when they happened would not collide. Once the timestamp is replaced by an interval, the same argument applies unchanged to the interval, so promoting the bounds is the consistent move rather than a new idea — and it is strictly better than the alternative of leaving the derivation alone and making "encode the window into your idempotency key" an emitter obligation, which pushes an identity concern onto every integrator and fails as a runtime conflict when they get it wrong. The namespace is deliberately left alone: it is a forever-fixed constant, and the derivation input is already changing.
```
capybutler: ## State this as a requirement, and move the derivation out of the PRD

Two separable issues with this FR — how it's framed, and how far down it goes.

### 1. It's written as a change request, not as a requirement

Almost every normative sentence here describes a *delta* rather than an *obligation*:

- the heading defines the FR by what it isn't — "(no new identity)"
- "**No new identifier is introduced.**"
- "**What changes is the input to the derivation, because the emitter-supplied timestamp `created_at` no longer exists**"
- "Two consequences **MUST** hold" — consequences of the change, not properties of the system
- "offline `corrects_id` pre-computation **now requires** … instead of its `created_at`"
- the Rationale argues the change is "the consistent move rather than a new idea"

A reader six months from now cannot learn what the system does from this FR without first knowing what it used to do. Compare `fr-idempotency` (`:213`) in the same document, which states its obligations directly and needs no history to be legible.

The change narrative is genuinely worth recording — just not here. It already has three proper homes: the PR description, §14 (`:1349`/`:1351`), and the replacement ADR this change requires. Note it's also currently duplicated into `fr-usage-windows` at `:497` ("Consequences for identity and dedup"), so the same delta is narrated twice.

### 2. The UUID derivation is ADR/DESIGN material, not PRD material

> `id = UUIDv5(NS, tenant_id ⟨0x1F⟩ gts_id ⟨0x1F⟩ window_start_micros ⟨0x1F⟩ window_end_micros ⟨0x1F⟩ idempotency_key)`, with the namespace constant unchanged

The algorithm (UUIDv5), the separator byte (`0x1F`), the field order, the microsecond encoding, and the namespace constant are all construction details. None is a product requirement, and all of them are exactly what `ADR-0013-deterministic-usage-record-id` and `ADR-0014-created-at-in-dedup-identity` exist to record.

This FR makes the point against itself at `:497`: "A PRD cannot retire an ADR: a replacement ADR **MUST** accompany this change." If the PRD can't retire the ADR that owns this derivation, it shouldn't restate the derivation either — otherwise the formula lives in two documents that can drift, with the non-authoritative copy being the one people read first.

The same leak reaches §9: `:1287` bakes "with the namespace constant unchanged" into an acceptance criterion.

What belongs at PRD level is the **observable property**, which is testable without naming an algorithm.

### Suggested rewrite

> #### Record Identity
>
> - [ ] `p1` - **ID**: `cpt-cf-usage-collector-fr-billing-record-identity`
>
> Every accepted usage record **MUST** carry a system-assigned identifier that is:
>
> - **server-derived** — never client-supplied;
> - **stable** — an exact-equality retry of the same submission yields the same identifier;
> - **unique per `(tenant_id, gts_id, idempotency_key, window_start, window_end)`** — records covering different windows are therefore distinct even under a single stable per-meter idempotency key;
> - **reproducible offline** by the emitter from that tuple alone, with no round-trip, so a correction reference can be computed before submission;
> - **addressable** — point lookup, deactivation, and the `corrects_id` correction pointer all resolve against it.
>
> Downstream billing **MUST** use this identifier as its deduplication and reference key, and every billing read path **MUST** carry it (`cpt-cf-usage-collector-fr-billing-fields-on-read`).
>
> The derivation algorithm and its namespace constant are recorded in ADR-0013 and its replacement, not in this document.

That keeps every property the FR actually needs — including both "consequences", which become plain properties of the uniqueness and reproducibility bullets rather than artifacts of a migration — and drops the construction.

### Minor

The heading says "for Billing", but nothing in the rewritten set is billing-specific: it's the record-identity contract for every UsageType. Worth dropping the qualifier (this also lines up with the separate thread on `fr-billing-usage-type-profile` — if the profile flag goes, the billing-scoping of this FR goes with it).

gears/system/usage-collector/docs/PRD.md
```diff
@@ -435,6 +440,198 @@ The system **MUST** treat its persisted data as one of three classes:
...
+
+**Acceptance instant.** The system **MUST** assign every record an `accepted_at`, a gear-assigned UTC instant recording when the gear accepted the record, and expose it on read. `accepted_at` **MUST NOT** be settable or overridable by the emitter. Late arrival is evaluated as `accepted_at` against `window_end`.
+
+**Consequences for identity and dedup.** The emitter-supplied timestamp `created_at` is a component of both the write-path dedup identity and the derived record `id` (ADR-0013 as amended by ADR-0014). Removing it **MUST** be accompanied by promoting the window bounds into both, so that the dedup identity becomes `(tenant_id, gts_id, idempotency_key, window_start, window_end)` and the `id` derivation covers the same tuple; see `cpt-cf-usage-collector-fr-billing-record-identity`. ADR-0014 recorded the opposite decision (folding `created_at` in), so adopting this change **would supersede** it. A PRD cannot retire an ADR: a replacement ADR **MUST** accompany this change, and ADR-0014 stays in the record marked superseded.
```
capybutler: ## Same framing issue here: state the end state, not the edit

Following the thread on `fr-billing-record-identity` — this FR has the same problem, and it's the more important of the two since this one applies to every UsageType.

To be clear about scope: the **Point events** block is the best normative writing in §5.9. The two selection cases and the month-close boundary argument are precise, testable, and belong exactly where they are. Nothing below touches them. The critique is confined to the normative text that describes an *edit to the document* rather than a *property of the system*, and it deliberately excludes the **Rationale** bullet — justification is where change history legitimately belongs.

### Where the normative text describes a change

- `:484` — "It **replaces** the record timestamp rather than adding a field alongside it."
- `:486` — "`created_at` — the single emitter-supplied event timestamp on the record **today** … — **MUST be removed** from the record model. The interval **replaces it**."
- `:488` — a point event is "the **direct replacement for the Phase-1 timestamp**"
- `:497` — the entire **Consequences for identity and dedup** paragraph.

"`created_at` MUST be removed" is a requirement addressed to whoever edits the schema, not to the system. It also under-specifies: it forbids the field existing *now*, but says nothing about a second emitter time field being added later. Stating the end state closes that and needs no history to read.

### `:497` is already recorded twice elsewhere

The Consequences paragraph is change governance — "Removing it MUST be accompanied by…", "ADR-0014 recorded the opposite decision", "adopting this change **would supersede** it", "a replacement ADR **MUST** accompany this change". That's real and important, but §14 already carries it verbatim at `:1351`:

> the record timestamp `created_at` is replaced by the interval `window_start` / `window_end` … This is wire-breaking and **would supersede ADR-0014**; a replacement ADR is required alongside this PRD change.

So the delta is stated in the document's designated changelog, in the PR description, and a third time as normative FR text — where it will still be sitting long after the change has shipped and stopped being a change. The dedup-tuple half of it is additionally duplicated in `fr-billing-record-identity` (see that thread).

### Suggested rewrite

> Every usage record **MUST** carry exactly one emitter-supplied time attribution: the half-open interval `[window_start, window_end)`, expressed as UTC instants with `window_start <= window_end`, describing the period the record's `value` covers. The record model **MUST NOT** carry any other emitter-supplied time field. The gear **MUST** reject timestamps without offset information and **MUST** normalize non-UTC offsets to UTC.
>
> **Point events.** A record with `window_start == window_end` is a **point event** — a single instant rather than a period, the shape used by emitters that observe discrete events. Because a zero-length half-open interval contains no instant, a point event **MUST NOT** be matched by interval-overlap logic. Query selection is therefore defined in two cases, and both **MUST** be implemented:
>
> *(two cases unchanged)*
>
> *(boundary paragraph unchanged)*
>
> **Acceptance instant.** *(unchanged — already stated as a property)*

Then delete `:497`, letting the dedup tuple live as the uniqueness property in `fr-billing-record-identity` and the ADR supersession live in §14 and the replacement ADR.

Note the rewritten first paragraph is strictly *stronger* than the original: "exactly one emitter-supplied time attribution" plus "MUST NOT carry any other emitter-supplied time field" rules out re-introducing a parallel timestamp, which "`created_at` MUST be removed" does not.

### Minor

The `:484` scope note's first clause is legitimate scoping, but it only exists because §5.9 is gated on the billing profile — if that flag goes (separate thread), the note can go with it and this simply becomes a record-model requirement in its own right.

gears/system/usage-collector/docs/PRD.md
```diff
@@ -435,6 +440,198 @@ The system **MUST** treat its persisted data as one of three classes:
...
+- **Actors**: `cpt-cf-usage-collector-actor-usage-source`, `cpt-cf-usage-collector-actor-platform-operator`
+
+#### Billing Fields on Read Paths
+
+- [ ] `p1` - **ID**: `cpt-cf-usage-collector-fr-billing-fields-on-read`
+
+For billing-relevant UsageTypes all read paths **MUST** return the full billing field set unstripped: `id`, `window_start`/`window_end`, `accepted_at`, `metering_unit`, declared `dimension_key` values, signed `value`, lifecycle status (active/inactive), the compensation/deactivation reference, and origin flag (`cpt-cf-usage-collector-fr-backfill`). Point lookup by record `id` **MUST** return the exact persisted fact.
+
+- **Depends on**: `cpt-cf-usage-collector-fr-query-raw`, `cpt-cf-usage-collector-fr-query-aggregation`
+- **Rationale**: Replay, backfill rating, and finance audits require the persisted fact with its billing identity intact; stripping fields forces consumers back into store internals.
```
capybutler: ## `metering_unit` is a UsageType property — it shouldn't be in the per-record field set

This field set mixes the two levels of the model. Everything else listed is a fact about the **record** — `id`, window bounds, `accepted_at`, `value`, lifecycle status, correction reference, origin, and the declared dimension values (which are the record's own metadata). `metering_unit` is a fact about the **UsageType**.

This isn't only a modelling preference — the rest of the document already places it on the type, in three separate places:

- `fr-usage-type-existence-and-semantics` (`:391`): "A UsageType is identified by a UsageTypeGtsId and described by its derived counter/gauge semantics and **an optional unit label**."
- `fr-metering-unit-binding` (`:517`): "The **declaration** of a billing-relevant UsageType **MUST** bind a `metering_unit`…" — bound at declaration, i.e. on the type.
- `domain-model.md:314`: `unit` is an open item whose two candidate homes are "a declared key in `metadata_fields`" or "a separate dedicated field **on the catalog row**". Both are type-level. The record was never a candidate.

`:573` is the only place in the document that treats the unit as record-level.

### The requirement is also ambiguous as written

Requiring `metering_unit` on a record leaves unspecified which of two different things is meant:

1. **A denormalized copy of the type's current binding** — redundant, since `gts_id` is already on every record and resolves the unit directly; or
2. **A snapshot of the binding in force when the record was accepted** — a genuinely stronger requirement that would make records self-describing across a unit rebinding.

These have different failure modes, and the PRD doesn't say whether a bound `metering_unit` can ever change. If it can't, (1) is pure redundancy with a second authority for the same fact. If it can, (2) needs stating explicitly, along with what happens to already-persisted records — and that's a real requirement currently written nowhere.

Worth noting the FR's own rationale asks for "the persisted fact with its billing identity intact". The metering unit isn't a persisted fact about the record; ingestion validates against it (`:517`) but never stores it on the row.

### On the consumer-convenience argument

The obvious defence is that a self-describing feed record saves the consumer a catalog lookup. But `gts_id` is already on every record, and the unit resolves **per type**, not per record — a bounded, trivially cached join over a small catalog, regardless of page size. Denormalizing a type attribute onto every row of a ≥10,000 records/sec stream to avoid a cache hit is a poor trade, and it introduces a second place the same fact can be read from and disagree.

### Suggested change

Drop `metering_unit` from the field set; keep everything else, including the declared dimension values, which are genuinely record-level. Something like:

> For billing-relevant UsageTypes all read paths **MUST** return the full billing field set unstripped: `id`, `window_start`/`window_end`, `accepted_at`, `gts_id`, declared dimension values, signed `value`, lifecycle status (active/inactive), the compensation/deactivation reference, and origin flag. The record's metering unit is resolved from its UsageType declaration (`cpt-cf-usage-collector-fr-metering-unit-binding`) and is not carried per record. Point lookup by record `id` **MUST** return the exact persisted fact.

If per-record capture *is* wanted, then the case to make is unit rebinding — and that should be specified head-on in `fr-metering-unit-binding`: whether a bound unit is immutable, and if not, what the snapshot semantics are. Right now it's neither ruled out nor specified, and `:573` quietly implies an answer.

Same applies to `:1299`, which restates the field set as an acceptance criterion.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Agreed on all three points, and the code references settle it — I built this FR on a premise I took from the §5.5 wording without checking `domain-model.md` or the implementation. Done in 9457030:

1. `fr-dimension-key` dropped.
2. `fr-record-metadata` now carries the grouping contract: any declared key of the queried UsageType, any combination and order, alongside the fixed fields, with admissibility computed per request from the type's declared set and rejection **before** plugin dispatch. I stated the rejection explicitly rather than just the permission, since that is the half that is currently missing.
3. The time-bucket claim is gone. You are right that asserting it hid a real gap — grouping by time period is specified nowhere and implemented nowhere, and a charging consumer that wants hourly buckets will need it. I have not added it here; it is a separate requirement, and I would rather raise it as one than smuggle it in as an existing capability.

On the cardinality point: I kept the bounding language but moved it off the grouping surface, since as you say the limit plus the mandatory bounded range already cover it, and per-key suitability belongs on the key's declaration.

Noted the missing admissibility check in `service.rs:1883` and the same gap on the `metadata.<key>` filter side-channel. That one matters to us directly: a silently-empty bucket for an undeclared key means a charge computed against an aggregate that quietly grouped everything together. Happy for that to be a follow-up, but I would treat it as a correctness bug rather than a nice-to-have.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Convinced. The flag is gone in 9457030, along with the enable-on-existing hole, and every requirement now applies to every UsageType.

You caught the flag with my own premise, which is fair — if "unreleased, no billing meter exists" justifies the break, it equally removes the legacy corpus the flag was protecting. The forked model was the real cost and I had underweighted it: two validation paths and two correction models in a gear that hasn't shipped is a mode switch whose wrong branch surfaces at first invoice.

On the two you flagged as genuinely operational, I took your framing and scoped them rather than universalising the numbers:

- **Feed inclusion** is now a **per-UsageType subscription** on the consumer side (`fr-billing-usage-feed`): a consumer declares the set of `gts_id` it reads, and the feed excludes everything else from its pages, watermark, and cursor. `nfr-replay-throughput` is re-anchored to `R_subscribed` rather than the gear-wide rate, so the obligation is per subscription. I kept your 180M/h figure in the threshold as the counterexample that motivates the scoping.
- **Retention** is now a **per-UsageType policy** honoured by the storage deployment, with the `max(replay horizon, backfill window)` floor applying only to meters a charging consumer reads.

Both feel like better placements than the profile: they are properties of what a meter is used for, not of whether a record is well-formed.

§9 collapsed accordingly and the regression guard is gone.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Taken as written — your rewrite is in 9457030 essentially verbatim, including dropping the "for Billing" qualifier and renaming to `fr-record-identity`.

You are right that the derivation does not belong here. Putting `UUIDv5(...)` and the namespace constant in a PRD states an implementation where an observable property was wanted, and it also made the requirement unverifiable in the way that matters: what a consumer depends on is that the identifier is stable, unique per tuple, and reproducible offline, not how those are achieved. The PRD now points at the identity ADR for the construction.

The two "consequences" did indeed dissolve into the uniqueness and reproducibility bullets, which is a good sign that they were artifacts of describing a migration rather than requirements.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Agreed, and your version is in 9457030 as written.

The point that changed my mind is the under-specification: "`created_at` MUST be removed" forbids the field now and says nothing about a parallel timestamp reappearing later, whereas "exactly one emitter-supplied time attribution … MUST NOT carry any other emitter-supplied time field" closes that permanently. It is a stronger requirement and shorter, which is a good trade.

The Consequences paragraph is deleted. Agreed it was change governance sitting in normative text, and stated three times over — §14 now carries it once, in two sentences, and points at the identity ADR for the superseding decision rather than narrating it.

Thanks for scoping the critique explicitly around the point-event block. That was the part I was least willing to lose, and it is untouched.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Right on both counts — the level confusion and the ambiguity behind it. Fixed in 9457030.

`metering_unit` is out of the read field set, `gts_id` is in, and the text now says the unit and the quantity semantics are resolved from the UsageType declaration rather than carried per record. Your cache argument is the decisive one: denormalising a type attribute onto every row of a 10k/s stream to avoid a bounded cached join buys nothing and creates a second authority for the same fact.

On the ambiguity you surfaced, I took the position rather than leaving it open, because the alternative is worse than it first looks: **a bound unit is now explicitly immutable**, and a meter that must change unit is a new UsageType. Without that, correctly reading a historical record requires knowing which unit was in force when it was accepted, which turns every read into a point-in-time catalog lookup and makes any rebinding a silent repricing of history. If you would rather have snapshot semantics instead, that is a real alternative, but it needs the snapshot field and the migration rule spelled out, and I would rather not pay for it before something needs it.

`fr-metering-unit-binding` and the `usecase-declare-meter` alternative flow both state the immutability now, and the acceptance criterion covers the rejection.

gears/system/usage-collector/docs/PRD.md
```diff
@@ -892,6 +1117,101 @@ The canonical endpoint surface is defined in `usage-collector-v1.yaml` (sibling
...
+**Preconditions**:
+
+- Actor is an authenticated billing consumer with PDP authorization for the requested scope
+- At least one UsageType is declared and the consumer's subscription includes it
+
+**Main Flow**:
+
+1. Consumer requests a page of the billing feed for a scope and time range, supplying either no cursor (first page) or the cursor returned with the previous page
+2. System authorizes the request via PDP; PDP-returned constraints define the authorization boundary and consumer-supplied filters only narrow it further
+3. System returns a page over a consistent snapshot: records ordered by the monotonic acceptance sequence scoped per `(tenant_id, gts_id)`, each carrying the unstripped billing field set (`id`, window bounds, `accepted_at`, `metering_unit`, dimension values, signed `value`, lifecycle status, correction reference, `origin`), together with a next cursor and an explicit watermark
```
capybutler: ## `metering_unit` survives here, contradicting the FR that removed it

The FR side of the earlier `metering_unit` thread landed cleanly: `fr-billing-fields-on-read` (`:556`) drops it from the read field set and states that the unit and quantity semantics "are resolved from its UsageType declaration through `gts_id` and are not carried per record"; `fr-metering-unit-binding` (`:510`) says the same in the other direction; and the acceptance criterion at `:1284` tests it.

Step 3 of this flow still hands it out per record:

> each carrying the unstripped billing field set (`id`, window bounds, `accepted_at`, **`metering_unit`**, dimension values, signed `value`, lifecycle status, correction reference, `origin`)

So the PRD now asserts both that the unit is not carried per record and that every feed record carries it. This is the copy that would get built: §8 is where an implementer reads the feed's page shape, and it is more concrete than the FR, so it wins ties in practice.

### Same line, smaller issue: "dimension values"

With `fr-dimension-key` dropped, **dimension** is no longer a defined term in this document. `:556` says "declared metadata values", and that is what this should say. (It also survives at `:1319` as "dimension breakdown", but that is prose about downstream evidence rather than a normative field set, so it is cosmetic by comparison.)

### Suggested change

> 3. System returns a page over a consistent snapshot: records ordered by the monotonic acceptance sequence scoped per `(tenant_id, gts_id)`, each carrying the unstripped field set of `cpt-cf-usage-collector-fr-billing-fields-on-read` — `id`, `gts_id`, window bounds, `accepted_at`, declared metadata values, signed `value`, lifecycle status, the correction reference with its `reason_code`, and `origin` — together with a next cursor and an explicit watermark

Worth pointing at the FR rather than restating the list: this is the second enumeration of that field set and `:1289` is a third, which is exactly how the copies drifted apart this time.

gears/system/usage-collector/docs/PRD.md
capybutler: Verified in `94570302` — all three land.

- `fr-dimension-key` is gone repo-wide: no dangling `Depends on`, and the membership validation that existed only to police the second list went with it.
- `fr-record-metadata` (`:230`) now carries the grouping contract, and you were right to state the rejection rather than just the permission — `:1287` mirrors it including the "before plugin dispatch" half, which is the part that's actually missing in code.
- The time-bucket claim is gone, and I agree with leaving hourly bucketing to its own requirement rather than folding it in here.

One leftover you didn't introduce, flagging rather than asking: `:1238`, in the Phase-1 acceptance criteria, still reads "Aggregation results can be grouped by any combination of time bucket, tenant, subject, and resource". It's inherited from `main`, so it's out of scope for this PR — but it's the same claim we just agreed hides a real gap, so it'll want fixing wherever the time-bucket requirement eventually lands.

On the `service.rs:1883` admissibility check — agreed it's a correctness bug, not a nice-to-have, and the PRD now has a normative requirement with no code behind it. Happy for it to be a follow-up.

Nothing further from me here. I only have triage on this repo so I can't resolve threads — would you mind closing this one out?

gears/system/usage-collector/docs/PRD.md
capybutler: Verified in `94570302`. The flag is gone repo-wide, the string "billing-relevant" no longer appears in the document, and the enable-on-existing hole went with it — `:455` now says the requirements apply to every UsageType, full stop.

The two operational concerns landed better than what I suggested. I proposed scoping them; you scoped them *and* fixed the number they rested on:

- Subscription at `:1291` and `:576`, with `nfr-replay-throughput` re-anchored to `R_subscribed` at `:746` and `:748` stating explicitly that it isn't the gear-wide rate. Keeping the 180M/h figure at `:751` as the counterexample is the right call — it records why the scoping exists rather than just asserting it.
- Retention as a per-UsageType policy with the floor applying only to meters a charging consumer reads (`:583`), which puts the cost on the meters that incur it.

`:732` reads as a plain plugin readiness gate now, with no per-type enablement to police. §9 collapsed and the regression guard is gone.

Nothing further from me. I've only got triage on this repo, so I can't resolve — this one's yours to close.

gears/system/usage-collector/docs/PRD.md
capybutler: Verified in `94570302`. Renamed to `fr-record-identity`, "for Billing" dropped, and the five properties are in at `:480`–`:486`.

The construction is fully out: no `UUIDv5`, no `0x1F`, no `window_start_micros`, and no namespace constant anywhere in the file — `:490` defers to the identity ADR, and `:1281`/`:1282` test the observable properties rather than the algorithm, so the acceptance criteria no longer carry "with the namespace constant unchanged" either.

Agreed on your read of the two "consequences" dissolving into the uniqueness and reproducibility bullets. That was the part I was least sure of when I wrote the rewrite, and it holding up is decent evidence the framing was the actual problem rather than the content.

Nothing further. Can't resolve threads here (triage only) — please close when you get a chance.

gears/system/usage-collector/docs/PRD.md
capybutler: Verified in `94570302`. `:461` carries the stronger formulation, and the Consequences paragraph is gone — §14 (`:1340`) states the record-model change once and defers the superseding decision to the identity ADR rather than narrating it in normative text.

The point-event block is untouched, and `:1279` now spells out the boundary case as an explicit test — a record at `00:00` on the first of the month, queried for that month, is returned. That wasn't in what I suggested and it's a real improvement: it turns the argument for why the point case is normative into something a test run can actually fail on.

Nothing further from me. Triage-only access here, so I can't resolve — yours to close.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: One factual addendum on the time-bucket half, after a closer look: "specified nowhere" turns out to be almost right rather than exactly right. The canonical PRD does carry it — §5.5 has "SHOULD allow consumers to narrow and group results by tenant, subject, resource, and time period", and §9 has the acceptance criterion "Aggregation results can be grouped by any combination of time bucket, tenant, subject, and resource". Neither `domain-model.md` nor the implementation has it, as you said (`AggregationDimension` carries no time variant). So this is a pre-existing PRD-vs-implementation gap on `main`, not something this PR introduces or resolves — I have left both canonical statements untouched.

For whether it matters to charging, our position after thinking it through: **it does not — the charging path aggregates downstream, from raw feed records.** Three reasons, in descending order of weight:

1. **A time-bucketed aggregate is not replay-stable.** The raw feed is append-only: replaying a cursor range yields the same records, with corrections arriving as new records. A bucket is not — a late arrival (`accepted_at > window_end`) or a compensation retroactively changes an already-read bucket with no marker that it changed. An invoice input that can silently change between reads is exactly what the snapshot/cursor design exists to prevent.
2. **Allocating a spanning window to billing periods is charging policy, not a metering fact.** A record covering `[10:30, 12:30)` against hourly buckets needs a proration rule, and that rule belongs to the consumer — same family as gauge integration, which this PRD already places downstream (`fr-quantity-semantics`).
3. **The evidence chain needs raw anyway.** A charge must be reconstructable to the record-`id` set that produced it (that assumption is what backs the 90-day operational retention floor); rating from buckets would break that chain.

So the unimplemented time grouping is a dashboard/reporting concern, not a charging blocker, and we are not asking for it in this phase. If you want, the §9 criterion could be downgraded to match §5.5's SHOULD or annotated as unimplemented — but that is canonical text on `main`, so your call, separate from this PR.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Our edits crossed mid-air: `aa5c16bf` had already swapped the list to `gts_id` + declared metadata values while you were writing this. But your version is better than that fix, so I took it — the deeper point stands regardless of which copy was current. Done in `ba1cdbff`:

- §8 step 3 now carries "the unstripped field set of `cpt-cf-usage-collector-fr-billing-fields-on-read`" instead of an enumeration.
- The §9 criterion is de-enumerated the same way, stating explicitly that the FR is the **single normative enumeration** — you called it the third copy, and a test author reading the AC now has to open the FR rather than trust a restatement.
- The cosmetic `:1319` "dimension breakdown" is now "metadata breakdown", so "dimension" no longer appears as a term anywhere in the document.

One field set, one home.

Artifizer (approved):

Corw1n-of-Amber:
**Status of the failing `Validate CFS Artifacts` check — not this PR's content.**

Measured against the actual merge state after the aggregation-folds rework landed on this branch. The run reports exactly **1 error**, and it is not in anything this PR touches:

```
✗ gears/system/account-management/…/m0005_tenant_metadata_indexes.rs:61 [marker-begin-no-end]
```

That `@cpt-begin` has no matching `@cpt-end` **on `main` today** (`git show origin/main` — 1 begin, 0 end; the closing marker appears lost in the SeaORM 2.0 upgrade, 5446288b). The workflow installs constructor-studio `1.6.3.dev5`, which flags it as an error, so every PR targeting `main` currently fails this check on that pre-existing condition.

This PR's own artifacts are clean: repo-wide `cfs validate` on the branch is **0 errors** (usage-collector PRD validates PASS; the ignore blocks for the six superseded-model artifacts are documented in the PRD's §13 with their removal condition), and DCO is green across all commits.

Fix opened separately: #4570 — one comment line restoring the marker balance. Once it merges, re-running this check here should go green with no changes on this branch.

capybutler (approved):

.cf-studio/config/artifacts.toml
```diff
@@ -60,6 +60,26 @@ reason = "Ignore legacy (pre-CFS) ADR files in oagw/docs root — converted copi
...
+reason = "Ignore usage-collector DESIGN, DECOMPOSITION, features and ADRs — they still describe the superseded model (counter/gauge usage kinds, event deactivation, signed compensation) and reference concepts the PRD no longer defines, pending the rewrite to aggregation folds and invalidation tracked in PRD section 13. PRD.md still validated via autodetect."
+patterns = [
+  "gears/system/usage-collector/docs/DESIGN.md",
+  "gears/system/usage-collector/docs/DECOMPOSITION.md",
+  "gears/system/usage-collector/docs/features/*",
+  "gears/system/usage-collector/docs/ADR/*",
+  "gears/system/usage-collector/usage-collector/src/*",
+  "gears/system/usage-collector/usage-collector/tests/*",
+  "gears/system/usage-collector/usage-collector-sdk/src/*",
+]
```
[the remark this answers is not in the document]
Corw1n-of-Amber: Tracked: the wire contract (usage-collector-v1.yaml) still describes the superseded record model and is part of the §13 rewrite backlog, alongside DESIGN and DECOMPOSITION — see the PR test plan item 'DESIGN and usage-collector-v1.yaml updated in a follow-up'. Resolving as tracked.

nonameffh (approved):

gears/system/usage-collector/docs/PRD.md
```diff
@@ -428,13 +494,220 @@ Deletion is available on both the REST and in-process SDK surfaces; surface deta
...
+- **addressable** — point lookup and the reference an invalidation entry carries both resolve against it.
+
+Downstream consumers **MUST** use this identifier as their deduplication and reference key, and every read path **MUST** carry it (`cpt-cf-usage-collector-fr-billing-fields-on-read`).
+
+An invalidation entry is identified by the same derivation over the same four attributes, which is well defined because [§5.6](#56-corrections) fixes the period it carries to that of the **Usage Record** it withdraws. Its identifier differs from its target's because the two carry different idempotency keys; an invalidation submitted under its target's own key collides on all four attributes and is therefore rejected as a same-key content mismatch (`cpt-cf-usage-collector-fr-idempotency`) rather than accepted as a second entry. Entry type is deliberately **not** part of the derivation: admitting it would let one key stand for both a measurement and its withdrawal, so an emitter defect that reused a key would silently produce both instead of surfacing a conflict.
+
+The derivation algorithm and its namespace constant are recorded in the identity ADR, not in this document.
```
Artifizer, line 545: This "identity ADR" is the replacement for ADR-0014 called out in the PR description ("A PRD cannot retire an ADR, so a replacement ADR is required alongside this change — it is not in this PR yet"). Worth being explicit here that `cpt-cf-usage-collector-fr-record-identity` and `cpt-cf-usage-collector-fr-usage-windows` are `p1` and change the dedup identity for *every* UsageType, not just billing ones — so this isn't a billing-scoped follow-up, it's a wire-breaking change to the whole gear currently governed by a document that doesn't exist yet.

Given that, would it be safer to either (a) land the replacement ADR in this same PR before merge, or (b) explicitly gate merge of this section on the ADR landing first, rather than tracking it only via the Test Plan checkbox? As written, a reader implementing against this PRD today has no normative source for the derivation algorithm or namespace constant it depends on.

---

gears/system/usage-collector/docs/PRD.md
```diff
@@ -738,251 +1074,383 @@ The canonical endpoint surface is defined in `usage-collector-v1.yaml` (sibling
...
 
 ## 13. Open Questions
 
-No open questions.
+- **ADR 0012 requires superseding.** ADR 0012 chose a plugin-owned GTS type catalog precisely to eliminate a second source of truth. [§5.7](#57-usage-record-typing) moves the catalog to `types-registry`, which reverses that decision — for the different reason that the second consumer is now another gear rather than a second store inside this one. The superseding ADR is outstanding, and DESIGN, DECOMPOSITION, the SDK and Plugin SPI contracts, the feature specifications, and the TimescaleDB plugin documents still describe the plugin-owned catalog.
```
Artifizer, line 1450: Two ADR gaps are named in this PR, but only one is tracked in the Test Plan checklist above ("Replacement ADR for ADR-0014 authored and reviewed") — this one (superseding ADR-0012's plugin-owned-catalog decision) isn't listed there at all, even though it's arguably the bigger reversal: ADR-0012 chose that model specifically to avoid a second source of truth, and this PRD undoes it.

Concretely, this leaves `DESIGN.md`, `DECOMPOSITION.md`, the SDK/Plugin SPI contracts, the feature specs, and the **running TimescaleDB plugin implementation** all still describing and coding against the catalog model this section retires — and the `.cf-studio/config/artifacts.toml` change in this same PR mutes the validator against that exact drift rather than closing it. Recommend adding the ADR-0012 supersession to the Test Plan checklist alongside ADR-0014's, so both are visible blockers rather than one tracked item and one open question.

---

Artifizer (commented):
Two inline notes on the two ADR-replacement gaps this PR itself calls out (ADR-0014 in §5.9's record-identity requirement, ADR-0012 in §13's open questions) — flagging where they're tracked (or not) relative to merge.

---

gears/system/usage-collector/docs/PRD.md
```diff
@@ -428,13 +494,220 @@ Deletion is available on both the REST and in-process SDK surfaces; surface deta
...
+- **Actors**: `cpt-cf-usage-collector-actor-usage-source`, `cpt-cf-usage-collector-actor-platform-operator`
+
+#### Billing Fields on Read Paths
+
+- [ ] `p1` - **ID**: `cpt-cf-usage-collector-fr-billing-fields-on-read`
+
+All read paths **MUST** return the following unstripped: the **Usage Record**'s identifier, the GTS type it references, the period it covers, its acceptance instant, its declared metadata values, its signed quantity, its entry type where it is an invalidation, the correction linkage in both directions — for an invalidation entry, the **Usage Record** it withdraws; for an invalidated **Usage Record**, the invalidation referring to it — with its reason code, and the origin marker that distinguishes a backfilled **Usage Record** (`cpt-cf-usage-collector-fr-backfill`). The metering unit and the aggregation fold are resolved from the GTS type declaration and are not carried per **Usage Record**. Point lookup by identifier **MUST** return the exact persisted fact.
```
Artifizer: Agreeing with an earlier CodeRabbit thread on this requirement ("scope the billing field set to record-level responses") — this doesn't look resolved in the current text.

"All read paths MUST return the following unstripped" lists per-record fields — identifier, covered period, acceptance instant, signed quantity, correction linkage — and the "Depends on" line below still names `cpt-cf-usage-collector-fr-query-aggregation` alongside `fr-query-raw` and `fr-record-identity`. But an aggregate response is a computed sum/count/max over many records; it has no single identifier, no single covered period, no single quantity to return. As written, the requirement can't literally be satisfied by the aggregate path.

Suggest scoping this explicitly to the raw query, point lookup, and feed paths, and dropping (or explaining) the dependency on `fr-query-aggregation`.

---

Artifizer (commented):
Follow-up inline note on `fr-billing-fields-on-read`, agreeing with an earlier CodeRabbit thread that this doesn't look resolved yet in the current text.

---

Artifizer (changes requested):

gears/system/usage-collector/docs/PRD.md
capybutler: We'll be handing DESIGN and ADRs in a separate PR once we merge the PRD changes.

gears/system/usage-collector/docs/PRD.md
capybutler: We'll be handing DESIGN and ADRs in a separate PR once we merge the PRD changes.

gears/system/usage-collector/docs/PRD.md
Corw1n-of-Amber: Agreed — the per-record field set cannot be satisfied by an aggregate response. `fr-billing-fields-on-read` is now scoped to raw query, point lookup, and the usage feed; the aggregate path is explicitly excluded, and the dependency on `fr-query-aggregation` is dropped. Same scoping in the §9 criterion and in `fr-record-identity`.

Artifizer (approved):

PR #4451, opened 2026-08-10
fix(oagw)!: relay WebSocket frames without parsing

fluiderson:
The OAGW gateway no longer reads WebSocket frames. Once the handshake is done, it simply copies bytes back and forth between the caller and the upstream server.

The old code took each frame apart and rebuilt it. That was fragile: the reading step could be interrupted partway through, losing the bytes it had already taken, after which everything that followed was misread and the connection eventually froze. Copying bytes has nothing to interrupt partway.

It is also more faithful. Frames now arrive exactly as they were sent, so compressed frames, unusual frame types and the client's own scrambling all survive the trip. The old code quietly altered some of that. Memory use no longer depends on how big a frame is.

Idle timeouts and graceful shutdown still behave the same way. Both tell each side politely that the connection is going away, with a time limit so a peer that has stopped responding cannot hold shutdown open.

One setting was dropped, the maximum frame size. Limiting a frame requires reading frames, and the setting never did what its documentation claimed anyway. Anyone who set it must remove it, or the service will refuse to start.

## Summary by CodeRabbit

* **Improvements**
  * WebSocket streaming now relays data transparently as raw bytes, preserving frames, fragmentation, masking, and protocol details.
  * Large streaming payloads are supported without per-frame or per-message size limits.
  * Activity tracking improves idle-timeout handling in both directions.
  * Shutdown uses bounded, independent teardown with graceful Close 1001 announcements and TCP half-closes.
  * Unexpected disconnects now propagate directly without synthesized close frames.
* **Configuration**
  * Removed the WebSocket maximum frame-size setting and oversized-frame closure behavior.

### DIFF
--- a/gears/system/oagw/docs/features/0007-cpt-cf-oagw-feature-streaming.md
+++ b/gears/system/oagw/docs/features/0007-cpt-cf-oagw-feature-streaming.md
@@ -134,25 +134,20 @@ Design constraints enforced: `cpt-cf-oagw-constraint-https-only`.
 5. [x] - `p1` - **ELSE** (upstream rejects upgrade) - `inst-ws-5`
    1. [x] - `p1` - **RETURN** 502 ProtocolError with `X-OAGW-Error-Source: gateway` — upstream rejected WebSocket upgrade - `inst-ws-5a`
 6. [x] - `p1` - **FOR EACH** message from caller - `inst-ws-6`
-   1. [x] - `p1` - **IF** message exceeds configured max frame size (default: no limit; pass-through) - `inst-ws-6a`
-      1. [x] - `p1` - Send Close frame (1009 Message Too Big) to caller and abort upstream - `inst-ws-6a1`
-   2. [x] - `p1` - Forward message to upstream (text or binary frame, preserve opcode) - `inst-ws-6b`
+   1. [x] - `p1` - Relay bytes to upstream verbatim — frames are never parsed, so opcode, masking, RSV bits and fragmentation are preserved by construction, and no per-message size limit is applied - `inst-ws-6b`
 7. [x] - `p1` - **FOR EACH** message from upstream - `inst-ws-7`
-   1. [x] - `p1` - **IF** message exceeds configured max frame size (default: no limit; pass-through) - `inst-ws-7a`
-      1. [x] - `p1` - Send Close frame (1009 Message Too Big) to upstream and close caller - `inst-ws-7a1`
-   2. [x] - `p1` - Forward message to caller (text or binary frame, preserve opcode) - `inst-ws-7b`
+   1. [x] - `p1` - Relay bytes to caller verbatim, with the same guarantees as `inst-ws-6b` - `inst-ws-7b`
 8. [x] - `p1` - **IF** either side sends Close frame - `inst-ws-8`
    1. [x] - `p1` - Forward Close frame to the other side with status code and reason - `inst-ws-8a`
-   2. [x] - `p1` - Wait for Close frame response (up to close timeout) - `inst-ws-8b`
-   3. [x] - `p1` - Release connection resources - `inst-ws-8c`
+   2. [x] - `p1` - Release connection resources once the byte copy ends — the gateway does not interpret the Close frame or wait for the peer's echo - `inst-ws-8c`
 9. [x] - `p1` - **IF** upstream connection drops unexpectedly - `inst-ws-9`
-   1. [x] - `p1` - Send Close frame (1006 Abnormal Closure) to caller - `inst-ws-9a`
+   1. [x] - `p1` - Half-close the caller's leg so it observes EOF with no Close frame; the caller's own stack reports 1006 Abnormal Closure, which the gateway does not synthesise - `inst-ws-9a`
    2. [x] - `p1` - Release connection resources - `inst-ws-9b`
 10. [x] - `p1` - **IF** caller disconnects unexpectedly - `inst-ws-10`
-    1. [x] - `p1` - Send Close frame to upstream - `inst-ws-10a`
+    1. [x] - `p1` - Propagate the disconnect to upstream as a TCP half-close (EOF); no Close frame is synthesised - `inst-ws-10a`
     2. [x] - `p1` - Release connection resources - `inst-ws-10b`
-11. [x] - `p1` - **IF** idle timeout exceeded (no messages in either direction within configured timeout) - `inst-ws-11`
-    1. [x] - `p1` - Send Close frame (1001 Going Away) to both sides - `inst-ws-11a`
+11. [x] - `p1` - **IF** idle timeout exceeded (no bytes in either direction within configured timeout) - `inst-ws-11`
+    1. [x] - `p1` - Send Close frame (1001 Going Away) to both sides, each announcement bounded independently by `websocket_close_timeout_secs` - `inst-ws-11a`
     2. [x] - `p1` - Release connection resources - `inst-ws-11b`
 
 ### WebTransport Proxy Flow
@@ -274,7 +269,7 @@ The system **MUST** detect SSE responses (`Content-Type: text/event-stream`) fro
 
 - [x] `p1` - **ID**: `cpt-cf-oagw-dod-websocket-streaming`
 
-The system **MUST** handle HTTP Upgrade (`Upgrade: websocket`) by negotiating the WebSocket handshake with both the caller and upstream. Messages (text and binary frames) **MUST** be forwarded bidirectionally with opcode preservation. **IF** a configurable max frame size is set, messages exceeding the limit **MUST** trigger Close frame 1009 (Message Too Big); by default, no per-message size limit is enforced (pass-through). Close frames **MUST** be propagated with status code and reason; Close frame reason strings **MUST NOT** include internal gateway details (use standard WebSocket status codes only). The system **MUST** handle unexpected disconnects: upstream drop (send 1006 Abnormal Closure to caller), caller drop (send Close to upstream), idle timeout (send 1001 Going Away to both sides). Auth and guard plugins **MUST** execute before the upgrade handshake. Transform plugins are NOT executed on individual WebSocket frames. Upstream rejection of the upgrade **MUST** return 502 ProtocolError with `X-OAGW-Error-Source: gateway`.
+The system **MUST** handle HTTP Upgrade (`Upgrade: websocket`) by negotiating the WebSocket handshake with both the caller and upstream. After the 101 the connection **MUST** be relayed as an opaque byte stream: frames are not parsed, so opcodes, masking, RSV bits and fragmentation are preserved verbatim, and no per-message or per-frame size limit is enforced. Close frames originated by a peer **MUST** be propagated with status code and reason; Close frame reason strings the gateway itself originates **MUST NOT** include internal gateway details (use standard WebSocket status codes only). The system **MUST** handle unexpected disconnects by propagating them as a half-close rather than a synthesised frame: an upstream drop leaves the caller observing EOF (its stack reports 1006 Abnormal Closure), and a caller drop reaches upstream as EOF. The gateway **MUST** originate Close 1001 (Going Away) toward both sides on idle timeout and on graceful shutdown, announcing to each side independently under its own `websocket_close_timeout_secs` budget so an unresponsive peer cannot suppress the other side's close. Auth and guard plugins **MUST** execute before the upgrade handshake. Transform plugins are NOT executed on individual WebSocket frames. Upstream rejection of the upgrade **MUST** return 502 ProtocolError with `X-OAGW-Error-Source: gateway`.
 
 **Implements**:
 - `cpt-cf-oagw-flow-streaming-websocket`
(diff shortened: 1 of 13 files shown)


gears/system/oagw/oagw/tests/proxy_integration.rs
```diff
@@ -4455,58 +4455,6 @@ async fn proxy_websocket_idle_timeout_sends_1001() {
     server_handle.abort();
 }
 
-// 14.7: Max frame size enforcement sends Close 1009 (Message Too Big).
```
striped-zebra-dev, line 4458: 14.7 - is not just a number it is scenario number, please check the scenarios, drop and add.
Correct connection interruption is a good candidate for a scenario.

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
striped-zebra-dev, line 1: @fluiderson 
Please divide this file into `websocket.rs` and `websocket_tests.rs`

striped-zebra-dev (changes requested):

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
```diff
@@ -1,495 +1,386 @@
...
+/// Every step is inside the timeout, deliberately. The peer that triggers an
+/// idle timeout is exactly the peer that may have stopped reading, so an
+/// unbounded `write_all` here would block forever — and since the bridge task
+/// is detached (`api::rest::handlers::proxy`), that would strand the task, the
+/// upstream stream and the active-session gauge for the process's lifetime.
+///
+/// The relay is *not* resumed afterwards: the gateway has told both peers to go
+/// away, so there is nothing left worth forwarding, and restarting the copy
+/// would append bytes after the Close frame it just wrote.
+async fn close_tunnel<C, U>(client: &mut C, upstream: &mut U, grace: Duration)
```
striped-zebra-dev, line 413: The shape of this function violates separation of concerns and introduces a hidden bug.
Please rewrite it keeping separation of concerns and composition in mind.

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
```diff
@@ -1,495 +1,386 @@
...
+///
+/// This is frame *construction* of one fixed shape, not parsing. Nothing about
+/// the frame is parameterised except masking, because the gateway only ever
+/// originates one close code: 1001 on idle timeout or shutdown. A peer's own
+/// Close is relayed verbatim by the byte copy and never passes through here.
+///
+/// `mask` must be `Some` for frames sent toward the upstream — the gateway is
+/// the client on that leg, and RFC 6455 §5.3 requires client-to-server frames
+/// be masked — and `None` for frames sent toward the caller.
+fn going_away_frame(mask: Option<[u8; 4]>) -> Vec<u8> {
```
striped-zebra-dev, line 45: Can we please avoid extra allocation?

```rust
fn going_away_frame(mask: Option<[u8; 4]>) -> Vec<u8> {
    let capacity = 2 + mask.map_or(0, |_| 4) + usize::from(CLOSE_PAYLOAD_LEN);
    let mut frame = Vec::with_capacity(capacity);
    frame.push(0x88); // FIN | Close

    let payload_bytes = CLOSE_GOING_AWAY
        .to_be_bytes()
        .into_iter()
        .chain(CLOSE_GOING_AWAY_REASON.bytes());

    match mask {
        Some(key) => {
            frame.push(0x80 | CLOSE_PAYLOAD_LEN);
            frame.extend_from_slice(&key);
            frame.extend(payload_bytes.enumerate().map(|(i, b)| b ^ key[i % 4]));
        }
        None => {
            frame.push(CLOSE_PAYLOAD_LEN);
            frame.extend(payload_bytes);
        }
    }
    frame
}
```

gears/system/oagw/oagw/src/infra/proxy/service.rs
```diff
@@ -73,7 +73,6 @@ pub struct DataPlaneServiceImpl {
     /// Timeout for the WebSocket Close frame handshake.
     websocket_close_timeout: Duration,
     /// Optional max WebSocket frame payload size (Close 1009 on exceed).
```
striped-zebra-dev: Leftover doc comment - this described `websocket_max_frame_size`, which was removed just above, but the comment stayed and is now sitting on top of `streaming_idle_timeout`. Same removal in `config.rs` cleaned its doc comment up fully; this spot got missed.

striped-zebra-dev:
Suggest adding a few scenarios to `gears/system/oagw/scenarios/protocols/websocket/` to cover the new relay design (numbers `14.5`-`14.7` are free in the current index):

**`negative-14.5-stalled-caller-does-not-block-upstream-close.md`**

```markdown
# Stalled caller does not block upstream's graceful close

## Setup

- WebSocket idle timeout (or graceful server shutdown) is configured as usual.
- Caller stops reading from its socket entirely after the handshake (its TCP
  receive buffer fills and stays full - it never acknowledges anything the
  gateway sends), while the connection itself stays open.

## Steps

1. Establish WebSocket connection, upstream accepts.
2. Caller stops draining its socket (simulates a hung/unresponsive client).
3. Trigger teardown: either let the idle timeout elapse, or initiate graceful
   server shutdown.

## Expected behavior

- Upstream still receives Close 1001 ("Going Away") and has its connection
  half-closed, within `websocket_close_timeout_secs` of teardown starting -
  regardless of whether the caller ever acknowledges its own Close frame.
- The caller being unresponsive costs only the caller's own close attempt the
  grace period; it does not consume the upstream's share of it.

## What to check

- Announcing close to each side is independent - one side stalling must not
  starve the other side's close notification.
```

**`positive-14.6-frame-relayed-intact-across-split-reads.md`**

```markdown
# WebSocket frame relayed intact when delivered across multiple reads

## Setup

- Standard WebSocket proxy connection, established and upgraded.

## Steps

1. Client sends one WebSocket frame's header bytes and payload bytes as two
   separate writes, with a short delay between them (simulating ordinary TCP
   segmentation - nothing here should require frame-aligned reads).
2. Upstream reads from its side of the tunnel.

## Expected behavior

- The frame arrives at upstream byte-identical to what was sent, uncorrupted
  and unsplit at any boundary other than the one the sender chose.

## What to check

- The gateway relays raw bytes without needing to see a whole frame at once -
  this must hold no matter how the underlying reads happen to chunk the bytes.
```

**`positive-14.7-fragmented-message-relayed-without-size-limit.md`**

```markdown
# Fragmented message of arbitrary size is relayed without a per-message cap

## Setup

- Standard WebSocket proxy connection. No per-message or per-frame size limit
  exists to configure (removed - see `cpt-cf-oagw-feature-streaming.md`).

## Steps

1. Client sends one logical message as many small continuation frames whose
   combined size is large (e.g., well beyond any size that used to be
   configurable via the old `websocket_max_frame_size_bytes` setting).
2. Upstream receives the reassembled stream.

## Expected behavior

- The full message is relayed successfully; no Close 1009 is ever sent, at any
  point, regardless of total size.

## What to check

- This is intentional, current behavior, not a gap: size limiting was already
  bypassable by fragmentation before this setting was removed (a single-frame
  limit was never a real message-size limit), so this scenario documents the
  behavior change rather than testing a regression.
```

gears/system/oagw/docs/features/0007-cpt-cf-oagw-feature-streaming.md
mattgarmon: Overall changes seem good, let's fix the docs to match updated implementation:

These steps are no longer aligned:
- `inst-ws-6a` - max frame size check no longer exists
- `inst-ws-6a1` - no Close 1009 to caller; that enforcement path is gone
- `inst-ws-7a` - same as 6a for the upstream→caller direction
- `inst-ws-7a1`- no Close 1009 to upstream; drop this substep
- `inst-ws-8b` - no close handshake anymore; Close frames are relayed as opaque bytes and the gateway no longer waits for a Close echo (`websocket_close_timeout` now only bounds the idle/shutdown teardown, not a handshake)
- `inst-ws-9a` - on abnormal upstream drop the gateway no longer synthesizes a Close frame; the caller observes EOF → 1006 (which is the correct signal for a connection that closed without a Close frame)
- `inst-ws-10a` - caller disconnect now propagates as a TCP half-close/EOF to upstream, not a synthesized WS Close frame

The gateway still originates Close 1001 on idle timeout (`inst-ws-11a`) and graceful shutdown - those are unchanged. Graceful peer-initiated closes also still pass through, so `inst-ws-8a` remains accurate.

mattgarmon: This DoD still states the max-frame-size→1009 MUST and "upstream drop → 1006 / caller drop → send Close".
Please update to match the updated pass-through model.

mattgarmon, line 317: These mentioned Acceptance Criteria are no longer aligned:
- [x] Upstream WebSocket drop sends 1006 Abnormal Closure to caller
- [x] Caller WebSocket disconnect sends Close frame to upstream
- [x] When a configurable max WebSocket frame size is set, oversized messages trigger Close frame 1009 (Message Too Big)

fluiderson: - upstream drop → now "half-closes the caller's leg, so the caller sees EOF and
  reports 1006 itself"
- caller disconnect → now "propagates to upstream as a TCP half-close (EOF),
  without a synthesised Close frame"
- max frame size → replaced with "messages of any size are relayed without a
  per-frame or per-message cap; Close 1009 is never originated by the gateway"
- added one criterion for the issue-1 fix: a peer that has stopped reading
  cannot suppress or delay the other side's Close 1001

fluiderson: Rewritten around the pass-through model: opaque byte relay after the 101, no
size-limit MUST, disconnects propagated as half-close rather than synthesised
frames, and Close 1001 on idle/shutdown announced per side under its own budget.
The unchanged parts (reason strings, plugin ordering, 502 on upgrade rejection)
are untouched.

fluiderson: - `inst-ws-6a`/`6a1`, `inst-ws-7a`/`7a1` — removed. Steps 6 and 7 now say bytes
  are relayed verbatim, with opcode/masking/RSV/fragmentation preserved by
  construction and no size limit applied.
- `inst-ws-8b` — dropped. `inst-ws-8c` now carries the note that the gateway
  neither interprets the Close frame nor waits for an echo.
- `inst-ws-8a` — left as-is per your note (peer-initiated closes still pass
  through).
- `inst-ws-9a` — reworded: caller's leg is half-closed, it observes EOF, and its
  own stack reports 1006; the gateway does not synthesise the frame.
- `inst-ws-10a` — reworded to TCP half-close (EOF), no synthesised Close.
- `inst-ws-11a` — kept, plus the clarification that each side's announcement is
  bounded independently by `websocket_close_timeout_seconds` (issue 1).

Retired IDs were not reassigned — surviving steps keep their original IDs so
traceability stays stable, which is why substep numbering starts at `6b`/`7b`.
Checked first that nothing outside this file references them.

gears/system/oagw/oagw/src/infra/proxy/service.rs
fluiderson: Removed. While there, two more docs describing the now-removed handshake wait
were corrected, since they were stale for the same reason:

- `service.rs` field doc → "Bound on announcing Close 1001 and half-closing each
  leg at teardown."
- `service.rs::with_websocket_close_timeout` → "Override the timeout for
  announcing a gateway-originated close."
- `config.rs::websocket_close_timeout_secs` → no longer claims the gateway waits
  for a Close response; states the per-leg announce budget.

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
fluiderson: Applied as suggested. The payload is now a `chain` iterator extended straight
into `frame`, so the intermediate `Vec` is gone and capacity is exact (mask
bytes only counted when masking). Existing layout tests
(`going_away_frame_unmasked_layout`, `..._masked_sets_bit_and_xors_payload`)
still pass unchanged.

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
fluiderson: Confirmed the bug: both announcements ran sequentially inside **one** shared
`timeout(grace)`. With a caller that has stopped reading, `client.write_all`
pended for the entire grace period, the timeout cancelled the block, and the
upstream got **nothing** — no Close 1001, no half-close. Only the tunnel drop
eventually cleaned it up.

Split into two functions:

- `announce_close(io, frame, side, grace)` — one peer, one frame, write → flush
  → `shutdown()`, bounded by its own `grace`. Takes `impl AsyncWrite + Unpin`
  (it never reads), and logs the three outcomes distinctly.
- `close_tunnel` — composes two `announce_close` calls under `tokio::join!`, so
  each side gets its own independent budget and a stalled peer cannot starve or
  delay the other. Bounds relaxed from `AsyncRead + AsyncWrite` to
  `AsyncWrite + Unpin`.

Regression test added:
`relay_announces_close_to_upstream_even_when_the_caller_has_stalled` — asserts
the upstream receives the masked Close 1001 *and* observes the half-close while
the caller is a `StalledIo`. Verified it fails against the old shared-budget
shape (hangs on the upstream read, panics at the 2s guard) and passes on the fix.

gears/system/oagw/oagw/src/infra/proxy/websocket.rs
fluiderson: Done.

gears/system/oagw/oagw/tests/proxy_integration.rs
fluiderson: It seems like this scenario wasn't documented in the first place. The only documented scenarios in group 14 are 14.1, 14.2, 14.3, and 14.4.

gears/system/oagw/scenarios/protocols/websocket/negative-14.5-stalled-caller-does-not-block-upstream-close.md
```diff
@@ -0,0 +1,28 @@
...
+3. Trigger teardown: either let the idle timeout elapse, or initiate graceful
+   server shutdown.
+
+## Expected behavior
+
+- Upstream still receives Close 1001 ("Going Away") and has its connection
+  half-closed, within `websocket_close_timeout_secs` of teardown starting -
+  regardless of whether the caller ever acknowledges its own Close frame.
+- The caller being unresponsive costs only the caller's own close attempt the
+  grace period; it does not consume the upstream's share of it.
```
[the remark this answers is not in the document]
fluiderson: This should be fixed in the latest revision.

striped-zebra-dev (approved):

PR #4455, opened 2026-08-10
ci(release): move release-plz to its documented push-only shape

MikeFalcon77:
- triggers: drop `pull_request`; keep `push: main` and `workflow_dispatch`.
- `release` runs on every push. No concurrency group, per upstream — "the concurrency block cancels the pending job if a new commit is pushed — we can't risk to skip a release" — and two concurrent publishes are harmless. Checkout needs no `ref:`: on push, `github.sha` is the commit that landed.
- `release-pr` gets `needs: release` so it never derives the next versions from crates.io mid-publish. That replaces the `gh run list` in-flight poll. Keeps the documented per-ref group with cancel-in-progress false, and `!cancelled()` so it still runs when `release` is skipped on a manual `release-pr` dispatch.
- `cleanup-caches` moves behind the publish, gated on the action's `releases_created` output (present in the pinned v0.3.155). It fired on the release path before; keying it to an actual release keeps that cadence instead of evicting caches open PRs are using on all 3-7 daily pushes, and takes it off the release's critical path.
- drop the advisory CI-report step. With ci.yml now running on push, the CI result lands on the published commit and is visible there; the step restated it in 75 lines of bash.
- ci.yml keeps the `push: branches: [main]` trigger added for this: PR CI only validates the ephemeral refs/pull/N/merge ref, so the integrated tip of main was tested nowhere, and CHANGELOG-only release PRs got no CI at all.

## Summary by CodeRabbit

- **Release Automation**
  - Releases now run automatically after changes reach the main branch.
  - Manual release and release-PR actions remain available when needed.
  - Retried releases can recover based on published package versions.
  - Release workflows include improved cleanup and completion reporting.

- **CI**
  - Continuous integration runs for pushes to the main branch, excluding documentation-only changes.

- **Documentation**
  - Updated release guidance explains the publishing flow, validation expectations, and manual release recommendations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,14 @@ on:
     paths-ignore:
       - '**/*.md'
       - 'docs/**'
+  # PR CI only ever validates the ephemeral refs/pull/N/merge ref, so without
+  # this trigger the integrated tip of main is tested nowhere: a semantic
+  # conflict between two PRs merged minutes apart is caught by neither PR's run.
+  push:
+    branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
   workflow_dispatch:
 
 # NOTE: workflow-level paths-ignore above still skips the entire workflow for
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    types: [ closed ]
-    branches: [ main ]
   workflow_dispatch:
     inputs:
       mode:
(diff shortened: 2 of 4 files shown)


.github/workflows/release-plz.yml
```diff
@@ -123,208 +179,45 @@ jobs:
...
-          fi
-
-          printf '%s\n' "$list" | awk -F'|' '{print "  "$3": "$1" "$2}'
-
-      - name: release-plz release
+      - name: release-plz release-pr
         uses: release-plz/action@52440b50d383aa252927de395c8b2c1e0a9cf8e9 # v0.3.155
         with:
-          command: release
+          command: release-pr
```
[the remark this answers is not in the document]
MikeFalcon77: The finding is not confirmed

docs/RELEASING.md
```diff
@@ -56,7 +85,9 @@ If you need a hotfix / manual release, prefer triggering the GitHub Actions work
 2. Go to GitHub → **Actions** → **Release (release-plz)** → **Run workflow**.
 3. Select `mode = release` (publishes crates + creates GitHub Releases).
 
-Note: the workflow already runs tests before publishing. Running tests locally is optional and just gives faster feedback.
+Note: `mode = release` publishes whatever is on the target branch and reports CI status
+without blocking on it, so confirm the branch is green first. Running the workspace
+tests locally gives faster feedback than waiting for CI:
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

docs/RELEASING.md
```diff
@@ -8,14 +8,26 @@ This repository uses **release-plz** to automate:
...
+   manual intervention.
+2. `release-plz release-pr` opens or updates a **Release PR** with:
+   - crate versions (per-crate, based on each crate's `Cargo.toml`)
+   - the root [`CHANGELOG.md`](../CHANGELOG.md)
+
+   It is ordered after `release` so it never derives the next versions from crates.io
+   while a publish is mid-flight.
+
+The Release PR is labelled **`release-plz`** automatically. Merging it is what puts the new
+versions on `main`; the publish then happens on that merge's push, like any other.
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

docs/RELEASING.md
```diff
@@ -42,11 +54,28 @@ release-plz publishes crates in the correct order for intra-workspace dependenci
...
 
-```bash
-cargo test --workspace --no-fail-fast --exclude cf-gears-toolkit-macros-tests --exclude cf-gears-toolkit-db-macros
-```
+What does verify a release:
+
+- [`ci.yml`](../.github/workflows/ci.yml) runs on every push to `main`, so the tip of
+  `main` gets tested — including the integrated result of merging, which PR CI never
+  sees. Pushes that only touch markdown or `docs/**` are skipped: they change no code
+  and inherit the previous commit's result.
```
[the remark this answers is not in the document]
MikeFalcon77: partially valid. fixed

.github/workflows/release-plz.yml
```diff
@@ -22,55 +19,114 @@ on:
...
-          (github.event_name == 'workflow_dispatch' && github.event.inputs.mode == 'release')
-          || (
-            github.event_name == 'pull_request'
-            && github.event.pull_request.merged == true
-            && contains(join(github.event.pull_request.labels.*.name, ','), 'release-plz')
-          )
-        )
-      }}
+    needs: release
+    if: ${{ needs.release.outputs.releases_created == 'true' }}
```
capybutler: **Filter this cleanup on `ref`, not on the key prefix — as written it now deletes `main`'s caches too.**

`ci.yml` hardcodes `pr-` into every `shared-key` regardless of trigger (`ci.yml:151`, `:240`, `:312`, `:421`, `:541`), so `rust-cache` writes `v0-rust-pr-…` keys. Before this PR that prefix was accurate, because CI only ran on `pull_request`. The new `push: branches: [main]` trigger changes it: `main`'s builds now write keys with the same `v0-rust-pr-` prefix, scoped to `refs/heads/main` — which is exactly the default-branch cache every PR inherits.

The filter at line 146 of this file uses the key prefix as a proxy for "belongs to a PR":

```
select(.key | startswith("v0-rust-pr-"))
```

That inference no longer holds, so every release deletes the shared `main` cache along with the PR ones. The PR creates that cache and then discards it — the next `main` build and every open PR start cold.

The current cache list confirms nothing in the key distinguishes the two. Everything `ci.yml` produces is `v0-rust-pr-Linux`, `v0-rust-pr-macOS`, `v0-rust-pr-Windows`, `v0-rust-pr-clippy`, `v0-rust-pr-Linux-integration`, `v0-rust-pr-Linux-nightly-…-llvm-cov`.

The caches API returns a `ref` per entry, so one extra filter fixes it — needed in both the count/size report and the delete loop:

```
select(.key | startswith("v0-rust-pr-")) | select(.ref | startswith("refs/pull/"))
```

Alternative: make the `shared-key` prefix depend on the trigger in `ci.yml` and leave this filter unchanged.

Separately worth weighing: `.github/workflows/cache_cleanup.yml` already prunes the same prefix weekly, keeping the newest 3 per bucket and only deleting entries older than 7 days. This job is a second, more aggressive retention policy over the same keys.

.github/workflows/release-plz.yml
```diff
@@ -22,55 +19,114 @@ on:
...
     permissions:
       actions: write
     steps:
```
capybutler: **Delete this job — `.github/workflows/cache_cleanup.yml` already prunes the same caches, with retention.**

`cache_cleanup.yml` runs weekly against the identical `v0-rust-pr-` prefix, but it groups keys into buckets, keeps the newest 3 per bucket, and only deletes entries untouched for 7 days (`cache_cleanup.yml:24-27`, `:69-100`).

This job deletes every match unconditionally, up to 500. That makes the retention policy in `cache_cleanup.yml` decorative: whatever it deliberately preserves, this job removes on the next release. Two cleanup implementations with conflicting policies on one key prefix is one too many. If the weekly cadence is too slow for the quota, tune the three constants in `cache_cleanup.yml` instead.

Related, and the reason this matters more after this PR: the `startswith("v0-rust-pr-")` filter cannot distinguish PR caches from `main`'s. `ci.yml` hardcodes the `pr-` prefix in every `shared-key` regardless of trigger (`ci.yml:151`, `:240`, `:312`, `:421`, `:541`), so the new `push: branches: [main]` trigger writes `v0-rust-pr-*` caches under `refs/heads/main` — the default-branch caches every PR inherits. This PR creates that shared cache and then wipes it on each release. Deleting this job fixes that too; if you keep it, add `select(.ref | startswith("refs/pull/"))` to the jq filter.

Gating on `releases_created` is a real improvement over firing on every push. It just doesn't change what the job does when it does fire.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,16 @@ on:
     paths-ignore:
       - '**/*.md'
       - 'docs/**'
+  # PR CI only ever validates the ephemeral refs/pull/N/merge ref, so without
+  # this trigger the integrated tip of main is tested nowhere: a semantic
+  # conflict between two PRs merged minutes apart is caught by neither PR's run.
+  # This is also what gives the release workflow something to report against —
+  # release PRs that only touch CHANGELOG.md get no PR CI at all.
```
capybutler: Both claims in these two lines are inaccurate.

**"gives the release workflow something to report against"** — this commit deletes the `Report release PR check status` step from `release-plz.yml`, so the consumer this points at no longer exists.

**"release PRs that only touch CHANGELOG.md get no PR CI at all"** — release PRs also touch every bumped `Cargo.toml` plus `Cargo.lock`; the last one (`2436373a`) touched 39 of them. `paths-ignore` only skips a run when *everything* changed is markdown or `docs/**`, and `Cargo.toml` is neither, so release PRs have always received full PR CI.

Suggest dropping lines 12-13. Lines 9-11 give the real justification — the integrated tip of `main` is tested nowhere — and it stands on its own.

.github/workflows/release-plz.yml
```diff
@@ -22,55 +19,114 @@ on:
...
           rustup show
 
-      - name: release-plz release-pr
+      # No test run here. ci.yml tests every push to main, and merging a release PR is
+      # the maintainer's release decision — re-testing could only override it, never
+      # inform it. `cargo publish` still compiles every crate, so a crate that does not
+      # build cannot be published.
+
+      - name: Clear cargo registry index (workaround gix slotmap overflow)
+        run: rm -rf ~/.cargo/registry/index
```
capybutler: This step is a no-op. The `Runner-local cache cleanup` step above (line 69) already runs `rm -rf ~/.cargo/registry`, and `index` lives inside that directory. Nothing in between repopulates it — free-disk-space, setup-go, setup-protoc and `rustup toolchain install` do not download crates.

Harmless to execute, but the step name asserts a live workaround, so the next person debugging a gix slotmap overflow will assume it is mitigating something. Either narrow the earlier `rm` so this step has something to do, or drop it.

Carried over unchanged from the previous `release` job — flagging it because the diff moves both blocks.

capybutler (commented):
Three notes on comments and a redundant step. None of them affect behaviour.

docs/RELEASING.md
```diff
@@ -8,14 +8,26 @@ This repository uses **release-plz** to automate:
...
+   - crate versions (per-crate, based on each crate's `Cargo.toml`)
+   - the root [`CHANGELOG.md`](../CHANGELOG.md)
+
+   It is ordered after `release` so it never derives the next versions from crates.io
+   while a publish is mid-flight.
+
+The Release PR is labelled **`release-plz`** automatically. Merging it is what puts the new
+versions on `main`; the publish then happens on that merge's push, like any other.
+
+Nothing here keys off the label or off the Release PR being merged — `release` decides for
```
capybutler, line 30: This is accurate, but `release-plz.toml:10` still states the opposite:

```toml
# Add a stable label to the release PR so the publish workflow can filter on it.
pr_labels = ["release-plz"]
```

That comment was correct until this commit removed the `contains(join(github.event.pull_request.labels.*.name, ','), 'release-plz')` gate. The label is now read by no workflow.

Worth keeping `pr_labels` — it is still useful for humans scanning the PR list — but the comment should say so, otherwise the two files contradict each other. (Commenting here because `release-plz.toml` is not in this diff.)

.github/workflows/release-plz.yml
MikeFalcon77: Confirmed, and fixed by deleting the job rather than filtering it (see your next comment) — the `v0-rust-pr-` inference no longer runs anywhere.

Verified the mechanic first: the caches API does carry `ref`, all 47 current entries are `refs/pull/*`, and `main`'s would have matched the old prefix-only filter once the push trigger landed.

.github/workflows/release-plz.yml
MikeFalcon77: Done — job removed, along with the `releases_created` job output and the `id: release` that fed it, whose only consumer was this job's `if`.

Recording the measurement for whoever hits this next: quota is at 9.93/10 GB. If PR builds start coming up cold, the lever is `cache_cleanup.yml`'s three constants or its cadence, not a second deleter.

.github/workflows/ci.yml
MikeFalcon77: Both lines dropped.

One correction for the record: release PRs have *not* always received full PR CI. 4 of the last 10 (#4370, #4337, #4335, #4283) changed only `CHANGELOG.md`, and #4370's head has zero `ci.yml` runs. The lines still had to go, but for a different reason than stated — the push trigger doesn't fix that case either: such a merge is markdown-only, so `paths-ignore` skips the push run too. The integrated-tip justification stands on its own, as you say.

.github/workflows/release-plz.yml
MikeFalcon77: Confirmed no-op and removed. It became one when the test step was dropped — that run was what repopulated the registry.

Moved the gix slotmap note onto `Runner-local cache cleanup`, which is the `rm` that now provides the mitigation, so the reasoning isn't lost with the step.

docs/RELEASING.md
MikeFalcon77: Fixed — pulled `release-plz.toml` into this PR. The comment now says the label is there for humans scanning the PR list and that no workflow filters on it. `pr_labels` kept.

docs/RELEASING.md
```diff
@@ -42,11 +56,32 @@ release-plz publishes crates in the correct order for intra-workspace dependenci
...
+  sees. Pushes touching only markdown or `docs/**` are filtered out by the workflow's
+  `paths-ignore`, so no run is created for them: such a commit has no CI result of its
+  own and inherits none. That is acceptable only because its code is identical to the
+  previous commit, which does have one. If `CI` is ever made a required status check,
+  this case needs an always-succeeding placeholder job — a requirement whose workflow
+  never runs is never satisfied.
+- The publish job runs on the pushed commit itself, so what goes to crates.io is exactly
+  what landed on `main`.
+- `cargo publish` compiles every crate, so a crate that does not build cannot be
+  published. It builds lib and bin targets only — it runs no tests.
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

capybutler (approved):

PR #4542, opened 2026-08-14
feat(cluster): cluster gear becomes deployable out of process

asmith987:
Makes the `cluster` gear runnable as its own process while staying embeddable in a monolith. The governing constraint is that **one consumer source file compiles and behaves identically in both deployments** — no `cfg`, no mode flag, no profile-specific code.

## Using cluster as a consumer

Two files. No wiring call, no endpoint, no `deps = [cluster]`.

**Declare a profile and resolve what you need:**

```rust
use cluster_sdk::{
    ClusterCacheV1, ClusterError, ClusterProfile, DistributedLockV1,
    cache::CacheCapability, lock::LockCapability,
};
use toolkit::client_hub::ClientHub;

#[derive(Clone, Copy)]
pub struct OrdersProfile;
impl ClusterProfile for OrdersProfile {
    const NAME: &'static str = "orders";
}
cluster_sdk::register_cluster_profile!(OrdersProfile);

pub struct OrdersCoordination {
    pub cache: ClusterCacheV1,
    pub lock: DistributedLockV1,
}

impl OrdersCoordination {
    pub async fn resolve(hub: &ClientHub) -> Result<Self, ClusterError> {
        let cache = ClusterCacheV1::resolver(hub)
            .profile(OrdersProfile)
            .require(CacheCapability::Linearizable)
            .resolve()
            .await?
            .scoped("orders")?;

        let lock = DistributedLockV1::resolver(hub)
            .profile(OrdersProfile)
            .require(LockCapability::Linearizable)
            .resolve()
            .await?
            .scoped("orders")?;

        Ok(Self { cache, lock })
    }
}
```

**Wire it into the gear:**

```rust
#[toolkit::gear(name = "orders", capabilities = [rest, stateful])]
#[derive(Default)]
struct Orders {
    hub: OnceLock<Arc<ClientHub>>,
    coord: OnceLock<OrdersCoordination>,
}

#[async_trait]
impl toolkit::Gear for Orders {
    async fn init(&self, ctx: &GearCtx) -> anyhow::Result<()> {
        drop(self.hub.set(ctx.client_hub()));   // capture only — do not resolve here
        Ok(())
    }
}

#[async_trait]
impl toolkit::contracts::RunnableCapability for Orders {
    async fn start(&self, _cancel: CancellationToken) -> anyhow::Result<()> {
        let hub = self.hub.get().expect("init ran");
        drop(self.coord.set(OrdersCoordination::resolve(hub).await?));
        Ok(())
    }
}

impl toolkit::contracts::RestApiCapability for Orders {
    fn healthcheck(&self, ctx: &GearCtx) -> Option<Arc<dyn toolkit::Healthcheck>> {
        Some(cluster_sdk::cluster_readiness(ctx.client_hub()))
    }
}
```

Then just use it: `self.coord.get().unwrap().lock.lock("batch-42", ttl).await?`.

## Why consumer code doesn't change

- **You name a profile, not an endpoint.** `"orders"` is a request parameter, resolved to a bound backend server-side. You never learn which provider serves it, so plugin linkage stays the cluster gear's concern.
- **The SDK wires the client, not you.** `cluster-sdk` submits its own `ConsumerRegistration` via `inventory`; the framework's proxy-wiring phase replays it *before any gear's `start`*. Embedded, the cluster gear registers a local client in its own `init` and local wins via `try_get_local`. Either way the hub holds exactly one `dyn ClusterClient` by the time you resolve — which is why there's no retry loop and no readiness wait in consumer code. Linking the SDK is the entire opt-in.
- **The facades are frozen**, and the embedded path gets the real backend `Arc` with no wrapper interposed — it isn't taxed for the remote path's existence.
- **No configuration exists.** No endpoint key, no mode flag, no timeouts. The out-of-process endpoint is derived by DNS convention.
- **Failures match too.** `ProfileNotBound` and `CapabilityNotMet` produce identical error text in both deployments, because the inline check and the deferred readiness verdict run the *same validator closure*.

## What you do write once

Adopting the SDK costs exactly three things:

1. **`.resolve()` → `.resolve().await`**, and the enclosing fn becomes `async`. The only SDK signature change; the compiler enumerates every site.
2. **One line in `healthcheck()`.** The framework collects healthchecks per *gear* — there's no inventory for them — so an SDK provably cannot self-register this one. Omit it and a requirement that becomes unmet *after* startup never gates your traffic.
3. **Resolve in `start`, never `init`.**

## Three things that bite

- **`deps = [cluster]` is fatal out-of-process.** It's a hard topo-sort edge, and `RegistryBuilder::build` fails the *whole registry* when the named gear isn't linked. You lose nothing by omitting it: ordering is free (cluster is `system`-priority) and the `/readyz` gate comes from the SDK's own registration.
- **The `healthcheck()` line is load-bearing**, per above.
- **`init`-time resolve fails asymmetrically** — `Ok` with an unbound facade out-of-process, `Err(ProfileNotBound)` embedded. It passes testing and fails deployment.

## Picking a profile name

A profile names a *backend binding*, not an environment or a tenant. Sharing one across gears is supported and tested — but a profile is not a trust boundary and partitions nothing, so sharing means sharing a keyspace, which is why `.scoped("orders")` above is load-bearing. Prefer workload-shaped names (`orders`) over environment-shaped ones (`production`): the environment is the deployment, and workload-shaped names keep readiness blast radius contained to the gear that owns the profile.


striped-zebra-dev:
Due to temporary absence of @asmith987 , I'll be taking over this work.
A new PR will be opened shortly.

Issue #4509, opened 2026-08-12
[toolkit] Two divergent retry policies: toolkit-http and toolkit-contract disagree on what is retryable


MikeFalcon77:
Part of #4474.

## Problem Statement

The two client paths in this repository retry differently, and neither doc explains why.

**Retryable set.** `toolkit-http` (`libs/toolkit-http/src/config.rs:210-231`):

- `always_retry` = `{429}`
- `idempotent_retry` = `{transport, timeout, 408, 500, 502, 503, 504}`, gated on `is_idempotent_method(method) || has_idempotency_key` (`config.rs:287-302`)

`toolkit-contract` (`libs/toolkit-contract/src/runtime/transport_error.rs:176-182`):

```rust
fn is_retryable_status(status: u16) -> bool {
    // Deliberately excludes 408 and 500 — a bare 500 is often a deterministic
    // server-side failure and blindly retrying it (especially a write) risks
    // duplicate side effects.
    matches!(status, 429 | 502 | 503 | 504)
}
```

No verb gating here at all — the gate is the `#[retryable]` attribute on the contract method (`libs/toolkit-contract-macros/src/rest_contract.rs:722-729`).

So the same 500 from the same upstream is retried by a hand-built `HttpClient` on a GET and not retried by a generated contract client on the same GET. Each policy is defensible in isolation; the divergence is not, and there is no document that says which one a gear author should expect.

**`Retry-After` parsing.** `toolkit-http` handles both delta-seconds and the HTTP-date form (`libs/toolkit-http/src/response.rs:23-47`). `toolkit-contract` handles delta-seconds only and explicitly drops the date form (`libs/toolkit-contract/src/runtime/http.rs:176-181`):

```rust
/// Only the numeric delta-seconds form is supported (the common case for
/// `429`/`503`); the HTTP-date form is ignored (returns `None`)
```

An upstream that answers `Retry-After: Wed, 21 Oct 2026 07:28:00 GMT` gets honored on one path and silently ignored on the other.

[...]

PR #4544, opened 2026-08-14
fix(toolkit-canonical-errors): emit Retry-After for service unavailable

### LINKED ISSUE #4541: bug(toolkit-canonical-errors): service-unavailable retry hint is not emitted as Retry-After
## Problem statement

`CanonicalError::service_unavailable().with_retry_after_seconds(n).create()` carries the retry hint in the RFC 9457 problem context, but neither `Problem::into_response` nor `toolkit::api::canonical_error_middleware` emits the corresponding HTTP `Retry-After` header.

This breaks the transport contract at the HTTP boundary:

- `toolkit-http` derives retry timing from the `Retry-After` response header (`libs/toolkit-http/src/response.rs::parse_retry_after`); it does not inspect the canonical problem body.
- `gears/system/event-broker/docs/PRD.md` explicitly says retry timing for 503 is additionally surfaced through the standard HTTP header and that the canonical-errors layer owns it.
- Callers already work around the missing behavior manually. For example, mini-chat constructs a service-unavailable canonical error and then inserts `Retry-After` on the response itself (`gears/mini-chat/mini-chat/src/api/rest/handlers/attachments.rs`). OAGW similarly owns custom header projection in its error response helper.

The result is that most `.with_retry_after_seconds(...)` call sites serialize a hint that standard HTTP clients cannot observe, while a few gears emit the header through local, inconsistent workarounds.

The behavior is unchanged on current `main`: `Problem::into_response` adds only `Content-Type`, and `canonical_error_middleware` preserves/rebuilds the response without projecting the retry hint into headers.

## Minimal reproduction

Add an axum integration test beside `libs/toolkit-canonical-errors/tests/axum_integration.rs`:

```rust
#[test]
fn service_unavailable_into_response_sets_retry_after_header() {
    let err = CanonicalError::service_unavailable()
        .with_retry_after_seconds(30)
        .create();

    let response = err.into_response();

    assert_eq!(response.status(), StatusCode::SERVICE_UNAVAILABLE);
    assert_eq!(response.headers().get(header::RETRY_AFTER).unwrap(), "30");
}
```

Today this fails because the header is absent.

## Expected behavior

At the canonical HTTP boundary, a `service_unavailable` occurrence with `retry_after_seconds: Some(n)` emits `Retry-After: n` in addition to retaining `context.retry_after_seconds` in the problem body.

Errors without a retry hint must not receive the header. Non-503 categories must not accidentally project similarly named context fields.

## Proposed solution

Project the typed `ServiceUnavailable` retry hint into the response in the canonical-errors axum integration (or in the canonical middleware if that is the intended single transport boundary). Keep the body field for canonical/out-of-process consumers.

Prefer one shared projection over gear-local helpers, then remove local workarounds opportunistically in follow-ups.

## Definition of done

- A canonical service-unavailable response with `Some(30)` includes `Retry-After: 30`.
- The RFC 9457 body still includes `context.retry_after_seconds: 30`.
- `None` produces no `Retry-After` header.
- Existing status/content-type/extension behavior remains unchanged.
- Tests cover both `Problem::into_response` and `CanonicalError::into_response` if both are public supported boundaries.

parviz-93:
## Summary

- Emit the Retry-After header from the canonical service unavailable retry hint.
- Apply the header to Problem and CanonicalError Axum responses.
- Use the canonical error category instead of the effective HTTP status when selecting the header.
- Ignore absent, malformed, and unrelated context values.

## Tests

- cargo test -p cf-gears-toolkit-canonical-errors --all-features

Closes #4541

## Summary by CodeRabbit

* **New Features**
  * Service-unavailable error responses now include a `Retry-After` header when a valid retry delay is provided.
  * The header remains available when the response’s HTTP status is customized.

* **Bug Fixes**
  * Retry guidance is omitted when the delay is missing or invalid, or when the error is not service-unavailable.

### DIFF
--- a/libs/toolkit-canonical-errors/src/problem.rs
+++ b/libs/toolkit-canonical-errors/src/problem.rs
@@ -613,18 +613,38 @@ impl TryFrom<Problem> for CanonicalError {
 // ---------------------------------------------------------------------------
 
 #[cfg(feature = "axum")]
+fn service_unavailable_retry_after_seconds(problem: &Problem) -> Option<u64> {
+    if category_from_problem_type(&problem.problem_type) != Some("service_unavailable") {
+        return None;
+    }
+
+    ServiceUnavailable::deserialize(&problem.context)
+        .ok()?
+        .retry_after_seconds
+}
+
+#[cfg(feature = "axum")]
 impl axum::response::IntoResponse for Problem {
     fn into_response(self) -> axum::response::Response {
         match serde_json::to_vec(&self) {
             Ok(body) => {
                 let status = http::StatusCode::from_u16(self.status)
                     .unwrap_or(http::StatusCode::INTERNAL_SERVER_ERROR);
-                (
+                let retry_after_seconds = service_unavailable_retry_after_seconds(&self);
+                let mut response = (
                     status,
                     [(http::header::CONTENT_TYPE, APPLICATION_PROBLEM_JSON)],
                     body,
                 )
-                    .into_response()
+                    .into_response();
+
+                if let Some(seconds) = retry_after_seconds {
+                    response
+                        .headers_mut()
+                        .insert(http::header::RETRY_AFTER, http::HeaderValue::from(seconds));
+                }
+
+                response
             }
             Err(e) => {
                 tracing::error!(
--- a/libs/toolkit-canonical-errors/tests/axum_integration.rs
+++ b/libs/toolkit-canonical-errors/tests/axum_integration.rs
@@ -3,7 +3,7 @@
 use axum::response::IntoResponse;
 use toolkit_canonical_errors::problem::APPLICATION_PROBLEM_JSON;
 use toolkit_canonical_errors::resource_error;
-use toolkit_canonical_errors::{CanonicalError, Problem};
+use toolkit_canonical_errors::{CanonicalError, Http, Problem};
 
 #[resource_error(gts_id!("cf.core.test.axum.v1~"))]
 struct AxumTestR;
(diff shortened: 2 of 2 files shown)


parviz-93:
@MikeFalcon77 
[PR #4171](https://github.com/constructorfabric/gears-rust/pull/4171) and this PR solve different problems.

- PR #4171 implements gateway throttling. It returns 429 Resource Exhausted and inserts Retry-After inside the throttling middleware.
- This PR fixes the shared canonical-errors Axum boundary. It projects retry_after_seconds only for canonical Service Unavailable errors.
- PR #4171 does not change toolkit-canonical-errors. This PR does not replace its throttling-specific headers.

The two changes only share the standard Retry-After header. They do not overlap in error category, HTTP status, or ownership.

libs/toolkit-canonical-errors/src/problem.rs
```diff
@@ -612,19 +612,42 @@ impl TryFrom<Problem> for CanonicalError {
 // axum integration (feature = "axum")
 // ---------------------------------------------------------------------------
 
+#[cfg(feature = "axum")]
+fn service_unavailable_retry_after_seconds(problem: &Problem) -> Option<u64> {
+    const SERVICE_UNAVAILABLE_TYPE: &str =
+        gts_uri!("cf.core.errors.err.v1~cf.core.err.service_unavailable.v1~");
```
MikeFalcon77: This re-hardcodes the service_unavailable GTS id, which already lives in `CanonicalError::gts_type()`. There's `category_from_problem_type()` a few hundred lines up (used by `TryFrom<Problem>`) that does the prefix/suffix stripping for exactly this:

```rust
if category_from_problem_type(&problem.problem_type) != Some("service_unavailable") {
    return None;
}
```

Same behavior, one less literal to keep in sync if the URI shape or `GTS_ID_PREFIX` handling changes.

libs/toolkit-canonical-errors/src/problem.rs
```diff
@@ -612,19 +612,42 @@ impl TryFrom<Problem> for CanonicalError {
...
+                let retry_after_seconds = service_unavailable_retry_after_seconds(&self);
+                let mut response = (
                     status,
                     [(http::header::CONTENT_TYPE, APPLICATION_PROBLEM_JSON)],
                     body,
                 )
-                    .into_response()
+                    .into_response();
+
+                if let Some(seconds) = retry_after_seconds {
```
MikeFalcon77, line 641: The PR description says the header is kept "when a 5xx transport status override exists", but there's no status check here — `Http::status_code()` takes any `u16` and isn't validated on the emit path, so an override to 200 or 404 also gets `Retry-After`. Harmless per RFC 9110, but either add the 5xx gate or drop that line from the description; right now the 502 test reads like the gate exists.

libs/toolkit-canonical-errors/tests/axum_integration.rs
```diff
@@ -25,6 +25,90 @@ fn problem_into_response_sets_status_and_content_type() {
...
+fn problem_into_response_sets_retry_after_header() {
+    let problem = Problem::from(
+        CanonicalError::service_unavailable()
+            .with_retry_after_seconds(30)
+            .create(),
+    );
+
+    let response = problem.into_response();
+
+    assert_eq!(response.status(), http::StatusCode::SERVICE_UNAVAILABLE);
```
MikeFalcon77, line 38: Worth also asserting the body still carries `context.retry_after_seconds: 30` — that's an explicit DoD item in #4541 and nothing pins it right now. Cheap here: deserialize the response body back into `Problem` and check the field.

libs/toolkit-canonical-errors/tests/axum_integration.rs
```diff
@@ -25,6 +25,90 @@ fn problem_into_response_sets_status_and_content_type() {
...
+    assert!(response.headers().get(http::header::RETRY_AFTER).is_none());
+}
+
+#[test]
+fn service_unavailable_status_override_keeps_retry_after_header() {
+    let response = CanonicalError::service_unavailable()
```
MikeFalcon77, line 112: Missing the `retry_after_seconds = 0` case. Not cosmetic: `toolkit-http`'s retry layer prefers `Retry-After` over its own backoff (`parse_retry_after` in `libs/toolkit-http/src/response.rs`), so `Retry-After: 0` makes clients retry immediately — see #4509. A test pinning `0` on the wire makes that a deliberate choice rather than a side effect.

MikeFalcon77 (commented):
.

libs/toolkit-canonical-errors/src/problem.rs
parviz-93: Updated in `141d5e35`. The helper now uses `category_from_problem_type()` and compares the result with `service_unavailable`.

libs/toolkit-canonical-errors/src/problem.rs
parviz-93: Updated the PR description. The intended behavior is category-based, not status-gated. The emit path does not add a 5xx gate. The 502 test verifies that a valid same-class override keeps the header.

libs/toolkit-canonical-errors/tests/axum_integration.rs
parviz-93: No new assertion added. Existing tests already pin this requirement. `service_unavailable_serialization` asserts the JSON field. `round_trip_service_unavailable_preserves_retry_after` runs CanonicalError through Problem and JSON, then asserts `Some(30)`. `Problem::into_response` serializes the same Problem before it adds the header.

libs/toolkit-canonical-errors/tests/axum_integration.rs
parviz-93: Added `service_unavailable_zero_retry_hint_emits_zero_header` in `141d5e35`. It asserts `Retry-After: 0`. A mutation check confirmed that the test fails when the emit path omits zero. Issue #4509 owns the client backoff floor.

MikeFalcon77 (approved):

PR #4562, opened 2026-08-15
chore: refactor local E2E runner, add gear-scoped workflows, and clean up example-server/test wiring

Artifizer:
## Summary

This PR updates the local build/test workflow around focused gear operations, manifest-driven E2E suites, and make benchmarking/matrix tooling.

Key changes:

- Adds gear-scoped Makefile workflows for build, check, test, lint, OpenAPI, and E2E paths.
- Introduces manifest-driven local E2E orchestration via `tools/scripts/run_e2e.py`.
- Renames E2E layout from `testing/e2e/gears` to `testing/e2e/suites` and updates suite manifests/docs.
- Adds Python helpers for make benchmarking and make matrix execution.
- Adds persisted benchmark scenario logs and clearer benchmark scenario naming.
- Adds a gear-scoped PR workflow.
- Updates example-server feature selection and Python tooling execution from the managed venv.
- Fixes OpenAPI generation bootstrapping by linking AM GTS envelope providers.
- Adds rustls corecrypto serial-test locking support for handshake smoke tests.

## Rebase

Rebased `make-optimization` onto current `origin/main` (`Artifizer/gears-rust`). No conflicts were reported during the rebase.

## Validation

Not run after the rebase in this session.


Makefile
```diff
@@ -445,57 +536,131 @@ web-docs-preview:
...
+GEAR_BUILD_ARGS ?=
+GEAR_TEST_ARGS ?=
+GEAR_RUN_ARGS ?=
+GEAR_FEATURES ?=
+
+GEAR_FEATURE_ARGS := $(if $(GEAR_FEATURES),--features $(GEAR_FEATURES),)
+GEAR_SERVER_BASE_FEATURES ?= static-tenants,static-authn,static-authz
+GEAR_SERVER_ALWAYS_LINKED ?= api-gateway gear-orchestrator types-registry tenant-resolver authn-resolver authz-resolver
+GEAR_SERVER_OPTIONAL_FEATURES := $(filter-out $(GEAR_SERVER_ALWAYS_LINKED),$(GEAR))
+GEAR_SERVER_FEATURES ?= $(GEAR_SERVER_OPTIONAL_FEATURES)$(if $(GEAR_SERVER_OPTIONAL_FEATURES),$(COMMA),)$(GEAR_SERVER_BASE_FEATURES)
```
MikeFalcon77, line 642: This passes the scope directory name straight through as a cargo feature of `cf-gears-example-server`, and for most scopes that feature doesn't exist.

`example-server`'s feature list has no `toolkit-db`, `cluster`, `event-broker`, `quota-enforcement`, `license-resolver`, `approval-service`, `llm-gateway`, `model-registry`, `serverless-runtime`, `infrastructure-resource-manager`, `simple-resource-registry`. `gears/bss/ledger` resolves to `ledger`, but the feature is `bss-ledger`.

So `.example-server-build` and `openapi` fail with "package does not have feature X", which means `make all GEAR=<scope>` fails outright. Of the 53 scope roots the workflow discovers, roughly 12 work. Every `libs/*` scope fails, even though the workflow header explicitly lists `libs/<gear>/` as supported.

Repro: `make all GEAR=toolkit-db`, `make all GEAR=cluster`, `make all GEAR=ledger`.

This needs an explicit scope → server-feature mapping (a field in the gear manifest would do), and `openapi` / `e2e-local` should no-op when a scope has no server feature, the way `run_e2e.py` already no-ops when no suite matches.

Makefile
```diff
@@ -445,57 +536,131 @@ web-docs-preview:
...
+GEAR_E2E_KEY := $(subst -,_,$(GEAR))
+GEAR_OPENAPI_TMP ?= target/openapi/$(GEAR).json
+GEAR_E2E_TARGET ?= testing/e2e/suites/$(GEAR_E2E_KEY)
+GEAR_PKG ?= cf-gears-$(GEAR)
+GEAR_SDK_PKG ?= $(GEAR_PKG)-sdk
+GEAR_SDK_FLAG := $(if $(wildcard gears/$(GEAR)/$(GEAR)-sdk),-p $(GEAR_SDK_PKG))
```
MikeFalcon77: Two problems with deriving the package set by convention here:

1. `cf-gears-$(GEAR)` is wrong for several crates — `gears/bss/ledger/ledger` is `bss-ledger`, and `users-info` / `calculator` don't carry the prefix either.
2. The wildcard only looks under `gears/<gear>/`, so for grouped gears (`bss/ledger`, `system/cluster`) and all of `libs/*` the SDK crate silently drops out of `build` and `clippy`. Silently — no error, just less coverage than the target name implies.

`resolve_gear_deps.py` already handles both cases correctly (multi-root lookup, real package names from `cargo metadata`). Worth routing `GEAR_PKGS` through it rather than maintaining a second, weaker resolver.

Makefile
```diff
@@ -251,20 +303,27 @@ CLIPPY_FLAGS := -- -D warnings -D clippy::perf
...
 
 clippy:
+	$(call print_target_banner)
 	$(call check_rustup_component,clippy)
+ifeq ($(GEAR),)
 	$(call check_tool,cargo-hack)
 	cargo clippy --workspace --all-targets --all-features $(CLIPPY_FLAGS)
 	cargo hack clippy $(CLIPPY_HACK_CRATES) --all-targets --each-feature $(CLIPPY_FLAGS)
+else
+	cargo clippy -p $(GEAR_PKG) $(GEAR_CLIPPY_ARGS)
```
MikeFalcon77: `test GEAR=` runs the gear's crates *plus* their transitive reverse deps via `resolve_gear_deps.py`, but `clippy` (and `fmt` above) run on `-p $(GEAR_PKG)` only.

That asymmetry matters most for exactly the scopes the new workflow targets: a change in `libs/toolkit-db` that breaks a lint in a downstream gear passes the scoped clippy lane and only surfaces in full CI. Since the goal is eventually to *replace* full CI for these PRs, clippy and fmt should use the same resolved package set as `test`.

.github/workflows/gear-scoped-ci.yml
```diff
@@ -0,0 +1,261 @@
...
+
+      - name: Cache Cargo/target
+        uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+        with:
+          cache-bin: false
+          cache-targets: false
+          shared-key: pr-gear-scoped-${{ needs.detect.outputs.scope_id }}-${{ hashFiles('rust-toolchain.toml') }}
+
+      - name: Mark required tools installed
+        run: touch .setup-stamp
```
MikeFalcon77, line 258: The job can't run `make all` with the tools it installs.

`make all GEAR=…` → `check` → `fmt cfs-validate clippy lychee security dylint gts-docs test`. This job installs protoc, Go, cargo-hack, nextest and cargo-llvm-cov. `check_tool` will fail on `cfs` (pipx install from the studio repo), `lychee`, `cargo-gears` (dylint), `gts-validator` (gts-docs), and `cargo-deny` — note `security` runs *before* `test`, and `test` is what pulls in `install-tools` where cargo-deny gets installed.

`touch .setup-stamp` only short-circuits the `setup` target; nothing here actually invokes `setup`, so it doesn't help. As written the scoped job would go red on the first PR that hits it. Either install the full tool set or point the fast path at a narrower target than `all`.

This hasn't shown up yet because the job has never executed — it's `skipping` on this PR, since the PR itself isn't scoped. Worth putting up a throwaway single-gear PR to exercise it before merge.

.github/workflows/gear-scoped-ci.yml
```diff
@@ -0,0 +1,261 @@
...
+      - name: Install cargo tools
+        uses: taiki-e/install-action@6ef672efc2b5aabc787a9e94baf4989aa02a97df # v2.70.3
+        with:
+          tool: cargo-hack, nextest, cargo-llvm-cov
+
+      - name: Cache Cargo/target
+        uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+        with:
+          cache-bin: false
+          cache-targets: false
```
MikeFalcon77, line 254: No sccache here, and target caching is off, so every scoped run is a cold build of the whole dependency graph.

Every Rust job in `ci.yml` sets up `mozilla-actions/sccache-action` and probes it into `RUSTC_WRAPPER` — that's what makes `cache-targets: false` workable there. Without it, the "fast path" plausibly ends up slower than the lanes it's meant to short-circuit, which would invert the whole point. The benchmark numbers in the description are warm local builds on a dev machine; ubuntu-latest from cold is a very different shape.

Also worth noting `shared-key` includes `scope_id`, so each gear gets its own cache bucket and none of them share the common dependency build.

.github/workflows/gear-scoped-ci.yml
```diff
@@ -0,0 +1,261 @@
...
+
+              cmd_parts = ["make", f"GEAR={gear}", f"GEAR_PKG={gear_pkg}"]
+              if sdk_pkg:
+                  cmd_parts.append(f"GEAR_SDK_PKG={sdk_pkg}")
+              cmd_parts.extend(make_target.split())
+
+              return scope_kind, gear, gear_pkg, " ".join(cmd_parts)
+
+          diff = subprocess.run(
+              ["git", "diff", "--name-only", os.environ["BASE_SHA"], os.environ["HEAD_SHA"]],
```
MikeFalcon77: Two-dot diff against `base.sha` compares the two tips, so anything merged into `main` after this PR branched shows up as changed. Given the gate requires *every* changed path to sit in one scope, that turns into spurious non-scoped verdicts that grow with branch age — on an already very low hit rate.

Three-dot (`$BASE_SHA...$HEAD_SHA`) or an explicit `git merge-base` gives the actual PR contents. `ci.yml` sidesteps this by using `dorny/paths-filter`, which handles the base resolution for you — might be worth reusing here for consistency.

MikeFalcon77 (commented):
### On the gear-scoped CI idea

The direction is right and the motivation is clear. But I think the gate as designed will almost never fire, and the scoped lane it guards doesn't work yet for most scopes.

**The gate fires on 2.7% of PRs.** I ran the detector logic from `gear-scoped-ci.yml` against the last 299 merge commits on `main`:

| Gate variant | PRs matched |
|---|---|
| As implemented (exactly one scope root) | **8 / 299 (2.7%)** |
| + ignore `*.md` and `docs/` | 12 (4.0%) |
| + allow more than one scope | 19 (6.4%) |

The reason is structural: the gate keys on *file location*, and 82% of merged PRs touch at least one workspace-global path (`Cargo.lock` in 205 PRs, root `Cargo.toml` in 182, plus `tools/` and `.github/`). Relaxing the gate doesn't fix it — even the most permissive variant is 6.4%.

Also, of those 8 matches, at least 4 (`libs/toolkit-canonical-errors`, `libs/rustls-fips-shim`, `gears/system/event-broker`, `gears/bss/ledger`) would have failed on the Makefile issue noted inline below. So the realistic hit rate today is ~1%, and full CI still runs anyway.

**Suggested alternative.** Key on the affected *package set* rather than on directories: changed files → owning packages → transitive reverse deps → run that set. Then there's no gate at all — a workspace-file change simply resolves to "all packages", and multi-scope PRs get partial narrowing instead of nothing. `resolve_gear_deps.py` already implements the hard part (rdep walk over `cargo metadata`); it mainly needs a `--changed-since <base>` mode. If you'd rather not hand-roll it, `guppy`'s `determinator` crate does exactly this for Cargo workspaces (used by Diem/Aptos/Sui for `x test --changed-since`), including mapping `Cargo.lock` diffs to specific affected packages instead of invalidating everything.

**One thing to settle now, not later.** Since this is a prerequisite for skipping full CI, the required-check story matters. `Scoped CI (${{ needs.detect.outputs.scope_path }})` has a dynamic name — it shows up unexpanded in `gh pr checks` and can't be made a required status check. The usual pattern is a fixed-name aggregating job (`if: always()`) that collapses the conditional jobs into one result. Worth adding before step 2, otherwise branches either block forever or the check goes green vacuously.

Artifizer:
> ### On the gear-scoped CI idea
> 
> The direction is right and the motivation is clear. But I think the gate as designed will almost never fire, and the scoped lane it guards doesn't work yet for most scopes.
> 
> **The gate fires on 2.7% of PRs.** I ran the detector logic from `gear-scoped-ci.yml` against the last 299 merge commits on `main`:
> Gate variant 	PRs matched
> As implemented (exactly one scope root) 	**8 / 299 (2.7%)**
> + ignore `*.md` and `docs/` 	12 (4.0%)
> + allow more than one scope 	19 (6.4%)
> 
> The reason is structural: the gate keys on _file location_, and 82% of merged PRs touch at least one workspace-global path (`Cargo.lock` in 205 PRs, root `Cargo.toml` in 182, plus `tools/` and `.github/`). Relaxing the gate doesn't fix it — even the most permissive variant is 6.4%.
> 
> Also, of those 8 matches, at least 4 (`libs/toolkit-canonical-errors`, `libs/rustls-fips-shim`, `gears/system/event-broker`, `gears/bss/ledger`) would have failed on the Makefile issue noted inline below. So the realistic hit rate today is ~1%, and full CI still runs anyway.
> 
> **Suggested alternative.** Key on the affected _package set_ rather than on directories: changed files → owning packages → transitive reverse deps → run that set. Then there's no gate at all — a workspace-file change simply resolves to "all packages", and multi-scope PRs get partial narrowing instead of nothing. `resolve_gear_deps.py` already implements the hard part (rdep walk over `cargo metadata`); it mainly needs a `--changed-since <base>` mode. If you'd rather not hand-roll it, `guppy`'s `determinator` crate does exactly this for Cargo workspaces (used by Diem/Aptos/Sui for `x test --changed-since`), including mapping `Cargo.lock` diffs to specific affected packages instead of invalidating everything.
> 
> **One thing to settle now, not later.** Since this is a prerequisite for skipping full CI, the required-check story matters. `Scoped CI (${{ needs.detect.outputs.scope_path }})` has a dynamic name — it shows up unexpanded in `gh pr checks` and can't be made a required status check. The usual pattern is a fixed-name aggregating job (`if: always()`) that collapses the conditional jobs into one result. Worth adding before step 2, otherwise branches either block forever or the check goes green vacuously.

Great summary

The gear-scoped-ci.yml is experimental for now. It's mostly for quick small fix CI as of now. The main intent of this PR is to offer developers the GEAR= flag to speed-up local testing before main CI

The later gear-scoped-ci.yml optimization to be done later as another big topic. I submitted an issue:
https://github.com/constructorfabric/gears-rust/issues/4585

Makefile
Artifizer: AI TODO: to confirm it's a valid concern. If yes - suggest a fix for it

Makefile
Artifizer: AI TODO: Investigate if it's valid concern carefully. I'm not sure here because `clippy` and `fmt` validates code that is not supposed to be in scope of current change. If so, a developer/CI will need to run it's own `test GEAR=xxx` checks

.github/workflows/gear-scoped-ci.yml
Artifizer: AI TODO: investigate if it's a valid concern carefully, suggest a fix

.github/workflows/gear-scoped-ci.yml
Artifizer: AI TODO: investigate if it's a valid concern carefully, suggest a fix

.github/workflows/gear-scoped-ci.yml
Artifizer: AI TODO: investigate if it's a valid concern carefully, suggest a fix

Makefile
Artifizer: AI TODO: to confirm it's a valid concern. If yes - propose a robust solution to fix the root cause. It must be fixed in a way that gears do not differ in their naming convention in this project. You will need to investigate what is the better option of doing that - either make a local per-project CI or maybe suggest a new rule in cargo-gears. But I'm not sure cargo-gears because other gears are not necessary coming from `cf` vendor and not necessary to be published. So maybe local make check and CI is OK

Makefile
Artifizer: **Resolved.** Valid concern — confirmed and fixed.

Server-dependent targets (`run`, `openapi`, `.example-server-build`) now validate `$(GEAR)` against `EXAMPLE_SERVER_ALL_FEATURES` (resolved via `cargo gears ls features --manifest`). When GEAR has no matching server feature (e.g. `toolkit-db`, `ledger`), these targets print a SKIP message and exit cleanly instead of failing with a Cargo feature error. Library-safe targets (`fmt`, `clippy`, `test`, `build`) continue to work for all gears.

See commit `8f0ebe8a`.

Makefile
Artifizer: **Resolved.** Valid concern — confirmed and fixed.

The convention-based `GEAR_PKG`/`GEAR_SDK_PKG`/`GEAR_SDK_FLAG` variables have been removed entirely. `GEAR_PKGS` is now resolved via:
```makefile
GEAR_PKGS := $(shell cargo gears ls packages --scope-dirs gears,libs --filter '$(GEAR_NAME_REGEXP)' -f cargo-flags)
```
This discovers packages by regex on Cargo package names from `cargo metadata` — handles `bss-ledger`, `cf-chat-engine`, `users-info`, grouped gears, and libs without any naming convention assumptions.

`resolve_gear_deps.py` has also been removed — fully replaced by `cargo gears ls packages`.

Tracking issue for crate renames: #4591.
cargo-gears enhancements: constructorfabric/cargo-gears#107.

See commit `11e08b07`.

Makefile
Artifizer: **Acknowledged — intentional asymmetry, documented.**

`fmt` and `clippy` now use `GEAR_PKGS` (gear's own packages: main + SDK, resolved via `cargo gears ls packages`). `test` uses the same command with `--include-rdeps` for transitive reverse-dep coverage.

The asymmetry is intentional:
- **fmt**: per-file, no value from reverse deps
- **clippy**: most lints are local; edge-case propagated-lint scenarios are caught by full CI which still runs
- **test**: reverse deps are critical — a breaking change causes test failures in consumers

If gear-scoped CI is later positioned as a full CI replacement, clippy can be extended to use `--include-rdeps` too.

.github/workflows/gear-scoped-ci.yml
Artifizer: **Resolved.** Valid concern — confirmed and fixed.

Added a lightweight `gear-ci` Makefile target that runs only `fmt`, `clippy`, and `test` — no extra tools beyond cargo, rustfmt, clippy, nextest, and cargo-gears. Updated `gear-scoped-ci.yml` to call `make gear-ci GEAR=<gear>` instead of `make all GEAR=<gear>`.

This eliminates the latent failure from 5 missing tools (`cfs`, `lychee`, `cargo-gears`, `gts-validator`, `cargo-deny`). Also simplified `classify_scope()` — removed the now-unnecessary `GEAR_PKG`/`GEAR_SDK_PKG` passthrough.

See commit `a4f8c6e7`.

.github/workflows/gear-scoped-ci.yml
Artifizer: **Resolved.** Valid concern — both points fixed.

1. **sccache**: Added `mozilla-actions/sccache-action` + `RUSTC_WRAPPER` probe, matching `ci.yml` pattern.
2. **Shared cache key**: Changed from `pr-gear-scoped-<scope_id>-<toolchain>` to `pr-gear-scoped-<toolchain>` — all gear-scoped runs now share one cache bucket.

See commit `7b6b04b3`.

.github/workflows/gear-scoped-ci.yml
Artifizer: **Resolved.** Valid concern — fixed.

Replaced two-dot `git diff BASE HEAD` with three-dot semantics via `git merge-base`:

```python
merge_base = subprocess.run(
    ["git", "merge-base", BASE_SHA, HEAD_SHA], ...
).stdout.strip()
diff = subprocess.run(
    ["git", "diff", "--name-only", merge_base, HEAD_SHA], ...
)
```

This gives the true PR diff regardless of how stale the branch is, preventing spurious non-scoped verdicts from unrelated merged commits.

See commit `7b6b04b3`.

PR #4564, opened 2026-08-16
test(rustls-corecrypto): harden openssl s_server spawn against load

Artifizer:
Fixes deterministic failures under `make make-benchmark` scenarios #1 (make all), #3 (make check), #4 (make test), all failing on the same 3 tests in `cf-gears-rustls-corecrypto-provider::handshake_smoke` with 'could not bind openssl s_server on an ephemeral port after 5 attempts'. This is test-harness flakiness under CPU saturation, not a crypto bug (tests pass in isolation, 11/11). Root cause: spawn_s_server waited only 1s (x5 retries) for openssl s_server to bind and connected via 'localhost' (may resolve to IPv6 ::1). Under the full nextest suite the machine is saturated by concurrent compile-fail tests spawning rustc, so openssl startup/RSA keygen exceeded the 1s window. Fix (test-only): bind wait 1s->10s, retries 5->10, bind/connect on explicit 127.0.0.1. Verified: handshake_smoke 11 passed; cargo fmt --check clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenSSL server smoke-test reliability by adding startup retries and longer readiness waits.
  * Standardized local server checks on explicit IPv4 loopback addresses.

### DIFF
--- a/libs/rustls-corecrypto-provider/tests/handshake_smoke.rs
+++ b/libs/rustls-corecrypto-provider/tests/handshake_smoke.rs
@@ -111,8 +111,19 @@ fn spawn_s_server(extra_args: &[&str]) -> (Child, u16, tempfile::TempDir) {
     // Reserve an ephemeral port from the OS, then release it immediately
     // so openssl can bind to the same number. There's a tiny TOCTOU race
     // window, but in practice the OS doesn't recycle ports that fast — and
-    // if a collision happens we retry up to 5 times.
-    for _ in 0..5 {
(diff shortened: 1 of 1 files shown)


Artifizer:
Superseded by #4565, which includes all changes from this PR plus #[serial] serialization of the 4 openssl-dependent tests — the timeout hardening alone wasn't sufficient, the benchmark still failed because 4 concurrent TOCTOU port races overwhelm the pattern under full-workspace load.

PR #4572, opened 2026-08-17
feat(file-parser): Add magika support

### LINKED ISSUE #4539: file-parser: content-based file type detection via Magika (feature-gated)
The file-parser gear decides which parser to use from the filename extension, falling back to a
client-supplied `Content-Type`. Nothing in the gear looks at file content.

Consequences today:

- `parse_local` rejects any file without an extension outright
  (`domain/service.rs:129-132`, error `unsupported_file_type("no extension")`).
- `parse_bytes` rejects uploads that have neither a filename nor a recognized `Content-Type`
  (`domain/service.rs:209-213`). `application/octet-stream` without a filename always fails,
  even though the SDK doc comment for `ParseBytesRequest` already names this case
  (`file-parser-sdk/src/models.rs:5-14`).
- A wrong extension routes to the wrong parser. `.pdf` on a DOCX goes to Kreuzberg with an
  `application/pdf` hint and fails, or worse, produces garbage.
- Three separate hardcoded extension to MIME tables exist and have already drifted:
  `domain/service.rs:14-27` (no xlsx/pptx), `infra/parsers/kreuzberg_parser.rs:44-57`
  (no docx/images), `infra/parsers/image_parser.rs:47-61` (images only).
- `ImageParser` accepts any `Content-Type` starting with `image/` without checking the bytes
  (`infra/parsers/image_parser.rs:69-73`). The gear is consumed by scan-gateway, so a lie in the
  header propagates into content scanning.

DESIGN.md:85 already states that format detection is the gateway's responsibility, not the
plugins'. So the gateway (`FileParserService`) is the right place to fix this.

## Proposal

Add content-based detection using the `magika` crate (https://crates.io/crates/magika,
Apache-2.0, Google). Magika is a small ONNX model that classifies file content into ~200 types.
The Rust API is minimal:

    let magika = magika::Session::new()?;
    let t = magika.identify_content_sync(&bytes)?;   // or identify_file_sync(path)
    t.info().label       // "pdf", "docx", "png", ...
    t.info().mime_type
    t.score()            // confidence

Integration points:

- New detector in `infra/` behind a trait defined in `domain/`, injected into
  `FileParserService` as `Option<Arc<dyn ContentTypeDetector>>` so the gateway compiles and
  behaves exactly as today when the feature is off.
- `parse_bytes`: run detection on the first N KiB. Use the detected type when there is no
  extension and no usable `Content-Type`. When the caller-supplied hint disagrees with the
  detected type above a confidence threshold, log a warning and prefer the detected type
  (exact policy is the ADR's call).
- `parse_local`: run detection when the path has no extension, instead of erroring.
- Map magika labels to the extensions the existing parsers already declare via
  `FileParserBackend::supported_extensions()` (`domain/parser.rs:9-26`). Keep
  `find_parser_by_extension` as the routing mechanism; do not change the plugin trait in the
  first iteration.
- Consolidate the three extension/MIME tables into one place while doing this.

## Feature gate

Magika pulls in `ort` (ONNX Runtime) and `ndarray`, adds ~2.8 MB of embedded model to the crate,
and requires the ONNX Runtime shared library at build or run time. That is too much to force on
every consumer of the gear, so it must be optional.

- Add to the root `[workspace.dependencies]`: `magika = { version = "1.1", optional = true }`.
  Version pinning and whether `ort` needs an explicit entry (it pins an rc: `2.0.0-rc.12`) are
  decisions for the ADR.
- Add a `[features]` block to `gears/file-parser/file-parser/Cargo.toml` - this gear has none
  today:

      [features]
      default = []
      # Content-based file type detection using the Magika ONNX model.
      # Off by default: pulls in ort (ONNX Runtime) + ndarray and ~3 MB of
      # embedded model weights, and needs libonnxruntime at build/run time.
      # Enable when callers cannot be trusted to send correct filenames or
      # Content-Type headers.
      magika = ["dep:magika"]

  Follow the existing convention: `default = []`, a comment above each feature explaining the
  trade-off. See `gears/chat-engine/chat-engine/Cargo.toml` (`k8s` feature) and
  `libs/toolkit-http/Cargo.toml` (`fips`, `otel`) for the house style.
- Behaviour with the feature off must be identical to current behaviour, including error
  messages. No `cfg` leakage into the public SDK types.
- CI: add a build of the gear with `--features magika` so the gated path does not rot. Check
  how the ONNX Runtime library is obtained in CI and in the container image; if
  `ort/download-binaries` is used, that is a network dependency at build time and needs an
  explicit decision.
- `deny.toml`: magika is Apache-2.0 and should pass the allow-list, but the transitive tree
  (`ort`, `ndarray`) must be checked. Any advisory suppression goes in the `ignore` list with a
  justification comment in the same style as the existing kreuzberg entries.

## ADR is required before implementation

The implementer must write and get an ADR merged before writing implementation code. This is a
dependency choice with runtime, packaging, and security consequences, and the gear has no ADRs
yet, so this creates the first one.

- Location: `gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md`
- Template: `docs/spec-templates/gears-sdlc/ADR/template.md` (MADR)
- Checklist to review against: `docs/checklists/ADR.md`
- ID: `cpt-cf-file-parser-adr-content-based-type-detection`

The ADR must at minimum cover:

- Options considered, not just Magika. At least: keep extension-only; magic-byte sniffing with
  a pure-Rust crate such as `infer` or `file-format`; delegate to Kreuzberg's own MIME
  detection; Magika. Include the "do nothing" option.
- Precedence rules between detected type, filename extension, and `Content-Type`, including
  what happens on disagreement and what the confidence threshold is.
- Whether a mismatch is a warning, a metric, or a hard error, and what scan-gateway expects.
- How many bytes are fed to the model and where detection sits relative to the max-file-size
  check.
- Performance: model load cost, whether the `Session` is created once at gear startup or lazily,
  inference latency per call, and whether inference needs to move off the async runtime
  (`spawn_blocking`) - magika's sync API is CPU-bound.
- Binary size, container image size, and how the ONNX Runtime library is provisioned in CI and
  in the runtime image.
- Failure mode when the feature is compiled in but the model or runtime cannot be loaded.
- Whether the SDK surface changes (for example, returning the detected type and confidence in
  the parse result, and what `UnsupportedFileType { extension }` in
  `file-parser-sdk/src/errors.rs:6-10` becomes when there is no extension).
- Mapping from ~200 magika labels to the small set of extensions the gear supports, and what
  happens for a confidently-detected type with no parser.
- FIPS / air-gapped deployment implications, if any.

## Acceptance criteria

- ADR merged before the implementation PR.
- Default build unchanged: no new dependencies in the dependency tree, identical behaviour and
  error messages.
- With `--features magika`: extensionless local files parse; uploads with no filename and
  `application/octet-stream` parse; a file with a wrong extension routes to the correct parser.
- Extension/MIME mapping lives in one place.
- Unit tests for the detector and for gateway precedence, with the feature both on and off.
- CI builds and tests the gear with the feature enabled.
- DESIGN.md updated (the gateway algorithm at DESIGN.md:181-187 changes).

## References

- https://github.com/google/magika
- https://crates.io/crates/magika - 1.1.0, Apache-2.0
- https://docs.rs/magika - `Session`, `Builder`, `identify_file_sync`, `identify_content_sync`
- Model weights are embedded in the crate via `include_bytes!`; the ONNX Runtime itself is not,
  and must be provided by the consumer.

jalankulkija:
## Add Magika content-based file type detection

Adds an optional `magika` feature to `file-parser` that detects file type from content (via ONNX-backed Magika model) instead of relying solely on the extension hint. Confident detections (`score >= 0.90`) override a wrong/missing extension; low-confidence or unmapped detections fall back to today's extension-hint behavior unchanged.

- `domain/detector.rs`: `ContentTypeDetector` trait + `Confidence` newtype (validated `[0.0, 1.0]`, rejects NaN/out-of-range at construction).
- `infra/magika_detector.rs`: `MagikaDetector` — pool of `magika::Session`s (round-robin, avoids serializing concurrent inference on one mutex), each behind its own poison-recovering `Mutex`. Errors, poisoned locks, and `spawn_blocking` panics are logged via `tracing::warn!` and swallowed to `None` rather than surfacing.
- `domain/service.rs`: wires detection into `parse_bytes`/`parse_local` precedence logic.
- `gear.rs`: eager fallible init at gear startup, bounded by a 30s timeout.
- Tests: `detection_precedence_tests.rs` (fake detector, no ORT needed), `magika_detector_tests.rs` (real-model smoke tests, needs `ORT_DYLIB_PATH`), unit test for the error-swallow path via a synthetic `magika::Error`.

## ONNX Runtime init deadlock workaround

`ort` 2.0.0-rc.12's `magika::S instead of erroring** when theloaded ONNX Runtime dylib is missing/incompatible — a reentrant `OnceLock` inside `ort::api()` blocks on itself.ing.

Mitigation in `init_magika_det
- Session init runs on a raw detached `std::thread::spawn`, **not** `tokio::task::spawn_blocking` awn_blocking` thread would hangthe whole Tokio runtime's shutdown; a raw thread just leaks silently.
- The async side awaits only tr a 30s `tokio::time::timeout`.
- Safe to abandon the thread because init failure here is fatal to the whole host
process — process exit reclaim, at most, the sub-second window before exit. **Caveat**: this safety only holds if the caller treats failure as fatal;
a future hot-reload/non-fatal de would leak for the process'sfull remaining lifetime.
- `TODO` left in-code to drop a`/`ort` ship a non-deadlockinginit path (watch `ort` upstream for a fix above `2.0.0-rc.12`).                      - The only automated test for s.rs`) was `#[ignore]`d andnever ran in CI; deleted in favor of the detailed in-code comment — a synthetic test would exercise the timeout plu` deadlock it guards against.

## Summary by CodeRabbit

* **New Features**
  * Added optional content-based file type detection for uploaded and local files.
  * High-confidence detection can correct missing or inaccurate filename and MIME hints.
  * Added controls to enable or skip detection, configure confidence thresholds, and tune runtime behavior.
  * Improved case-insensitive extension and MIME recognition.
  * Added bounded local-file reads to enforce size limits.

* **Documentation**
  * Documented detection behavior, fallback rules, configuration, and supported formats.

* **Tests**
  * Added coverage for routing precedence, fallback behavior, runtime initialization, cancellations, and timeouts.

### DIFF
--- a/gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
+++ b/gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
@@ -1,0 +1,151 @@
+---
+status: proposed
+date: 2026-08-13
+decision-makers: file-parser gear maintainers
+---
+
+# Add feature-gated, Magika-based content sniffing to resolve file type when filename/Content-Type are absent or wrong
+
+**ID**: `cpt-cf-file-parser-adr-content-based-type-detection`
+
+<!-- toc -->
+
+- [Context and Problem Statement](#context-and-problem-statement)
+- [Decision Drivers](#decision-drivers)
+- [Considered Options](#considered-options)
+- [Decision Outcome](#decision-outcome)
+  - [Consequences](#consequences)
+  - [Confirmation](#confirmation)
+- [Pros and Cons of the Options](#pros-and-cons-of-the-options)
+  - [Do nothing — keep extension/`Content-Type`-only detection](#do-nothing--keep-extensioncontent-type-only-detection)
+  - [Pure-Rust magic-byte sniffing (`infer` or `file-format`)](#pure-rust-magic-byte-sniffing-infer-or-file-format)
+  - [Delegate to Kreuzberg's own MIME detection](#delegate-to-kreuzbergs-own-mime-detection)
+  - [Magika (ONNX ML content classifier), feature-gated behind `magika`](#magika-onnx-ml-content-classifier-feature-gated-behind-magika)
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+## Context and Problem Statement
+
+`FileParserService` (`domain/service.rs`) currently routes every request to a parser backend using only the filename extension, falling back to a client-supplied `Content-Type` header. It never inspects file bytes. This means `parse_local` hard-rejects extensionless files, `parse_bytes` rejects uploads that have neither a filename nor a recognized `Content-Type` (a case the SDK's own `ParseBytesRequest` doc comment already anticipates), a wrong extension silently routes to the wrong parser, and `ImageParser` trusts an `image/*` `Content-Type` header without checking the bytes — a trust boundary problem, since a caller-supplied header is never a substitute for verifying what the bytes actually are. `DESIGN.md:181-187` already assigns format detection to the gateway, not the plugins, so the gateway is where this must be fixed. How should the gateway determine file type when the filename/`Content-Type` hints are missing or wrong, without imposing a new mandatory runtime dependency on every consumer of the gear? Note that this ADR only *mitigates* the `ImageParser` trust-boundary problem rather than eliminating it — see the "Consequences" section below.
+
+## Decision Drivers
+
+* Callers cannot always be trusted to supply a correct filename or `Content-Type`
+* The three existing extension→MIME tables have already drifted and must collapse to one
+* Default build behavior and error messages must stay byte-for-byte identical for gear consumers who never opt in, with one intentional, narrowly-scoped exception from collapsing the three drifted tables into one canonical table (see "Consequences"): feature-off `Content-Type`-only routing now also resolves `xlsx`/`xls`/`xlsm`/`xlsb`/`pptx`. The canonical table is used only to fill in a *missing* `Content-Type` — an explicitly supplied `Content-Type` always passes through to the backend unchanged, regardless of whether the resolved extension has a canonical-table entry (an explicit header is a stronger signal than a filename extension, so the table must never silently override it). Both halves are pinned by tests in `detection_precedence_tests.rs`.
+* Callers that already know the exact type they're passing (in particular, in-process callers going through `ClientHub`/the SDK, e.g. `FileParserLocalClient::parse_bytes`) can set `ParseBytesRequest::detection` to `Detection::Skip` to bypass content-based detection entirely and route purely off `filename`/`content_type`, even when a detector is registered — detection there buys nothing but latency and a chance of misrouting a type the caller is certain of. The public REST upload endpoints do not expose this — every request through them is always detected, per the "callers are not trusted" driver above.
+* Detection must add bounded, predictable latency — parsing runs in a request path, not a batch job
+* Whatever mechanism is chosen must resolve the actual formats this gear supports today (PDF, HTML, DOCX, XLSX, PPTX, PNG/JPEG/WEBP/GIF, incl. OOXML formats that share a ZIP container)
+* No new dependency, binary size, or native-library requirement may land on the default build path
+* The gear has no ADRs yet; this sets the precedent for how model/runtime dependencies get introduced
+
+## Considered Options
+
+* Do nothing — keep extension/`Content-Type`-only detection
+* Pure-Rust magic-byte sniffing (e.g. `infer` or `file-format` crates)
+* Delegate to Kreuzberg's own MIME/format detection
+* Magika (ONNX ML content classifier), feature-gated behind `magika`
+
+## Decision Outcome
+
+Chosen option: "Magika (ONNX ML content classifier), feature-gated behind `magika`", because it is the only option that reliably distinguishes the OOXML-family formats this gear must support (DOCX/XLSX/PPTX are all ZIP containers that magic-byte sniffing cannot tell apart without deep-parsing the archive), and feature-gating confines its runtime/binary cost to deployments that actually need it, per `DESIGN.md:85`'s assignment of detection to the gateway.
+
+### Consequences
+
+* `infra/` gains a `MagikaDetector` implementing a new `domain`-defined `ContentTypeDetector` trait, injected into `FileParserService` as `Option<Arc<dyn ContentTypeDetector>>`; with the `magika` feature off, this is always `None` and every code path is identical to today's behavior, including error message text.
+* `EXTENSION_MIME_MAPPINGS` (`domain/service.rs:14-27`), the Kreuzberg-parser table (`infra/parsers/kreuzberg_parser.rs:44-57`), and the image-parser table (`infra/parsers/image_parser.rs:47-61`) collapse into one canonical extension↔MIME table in `domain/`; all three call sites (and the new Magika label→extension map) read from it. This is the union of the three previous tables, so it also changes feature-off routing: a `Content-Type`-only request (no filename) for `xlsx`, `xls`, `xlsm`, `xlsb`, or `pptx` now resolves to a MIME type and routes to `KreuzbergParser` even when the `magika` feature is not compiled in, whereas previously only `pdf`/`html`/`htm` resolved this way at the gateway. This is intentional: these formats were already routable via filename extension and via Kreuzberg's own table; the canonical table just makes `Content-Type`-only requests consistent with that.
+* Detection precedence becomes: (1) if the `magika` feature is compiled in and the detector loaded successfully, run content detection on every request, not only when the extension/`Content-Type` are missing; (2) if the detected label maps to a supported extension with confidence ≥ 0.90, that extension wins routing — this is what lets a wrong extension resolve to the correct parser; (3) below 0.90 confidence, or when the label maps to no supported extension, fall back to the extension/`Content-Type`-derived value exactly as today; (4) with the feature off, or with it on but the extension/`Content-Type` agreeing with a confident detection, behavior and routing are unchanged from today.
+* This ADR mitigates, but does not eliminate, the `ImageParser` trust-boundary problem: a caller-supplied `image/*` `Content-Type` hint still selects `ImageParser` without byte validation whenever Magika is not compiled in, doesn't run, or scores below 0.90 confidence on the actual bytes. Fully closing this gap (e.g. mandatory content validation before `ImageParser` selection regardless of detector confidence) is deferred to a future ADR.
+* Every disagreement between the detected type and the caller-supplied extension/`Content-Type` at ≥ 0.90 confidence is logged at `WARN` with both values, so operators can see when clients are sending an incorrect type hint without the gateway hard-failing the request. A metric on this signal is deferred to a future iteration.
+* `parse_bytes` runs detection on the full in-memory buffer after the existing `max_file_size_bytes` check, never before — the size check is cheap and must reject oversized uploads before any CPU is spent on inference.
+* `parse_local` runs detection whenever a detector is registered — not only when the extension is absent — mirroring `parse_bytes`'s precedence, so a present-but-wrong extension is corrected the same way for local files. Detection does **not** buffer the file: it goes through `ContentTypeDetector::detect_path`, backed by `magika::Session::identify_file_sync`, which seeks to the head/tail blocks the model needs instead of reading the whole file (and avoids duplicating the read the backend then performs itself).
+* The `max_file_size_bytes` limit is enforced in `parse_local` unconditionally, before a backend is selected and regardless of whether a detector is registered — a feature flag must not change the size-limit contract. That check reads `metadata().len()`, so it is not atomic against a file growing afterwards; it exists to reject oversized files early with a message naming the actual size. The actual memory bound is each backend's own bounded read (`infra::parsers::bounded_read::read_bounded`, a `Read::take` against a single open handle). `KreuzbergParser` is the exception: it hands the path to `kreuzberg`, which reads the file itself, so only the metadata check applies there.
+* Because `ort`'s inference call is synchronous and CPU-bound, `MagikaDetector` executes it inside `spawn_blocking`; the `magika::Session` is constructed once, eagerly, at gear startup (not lazily on first request), so cold-load latency and any missing-runtime/model failure surface at boot, not on a user request.
+* With the `magika` feature compiled in, failure to load the ONNX Runtime shared library or the embedded model is a **startup failure**, not a silent fallback to extension-only routing — a gear that silently disabled detection would reintroduce exactly the spoofable-`Content-Type` risk this ADR exists to close. Operators who need best-effort behavior must build without the feature.
+* `ort` 2.0.0-rc.12 does not fail cleanly on its own: when the runtime it loads is missing, the wrong architecture, or older than the ONNX Runtime C API level (`api-24`, i.e. the API surface shipped starting with ONNX Runtime 1.24.x — CI provisions 1.29.0) the `ort` build was compiled against, the call **hangs forever instead of returning an error**. Reproduced directly against a nonexistent `ORT_DYLIB_PATH`: `magika::Session::new()` returns neither `Ok` nor `Err`, and does not panic, within 45s. We deliberately do *not* claim a mechanism: an earlier version of this ADR attributed it to `ort`'s error-reporting path recursively re-entering the `OnceLock` behind `ort::api()`, but that explanation is contradicted by `ort::init_from(path)` — which never touches that lock — hanging on the same input, also verified at 45s. Treat it as an observed property of an unloadable `ORT_DYLIB_PATH`.
+* There is consequently **no pre-flight validation available**. `ort::init_from(path)` is the obvious candidate (it takes a dylib path and returns `Result`) and does not work; every `ort` entry point funnels through the same lazy dylib initialization. So the timeout-and-abandon workaround cannot be replaced by a cheap "check the runtime first" call, and anyone attempting that simplification should re-run the two 45s probes before concluding otherwise. No upstream issue has been filed as of 2026-08-18; the `init_from` reproduction is the useful thing to report against `pykeio/ort`.
+* `MagikaDetector`'s constructor (`with_config`) is therefore never called directly from `gear.rs`; it runs on a detached `std::thread` (not `tokio::task::spawn_blocking`, whose threads Tokio joins on runtime shutdown — a wedged one would hang shutdown too), with the result awaited through a `oneshot` channel under a 30s `tokio::time::timeout`. A broken runtime surfaces as a startup error after 30s instead of an indefinite hang, at the cost of leaking one OS thread for the life of the process in that failure case. That leak is only bounded because gear-init failure terminates the host; `MagikaInitError::leaked_init_thread()` encodes which failures carry the obligation, so a future non-fatal caller has to confront it in code rather than in a comment.
+* Exclusive access per inference is forced by `magika`'s API, not by ONNX Runtime. `ort::session::Session` is `Send + Sync` and ONNX Runtime is designed for one session shared across threads (the `&mut self` on `ort`'s `run` is a borrow-lifetime device for the returned outputs), but `magika::Session` keeps its inner session private and exposes only `&mut self` methods, so none of that is reachable. A mutex is unavoidable while going through `magika`; if detection throughput ever matters, the fix is upstream, not more sessions. The detector therefore holds **one** session, serialized behind a fair FIFO mutex, and exposes ONNX Runtime's intra-op thread count (`magika_intra_op_threads`) as the CPU knob instead of a session count. Measured: ~3.5 ms per detection and ~6 extra OS threads at ONNX Runtime's default, versus ~9.3 ms pinned to a single thread — so one session sustains ~285 detections/second against document extraction costing tens to hundreds of milliseconds on the same request.
+* `magika`'s async API (`identify_content_async` / `identify_file_async`) must **not** be used, despite being the obvious way to avoid `spawn_blocking`. `ort`'s `InferenceFut::drop` calls `RunOptions::terminate()`, which sets a sticky flag that only `unterminate()` clears, and `magika` passes a single process-wide `static OnceLock<RunOptions>` to every async run without ever unterminating. The first dropped async inference future would therefore disable async inference for the entire remaining process lifetime — in a request path, one client disconnect would silently kill detection. This is an upstream `magika` bug; until it is fixed, the `_sync` methods on `spawn_blocking` are mandatory rather than merely conventional.
+* `magika` 1.1.0's own manifest already pins `ort = "=2.0.0-rc.12"` with `default-features = false` and only `features = ["ndarray", "std"]` — `ort`'s `download-binaries` default is already off for every consumer of `magika`, with no action needed on our side to keep the network fetch out of `cargo build`. That resolved dependency set enables **neither** `download-binaries` **nor** `load-dynamic`, so by default `ort` falls back to build-time static linking against a system ONNX Runtime install, located via `ORT_LIB_LOCATION` (or `pkg-config`) — meaning the ONNX Runtime dev headers/libs would need to be present in *every* environment that compiles `file-parser` with `--features magika`, not just wherever it eventually runs.
+* We add an explicit `[workspace.dependencies]` entry, `ort = { version = "=2.0.0-rc.12", default-features = false, features = ["load-dynamic", "api-24"] }`, layered on top of Magika's own feature set — `ndarray` and `std` are not declared directly here because `magika` already contributes them transitively, while `api-24` **must** be declared (without it `ort`'s unconditionally-compiled `ep/vitis.rs` references an `OrtApi` field that only exists at that ABI level). Copy the entry as written; dropping `api-24` breaks the build. `load-dynamic` makes `ort` `dlopen` the ONNX Runtime shared library at process start (path supplied via the `ORT_DYLIB_PATH` env var, or a well-known default search path) instead of linking against it at compile time. This decouples "who builds the gear" from "who has the ONNX Runtime installed": CI here runs on plain GitHub-hosted `ubuntu-latest`/macOS runners with no per-gear container or custom base image (`.github/workflows/ci.yml`), so requiring onnxruntime dev headers at *build* time on every such runner is a heavier, more fragile ask than requiring the shared library only where the `magika`-enabled binary actually runs.
+* `ort/download-binaries` was considered and rejected for CI, not just for production: it always links **statically** (verified by reading `ort-sys`'s build script), which is incompatible with `load-dynamic` — already active for every `magika` build via the workspace `ort` entry — and reliably reproduces the deadlock above (a `dlopen` attempt against a static `.a` fails, triggering the same recursive-`OnceLock` hang). This isn't a macOS-only artifact of local testing; `ort-sys` links statically via `download-binaries` on every platform, so the originally-proposed CI job would have hung on Linux runners too.
+* CI instead downloads the official ONNX Runtime shared-library release directly (`onnxruntime-linux-x64-<version>.tgz` from `microsoft/onnxruntime`'s GitHub releases, pinned to 1.29.0, which satisfies the `api-24` C API level `ort` requires) and points `ORT_DYLIB_PATH` at the extracted `.so` — the same `load-dynamic` mechanism production uses, just with CI choosing to fetch the library itself rather than assuming it's preinstalled. This was verified end-to-end locally (macOS build, official `onnxruntime-osx-arm64` release): real Magika inference against a PDF byte stream succeeds with high confidence. The `cargo test` step also carries a `timeout-minutes` job-level backstop, since a future regression reintroducing the hang above would otherwise consume the runner indefinitely. Provisioning the shared library for actual deployments (which pinned release, fetched/installed how, on which runtime images) remains a separate, deploy-time decision out of scope for this ADR.
+* `deny.toml` needs no new `ignore` entries for Magika itself (Apache-2.0, already allow-listed via the existing Kreuzberg precedent), but `cargo deny check` must be run with `--features magika` in CI to catch any advisory or license issue in the `ort`/`ndarray` transitive tree; any suppression found gets a justification comment in the same style as the existing `kreuzberg` entries in `deny.toml:41-43`.
+* `file-parser-sdk`'s public types are unaffected, and so is rendered error text: an unsupported detected label never reaches routing at all. `MagikaDetector` intersects Magika's labels with the registered `supported_extensions()` and yields `None` for anything outside that set, and `FileParserService::reconcile_extension` independently requires a registered parser for the detected extension before letting it win — so a confident `"zip"` detection falls back to the caller's hint rather than producing an error naming `"zip"`. Golden error-message tests are therefore unchanged in **both** feature states.
+Surfacing detection confidence/label in the SDK response (`ParsedText`) is explicitly deferred to a future iteration/ADR — this decision does not change the plugin trait (`FileParserBackend`) or the SDK response shape.
+* Magika's ~200 output labels map through a single new `magika_label -> extension` table to only the extensions this gear's registered plugins already declare via `supported_extensions()`; a confidently-detected label with no entry in that table (e.g. `"zip"`, `"mp4"`) is treated as "detection did not help": it is discarded and routing falls back to the caller's extension / `Content-Type` hint, so such a request behaves exactly as it does today with the feature off, rather than producing a new error class.
+* `DESIGN.md:181-187` (the gateway routing algorithm) must be updated to describe the detector as an optional step 0 ahead of today's steps 1–3.
+
+### Confirmation
+
+* Unit tests for `MagikaDetector` (label→extension mapping, confidence threshold behavior) and for `FileParserService` precedence logic, run twice in CI — once with the `magika` feature off (detector absent, behavior identical to pre-ADR baseline) and once with it on.
+* Integration tests covering the three acceptance scenarios from the tracking issue: an extensionless local file parses; a byte upload with no filename and `Content-Type: application/octet-stream` parses; a file with a wrong extension (e.g. a DOCX saved with a `.pdf` name) routes to the DOCX parser and succeeds.
+* `cargo deny check --features magika` passes in CI with no new unjustified advisories.
+* Code review confirms the default-feature build has no new entries in `cargo tree` and that golden error-message tests (already covering `unsupported_file_type`/`no_parser_available`) pass unmodified **for default-feature (`magika` off) runs**; with the feature on, an unsupported *detected* type populates `extension` with the detected label rather than the literal `"no extension"`, per "Consequences" — that is an expected, feature-on-only change in rendered error text, not a regression.
+* **What would validate the 0.90 threshold, and what does not exist yet.** Nothing in this iteration produces the signal needed to tune it: the per-request `WARN` on a detected/hinted disagreement cannot be aggregated or alerted on, and the `file_parser_type_mismatch_total` counter this ADR originally committed to is explicitly deferred (see "Consequences"). Until a metric lands, the threshold is set by judgement, not evidence, and the honest position is that we do not know its false-override rate in production. What *does* exist is a `file_parser.detect` tracing span around every detection and a `magika.inference` span inside the detector, so detection's latency contribution is attributable in a trace without enabling `debug` logging — spans were chosen over a metric because this gear emits no metrics at all and adding that infrastructure is out of scope here. Revisiting the threshold should begin by adding the mismatch counter.
+
+## Pros and Cons of the Options
+
+### Do nothing — keep extension/`Content-Type`-only detection
+
+Leave `parse_local`/`parse_bytes` as they are; only consolidate the three MIME tables.
+
+* Good, because zero new dependencies, zero new runtime/CI/container work
+* Good, because zero risk of a new failure mode (model/runtime load failure)
+* Neutral, because the table consolidation (already useful on its own) still happens
+* Bad, because it does not fix any of the four concrete failure cases in the tracking issue — extensionless files, `octet-stream`-with-no-filename uploads, wrong-extension misrouting, and unchecked image `Content-Type` all remain unresolved
+* Bad, because the unresolved image-`Content-Type` trust gap leaves this gear returning parsed content for whatever type the caller *claims* the bytes are, not what they actually are — the primary reason this issue was filed
+
+### Pure-Rust magic-byte sniffing (`infer` or `file-format`)
+
+Sniff a small byte prefix against a table of known magic-number signatures; no ML model, no native runtime, pure Rust.
+
+* Good, because it is a pure-Rust dependency: no `ort`, no native shared library, no embedded model weights, trivially usable as a default (non-feature-gated) dependency
+* Good, because inference is a table lookup — sub-microsecond, no `spawn_blocking` needed
+* Good, because it fixes the extensionless-file and `octet-stream`-upload cases for formats with a distinct magic number (PDF, PNG, JPEG, GIF, WEBP)
+* Bad, because DOCX/XLSX/PPTX are all `PK\x03\x04` ZIP containers at the byte level; magic-byte sniffing alone cannot tell them apart, and these three formats are core to this gear's supported set — a signature crate would need bespoke ZIP-central-directory parsing to distinguish them, which is materially the same integration cost as adopting a purpose-built classifier
+* Bad, because plain-text-ish formats (HTML) have no reliable magic number and would still fall back to extension/`Content-Type` heuristics
+* Bad, because it does not close the `ImageParser` trust gap any better than "good enough for images" — it *would* fix that specific case, but leaves the office-document cases (the majority of this gear's traffic per `DESIGN.md`) unresolved
+
+### Delegate to Kreuzberg's own MIME detection
+
+Kreuzberg (already a dependency, MIT/Elastic-2.0-licensed, used for PDF/DOCX/etc.) reportedly does some internal format sniffing; call into that instead of adding a new dependency.
+
+* Good, because no new dependency if Kreuzberg's internal detection is reachable and licensable for this purpose
+* Bad, because Kreuzberg's detection is invoked *after* a parser is already selected — using it to *select* the parser would require restructuring Kreuzberg's own entry points or duplicating its internal logic outside the crate boundary, which is not something this gear controls (Kreuzberg is an upstream Elastic-2.0 dependency, not code we can extend)
+* Bad, because it does nothing for the `ImageParser`/non-Kreuzberg formats (PNG/JPEG/WEBP/GIF), which are handled by a separate, unrelated backend
+* Bad, because it couples this gear's routing logic to an upstream library's internal (and undocumented, for this purpose) behavior rather than a purpose-built, versioned API
+
+### Magika (ONNX ML content classifier), feature-gated behind `magika`
+
+Google's small ONNX model, ~200 content-type labels, wrapped by the `magika` crate (Apache-2.0). Chosen option — added as an optional dependency behind a new `magika` Cargo feature, off by default.
+
+* Good, because it distinguishes OOXML formats (DOCX/XLSX/PPTX) and the image/PDF/HTML formats this gear supports from content alone, addressing every case in the tracking issue including — per the confidence threshold and fallback behavior discussed in "Consequences" — mitigating (not eliminating) the `ImageParser` `Content-Type`-trust gap
+* Good, because feature-gating means the default build — and every consumer who does not opt in — pays zero cost: no `ort`, no `ndarray`, no embedded model bytes, and behavior/error messages identical to before this ADR (with the one exception noted in "Decision Drivers"/"Consequences": the canonical MIME table also widens feature-off `Content-Type`-only routing for `xlsx`/`xls`/`xlsm`/`xlsb`/`pptx`)
+* Good, because the model ships embedded in the crate (`include_bytes!`); no network fetch of weights at build or run time
+* Neutral, because it is the first ONNX/`ort` dependency in this workspace, setting precedent for how such dependencies are vetted, built, and shipped (this ADR is that precedent)
+* Bad, because it requires the ONNX Runtime shared library to be present wherever the `magika`-enabled binary runs (via `ort/load-dynamic`, resolved at process start) — a new ops dependency for the runtime/container image that `infer`/`file-format` would not need
+* Bad, because it adds ~2.8 MB of embedded model weight and pulls in `ort` + `ndarray` for any consumer who does enable the feature
+* Bad, because inference is CPU-bound and synchronous, requiring `spawn_blocking` discipline in the gateway that a pure lookup-table approach would not need
+* Bad, because confidence-threshold and mismatch-handling policy (this ADR's Consequences section) is new decision surface that a magic-byte or do-nothing approach would not introduce
+
+## More Information
+
+The 0.90 confidence threshold and the "log, don't hard-fail" mismatch policy are initial values, not empirically tuned; `TEST-ADR-002`-style validation (real-world mismatch-rate telemetry, were a metric added in a future iteration per "Consequences") may justify revisiting them without requiring a new ADR — see Confirmation. Because the threshold is expected to move, it is a configuration key (`detection_confidence_threshold`, default `0.90`) rather than a constant, so tuning it per deployment does not require a recompile; `Gear::init` rejects values outside `[0.0, 1.0]` at startup rather than clamping a value an operator plainly did not mean.
+
+## Traceability
+
+- **PRD**: [PRD.md](../PRD.md)
+- **DESIGN**: [DESIGN.md](../DESIGN.md)
+
+This decision directly addresses the following requirements or design elements:
+
+* `cpt-cf-file-parser-fr-formats` — content-based detection lets uploads reach the correct backend for a supported format even when the filename/`Content-Type` hint is absent or wrong, which is this requirement's actual intent
+* `cpt-cf-file-parser-fr-local-path-security` — detection for extensionless local paths runs only after the existing `..`-rejection/canonicalization/base-dir checks in `parse_local`; this decision does not relax that ordering
+* `cpt-cf-file-parser-fr-plugin-extensibility` — the `magika_label -> extension` map is validated against each registered plugin's `supported_extensions()`, not a hardcoded parser list, so new plugins are picked up without touching the detector
+* `cpt-cf-file-parser-nfr-response-time` — the eager, once-at-startup `Session` load and `spawn_blocking`-wrapped, per-request inference are chosen specifically to keep steady-state request latency bounded and predictable
+* `cpt-cf-file-parser-component-parser-service` (`fdd-file-parser-component-parser-v1`) — `FileParserService`'s routing algorithm gains an optional detection step; `DESIGN.md:181-187` must be updated to reflect it
+* `cpt-cf-file-parser-component-parser-backend` (`fdd-file-parser-component-backend-v1`) — type *resolution* stays entirely in the gateway, consistent with `DESIGN.md:85`, but the trait does change: `parse_local_path` gains a `resolved_content_type: Option<&str>` parameter so the gateway can hand a backend the type it resolved instead of the backend re-deriving one from a possibly-wrong on-disk extension. Implementations must accept the updated local-path signature; those that route purely by filename may ignore the argument
--- a/gears/file-parser/file-parser/Cargo.toml
+++ b/gears/file-parser/file-parser/Cargo.toml
@@ -18,6 +18,15 @@ name = "file_parser"
 [lints]
 workspace = true
 
+[features]
+default = []
+# Content-based file type detection using the Magika ONNX model. Off by
+# default: pulls in ort (ONNX Runtime) + ndarray and ~3 MB of embedded model
+# weights, and needs libonnxruntime at runtime (dlopen via `ORT_DYLIB_PATH`).
+# Enable when callers cannot be trusted to send correct filenames or
+# Content-Type headers.
+magika = ["dep:magika", "dep:ort", "dep:tokio-util"]
+
 [dependencies]
 # Local dependencies
 file-parser-sdk = { package = "cf-gears-file-parser-sdk", version = "0.1.0", path = "../file-parser-sdk" }
@@ -61,6 +70,11 @@ tempfile = { workspace = true }
 
 # Document parsing libraries
 kreuzberg = { workspace = true }
+
+# Content-based file type detection (optional, behind the `magika` feature)
+magika = { workspace = true, optional = true }
+ort = { workspace = true, optional = true }
+tokio-util = { workspace = true, optional = true }
 
 # MIME type parsing
 mime = { workspace = true }
--- a/gears/file-parser/file-parser/src/config.rs
+++ b/gears/file-parser/file-parser/src/config.rs
@@ -1,3 +1,4 @@
+use std::num::NonZeroUsize;
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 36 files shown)


MikeFalcon77 (commented):
The cargo gating inside the crate is right (feature, `infra` module, `gear.rs` init and both magika test files are all correctly `cfg`-gated, and `make clippy` uses `--all-features` so the gated code is linted). Two things sit above the line-level comments, though:

1. **No binary can enable it.** `apps/cf-gears-example-server` depends on `file_parser` unconditionally with no passthrough feature, so `magika` is reachable only from the new `cargo test -p cf-gears-file-parser --features magika` job.
2. **The resolved type never reaches the parser backend.** The service picks a backend by the detected extension and then hands it the original hints. With the backends `gear.rs` actually registers, the headline case (`parse_bytes(None, None, pdf_bytes)`) returns "Cannot determine MIME type" rather than parsing. Only `PlainTextParser`, `StubParser` and `DocxParser` are unaffected — which is also why the new precedence tests pass: they register exactly the backends that ignore the hint.

So end-to-end the feature does not work yet for pdf/html/xlsx/pptx or any image format. Everything else below is smaller.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -185,31 +216,42 @@ impl FileParserService {
...
+        // wrong hint, not just fill in a missing one. With no detector
+        // registered (including always, when built without the `magika`
+        // feature), this is a no-op and behavior is unchanged.
+        let detected = match &self.detector {
+            Some(detector) => detector.detect(bytes.clone()).await,
+            None => None,
+        };
+
+        let Some(extension) =
+            Self::reconcile_extension(hint_extension.as_deref(), detected.as_ref())
```
MikeFalcon77: **CRITICAL**

The detected extension picks the parser, but line 265 still passes the original `filename_hint` and `content_type` to `parser.parse_bytes(...)`, so the backend never learns about the correction.

This breaks the feature's main use case. Running these scenarios against the backend list `gear.rs` actually registers:
- `parse_bytes(Some("report.doc"), None, PNG bytes)` with detected `png@0.99` -> `ImageParser` returns `UnsupportedFileType { extension: "Unable to determine image MIME type" }`
- same call with detected `pdf@0.99` -> `KreuzbergParser` returns `ParseError { message: "Cannot determine MIME type: no content_type or recognized filename extension provided" }`
- `parse_bytes(Some("notes.txt"), Some("text/plain"), PDF bytes)` -> routes to Kreuzberg and hands it `text/plain` for PDF bytes

Only `PlainTextParser`, `StubParser` and `DocxParser` survive, because they ignore `content_type` and use the filename for metadata only. That covers 5 of the 18 registered extensions.

When `reconcile_extension` returns the detected extension rather than the hint, pass the resolved type down: `parser.parse_bytes(Some(&format!("file.{extension}")), mime_table::mime_for_extension(&extension), bytes)`, or add a resolved-extension argument to `FileParserBackend::parse_bytes`.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -125,16 +138,34 @@ impl FileParserService {
...
 
         // Find parser
         let parser = self
-            .find_parser_by_extension(extension)
-            .ok_or_else(|| DomainError::no_parser_available(extension))?;
+            .find_parser_by_extension(&extension)
+            .ok_or_else(|| DomainError::no_parser_available(&extension))?;
 
         // Parse the file
         let document = parser.parse_local_path(&canonical).await.map_err(|e| {
```
MikeFalcon77: **HIGH**

`parse_local_path` only receives the path, so the reconciled extension is used for routing and then discarded.

`ImageParser::parse_local_path` re-reads `path.extension()` and returns `unsupported_file_type` when it is not png/jpg/jpeg/webp/gif; `KreuzbergParser::parse_local_path` derives its MIME hint the same way. So for local files a confident detection can only mis-route, never fix: verified: `parse_local("report.doc")` holding PNG bytes with detected `png@0.99` routes to `ImageParser` and returns `UnsupportedFileType { extension: "doc" }` — without detection it would have gone to `StubParser` and succeeded.

Add a method to `FileParserBackend` that takes the resolved extension alongside the path (defaulting to the current `parse_local_path`), and call it with the reconciled extension here.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -235,18 +277,66 @@ impl FileParserService {
...
+
+        let mut file = tokio::fs::File::open(path)
+            .await
+            .map_err(|e| DomainError::io_error(format!("Cannot open file for detection: {e}")))?;
+
+        let limit = self.config.max_file_size_bytes as u64;
+        let mut buf = Vec::new();
+        let read = (&mut file)
+            .take(limit + 1)
+            .read_to_end(&mut buf)
```
MikeFalcon77: **HIGH**

Detection reads the whole file into a `Vec<u8>` — up to `max_file_size_bytes`, 100 MB by default — and then the backend reads the same file from disk again.

That is up to 100 MB of heap per concurrent `parse_local` call plus a duplicated read, for a model that only looks at the beginning, middle and end blocks of the file.

`magika` 1.1 exposes `Session::identify_file_sync(&mut self, path)`, which seeks instead of slurping. Add a path-based method to `ContentTypeDetector` (e.g. `detect_path(&self, path: &Path)`) backed by `identify_file_sync`, and drop the read-to-buffer here.

gears/file-parser/file-parser/Cargo.toml
```diff
@@ -18,6 +18,15 @@ name = "file_parser"
...
 workspace = true
 
+[features]
+default = []
+# Content-based file type detection using the Magika ONNX model. Off by
+# default: pulls in ort (ONNX Runtime) + ndarray and ~3 MB of embedded model
+# weights, and needs libonnxruntime at runtime (dlopen via `ORT_DYLIB_PATH`).
+# Enable when callers cannot be trusted to send correct filenames or
+# Content-Type headers.
+magika = ["dep:magika", "dep:ort"]
```
MikeFalcon77: **HIGH**

Nothing can turn this feature on for a real build.

`apps/cf-gears-example-server/Cargo.toml:76` depends on `file_parser` unconditionally and declares no passthrough feature, so `magika` is reachable only via `cargo test -p cf-gears-file-parser --features magika` — the CI job added in this PR. Every other optional gear capability in that manifest has one: `k8s = ["mini-chat/k8s", "chat-engine?/k8s"]`, `fips = ["toolkit/fips", ...]`, `file-storage = ["dep:file_storage"]`.

Add `magika = ["file_parser/magika"]` to `apps/cf-gears-example-server/Cargo.toml` next to the other feature flags.

gears/file-parser/file-parser/tests/detection_precedence_tests.rs
```diff
@@ -0,0 +1,334 @@
...
+
+/// Registers `PlainTextParser` (txt/log/md) and `StubParser` (doc/rtf/...)
+/// so tests can tell which one handled a request from `meta.is_stub`.
+fn service_with_detector(
+    detected: Option<DetectedType>,
+    base_dir: std::path::PathBuf,
+) -> FileParserService {
+    let parsers: Vec<Arc<dyn FileParserBackend>> = vec![
+        Arc::new(PlainTextParser::new()),
+        Arc::new(StubParser::new()),
```
MikeFalcon77, line 122: **HIGH**

These tests register only `PlainTextParser` and `StubParser` — the two backends that ignore `content_type` and never derive a MIME from the filename hint — so every override test passes while the backends `gear.rs` actually registers reject the same requests.

Re-running these exact scenarios with `ImageParser`/`KreuzbergParser` registered fails: detected `png@0.99` over a `report.doc` hint gives `UnsupportedFileType { extension: "Unable to determine image MIME type" }`, detected `pdf@0.99` gives `ParseError { message: "Cannot determine MIME type: ..." }`, and the `parse_local` equivalent gives `UnsupportedFileType { extension: "doc" }`. The `!doc.meta.is_stub` assertions are also only a proxy for "PlainTextParser handled it" and never assert the resolved extension.

Register `ImageParser` and `KreuzbergParser` to match `gear.rs`'s list and assert a successful parse after a confident override, or use a fake backend that asserts on the `filename_hint`/`content_type` it receives.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+    async fn detect(&self, bytes: bytes::Bytes) -> Option<DetectedType> {
+        // Magika's inference is synchronous and CPU-bound; run it on the
+        // blocking pool rather than the async worker thread. `bytes` and the
+        // session pool must be owned by the closure since spawn_blocking
+        // requires `'static`; cloning `Bytes` is a ref-count bump, not a
+        // buffer copy. Round-robin across the pool so concurrent requests
+        // aren't all serialized on a single session's mutex.
+        let sessions = Arc::clone(&self.sessions);
+        let supported_extensions = self.supported_extensions.clone();
+        let index = self.next_session.fetch_add(1, Ordering::Relaxed) % sessions.len();
```
MikeFalcon77: **HIGH**

Every `detect` call spawns a blocking task that then parks on `sessions[index].lock()`, so concurrent requests consume blocking threads with no permit or backpressure, and the round-robin index can pick a busy session while another sits idle.

There is no `max_blocking_threads` override anywhere in the workspace, so this is Tokio's default 512-thread pool — shared with `tokio::fs` in `detect_from_path` and with every other gear in the host. N concurrent uploads pin N blocking threads, each holding its `Bytes` (up to 100 MB), while only `pool_size` make progress. The blind `fetch_add` also defeats the pool's stated purpose in the type doc: a request assigned to a slot running a long inference waits behind it instead of taking a free session.

Hand sessions out over an idle-session queue — move each `magika::Session` into a `tokio::sync::mpsc` channel and take/return one per request — and `.await` the receive before calling `spawn_blocking`, so queuing happens on the async side and at most `pool_size` blocking threads are ever used.

gears/file-parser/file-parser/src/infra/parsers/kreuzberg_parser.rs
```diff
@@ -42,18 +42,7 @@ impl KreuzbergParser {
...
-            "xlsx" => Some("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"),
-            "xls" => Some("application/vnd.ms-excel"),
-            "xlsm" => Some("application/vnd.ms-excel.sheet.macroEnabled.12"),
-            "xlsb" => Some("application/vnd.ms-excel.sheet.binary.macroEnabled.12"),
-            "pptx" => {
-                Some("application/vnd.openxmlformats-officedocument.presentationml.presentation")
-            }
-            _ => None,
-        }
+        crate::domain::mime_table::mime_for_extension(ext)
```
MikeFalcon77, line 58: **HIGH**

Delegating to the canonical table silently widened this function: it now returns `Some` for `txt`, `docx`, `png`, `jpg`, `jpeg`, `webp` and `gif`, none of which `supported_extensions()` declares, where the deleted `match` returned `None`.

`parse_bytes` uses this as the extractor MIME hint when `content_type` is absent, so the new detector path makes a wrong MIME reachable: a confident `pdf` detection routes to this backend, and a `photo.png` filename hint now yields `image/png` instead of `None`, so Kreuzberg is asked to extract PDF bytes as a PNG. Previously that produced an explicit "Cannot determine MIME type" error; now it produces garbage output. `ImageParser::mime_type_from_extension` guards against exactly this with a `SUPPORTED_EXTENSIONS` filter — this one does not.

Filter by `self.supported_extensions()` before delegating, mirroring `ImageParser::mime_type_from_extension`, so `mime_for_ext` keeps returning `None` for formats this backend cannot extract.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -235,18 +277,66 @@ impl FileParserService {
...
+            {
+                warn!(
+                    hinted_extension = hint,
+                    detected_extension = %detected.extension,
+                    confidence = %detected.confidence,
+                    "file-parser: caller-supplied type hint disagrees with detected content \
+                     type; routing by detected type"
+                );
+            }
+            return Some(detected.extension.clone());
```
MikeFalcon77, line 376: **MEDIUM**

A confident detection replaces the hint without checking that a parser exists for the detected extension.

DESIGN.md states the gateway overrides routing "only when its label maps to a registered/supported extension", but that check lives in `MagikaDetector`, not here. `with_detector` is public, so any other implementation that gets the invariant wrong turns a request that would have succeeded via the hint into `no_parser_available`.

Make this a method on `&self` and take the detected extension only when `self.find_parser_by_extension(&detected.extension).is_some()`; otherwise fall back to the hint.

gears/file-parser/file-parser/src/config.rs
```diff
@@ -14,8 +14,20 @@ pub struct FileParserConfig {
...
     /// the path cannot be resolved.
     pub allowed_local_base_dir: PathBuf,
+
+    /// Number of `magika::Session` instances to keep in the content-type
+    /// detector's pool. Each session holds its own ONNX runtime state, so
+    /// this trades memory at startup for inference concurrency: requests are
+    /// round-robined across the pool instead of serializing on one session.
+    /// Only used when the gear is built with the `magika` feature.
+    #[serde(default = "default_magika_session_pool_size")]
+    pub magika_session_pool_size: usize,
```
MikeFalcon77: **MEDIUM**

`magika_session_pool_size` is a bare `usize` with no range check, so the type permits values the code cannot honor.

`0` deserializes fine and is then silently corrected by `pool_size.max(1)` deep inside `MagikaDetector::new`, with no log — an operator who configures `0` gets a one-session pool and no indication it was overridden. At the other end, a large value allocates that many ONNX sessions at startup, each with its own copy of the model, while `MAGIKA_INIT_TIMEOUT` stays a fixed 30s, so a legitimately slow but working init fails gear startup. `Gear::init` validates `allowed_local_base_dir` but not this.

Type the field as `std::num::NonZeroUsize` with a `NonZeroUsize` default so `0` fails deserialization, cap it in `Gear::init` at something like `std::thread::available_parallelism()`, and scale `MAGIKA_INIT_TIMEOUT` with the pool size.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+impl ContentTypeDetector for MagikaDetector {
+    async fn detect(&self, bytes: bytes::Bytes) -> Option<DetectedType> {
+        // Magika's inference is synchronous and CPU-bound; run it on the
+        // blocking pool rather than the async worker thread. `bytes` and the
+        // session pool must be owned by the closure since spawn_blocking
+        // requires `'static`; cloning `Bytes` is a ref-count bump, not a
+        // buffer copy. Round-robin across the pool so concurrent requests
+        // aren't all serialized on a single session's mutex.
+        let sessions = Arc::clone(&self.sessions);
+        let supported_extensions = self.supported_extensions.clone();
```
MikeFalcon77: **MEDIUM**

`self.supported_extensions.clone()` deep-copies the whole `HashSet<String>` on every detect call.

Detection runs on every upload, so this allocates one `String` per registered extension per request for data that never changes after construction.

Store the field as `Arc<HashSet<String>>` and `Arc::clone` it here, the same way `sessions` is handled on the line above.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -235,18 +277,66 @@ impl FileParserService {
...
+        let limit = self.config.max_file_size_bytes as u64;
+        let mut buf = Vec::new();
+        let read = (&mut file)
+            .take(limit + 1)
+            .read_to_end(&mut buf)
+            .await
+            .map_err(|e| DomainError::io_error(format!("Cannot read file for detection: {e}")))?;
+
+        if read as u64 > limit {
+            return Err(DomainError::invalid_request(format!(
```
MikeFalcon77, line 430: **MEDIUM**

This makes `max_file_size_bytes` apply to `parse_local` only when a detector is registered.

Without the `magika` feature `parse_local` has no service-level size limit at all; with it, oversize local files start failing with `invalid_request`, and the message differs from the one `parse_bytes` produces ("File size {n} exceeds maximum of {m} bytes"). A feature flag should not change the size-limit contract.

Move the limit check out of `detect_from_path` into `parse_local` so it runs unconditionally, and reuse `parse_bytes`'s message text.

This path is also untested — every test uses a 10 MB limit with 5-byte payloads, so the guard never trips.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+        let index = self.next_session.fetch_add(1, Ordering::Relaxed) % sessions.len();
+        match tokio::task::spawn_blocking(move || {
+            let mut session = sessions[index].lock().unwrap_or_else(|poisoned| {
```
MikeFalcon77: **MEDIUM**

A poisoned session mutex is recovered with `poisoned.into_inner()` and the `magika::Session` is put straight back into service.

The poison means a panic unwound out of the middle of `identify_content_sync`, so the session's ONNX state is unknown. Recovering it keeps a possibly-corrupt session in the pool permanently: 1 in `pool_size` of all subsequent requests round-robins to that slot and may return wrong detections or re-panic, with only a `warn!` line as the signal.

Store slots as `Mutex<Option<magika::Session>>` and on a poisoned lock drop the old session and rebuild it with `magika::Session::new()`, or retire the slot and return `None` for that request.

gears/file-parser/file-parser/src/gear.rs
```diff
@@ -31,6 +31,84 @@ impl Default for FileParserGear {
...
+    std::thread::spawn(move || {
+        // The receiver may already be gone if we timed out and moved on; a
+        // dropped receiver just means `send` fails harmlessly.
+        drop(tx.send(crate::infra::MagikaDetector::new(
+            supported_extensions,
+            session_pool_size,
+        )));
+    });
+
+    let detector = tokio::time::timeout(MAGIKA_INIT_TIMEOUT, rx)
```
MikeFalcon77: **MEDIUM**

Startup waits up to 30s for the detector and spawns a detached thread without ever consulting `ctx.cancellation_token()`.

`GearCtx::cancellation_token()` exists (`libs/toolkit/src/context.rs:166`), but a shutdown signal arriving during init cannot interrupt this wait, so the host blocks for the full `MAGIKA_INIT_TIMEOUT` before it can start graceful shutdown. The checklist requires new startup/background work to honour the framework's cancellation contract.

Pass `ctx.cancellation_token()` into `init_magika_detector` and race it in a `tokio::select!` against the `timeout(MAGIKA_INIT_TIMEOUT, rx)` arm, returning an early cancellation error when the token fires.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -235,18 +277,66 @@ impl FileParserService {
...
+    /// disagreement is logged. Below the threshold (or with no detection),
+    /// the hint is used unchanged.
+    fn reconcile_extension(hint: Option<&str>, detected: Option<&DetectedType>) -> Option<String> {
+        if let Some(detected) = detected
+            && detected.confidence.get() >= DETECTION_CONFIDENCE_THRESHOLD
+        {
+            if let Some(hint) = hint
+                && !hint.eq_ignore_ascii_case(&detected.extension)
+            {
+                warn!(
```
MikeFalcon77, line 368: **MEDIUM**

The mismatch is logged but the counter the ADR commits to is not implemented.

ADR-0001 line 46 makes the accepted decision "logged at `WARN` with both values **and** increments a `file_parser_type_mismatch_total{detected, hinted}` counter, so operators can see when clients are sending an incorrect type hint without the gateway hard-failing the request". `file_parser_type_mismatch_total` does not exist anywhere in the workspace, and this gear emits no metrics at all. Since the ADR deliberately chose not to hard-fail on a mismatch, that counter is the only aggregatable signal operators would have; a per-request WARN line cannot be alerted on or rate-tracked.

Emit the counter here alongside the `warn!`, with `detected` and `hinted` labels as named in the ADR — or amend the ADR if the metric is intentionally deferred.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -185,31 +216,42 @@ impl FileParserService {
...
+            (None, Some("no extension and no content-type"))
+        };
+
+        // Run content-based detection on every request when a detector is
+        // registered — this is what lets a confident detection correct a
+        // wrong hint, not just fill in a missing one. With no detector
+        // registered (including always, when built without the `magika`
+        // feature), this is a no-op and behavior is unchanged.
+        let detected = match &self.detector {
+            Some(detector) => detector.detect(bytes.clone()).await,
```
MikeFalcon77: **MEDIUM**

The detector invocation records no span, event, or timing.

ONNX inference now runs on every `parse_bytes` and `parse_local` call when the feature is on, but nothing logs whether detection ran, what it returned, or how long it took — the only output on the whole detection path is the disagreement `warn!` in `reconcile_extension`. There is no way to attribute request latency to detection or to confirm from logs that it is working.

Emit a `debug!` with the detected extension, confidence and elapsed duration around this call, and add `#[instrument]` plus a latency metric on `MagikaDetector::detect`.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn inference_error_is_swallowed_to_none() {
```
MikeFalcon77, line 238: **MEDIUM**

`map_inference_result` was factored out specifically so its swallow-to-`None` logic could be unit-tested, but the only test covers the `Err` branch.

The label-to-extension intersection, the `to_lowercase()` normalization, the unmapped-label path and the `NaN`-score path are all untested, so a regression that dropped the `supported_extensions.contains(...)` filter — violating the contract `detector.rs` documents — would still pass this suite. These branches need no ONNX Runtime: `magika::FileType::Inferred(InferredType { content_type: None, inferred_type: ContentType::Pdf, score: 0.97 })` can be constructed directly, and its `.info().extensions` is `["pdf"]`.

Add cases for (a) an extension present in `supported_extensions`, asserting the returned extension and confidence, (b) one absent from it, asserting `None`, and (c) a `NaN` score, asserting `None`.

gears/file-parser/file-parser/tests/magika_timeout_tests.rs
```diff
@@ -0,0 +1,44 @@
...
+    // call when the runtime is incompatible) would hang this test's own
+    // runtime teardown even after the timeout below correctly fires. See the
+    // matching comment in `gear.rs`.
+    let (tx, rx) = tokio::sync::oneshot::channel();
+    std::thread::spawn(move || {
+        drop(tx.send(MagikaDetector::new(Vec::<String>::new(), 1)));
+    });
+
+    let timeout = Duration::from_secs(10);
+    let result = tokio::time::timeout(timeout, rx).await;
```
MikeFalcon77: **MEDIUM**

This test re-implements the mitigation inline instead of calling `init_magika_detector`, so it asserts that `tokio::time::timeout` works rather than that the production startup path is bounded.

The module doc claims it "Proves the startup-timeout mitigation in `gear.rs` actually bounds" the hang, but no code in `gear.rs` is executed. If `init_magika_detector` regressed to `spawn_blocking`, lost its 30s timeout, or called `MagikaDetector::new` directly on the async runtime, this test would still pass — and being `#[ignore]`d it never runs in the new CI job either.

Make the bounded init reachable from tests (`pub(crate)` plus a thin wrapper, or a `#[cfg(test)]` test inside `gear.rs`) and assert that the production function returns an `Err` mentioning the timeout.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+            supported_extensions: supported_extensions
+                .into_iter()
+                .map(|ext| ext.into().to_lowercase())
+                .collect(),
+        })
+    }
+}
+
+/// Turns a raw `identify_content_sync` outcome into the public
+/// [`DetectedType`], logging (and discarding) an `Err`, an unmapped label, or
```
MikeFalcon77: **LOW**

The doc says this logs and discards an unmapped label and an out-of-range confidence; neither is accurate.

The unmapped label is dropped by the `?` on the `find` with no log at all — and that is the routine "detection did not help" outcome, so the most common reason detection stops influencing routing produces zero output. An out-of-range confidence cannot occur either: `Confidence::new` clamps into range and rejects only `NaN`, so the `?` on line 89 fires only for a `NaN` score.

Emit a `debug!` with the Magika label and its candidate extensions on the unmapped-label path, a `warn!` on the `NaN` path, and reword the doc to say `NaN` score instead of out-of-range confidence.

gears/file-parser/file-parser/src/domain/detector.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl Confidence {
+    /// Builds a `Confidence` from `value`, clamping it into `[0.0, 1.0]`.
+    /// Returns `None` if `value` is `NaN`.
+    #[must_use]
+    pub fn new(value: f32) -> Option<Self> {
+        if value.is_nan() {
+            return None;
+        }
+        Some(Self(value.clamp(0.0, 1.0)))
```
MikeFalcon77, line 40: **LOW**

Clamping an out-of-range score upward turns a broken detector into a maximally confident one.

`Confidence::new(5.0)` yields `1.0`, which is above `DETECTION_CONFIDENCE_THRESHOLD`, so a detector with a scoring bug silently gains the power to override a correct caller-supplied extension. The doc justifies clamping by "floating-point drift", but drift is a few ULPs, not an arbitrary value.

Return `None` for values outside `[0.0, 1.0]` by more than a small epsilon, and clamp only within that epsilon.

Neither branch has a test: the only caller in the suite is the `detected()` helper, which always passes in-range values.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    Some(DetectedType {
+        extension: extension.to_lowercase(),
+        confidence: Confidence::new(file_type.score())?,
+    })
+}
+
+#[async_trait]
+impl ContentTypeDetector for MagikaDetector {
+    async fn detect(&self, bytes: bytes::Bytes) -> Option<DetectedType> {
```
MikeFalcon77: **LOW**

`detect` carries no cancel-safety annotation, and dropping its future does not stop the spawned inference.

The Rust guideline for this review requires every async fn that can run inside `select!`/`timeout`/an abortable task to carry a `// cancel-safe:` or `// NOT cancel-safe:` comment, "No exceptions". Here the non-obvious consequence is that a client disconnect drops the request future but releases no detector resources: the `spawn_blocking` inference runs to completion still holding a blocking thread and a pool session.

Add a `// NOT cancel-safe:` comment on `detect` stating that a dropped future leaves the inference running until it finishes, and the same on `FileParserService::detect_from_path`.

gears/file-parser/file-parser/src/domain/mime_table.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[test]
+    fn mime_for_extension_unknown_returns_none() {
+        assert_eq!(mime_for_extension("zip"), None);
+    }
+
+    #[test]
+    fn extension_for_mime_round_trips_for_every_registered_extension() {
+        for (ext, mime) in EXTENSION_MIME_TABLE {
+            let round_tripped = extension_for_mime(mime).expect("mime should resolve");
+            assert_eq!(
```
MikeFalcon77, line 83: **LOW**

This test asserts a property that holds by construction and cannot fail for the current table.

`extension_for_mime(mime)` returns the first entry whose MIME equals `mime`, so `mime_for_extension` on that key necessarily returns the same entry's MIME. It can only break if someone adds a duplicate extension key, and it deliberately sidesteps the property that actually matters — that each extension round-trips to itself — which the `jpg`/`jpeg` and `htm`/`html` aliases would expose.

Assert the aliasing contract instead: for each MIME, assert `extension_for_mime(mime)` equals the canonical extension you expect, spelled out per entry.

jalankulkija: Resolved — the round-trip assertion is replaced by `extension_for_mime_resolves_aliases_to_their_canonical_extension`, which spells out the expected canonical extension per MIME type, so the `jpg`/`jpeg` and `htm`/`html` aliasing is now pinned rather than sidestepped.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — but worth splitting: the `parse_bytes` half was already handled at the reviewed commit. `service.rs` computes `let resolved_content_type = mime_table::mime_for_extension(&extension).or(content_type)` and passes *that* to `parser.parse_bytes`, so the scenarios listed here (detected `png` over a `report.doc` hint reaching `ImageParser`, detected `pdf` reaching `KreuzbergParser`) already resolved correctly.

The `parse_local` half was real and is now fixed — see the separate thread on `parse_local_path`. `FileParserBackend::parse_local_path` takes a `resolved_content_type: Option<&str>` and the service threads the resolved MIME through it.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — `FileParserBackend::parse_local_path` now takes `resolved_content_type: Option<&str>`, and `parse_local` passes `mime_table::mime_for_extension(&extension)` so the reconciled type reaches the backend instead of being discarded after routing. `ImageParser` and `KreuzbergParser` consume it; the hint-agnostic backends ignore it.

Covered by `detected_type_is_passed_to_the_backend_for_local_files_too`, which registers the real `ImageParser`, writes PNG content to `report.doc`, and asserts `content_type == image/png`.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — `ContentTypeDetector` gained `detect_path`, backed by `magika::Session::identify_file_sync`, and `parse_local` uses it. The read-to-`Vec` and the duplicated read are both gone.

The size limit that used to be enforced by that bounded read is now a `metadata()` check in `parse_local` (unconditional), plus real bounded reads in the backends — see the separate `max_file_size_bytes` thread.

gears/file-parser/file-parser/Cargo.toml
jalankulkija: Not reproducible at this commit — the passthrough is already present. `apps/cf-gears-example-server/Cargo.toml` carries `magika = ["file_parser/magika"]` alongside the other feature flags, added by this PR's own commit. Resolving as already-satisfied rather than as changed.

gears/file-parser/file-parser/tests/detection_precedence_tests.rs
jalankulkija: Resolved, and this was the sharpest comment in the review — the old tests genuinely could not have caught the `parse_local` bug.

Added: `detected_html_overrides_a_wrong_extension_and_still_parses_via_kreuzberg` (real `KreuzbergParser`, asserts a successful extraction and `text/html`), `detected_type_is_passed_to_the_backend_for_local_files_too` (real `ImageParser`), and a `RecordingParser` fake that captures the `filename_hint`/`content_type` it was handed so tests assert on what the backend actually received rather than inferring it from output.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved, though not via the idle-session queue suggested here — that shape was implemented first and turned out to have its own defect, so the pool is gone entirely.

The queue version moved the session *out* of the pool into the `spawn_blocking` closure and returned it by hand after the `.await`. A dropped caller future skipped that return and destroyed the session without decrementing the live count; enough cancellations drained the pool and later requests parked forever on `recv()`, which never yields `None` because the struct owns the sender for life.

Now: one `Arc<Mutex<Option<Session>>>`. `lock_owned()` and the *guard* moves into the closure, so the session never leaves the mutex and release happens in `Drop` — cancellation cannot strand it. Requests queue on tokio's fair FIFO mutex, so at most one inference is in flight and at most one blocking thread is used, which is what this comment asked for. Regression test `detector_survives_cancelled_detections`; verified non-vacuous by mutation (adding `mem::forget(guard)` makes it fail while the pre-existing smoke tests still pass).

Measurement behind dropping the pool rather than fixing it: each session carries its own ONNX intra-op thread pool (~6 threads on a 10-core host) and its own model copy, while detection measures ~3.5 ms against document extraction costing tens to hundreds of ms on the same request.

gears/file-parser/file-parser/src/infra/parsers/kreuzberg_parser.rs
jalankulkija: Resolved — `SUPPORTED_EXTENSIONS` is hoisted to a module const and `mime_for_ext` filters against it before delegating to the shared table, mirroring `ImageParser::mime_type_from_extension`. `photo.png` yields `None` again rather than `image/png`.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — `reconcile_extension` is now a `&self` method and additionally requires `self.find_parser_by_extension(&detected.extension).is_some()` before a detection wins routing, so the invariant lives in the gateway rather than only inside `MagikaDetector`. `DESIGN.md` was updated to say where the check lives and why it matters given `with_detector` is public API.

gears/file-parser/file-parser/src/config.rs
jalankulkija: Resolved by removing the setting — `magika_session_pool_size` no longer exists. The detector holds a single session, so there is no pool size to validate.

One note for the record: the `NonZeroUsize` half was applied first, but capping at `available_parallelism()` and erroring turned out to be actively harmful. The default was `4`, and `available_parallelism()` honours cgroup limits, so the gear failed to start with its own default config on any host with fewer than 4 usable cores — an ordinary 2-vCPU pod.

Replaced by `magika_intra_op_threads: Option<NonZeroUsize>`, which is the setting that actually governs detection's CPU use. `None` keeps ONNX Runtime's default (~3.5 ms/detection); lowering it costs measured latency (~5.6 ms at 2, ~9.3 ms at 1), and that trade-off is documented on the field.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved — the field is `Arc<HashSet<String>>` and the detect path does `Arc::clone`, so no per-request `String` allocations.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — the check moved into `parse_local` unconditionally, so the limit no longer depends on whether a detector is registered, and both entry points now share `FileParserService::check_size_limit` so the threshold and message cannot drift.

One addition beyond the suggestion: a `metadata()` check is not atomic against the file growing afterwards, which is exactly why the previous code used a bounded read. So the metadata check stays for the early, precise error, and the actual memory bound is a new shared `read_bounded` helper (`Read::take` on a single open handle) used by `PlainTextParser`, `StubParser` and `ImageParser`. `KreuzbergParser` hands the path to `kreuzberg`, which reads the file itself, so only the metadata check applies there — noted in the ADR rather than left implied.

The untested-path point is covered by `local_file_over_the_size_limit_is_rejected_even_without_a_detector`, plus unit tests on `read_bounded` for at-limit, over-limit and missing-file.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved — there is no poisoning to recover from any more. The session lives in a `tokio::sync::Mutex<Option<Session>>`, which does not poison, and the blocking closure `take()`s the session before inference. A panic therefore leaves the slot empty rather than returning a session with unknown ONNX state, and the next caller rebuilds it. That is the "drop the old session and rebuild" behaviour suggested here, reached by construction rather than by handling a poison error.

gears/file-parser/file-parser/src/gear.rs
jalankulkija: Resolved — `init_magika_detector` takes `&CancellationToken` and races it against the bounded wait in a `tokio::select!`, returning `MagikaInitError::Cancelled` when it fires first.

Also now tested, which it previously was not: `cancellation_during_init_returns_early` passes an already-cancelled token and asserts the early return. That arm needs no ONNX Runtime, so unlike the timeout test it actually runs in CI.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved by taking the second option this comment offered — the design doc is amended rather than the metric added. The ADR now states that a metric on the mismatch signal is deferred, so the doc and code agree; `file_parser_type_mismatch_total` no longer appears anywhere as a commitment.

The Confirmation section additionally states plainly what would validate the 0.90 threshold and that it does not exist yet: the per-request `WARN` cannot be aggregated, so the threshold is currently set by judgement rather than evidence, and revisiting it starts with adding that counter. Related: the threshold is now configurable via `detection_confidence_threshold` so tuning it does not need a recompile.

gears/file-parser/file-parser/src/domain/service.rs
jalankulkija: Resolved — `#[instrument(name = "file_parser.detect")]` on the shared detection helper and `#[instrument(name = "magika.inference", fields(intra_op_threads))]` on the detector's inference path, plus a `debug!` carrying the detected extension, confidence, elapsed ms, and a `source` field distinguishing the byte-buffer path from the local-path one.

Spans rather than a metric, deliberately: `debug` is normally off in production, and this gear emits no metrics at all, so standing that infrastructure up was out of scope here. The ADR records that choice.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved — four tests added using a directly-constructed `magika::FileType::Inferred`, exactly as suggested, so none of them need ONNX Runtime: extension present in `supported_extensions` (asserts the returned extension and confidence), extension absent (asserts `None`), `NaN` score (asserts `None`), and lowercase normalisation.

gears/file-parser/file-parser/tests/magika_timeout_tests.rs
jalankulkija: Resolved — the test now calls `init_magika_detector` itself and asserts the error message, so a regression in the production timeout wrapping is actually caught. Reaching it from an integration test required `#[doc(hidden)] pub` on the function, documented as not being public API.

One part stands open by design and is now stated rather than left implied: the timeout arm is still `#[ignore]`d, because the `test-magika` job provisions a *compatible* runtime and so cannot trigger the hang. The module docs say so explicitly and note that fixing it needs a CI fixture pinning a deliberately old runtime. The cancellation arm does run in CI.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved — `debug!` with the Magika label and its candidate extensions on the unmapped-label path, `warn!` on the `NaN` path, and the doc reworded to say `NaN` score rather than out-of-range confidence.

gears/file-parser/file-parser/src/domain/detector.rs
jalankulkija: Resolved — `Confidence::new` now clamps only within `CLAMP_EPSILON` (1e-4) of the range and rejects anything further out, so `Confidence::new(5.0)` is `None` and a detector with a scoring bug cannot acquire override authority. `CLAMP_EPSILON` is `pub` since it is the whole difference between a clamped score and a rejected one.

Both branches are now tested: in-range pass-through, tiny drift clamped, far-out-of-range rejected, and `NaN` rejected.

gears/file-parser/file-parser/src/infra/magika_detector.rs
jalankulkija: Resolved — annotations added on both `detect` and `detect_path`, and on `FileParserService::detect_from_path`.

The hazard itself is also gone rather than just documented. Dropping the future still does not stop the in-flight inference, but it can no longer release nothing: the mutex guard is owned by the blocking task and released in `Drop`, so the session returns for the next caller regardless of what the caller does. The annotations say that, instead of the weaker "result discarded".

gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
```diff
@@ -0,0 +1,143 @@
...
+* `deny.toml` needs no new `ignore` entries for Magika itself (Apache-2.0, already allow-listed via the existing Kreuzberg precedent), but `cargo deny check` must be run with `--features magika` in CI to catch any advisory or license issue in the `ort`/`ndarray` transitive tree; any suppression found gets a justification comment in the same style as the existing `kreuzberg` entries in `deny.toml:41-43`.
+* `file-parser-sdk`'s public types are unaffected: `FileParserClientError::UnsupportedFileType { extension }` keeps its current shape; when detection is enabled and no parser matches a confidently-detected type, `extension` is populated with the detected label (e.g. `"zip"`) instead of the literal string `"no extension"`, but the error variant and its `#[error(...)]` message format do not change. Surfacing detection confidence/label in the SDK response (`ParsedText`) is explicitly deferred to a future iteration/ADR — this decision does not change the plugin trait (`FileParserBackend`) or the SDK response shape.
```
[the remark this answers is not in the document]
jalankulkija: Resolved — the Confirmation section already distinguishes the two feature states: golden error-message tests pass unmodified **for default-feature (`magika` off) runs**, while with the feature on an unsupported *detected* type populates `extension` with the detected label rather than the literal `"no extension"`, which is an expected feature-on-only change in rendered error text rather than a regression.

The feature-off contract has since been tightened further and pinned by tests — see the widened Consequences carve-out covering canonical-MIME precedence over a conflicting caller `Content-Type`.

gears/file-parser/file-parser/src/domain/service.rs
```diff
@@ -235,18 +279,66 @@ impl FileParserService {
...
+            .take(limit + 1)
+            .read_to_end(&mut buf)
+            .await
+            .map_err(|e| DomainError::io_error(format!("Cannot read file for detection: {e}")))?;
+
+        if read as u64 > limit {
+            return Err(DomainError::invalid_request(format!(
+                "File size exceeds maximum of {limit} bytes"
+            )));
+        }
```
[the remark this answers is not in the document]
jalankulkija: Resolved — good catch, and it now applies to different code. The `limit + 1` bounded read in `detect_from_path` is gone entirely (detection uses `ContentTypeDetector::detect_path`, which seeks rather than buffering).

The equivalent bound now lives in the new shared `read_bounded` helper, and it uses `max_bytes.saturating_add(1)` — so the overflow path this comment identified cannot occur there.

gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
```diff
@@ -0,0 +1,149 @@
...
+decision-makers: file-parser gear maintainers
+---
+
+# Add feature-gated, Magika-based content sniffing to resolve file type when filename/Content-Type are absent or wrong
+
+**ID**: `cpt-cf-file-parser-adr-content-based-type-detection`
+
+## Table of Contents
+
+<!-- generated by `cfs toc` -->
```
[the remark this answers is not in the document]
jalankulkija: Resolved (pushed shortly) — both halves addressed, and the diagnosis differs slightly from the report.

**The DESIGN reference** was the real blocker: `cfs validate` reported `Errors: 1`, this one, which is what made the job exit 2. `DESIGN.md` §1.2 Architecture Drivers now carries an `#### Architecture Decisions` subsection referencing `cpt-cf-file-parser-adr-content-based-type-detection` by id, in the `design-arch-overview-drivers` section CFS validates.

**The stale TOC** was a warning, not the error, and the TOC *contents* were already correct — every heading present, in order, with matching anchors. The actual problem was the markers: this ADR used `## Table of Contents` plus a bare `<!-- generated by \`cfs toc\` -->`, while every other ADR in the repo wraps the list in `<!-- toc -->` / `<!-- /toc -->` with no heading. CFS could not locate the managed block. Normalised to match the sibling ADRs.

Caveat: `cfs` is a pipx tool that isn't installed here, so the TOC change is matched against repo convention rather than verified locally. CI will confirm.

gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
```diff
@@ -0,0 +1,149 @@
...
+* We add an explicit `[workspace.dependencies]` entry, `ort = { version = "=2.0.0-rc.12", default-features = false, features = ["ndarray", "std", "load-dynamic"] }`, layered on top of Magika's own feature set, and forward `load-dynamic` from the gear's `magika` feature. `load-dynamic` makes `ort` `dlopen` the ONNX Runtime shared library at process start (path supplied via the `ORT_DYLIB_PATH` env var, or a well-known default search path) instead of linking against it at compile time. This decouples "who builds the gear" from "who has the ONNX Runtime installed": CI here runs on plain GitHub-hosted `ubuntu-latest`/macOS runners with no per-gear container or custom base image (`.github/workflows/ci.yml`), so requiring onnxruntime dev headers at *build* time on every such runner is a heavier, more fragile ask than requiring the shared library only where the `magika`-enabled binary actually runs.
```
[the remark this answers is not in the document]
jalankulkija: Resolved (pushed shortly) — correct, and the error mattered more than "minor": the ADR listed `["ndarray", "std", "load-dynamic"]` while the workspace actually declares `default-features = false, features = ["load-dynamic", "api-24"]`.

Anyone copying the documented list would have dropped `api-24`, which this same ADR elsewhere says is mandatory — without it `ort`'s unconditionally-compiled `ep/vitis.rs` references an `OrtApi` field that only exists at that ABI level, so the build breaks. The text now states the actual direct configuration, notes that `ndarray`/`std` arrive transitively from `magika`, and warns explicitly against dropping `api-24`.

gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
```diff
@@ -0,0 +1,149 @@
...
+* `file-parser-sdk`'s public types are unaffected: `FileParserClientError::UnsupportedFileType { extension }` keeps its current shape; when detection is enabled and no parser matches a confidently-detected type, `extension` is populated with the detected label (e.g. `"zip"`) instead of the literal string `"no extension"`, but the error variant and its `#[error(...)]` message format do not change. Surfacing detection confidence/label in the SDK response (`ParsedText`) is explicitly deferred to a future iteration/ADR — this decision does not change the plugin trait (`FileParserBackend`) or the SDK response shape.
+* Magika's ~200 output labels map through a single new `magika_label -> extension` table to only the extensions this gear's registered plugins already declare via `supported_extensions()`; a confidently-detected label with no entry in that table (e.g. `"zip"`, `"mp4"`) is treated identically to an already-unsupported extension today — it produces the existing `no_parser_available` error, not a new error class.
```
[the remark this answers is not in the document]
jalankulkija: Resolved (pushed shortly) — you're right, and this was the most useful of the five, because two adjacent bullets both described behaviour the code does not have.

The ADR claimed an unsupported detected label populates `UnsupportedFileType { extension }` with the detected label, and separately that it "produces the existing `no_parser_available` error". Neither happens: `map_inference_result` intersects Magika's labels with the registered `supported_extensions()` and yields `None` for anything outside the set, and `reconcile_extension` independently requires a registered parser before a detection wins. A confident `"zip"` detection is therefore discarded and routing falls back to the caller's hint.

Both bullets are rewritten to say that, which also makes the answer to the golden-tests question simpler than the one previously documented: error text is unchanged in **both** feature states, not just with the feature off. I had cited that now-corrected bullet when resolving your earlier "clarify which golden tests remain unchanged" thread — that reply's conclusion still holds, but its reasoning was wrong, and the ADR is the thing that was at fault. Thanks for catching it.

gears/file-parser/docs/ADR/0001-cpt-cf-file-parser-adr-content-based-type-detection.md
```diff
@@ -0,0 +1,149 @@
...
+* `cpt-cf-file-parser-fr-local-path-security` — detection for extensionless local paths runs only after the existing `..`-rejection/canonicalization/base-dir checks in `parse_local`; this decision does not relax that ordering
+* `cpt-cf-file-parser-fr-plugin-extensibility` — the `magika_label -> extension` map is validated against each registered plugin's `supported_extensions()`, not a hardcoded parser list, so new plugins are picked up without touching the detector
+* `cpt-cf-file-parser-nfr-response-time` — the eager, once-at-startup `Session` load and `spawn_blocking`-wrapped, per-request inference are chosen specifically to keep steady-state request latency bounded and predictable
+* `cpt-cf-file-parser-component-parser-service` (`fdd-file-parser-component-parser-v1`) — `FileParserService`'s routing algorithm gains an optional detection step; `DESIGN.md:181-187` must be updated to reflect it
+* `cpt-cf-file-parser-component-parser-backend` (`fdd-file-parser-component-backend-v1`) — the `FileParserBackend` trait itself is explicitly *not* changed by this decision; detection lives entirely in the gateway, consistent with `DESIGN.md:85`
```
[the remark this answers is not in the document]
jalankulkija: Resolved (pushed shortly) — the Traceability entry claimed the `FileParserBackend` trait is "explicitly *not* changed by this decision", which stopped being true when `parse_local_path` gained `resolved_content_type: Option<&str>`.

Rewritten to draw the distinction you describe: type *resolution* stays entirely in the gateway (still consistent with `DESIGN.md:85`), but the trait signature does change, and implementations must accept the updated local-path contract — those routing purely by filename may ignore the argument. `DESIGN.md`'s plugin-contract section was already updated with the new signature.

gears/file-parser/file-parser/src/infra/magika_detector.rs
```diff
@@ -0,0 +1,285 @@
...
+/// an unmapped label (the routine "detection did not help" case) and a `NaN`
+/// score rather than propagating them. Split out so it can be unit-tested with
+/// synthetic values, without a real `Session` or ONNX Runtime.
+fn map_inference_result(
+    result: magika::Result<magika::FileType>,
+    supported_extensions: &HashSet<String>,
+) -> Option<DetectedType> {
+    let file_type = result
+        .inspect_err(|e| warn!(error = %e, "magika identify_content_sync failed"))
+        .ok()?;
```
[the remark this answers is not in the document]
jalankulkija: Resolved (pushed shortly) — correct, `map_inference_result` serves both routes, so a failure from `identify_file_sync` was being logged as `identify_content_sync`. Now a neutral `"magika inference failed"`, with a comment recording why it must stay neutral.

MikeFalcon77:
Two things I'd like changed before this lands.

## 1. There is no way to set the file type manually

Detection runs on every request once the feature is compiled in, and nothing can turn it off. `UploadQuery` has only `render_markdown` and `filename`; `ParseBytesRequest` has only `filename`, `content_type`, `bytes`. A detection at or above the threshold beats both the filename extension and the `Content-Type`, so a caller that knows the format exactly cannot force it.

The ADR's argument — callers are not trusted — holds for public upload. It does not hold for every caller. The in-process path through ClientHub (`FileParserLocalClient::parse_bytes`) goes through the same `FileParserService::parse_bytes` and gets sniffed as well, even though the calling gear normally knows the type it is passing. There detection is ~3.5 ms per request plus a chance of misrouting, and buys nothing.

Add an explicit way to pin the type: an optional field on `ParseBytesRequest` and a query parameter on the upload endpoint, which skips detection when set. If you do not want it on public REST, expose it on the SDK/ClientHub path only and leave REST always-detect.

## 2. Behavior changes with the feature off

`gears/file-parser/file-parser/src/domain/service.rs:297`:

```rust
let resolved_content_type = mime_table::mime_for_extension(&extension).or(content_type);
```

This runs regardless of the `magika` feature. When the extension is in the canonical table, the backend receives the table's MIME instead of the `Content-Type` the caller sent — `report.html` sent with `Content-Type: text/plain` now reaches the backend as `text/html`.

The ADR does list this as an intentional exception (line 37), so it is not an accident. My objection is to the choice, not the omission. An explicitly supplied `Content-Type` is now silently replaced by a value derived from the filename, and this hits every existing consumer, including those who never enable `magika`. The filename is the weaker of the two signals — that is the premise of this whole PR — so preferring it over an explicit header is backwards.

Use the table to fill in a missing `Content-Type`, and keep an explicitly supplied one. If the override is genuinely wanted, put it behind the feature so the default build is unchanged.

MikeFalcon77:
failed CI is not related to this PR. A new issues is created: https://github.com/constructorfabric/gears-rust/issues/4647

Issue #4289, opened 2026-07-29
CORE - Usage Collector plugin for ClickHouse


Artifizer:
**Description**: simple plugin to store/fetch data into ClickHouse DB
**Is Plugin**: YES
**Has Plugins**: NO
**Has Extension Points**: NO

PR #4590, opened 2026-08-18
feat(usage-collector): ClickHouse storage plugin for usage-collector

CommName:
## Summary

Adds the ClickHouse storage-backend plugin for the Usage Collector ([#4289](https://github.com/constructorfabric/gears-rust/issues/4289)). One logical part:

1. **New plugin** — adds `clickhouse-usage-collector-plugin` as a workspace member, aligned with this repo’s plugin conventions: crate manifest, lint rules, docs layout (`README.md`, `PRD.md`, `DESIGN.md`, `DECOMPOSITION.md`, feature specs), SQL migrations, config, domain adapter/ports, storage layer (record store, catalog store, query translate/keyset/aggregate, pool), cluster-backed per-`gts_id` coordination lock, metrics, unit tests, and ClickHouse integration tests. Wires an optional `clickhouse-usage-collector` feature on `cf-gears-example-server` and extends `make e2e-usage-collector` to run the shared usage-collector E2E suite against TimescaleDB and ClickHouse (separate binaries + containers).

## Type of Change

- [ x] Documentation update

## Testing

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing completed

Unit tests added across config, metrics, storage (record/catalog/mapper/pool/error/entity/query translate/keyset/aggregate/bind), lock manager, gear, and domain adapter. ClickHouse integration tests (feature-gated `clickhouse`, Docker): catalog, records_ingest, records_query. E2E: shared usage-collector suite runs twice via `UC_E2E_BACKEND` (TimescaleDB, then ClickHouse).

## Documentation

- [x] README updated (if applicable)
- [ ] API documentation updated (if applicable)

Adds plugin `README.md`, `PRD.md`, `DESIGN.md`, `DECOMPOSITION.md`, and feature specs; updates `docs/TESTING.md`, e2e workflow label, and e2e sidecar/config for ClickHouse.

## Checklist

- [x] Self-review completed

### New Features
- Added a ClickHouse-backed Usage Collector storage plugin for usage records and usage types.
- Implements `UsageCollectorPluginV1` with `ReplacingMergeTree(version)` + `FINAL` for dedup/deactivation, TTL retention on `usage_records`, keyset pagination, and pushed aggregation.
- Emulates referential integrity via a per-`gts_id` exclusive cluster lock (`DistributedLockV1`, profile `usage-collector`).
- Added backend observability metrics for the plugin.
- Extended usage-collector E2E to cover the ClickHouse backend alongside TimescaleDB.

## Summary by CodeRabbit

* **New Features**
  * Added ClickHouse as an alternative Usage Collector storage backend.
  * Added support for usage-record ingestion, querying, aggregation, pagination, catalog management, retention, and coordinated operations with ClickHouse.
  * Added configurable ClickHouse connection, TLS, timeout, locking, and retention settings.
* **Testing**
  * Usage Collector end-to-end tests now run against both TimescaleDB and ClickHouse.
* **Documentation**
  * Added ClickHouse setup, configuration, behavior, and testing documentation.


CommName:
@coderabbitai review

Cargo.toml
```diff
@@ -524,6 +535,8 @@ humantime = "2.3.0"
 # URL parsing
 url = "2.5"
 urlencoding = "2.1"
+# Percent-decode URL userinfo before applying credentials (e.g. ClickHouse DSN).
+percent-encoding = "2"
```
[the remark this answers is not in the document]
CommName: No finding found, this is false report as far as I see it

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/migrations/0001_init.sql
```diff
@@ -0,0 +1,117 @@
...
+    metadata        Map(String, String)         COMMENT 'Caller metadata; Map(String, String) chosen over JSON for efficient metadata[key] push-down at query time',
+    ingested_at     DateTime64(6)               COMMENT 'Server insert timestamp',
+    version         UInt64                      COMMENT 'ReplacingMergeTree version column; higher value wins on merge / FINAL resolution',
+
+    INDEX idx_records_id id TYPE bloom_filter GRANULARITY 1,
+    INDEX idx_records_corrects_id corrects_id TYPE bloom_filter GRANULARITY 1
+)
+ENGINE = ReplacingMergeTree(version)
+ORDER BY (tenant_id, gts_id, created_at, id)
+TTL toDateTime(created_at) + INTERVAL 31536000 SECOND DELETE;
```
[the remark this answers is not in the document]
CommName: Updated

Makefile
```diff
@@ -648,14 +648,28 @@ e2e-mini-chat:
...
+
+## Run usage-collector E2E tests against ClickHouse
+## (dedicated binary + ClickHouse container; Docker required)
+e2e-usage-collector-clickhouse:
+	cargo build --bin cf-gears-example-server --features=$(UC_E2E_FEATURES),clickhouse-usage-collector
+	E2E_BINARY=target/debug/cf-gears-example-server UC_E2E_BACKEND=clickhouse \
 		$(PYTHON) -m pytest testing/e2e/gears/usage_collector/ -vv
 
+## Run usage-collector E2E tests against both storage backends
+e2e-usage-collector: e2e-usage-collector-timescaledb e2e-usage-collector-clickhouse
```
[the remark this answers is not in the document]
CommName: Tests run linear, not valid for our pipelines

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/config.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub priority: i16,
+}
+
+impl Default for ClickHousePluginConfig {
+    fn default() -> Self {
+        Self {
+            database_url: SecretFromEnv::default(),
+            allow_insecure_http: false,
+            request_timeout_secs: 30,
+            lock_ttl_secs: 30,
```
nonameffh: `lock_ttl_secs = 30`, with client deadline calculated as `request_timeout_secs + CLIENT_DEADLINE_GRACE_SECS = 35 s`, can result in exceeding the lock’s lifetime.

Could we adjust the default TTL and make the relation explicit?

For example
```rust
if self.lock_ttl_secs <= self.client_deadline().as_secs() {
    return Err(format!(
        "lock_ttl_secs ({}) must exceed the client deadline ({}s = \
         request_timeout_secs + {CLIENT_DEADLINE_GRACE_SECS}): a single ClickHouse \
         round-trip inside the coordination lock must not outlive its lease",
        self.lock_ttl_secs,
        self.client_deadline().as_secs(),
    ));
}
```

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/infra/storage/record_store.rs
```diff
@@ -0,0 +1,1309 @@
...
+                return Err(e);
+            }
+        }
+
+        result
+    }
+
+    // @cpt-flow:cpt-cf-uc-ch-plugin-seq-ingest-batch
+    #[instrument(skip(self, records), fields(batch_size = records.len()))]
+    async fn create_batch(
```
nonameffh, line 1072: Phase 1. Acquisition is sequential and every early lease is held while the remaining locks are taken; the loop also keeps acquiring after a partition times out. With a 100-record batch of 100 distinct gts_id and the 5s per-lock timeout, the acquisition phase alone is unbounded relative to the TTL, and it is measured against a documented 500 ms p95 budget for that batch size.

From my perspective the cheapest shape that keeps everything intact is an overall acquisition budget rather than a partition cap:

```rust
let budget = self.lock_ttl.saturating_sub(self.request_timeout);
let started = Instant::now();
for (p_idx, (partition_gts_id, idxs)) in partitions.iter().enumerate() {
    let Some(remaining) = budget.checked_sub(started.elapsed()) else {
        // Budget spent
        for &idx in idxs {
            outcomes[idx] = Some(Err(UsageCollectorPluginError::transient(
                "create-batch lock acquisition budget exhausted",
            )));
        }
        continue;
    };
    // ... acquire with min(lock_timeout, remaining)
}
```

Partitions that never get a lock take exactly the existing acquire-failure path, so the single combined INSERT, the per-record outcome ordering, the version-offset determinism, and "a partition failure MUST NOT affect other partitions" all stay as documented.

CommName: Refactored the implementation to remove the locks entirely, as they are only required for a very rare use case: deleting a usage type.

nonameffh (approved):

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/tests/common/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// Bring up the harness, or print a Docker-unavailable notice and return
+/// `None` for the caller to skip its own test body.
+///
+/// Skipping is silent to the test harness (the test still reports `ok`), which
+/// makes a Docker-less run indistinguishable from a real one — including in a
+/// coverage report, where every gated line stays red while the suite claims
+/// success. Set `CH_REQUIRE_DOCKER=1` to turn a failed bring-up into a panic
+/// instead; coverage runs MUST set it.
+pub async fn bring_up_or_skip() -> Option<ChHarness> {
```
capybutler, line 150: **HIGH**

No CI job runs the ClickHouse integration tests.

These tests have two gates: the `clickhouse` feature and `#[ignore]` on every test. A run needs both `--features clickhouse` and `--run-ignored all`. With the feature flag alone, 67 tests still do not run. They are the tests that prove the dedup, the deactivation cascade, and the insert settings.

Add a Makefile target and a `ci.yml` step:

```makefile
test-usage-collector-ch: install-tools
	CH_REQUIRE_DOCKER=1 cargo nextest run -p cf-gears-clickhouse-usage-collector-plugin \
		--features clickhouse --run-ignored all --no-fail-fast
```

Keep `CH_REQUIRE_DOCKER=1`. Without it, `bring_up_or_skip` reports `ok` when Docker is absent.

Makefile
```diff
@@ -1026,10 +1026,27 @@ e2e-mini-chat:
 	$(call print_target_banner)
 	$(MAKE) e2e-local SUITE=mini-chat
 
-## Run usage-collector E2E tests (alias for focused local E2E; Docker required)
-e2e-usage-collector:
-	$(call print_target_banner)
-	$(MAKE) e2e-local SUITE=usage-collector
+# Shared by both backends; each build appends its own storage-plugin feature.
+# The plugins cannot share a binary: every linked gear is initialized, and both
```
capybutler: **HIGH**

`make run` and `make example` fail on this branch. The build succeeds. The server then stops during gear init, because the ClickHouse plugin returns this error:

```
invalid clickhouse-usage-collector-plugin config: database_url must not be empty
```

`Makefile:27` holds the exclude list: the features that the example server must not link. The new `clickhouse-usage-collector` feature is not in that list. `Makefile:29` subtracts the exclude list from all example-server features, so the build links the plugin by default.

`make -n run` shows it:

```
cargo run --bin cf-gears-example-server --features …,clickhouse-usage-collector,… -- --config config/quickstart.yaml run
```

`config/quickstart.yaml` has no usage-collector section. The plugin gets the default configuration, so `database_url` is empty. `validate()` rejects it at `config.rs:280`, and `gear.rs:40` returns the error above.

`timescaledb-usage-collector` is in the exclude list for this reason. The comment two lines above states it: "every linked gear is initialized, and both fail init without their own live database".

Add the feature to the exclude list:

```diff
-EXAMPLE_SERVER_FEATURE_EXCLUDES ?= default fips k8s otel oop-example timescaledb-usage-collector
+EXAMPLE_SERVER_FEATURE_EXCLUDES ?= default fips k8s otel oop-example timescaledb-usage-collector clickhouse-usage-collector
```

This defect does not affect CI. No CI job runs `make run` or `make example`.

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/config.rs
```diff
@@ -0,0 +1,351 @@
...
+            allow_insecure_http: false,
+            request_timeout_secs: 30,
+            // On by default: the write path is one INSERT per request, so
+            // without server-side coalescing a request-shaped ingest stream
+            // writes one small part per record.
+            async_insert: true,
+            // Same window the migration DDL bakes in, so a config-less start
+            // needs no `MODIFY TTL` reconciliation at startup.
+            retention_period_secs: crate::infra::storage::pool::DEFAULT_RETENTION_SECS,
+            vendor: "cyberfabric".to_owned(),
```
capybutler: **HIGH**

Change the default `vendor` from `cyberfabric` to `constructorfabric`.

`vendor` is the key that binds the gear to a storage plugin. The gear default is `constructorfabric` (`usage-collector/src/config.rs:38`). This plugin uses `cyberfabric`, so the two names do not match.

The gear resolves the plugin lazily, on the first dispatch. Startup stays clean: the plugin migrates the database, registers its instance, and reports ready. The first request then finds no instance for `constructorfabric`. The gear maps this to `plugin_unavailable()` and returns **HTTP 503**. Every request after it returns 503.

A deployment that sets `database_url` and omits `vendor` gets a healthy server that returns 503 to everything.

`cyberfabric` is also a legacy name. The project renamed Cyber Fabric to Constructor Fabric, and `.github/forbidden-words.json:38` bans the term. No other gear uses it as a live value.

Sites:

- `src/config.rs:211` — the default
- `src/config_tests.rs:6` — asserts the wrong default
- `src/config.rs:327` — example text in the validation error
- `README.md:16` and `README.md:27`
- `testing/e2e/suites/usage_collector/config-clickhouse.yaml:180` and `:184`

The comment at `config-clickhouse.yaml:174` says the gear and the plugin "happen to share the same crate default today (`cyberfabric`)". That is not correct. The gear default is `constructorfabric`.

The E2E config sets `vendor` on both sides, so the suite never uses the default. Remove both lines to cover that path.

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/infra/storage/pool.rs
```diff
@@ -0,0 +1,666 @@
...
+    });
+
+    // `send_timeout` and `receive_timeout` are standard `ClickHouse` HTTP API
+    // settings accepted as URL query parameters or per-request headers.
+    // The `clickhouse` crate passes them as request headers on every query.
+    // Both are set to `request_timeout_secs` — the crate exposes a single
+    // combined timeout knob rather than separate send/receive splits.
+    let timeout_str = cfg.request_timeout_secs.to_string();
+
+    let mut client = clickhouse::Client::default()
```
capybutler: **HIGH**

The ClickHouse client uses its own crypto provider. A FIPS build therefore sends ClickHouse traffic outside the FIPS boundary.

`Cargo.toml:643` enables the `rustls-tls` feature. That feature makes the `clickhouse` crate build its own TLS connector with a hardcoded provider (`clickhouse-0.15.1/src/http_client.rs:63`):

```rust
prepare_hyper_rustls_connector(connector, rustls::crypto::aws_lc_rs::default_provider())
```

This is plain `aws-lc-rs`, not `default_fips_provider()`. The crate never calls `CryptoProvider::get_default()`, so it ignores the provider that `init_crypto_provider` installs (`libs/toolkit/src/bootstrap/crypto.rs:113`). `pool.rs:179` reaches it through `clickhouse::Client::default()`.

Under `--features fips` the process installs a validated provider, and ClickHouse still uses the plain one. On macOS and Windows the installed provider is Apple corecrypto or Windows CNG, not aws-lc. The connection carries the usage records and the database password.

Three checks miss this:

- `FipsProviderConflict` (`crypto.rs:161`) fires only when a second provider calls `install_default()`. The crate never calls it.
- `deny-fips.toml:113` still lists non-FIPS `aws-lc-rs` as Phase B, so `make fips-policy` passes. That check reads the dependency graph, and this is a source-level bypass.
- `make test-fips` covers `cf-gears-toolkit`, `cf-gears-toolkit-http` and `cf-gears-oagw`. This plugin is not one of them.

A FIPS build logs "FIPS-140-3 crypto provider installed", passes its own `provider.fips()` witness at `crypto.rs:183`, and still uses non-validated crypto for ClickHouse.

Build the connector from the installed provider. Pass it to `clickhouse::Client::with_http_client` (`clickhouse-0.15.1/src/lib.rs:121`) in place of `Client::default()`:

```rust
let provider = rustls::crypto::CryptoProvider::get_default()...;
hyper_rustls::HttpsConnectorBuilder::new()
    .with_provider_and_webpki_roots((**provider).clone())?
```

Keep the settings that `Client::default()` gives today: `enforce_http(false)`, `https_or_http()`, `enable_http1()`, TCP keepalive 60s, and `pool_idle_timeout(2s)` (`http_client.rs:46-76`).

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/infra/storage/record_store.rs
```diff
@@ -0,0 +1,1717 @@
...
+                .map(str::to_owned)
+                .unwrap_or_default()
+        })
+        .collect();
+    let value = match obj.get("agg") {
+        None | Some(serde_json::Value::Null) => None,
+        Some(v) => {
+            let s = match v {
+                serde_json::Value::String(s) => s.clone(),
+                serde_json::Value::Number(n) => n.to_string(),
```
capybutler, line 1702: **HIGH**

`SUM`, `MIN` and `MAX` lose precision. Aggregate values pass through an `f64`.

`aggregate` reads the result as `JSONEachRow` (`record_store.rs:1477`) and parses each value at `record_store.rs:1699`:

```rust
serde_json::Value::Number(n) => n.to_string(),
```

ClickHouse sends a `Decimal128(9)` as an **unquoted** JSON number. `serde_json` has no `arbitrary_precision` feature (root `Cargo.toml:418`), so it stores an unquoted fractional number as an `f64`. `BigDecimal` then parses a number that is already wrong.

Measured against ClickHouse 25.6, the pinned tag:

```
SELECT sum(value) AS agg FROM t FORMAT JSONEachRow
{"agg":1234567890.123456789}

SELECT name, value FROM system.settings WHERE name = 'output_format_json_quote_decimals'
{"name":"output_format_json_quote_decimals","value":"0"}
```

Those exact bytes through the parse path above:

```
{"agg":12345678.123456789}      -> 12345678.123456787
{"agg":1234567890.123456789}    -> 1234567890.1234567
{"agg":"1234567890.123456789"}  -> 1234567890.123456789
```

A sum above about 10 million already returns a digit that was never in the data. `COUNT(*)` is safe: `output_format_json_quote_64bit_integers` defaults to `1`, so counts arrive quoted.

Add one setting to the aggregate query:

```rust
.with_setting("output_format_json_quote_decimals", "1")
```

The value then arrives quoted and takes the `Value::String` branch, which is exact. Measured:

```
... FORMAT JSONEachRow SETTINGS output_format_json_quote_decimals=1
{"agg":"1234567890.123456789"}
```

`AVG` needs a separate fix. ClickHouse `avg()` returns `Float64`, so `ROUND(AVG(value), 6)` (`query/aggregate.rs:47`) is already a float on the server and the setting does not apply:

```
... SETTINGS output_format_json_quote_decimals=1
{"agg":1234567890.123457}
```

Cast it, as in `toDecimal128(avg(value), 9)`.

No test covers this. The unit tests feed `"agg":"1"` and `"agg":2` (`record_store_tests.rs:1003,1009`), and every integration assertion is a whole number. An unquoted fractional number is the one shape that fails, and no test uses it.

Makefile
```diff
@@ -1026,10 +1026,27 @@ e2e-mini-chat:
...
+# fail init without their own live database. Hence two builds, two runs.
+UC_E2E_FEATURES = usage-collector,static-tenants,static-authn,static-authz
+
+## Run usage-collector E2E tests against TimescaleDB
+## (dedicated binary + TimescaleDB container; Docker required)
+e2e-usage-collector-timescaledb:
```
capybutler: **MEDIUM**

The two new E2E targets call `$(PYTHON)` without the `py-env` prerequisite.

`$(PYTHON)` is `.venv/bin/python` (`Makefile:10`). Only `py-env` creates that virtual environment (`Makefile:195-204`). Every other E2E target declares it:

```
976:  e2e-docker: py-env
1004: e2e-local: py-env
1009: e2e-tr-authz: py-env
1014: e2e-local-smoke: py-env
1036: e2e-usage-collector-timescaledb:      <- missing
1043: e2e-usage-collector-clickhouse:       <- missing
```

The old recipe was `$(MAKE) e2e-local SUITE=usage-collector`, so it inherited the prerequisite. The rewrite dropped it.

On a clean checkout, `make -n` shows the difference:

```
$ make -n e2e-local
/opt/homebrew/bin/python3 -m venv .venv
.venv/bin/python -m pip install -r testing/requirements.txt -r testing/e2e/requirements.txt

$ make -n e2e-usage-collector-clickhouse
cargo build --bin cf-gears-example-server --features=...
.venv/bin/python -m pytest testing/e2e/suites/usage_collector/ -vv
```

The second target runs the full cargo build and then stops, because the interpreter does not exist.

CI passes because an earlier step runs `make e2e-local` in the same job (`.github/workflows/e2e.yml:101`), which creates the virtual environment first. `make e2e-usage-collector` alone is a documented command (`docs/TESTING.md:49`, `testing/e2e/README.md:473`), and it fails on a new clone.

```diff
-e2e-usage-collector-timescaledb:
+e2e-usage-collector-timescaledb: py-env
+	$(call print_target_banner)

-e2e-usage-collector-clickhouse:
+e2e-usage-collector-clickhouse: py-env
+	$(call print_target_banner)
```

Two smaller losses from the same rewrite: every other target opens with `$(call print_target_banner)`, and `e2e-local` forwards `-- $(E2E_TARGET)` (`Makefile:1006`). With neither, `make e2e-usage-collector E2E_TARGET="-k test_foo"` runs the whole suite: two image pulls, two container starts, two server builds.

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/config.rs
```diff
@@ -0,0 +1,351 @@
+use serde::Deserialize;
+use toolkit::var_expand::{ExpandVars as ExpandVarsTrait, ExpandVarsError};
+
+/// Wrapper around a config string whose final value is a secret.
+///
+/// Holds a plain `String` for `Deserialize` + `ExpandVars` compatibility
+/// (toolkit's `#[expand_vars]` derive substitutes `${VAR}` placeholders on
+/// `String` fields; `secrecy::SecretString` is not `ExpandVars`-aware), while
```
capybutler: **LOW**

`SecretFromEnv` works around a limitation that does not exist.

`config.rs:6-8` gives the reason for the wrapper:

> `secrecy::SecretString` is not `ExpandVars`-aware

The toolkit ships that impl. `libs/toolkit-utils/src/var_expand.rs:148-153`:

```rust
impl ExpandVars for secrecy::SecretString {
    fn expand_vars(&mut self) -> Result<(), ExpandVarsError> {
        use secrecy::ExposeSecret;
        let expanded = expand_env_vars(self.expose_secret())?;
        *self = secrecy::SecretString::from(expanded);
        Ok(())
    }
}
```

`SecretString` also deserializes: the workspace enables the feature (`Cargo.toml:433`, `secrecy = { version = "0.10", features = ["serde"] }`). Two plugins already use it as a config field — `static-authn-plugin/src/config.rs:91` and `http-json-plugin/src/config.rs:74`.

So this works:

```rust
#[expand_vars]
pub database_url: SecretString,
```

It gives the same redaction, plus zeroize on drop, and removes about 35 lines.

The redaction in `SecretFromEnv` is correct, so this is not a security defect. Two things make it worth doing. The wrapper is copied into three crates with the same wrong comment. A plain `String` keeps the password readable in the heap after it is dropped.

Zeroize helps only in part. `clickhouse-0.15.1/src/lib.rs:87-91` holds the password as a plain `String` for the life of the client, so one copy stays either way.

CommName: Updated, just so you are aware timesacledb plugin has the same issue

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/infra/storage/pool.rs
CommName: Updated: thanks for suggestion on how to fix it

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/tests/common/mod.rs
CommName: Updated, I added make command and updated ci.yml step

Makefile
CommName: Missed this path, my bad, it should be fixed now

Makefile
CommName: Updated, 
Thanks for suggestion on how to fix it

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/config.rs
CommName: Not sure where I read cyberfabric, but you are right. I updated it everywhere to constructorfabric

gears/system/usage-collector/plugins/clickhouse-usage-collector-plugin/src/infra/storage/record_store.rs
CommName: Forgot to comment on this, I updated it but with suggestion from llm

> toDecimal128(avg(value), 9) would not work. avg(value) evaluates to Float64 first, so the cast converts an already-lossy float — it fixes the transport hop but not the computation hop. Dividing the two exact Decimal aggregates (Decimal128(9) / UInt64 → Decimal(38, 9)) is the fix that actually preserves precision, which is what's in the tree.

capybutler (approved):

PR #4598, opened 2026-08-18
feat(contract): support PlatformSecurityContext as an outbound contract parameter

### LINKED ISSUE #4577: Support `PlatformSecurityContext` as a contract-function parameter (outbound)
### Summary
Let gears express platform-plane (non-tenant, workload-authenticated) calls through the contract macros, the same way tenant-plane calls already work. Today a contract method that takes `PlatformSecurityContext` cannot be used: REST rejects it at compile time, and gRPC silently mishandles it (the parameter is treated as request payload, no credential is attached, and it usually fails to compile). Every platform-plane call must therefore be hand-wired with a manual client and a manually attached internal-auth interceptor - defeating the contract-first workflow, and leaving an accepted design ([ADR-0006](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md) / [ADR-0008](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md)) unimplemented at the codegen layer.

### Problem
The tenant and platform planes differ in *where the credential comes from*:

- **Tenant**: the `SecurityContext` argument carries the caller's bearer token; the client just forwards it.
- **Platform**: `PlatformSecurityContext` is a validated identity with no secret inside it. The credential to present is the **calling process's own** internal credential (`InternalCredential`) — the same regardless of which service is called.

So the parameter can only ever be a compile-time marker for "this is a platform call." The credential must be sourced from the runtime, not the argument.

### Design (per [ADR-0006](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md) / [ADR-0008](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md))
This is the split the accepted ADRs already mandate - the issue implements it at the codegen layer:

- **The context type *is* the plane marker; carrier selection is per-method and compile-time.** A method whose first argument is `&PlatformSecurityContext` is a platform call and maps to `X-ToolKit-Internal-Token`; `&SecurityContext` maps to `Authorization: Bearer`. Codegen MUST NOT default a platform-plane method to `Authorization`. Plane is a property of the individual method signature, not of the client.
- **Credential population happens below the contract layer.** The runtime attaches the actual secret via `attach_internal_auth`, sourced from the bootstrap-selected `InternalCredential`. The contract macro's only jobs are: recognize `PlatformSecurityContext` as the plane marker, keep it off the wire, route the method to the internal-token carrier, and remove the REST compile-time rejection. Gear code never sets an auth header.
- **No credential configured ⇒ attach nothing.** `InternalCredential::None` (in-process / single-process) is legitimate, so the client stays permissive rather than failing when no credential is present; requirement is enforced server-side.

### Open questions
- **Per-method carrier signalling.** How codegen conveys the per-call plane to the transport (a request marker the interceptor reads, vs. a per-method attach that pulls the runtime credential) rather than assuming a plane-uniform channel.
- **Transport security.** Whether attaching an internal token requires TLS, or continues to rely on the in-mesh trust boundary as tenant traffic does today.

### References
- [ADR-0008 — Two-Plane Authentication Model](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md)
- [ADR-0006 — Platform-Plane Authentication](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md)

mattgarmon:
Closes https://github.com/constructorfabric/gears-rust/issues/4577

## Summary by CodeRabbit

* **New Features**
  * Added platform-plane authentication support for REST and gRPC clients.
  * Platform credentials use the `X-ToolKit-Internal-Token` header.
  * Added support for static and rotating internal tokens, including refresh during retries and streaming reconnects.
  * Platform security contexts are accepted without appearing in request payloads.
  * Requests without configured credentials continue without authentication headers.

* **Bug Fixes**
  * Tenant requests use bearer authentication while platform requests use internal tokens.
  * TLS requirements are now enforced for secure connections.


libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -594,18 +621,29 @@ fn body_param_ident(method: &GrpcMethodModel) -> Option<syn::Ident> {
...
+/// Whether the method's security-context parameter is the *platform* plane
+/// (`PlatformSecurityContext`). Platform-plane methods carry the runtime
+/// internal credential (`x-toolkit-internal-token`) rather than a tenant bearer
+/// token (`cpt-cf-adr-two-plane-auth`).
+fn security_context_is_platform(method: &GrpcMethodModel) -> bool {
+    method
+        .params
+        .iter()
+        .any(|p| is_platform_security_context_type(&p.ty))
+}
```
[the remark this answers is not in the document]
mattgarmon: Fixed

libs/toolkit/src/bootstrap/run.rs
```diff
@@ -87,6 +87,7 @@ pub async fn run_server(config: AppConfig) -> Result<()> {
         instance_id,
         oop: oop_options,
         shutdown_deadline: None,
+        internal_token_provider: None,
```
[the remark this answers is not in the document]
mattgarmon: Internal auth not needed for the embedded run, everything is in the same process

libs/toolkit-contract/Cargo.toml
```diff
@@ -30,6 +30,7 @@ runtime-client = [
     "dep:futures-core",
     "dep:rand",
     "dep:async-stream",
+    "dep:secrecy",
```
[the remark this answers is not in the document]
mattgarmon: Not needed

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -1107,28 +1133,71 @@ fn build_query_string(method: &RestMethodModel, support: &TokenStream) -> TokenS
...
+///   sourced from the runtime, **never** from the argument (which carries no
+///   secret). Permissive: with no provider configured
+///   (`InternalCredential::None`) nothing is attached — the requirement is
+///   enforced server-side.
+/// - No context parameter → both `None`.
+fn capture_auth_credentials(method: &RestMethodModel, support: &TokenStream) -> TokenStream {
+    let secctx = method
+        .params
+        .iter()
+        .find(|p| is_security_context_type(&p.ty));
```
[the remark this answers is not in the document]
mattgarmon: Added

striped-zebra-dev (approved):

libs/toolkit-contract-macros/src/grpc_contract_parse.rs
```diff
@@ -314,6 +314,26 @@ fn parse_params(method: &TraitItemFn) -> syn::Result<Vec<GrpcParam>> {
...
+
+    // Exactly one security-plane context per method: the client codegen
+    // (`security_context_param` / plane detection in `grpc_contract.rs`)
+    // picks a single param to source the auth metadata from. A second
+    // context param (e.g. a stray `PlatformSecurityContext` alongside a
+    // tenant `SecurityContext`) would silently be dropped from the wire
+    // without ever being consulted for token attachment, mixing planes.
+    let secctx_count = params
+        .iter()
+        .filter(|p| crate::projection::is_security_context_type(&p.ty))
```
MikeFalcon77, line 326: `PlatformSecurityContext` is now a plane marker here, but `determine_param_role` in `parse.rs` was not updated — it still matches only the exact ident `SecurityContext`:

```rust
if name == "ctx" && type_last_segment_is(ty, "SecurityContext") {
    return ParamRole::SecurityContext;
}
ParamRole::Wire
```

So `#[contract]` records a platform context as `FieldRole::Wire`. `method_input_type` in toolkit-contract-protogen keeps exactly the `Wire` fields, so the generated `.proto` request message for a platform-plane RPC gets a `ctx` field typed `PlatformSecurityContext`, which is not a defined proto message. If the context is the method's only other param, it becomes the whole request type.

The plane marker ends up on the wire — the opposite of what this PR intends. The new gRPC test hand-writes its stubs, so it never runs protogen and does not catch this.

Change `determine_param_role` to use `crate::projection::is_security_context_type(ty)`.

libs/toolkit-contract-macros/src/rest_contract_parse.rs
```diff
@@ -327,23 +327,14 @@ fn parse_method(method: &TraitItemFn) -> syn::Result<RestMethodModel> {
...
-            return Err(syn::Error::new_spanned(
-                &first.ty,
-                "platform-plane REST projections are not supported yet: the generated \
-                 client cannot source the internal token, so it would emit an \
-                 unauthenticated request. Serve this contract over gRPC \
-                 (`#[toolkit::grpc_contract]` + `InternalAuthInterceptor`) or write a \
-                 manual client. See DESIGN section 2.2 (cpt-cf-binding-constraint-security-context)",
-            ));
-        }
+        // Either plane is accepted. The context type is the compile-time plane
```
MikeFalcon77, line 330: Accepting platform-plane REST methods here is only half the change — the server-side route codegen was not updated. `generate_route_registration` picks the auth axis from `method.anonymous` alone:

```rust
let auth_registration = if method.anonymous {
    quote! { .anonymous() }
} else {
    quote! { .authenticated().no_license_required() }
};
```

A `PlatformSecurityContext` method without `#[anonymous]` therefore registers a route that demands a tenant JWT. `security_context_middleware` rejects it with 401 before the handler runs, and the handler's `Extension<PlatformSecurityContext>` is only inserted by `internal_auth_middleware`. A legitimate `X-ToolKit-Internal-Token` caller carrying no JWT cannot reach the endpoint.

The new REST test uses `#[server_manual]`, so no route is generated and this is untested.

Select the auth axis from the plane marker: emit `.anonymous()` for platform-plane methods, or require `#[server_manual]` on them until `OperationBuilder` has a platform auth axis.

libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -468,13 +477,26 @@ fn generate_one_shot_unary_method(
...
     let method_ident = &method.ident;
     let attach_metadata = match ctx_ident {
+        // Platform plane: source the credential from the runtime provider on
+        // `self.config`, not from the `PlatformSecurityContext` argument (which
+        // carries no secret). The marker stays off the wire; `let _ = &#ctx`
+        // consumes it so it is not an unused parameter. Permissive when no
+        // provider is configured.
+        Some(ctx) if is_platform => quote! {
+            let _ = &#ctx;
+            #support::grpc::attach_internal_token(
```
MikeFalcon77, line 538: This attaches the process-wide internal token, but the generated `connect` (same file) builds the channel with no TLS:

```rust
let endpoint = ::tonic::transport::Endpoint::from_shared(config.base_url.clone())?;
let endpoint = endpoint.timeout(config.timeout);
let channel = endpoint.connect().await?;
```

`config.require_tls` is never read and `ClientTlsConfig` is never applied, so the credential goes out over cleartext h2c and there is no way to force otherwise. Unlike a tenant JWT this token is long-lived and process-scoped — it authenticates the gear to the whole platform plane.

Apply `ClientTlsConfig` in `connect` when `config.require_tls` is set, and refuse to attach the token when the channel scheme is not `https`.

libs/toolkit-contract-macros/src/provides.rs
```diff
@@ -311,8 +311,14 @@ pub fn generate(attr: &ProvidesAttr, item: &ItemStruct) -> SynResult<TokenStream
         quote! {
             ::toolkit_contract::wiring::ClientWiring::Rest { endpoint, tuning } => {
                 #rest_policy_warning
+                // Thread the process's platform-plane credential source onto the
+                // client config so platform-plane (`PlatformSecurityContext`)
+                // methods attach `X-ToolKit-Internal-Token`. `None` (Profile 1 /
+                // no credential) attaches nothing (cpt-cf-adr-two-plane-auth).
+                let __cfg = tuning.apply_to(endpoint)
```
MikeFalcon77, line 318: This threads the provider through the `#[toolkit::provides]` Rest and Grpc arms only. The other ClientHub wiring path — the `<Contract>RestResolvingClient` proxy from `#[toolkit::consumes]` — still builds its config without one, in `resolving.rs`:

```rust
let cfg = self.tuning.apply_to(&uri);
```

The directory-resolving consumer is the main Profile 2/3 inter-gear path, and its generated wire fn receives only `&ClientHub` and an `EndpointResolver` — no `GearCtx` — so it has no way to reach a provider. Platform-plane calls through a ClientHub proxy attach nothing and get rejected server-side. The DESIGN.md change in this PR says the platform plane is fully wired over REST and gRPC; that is not true for this path.

Carry the provider through `ClientTuning` / `DirectoryResolvingClient::new` so the rebuild closure calls `.with_internal_token_provider(..)`, and extend the `#[toolkit::consumes]` wire fn to receive it.

libs/toolkit/src/bootstrap/run.rs
MikeFalcon77: The in-process host never builds a provider, while `run_oop_with_options` does. The config is reachable from here — `AppConfig.oop_http.internal_auth`, whose own doc says it drives "both the *inbound* HTTP validator ... and the *outbound* credential".

So a process started this way validates inbound platform credentials but sends outbound platform-plane calls unauthenticated. Since this PR also removed the compile-time rejection of platform-plane REST, there is now no compile-time and no runtime signal — just a 401 from the peer.

Move `build_internal_token_provider` into a shared bootstrap module and call it here with `config.oop_http.as_ref().and_then(|h| h.internal_auth.as_ref())`. If the host is meant to have no credential, say so in a comment and log it.

libs/toolkit-contract/tests/grpc_client_platform_secctx.rs
```diff
@@ -0,0 +1,135 @@
...
+async fn platform_grpc_client_builds_with_internal_token_provider() {
+    let channel = tonic::transport::Channel::from_static("http://127.0.0.1:1").connect_lazy();
+    let config = ClientConfig::new("http://127.0.0.1:1").with_internal_token_provider(
+        InternalTokenProvider::from_token(SecretString::from("sa.jwt.token")),
+    );
+
+    let client = RegApiGrpcClient::new(channel, config);
+    // Both the base trait (real bodies) and the projection trait (delegating
+    // defaults) must be object-safe views of the generated client.
+    let _as_base: Arc<dyn RegApi> = Arc::new(client);
```
MikeFalcon77, line 147: This test has no assertions. It builds a client, coerces it to `Arc<dyn RegApi>`, and ends — so it passes as long as the code compiles, which the UI pass fixture already covers. It is named after the internal-token provider but never checks the provider is used.

The stub also drops `_request` and returns `Status::unimplemented`, so the gRPC token attachment is not verified at runtime anywhere in this PR. Attaching nothing, attaching under the wrong key, or attaching as `authorization` metadata would leave every test green.

Capture the metadata in the stub (`Arc<Mutex<Option<MetadataMap>>>` filled from `request.metadata().clone()`), invoke `RegApi::register`, then assert `x-toolkit-internal-token` is `sa.jwt.token` and there is no `authorization` key.

libs/toolkit-contract/tests/rest_client_platform_secctx.rs
```diff
@@ -0,0 +1,165 @@
...
+    #[idempotency(SafeRead)]
+    async fn whoami(
+        &self,
+        ctx: &PlatformSecurityContext,
+        id: String,
+    ) -> Result<HeaderEcho, DemoError>;
+}
+
+#[rest_contract(base_path = "/api/sys/v1")]
+pub trait SysApiRest: SysApi {
```
MikeFalcon77, line 93: Nothing tests the tenant plane in the other direction: that a `SecurityContext` method still sends `Authorization: Bearer` and does not send `X-ToolKit-Internal-Token`.

`capture_auth_credentials` was rewritten in this PR and now emits both `__bearer` and `__internal_token_provider` bindings for the tenant arm too. No test in the workspace checks that a generated client actually emits `Authorization`, so silently dropping tenant auth would go unnoticed.

Add a second method to `SysApiRest` taking `ctx: &SecurityContext`, call it on a client that also has an internal-token provider configured, and assert `authorization` is `Some` and `internal_token` is `None`.

libs/toolkit-contract/src/grpc/mod.rs
```diff
@@ -74,6 +87,36 @@ where
...
+/// sourced from the runtime (`provider`), never from the contract argument —
+/// `PlatformSecurityContext` carries no secret. The call is **permissive**: a
+/// `None` provider, or a provider that currently yields no token, attaches
+/// nothing (legitimate for Profile 1 / `InternalCredential::None`); the
+/// requirement is enforced server-side.
+///
+/// # Errors
+/// Returns a [`TransportError::Grpc`] when the token cannot be represented as an
+/// ASCII metadata value.
+pub fn attach_internal_token(
```
MikeFalcon77, line 99: `attach_internal_token` has no unit test. Nothing checks the token lands under `x-toolkit-internal-token`, that a `None` provider is a no-op, or that a token that is not valid ASCII metadata maps to `TransportError::Grpc`.

`attach_bearer` right above it is tested (`attach_bearer_returns_transport_error_on_failure`), so the platform plane is the only auth path in this file whose metadata write and error mapping are unverified.

Add three cases to the existing `mod tests`: a provider yielding a token (assert the key and that `authorization` is absent), a `None` provider and a provider yielding `None` (assert the map is untouched), and a token containing `\n` (assert `Err(TransportError::Grpc { .. })`).

libs/toolkit-contract-macros/src/provides.rs
```diff
@@ -311,8 +311,14 @@ pub fn generate(attr: &ProvidesAttr, item: &ItemStruct) -> SynResult<TokenStream
...
+                    .with_internal_token_provider(ctx.internal_token_provider());
```
MikeFalcon77, line 319: The provider is moved onto the config for a directory-resolved `endpoint` with no scheme check. `ClientConfig::require_tls` defaults to `false`, so `build_default_http_client` runs with `TransportSecurity::AllowInsecureHttp`.

An `http://` endpoint from discovery — or a spoofed directory registration — receives the gear's raw platform credential in plaintext. There is no scheme gate or allow-list bounding which resolved host the secret may go to.

Attach the provider only when the resolved endpoint is `https://` or TLS is required, and fail the wiring otherwise, with an explicit opt-out for local dev.

libs/toolkit-http/src/request.rs
```diff
@@ -201,6 +201,38 @@ impl RequestBuilder {
...
+    /// sensitive.
+    pub fn internal_token_auth(mut self, token: &str) -> Self {
+        if self.error.is_some() {
+            return self;
+        }
+        match http::header::HeaderValue::try_from(token) {
+            Ok(mut value) => {
+                value.set_sensitive(true);
+                self.headers.push((
+                    http::header::HeaderName::from_static(
```
MikeFalcon77, line 222: The doc comment claims the same guarantee as `attach_internal_token_http`, but `x-toolkit-internal-token` is not in `SENSITIVE_HEADERS` in `layers/redirect.rs`:

```rust
const SENSITIVE_HEADERS: &[header::HeaderName] = &[
    header::AUTHORIZATION,
    header::COOKIE,
    header::PROXY_AUTHORIZATION,
];
```

`strip_sensitive_headers` therefore leaves it in place. With `RedirectConfig::permissive()` or any `allowed_redirect_hosts` entry, a 3xx forwards the platform credential to a host the responding server picks. `set_sensitive(true)` covers logging, not redirects.

Add `toolkit_security::constants::INTERNAL_TOKEN_HEADER` to `SENSITIVE_HEADERS`.

libs/toolkit-contract/src/grpc/mod.rs
```diff
@@ -74,6 +87,36 @@ where
...
+    metadata: &mut MetadataMap,
+    provider: Option<&crate::runtime::config::InternalTokenProvider>,
+) -> Result<(), TransportError> {
+    let Some(token) = provider.and_then(crate::runtime::config::InternalTokenProvider::current)
```
MikeFalcon77: This returns `Ok(())` and attaches nothing in two very different cases: no provider configured (Profile 1, intended), and a configured provider that currently yields no token (broken credential source — e.g. the projected token file went empty). Neither emits any trace event.

This is the runtime stand-in for the compile-fail guard the PR deletes, and it provides no detection at all. The only symptom is a remote `PermissionDenied` with nothing local tying it to a missing credential.

Split the branches: stay silent when `provider.is_none()`, and `tracing::warn!` (naming the plane and RPC, never the token) when a provider is present but yields nothing. Consider an opt-in strict flag on `ClientConfig` that turns that case into a `TransportError`.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -745,6 +756,38 @@ async fn build_internal_authenticator(
...
+    if let Some(secret) = cfg.shared_secret() {
+        return Ok(Some(InternalTokenProvider::from_token(secret)));
+    }
+    if let Some(path) = cfg.kube_token_path() {
+        let reader = toolkit_transport_grpc::ServiceAccountTokenReader::new(path)
+            .await
+            .context("failed to read projected service-account token for outbound credential")?;
+        return Ok(Some(InternalTokenProvider::new(reader.token_provider())));
+    }
+    Ok(None)
```
MikeFalcon77: This probes `InternalAuthConfig` through the `shared_secret()` / `kube_token_path()` accessors and falls through to `Ok(None)`, so a config matching neither yields no outbound credential — silently, with no log.

`build_internal_auth_interceptor` matches `InternalAuthConfig` exhaustively, so a new variant there is a compile error; `build_internal_authenticator` deliberately `bail!`s rather than allow a silent downgrade. The accessor chain reintroduces that downgrade for the planned mTLS/SPIFFE variant.

Match exhaustively on `InternalAuthConfig` so a new variant fails to compile, and `warn!` on the `kube`-without-`token_path` arm that platform-plane calls will carry no credential.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -745,6 +756,38 @@ async fn build_internal_authenticator(
...
+async fn build_internal_token_provider(
+    cfg: &toolkit_security::InternalAuthConfig,
+) -> Result<Option<toolkit_contract::runtime::config::InternalTokenProvider>> {
+    use toolkit_contract::runtime::config::InternalTokenProvider;
+
+    if let Some(secret) = cfg.shared_secret() {
+        return Ok(Some(InternalTokenProvider::from_token(secret)));
+    }
+    if let Some(path) = cfg.kube_token_path() {
+        let reader = toolkit_transport_grpc::ServiceAccountTokenReader::new(path)
```
MikeFalcon77: This builds a second `ServiceAccountTokenReader` for the same `kube_token_path` that `build_internal_auth_interceptor` (line 549) already built one for — duplicating the initial file read, the 60s background refresh task, and the cached token.

Two readers poll the same file on independent timers and hold two separately cached values that can disagree right after a rotation, so the gRPC interceptor and the generated contract clients can present different tokens.

Build one reader inside the `if let Some(cfg) = internal_auth_cfg` block and derive both from it: the interceptor for `DirectoryGrpcClient::connect_with_interceptor`, and `InternalTokenProvider::new(reader.token_provider())` for `RunOptions`.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -745,6 +756,38 @@ async fn build_internal_authenticator(
...
+            .await
+            .context("failed to read projected service-account token for outbound credential")?;
+        return Ok(Some(InternalTokenProvider::new(reader.token_provider())));
```
MikeFalcon77: The refresh task behind this provider is spawned by `ServiceAccountTokenReader::new` with a bare `tokio::spawn` and is never tied to the bootstrap `cancel` token that governs the rest of the OoP lifecycle.

Its loop awaits `sleep(interval)` before re-checking its `Weak` handle, so after shutdown drops the last provider clone it can linger for up to a full refresh interval (60s) and re-read the token file during drain.

Add a cancellation-aware constructor whose loop is `tokio::select! { _ = sleep(interval) => {}, _ = token.cancelled() => break }` and pass the bootstrap `cancel` token in from here.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -1,7 +1,65 @@
...
+/// `InternalCredential`), never from the contract argument, because
+/// `PlatformSecurityContext` is a validated identity with no secret inside it.
+///
+/// The provider is invoked on every outgoing platform-plane call so a rotating
+/// credential (e.g. a projected Kubernetes `ServiceAccount` token re-read at the
+/// bootstrap layer) is always attached in its current form. Returning `None`
+/// attaches nothing — legitimate for Profile 1 / in-process
+/// (`InternalCredential::None`); the requirement is enforced server-side.
+#[derive(Clone)]
+pub struct InternalTokenProvider(Arc<dyn Fn() -> Option<SecretString> + Send + Sync>);
```
MikeFalcon77: `Option<SecretString>` gives the provider no way to distinguish "no credential is configured" from "a credential is configured but is currently unavailable", and no way to report an error at all.

`ServiceAccountTokenReader::token_provider` returns `None` when the token file is empty or the refresh loop has not yet populated it, so a transient failure is indistinguishable at the call site from an intentionally unauthenticated deployment. This is the root cause of the silent no-credential paths in `grpc::attach_internal_token` and the REST codegen — no call site can tell the two apart.

Use a three-state return, e.g. `Fn() -> Result<Option<SecretString>, CredentialError>` or an explicit `enum CredentialState { NotConfigured, Token(..), Unavailable(..) }`, and have the attach sites warn or fail on unavailable while staying permissive on not-configured.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -1,7 +1,65 @@
...
+
+impl InternalTokenProvider {
+    /// Build a provider whose credential is resolved by `provider` on each call
+    /// (supports rotation; `None` attaches no credential).
+    #[must_use]
+    pub fn new(provider: impl Fn() -> Option<SecretString> + Send + Sync + 'static) -> Self {
```
MikeFalcon77: This accepts an arbitrary synchronous `Fn` that generated clients call on every outbound request from inside async request-build paths, with no documented non-blocking requirement.

The stated use case is reading a rotating credential, which invites the obvious implementation — `std::fs::read_to_string`, a `std::sync::Mutex`, or a network fetch — any of which stalls a runtime worker thread once per platform-plane request. The one in-repo implementation avoids this only because it happens to read a `parking_lot::RwLock`-cached value refreshed in the background.

Document that the closure must not block, do I/O, or take a contended lock, and point at `ServiceAccountTokenReader::token_provider` as the reference pattern for rotating credentials.

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -1107,28 +1133,71 @@ fn build_query_string(method: &RestMethodModel, support: &TokenStream) -> TokenS
...
-    if let Some(ident) = ctx_ident {
-        quote! {
-            let __bearer: ::std::option::Option<::std::string::String> = #ident
-                .bearer_token()
-                .map(|__t| {
-                    use ::secrecy::ExposeSecret as _;
-                    __t.expose_secret().to_owned()
-                });
+    match secctx {
+        Some(p) if is_platform_security_context_type(&p.ty) => {
```
MikeFalcon77, line 1167: The platform plane is selected purely from the last path segment of the type name (`is_platform_security_context_type`), and the generated code only does `let _ = &#ident;` — nothing ever proves the argument is the real `toolkit_security::PlatformSecurityContext`.

The gRPC path constrains the same decision with a `SecurityContextMarker` bound, which this PR extends to `PlatformSecurityContext` for exactly this reason. On REST, any local type named `PlatformSecurityContext` compiles and silently switches a method off the tenant bearer.

Emit a compile-time assertion in this branch mirroring the gRPC guard, e.g. `const _: fn() = || { fn assert_platform<T: ::toolkit_security::PlatformPlaneMarker>() {} assert_platform::<#ty>(); };`, adding the sealed marker trait if there isn't one.

libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -468,13 +477,26 @@ fn generate_one_shot_unary_method(
     return_ty: &TokenStream,
     body_ident: &syn::Ident,
     ctx_ident: Option<&syn::Ident>,
+    is_platform: bool,
```
MikeFalcon77: Plane selection is threaded through three codegen functions as a bare `is_platform: bool` sitting next to a separate `ctx_ident: Option<&syn::Ident>`. The two can be passed inconsistently and the invariant is held only by the doc comment on `security_context_param` saying callers must derive both from the same param.

This boolean decides which credential the generated client attaches, and `Some(ctx) if is_platform` versus `Some(ctx)` is invisible at the three call sites that just pass `is_platform,`.

Replace the pair with one value: `enum AuthPlane<'a> { None, Tenant(&'a syn::Ident), Platform(&'a syn::Ident) }` returned by `security_context_param`, matched in the three generators.

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -790,15 +791,28 @@ fn generate_client_method(
...
+                    // credential as a sensitive `X-ToolKit-Internal-Token`
+                    // (never `Authorization`); at most one of the two is Some.
+                    // The platform credential is resolved fresh on every
+                    // attempt (not once before the retry loop) so a
+                    // rotated/refreshed token is used on retries.
                     let mut __builder = self.http.#verb_call(&__url);
                     if let Some(ref __t) = __bearer {
                         __builder = __builder.bearer_auth(__t);
                     }
+                    if let Some(__t) = __internal_token_provider
```
MikeFalcon77: This resolve-and-attach sequence is inlined here and again verbatim in the SSE reconnect factory (around line 960), rather than going through a runtime helper.

The gRPC generator routes both its unary and streaming paths through one audited `#support::grpc::attach_internal_token`. The REST side now has two hand-maintained copies of a security-critical snippet, with no single place to add instrumentation or handle a credential that is configured but unavailable.

Add `toolkit_contract::runtime::http::attach_internal_token(builder, provider)` (or a `RequestBuilder::internal_token_from(provider)`) and emit one call to it from both `__attempt` and the SSE factory.

libs/toolkit/src/runtime/runner.rs
```diff
@@ -128,6 +128,12 @@ pub struct RunOptions {
     /// See `HostRuntime::with_shutdown_deadline` for details on the relationship
     /// with `WithLifecycle::stop_timeout`.
     pub shutdown_deadline: Option<std::time::Duration>,
+    /// Process-wide platform-plane credential source (the selected
+    /// `InternalCredential`), threaded into every [`GearCtx`] so
+    /// `#[toolkit::provides]`-generated clients attach `X-ToolKit-Internal-Token`
+    /// on platform-plane (`PlatformSecurityContext`) methods. `None` (Profile 1
+    /// / in-process, or no credential configured) attaches nothing.
+    pub internal_token_provider: Option<toolkit_contract::runtime::config::InternalTokenProvider>,
```
MikeFalcon77, line 143: `RunOptions` has all-public fields and no `#[non_exhaustive]`, no `Default`, and no builder, so adding this field breaks every struct-literal construction site — 16 test sites plus both bootstrap paths in this PR — and breaks any out-of-repo consumer with no deprecation path.

`RunOptions` is exported from `libs/toolkit/src/lib.rs`, and the sibling `OopRunOptions` already carries a `Default` impl for this reason. The next optional field will cause the same churn.

Mark `RunOptions` `#[non_exhaustive]` and add `RunOptions::new(gears_cfg, db, shutdown, instance_id)` plus `with_*` setters, then migrate the bootstrap paths and tests.

libs/toolkit-http/src/request.rs
```diff
@@ -201,6 +201,38 @@ impl RequestBuilder {
...
+    ///
+    /// The token is carried raw (no `Bearer` scheme) and **never** on
+    /// `Authorization`, to avoid colliding with the tenant-plane user JWT
+    /// (`cpt-cf-adr-two-plane-auth`). The header value is marked sensitive
+    /// (`HeaderValue::set_sensitive(true)`) so header-logging / redaction layers
+    /// never emit it — the same guarantee as
+    /// [`crate::security::attach_internal_token_http`]. Prefer this over
+    /// `.header(..)` for the internal credential, which cannot mark the value
+    /// sensitive.
+    pub fn internal_token_auth(mut self, token: &str) -> Self {
```
MikeFalcon77: This takes the credential as `&str`, forcing every caller to call `expose_secret()` before the transport boundary — the generated REST client does exactly that at two separate sites. The sibling helper the doc comment cites as equivalent, `security::attach_internal_token_http`, takes `&SecretString` and exposes it internally.

`toolkit-http` already depends on `secrecy`, so the weaker type buys nothing and widens the window in which a bare `&str` view of the internal token exists in caller frames.

Change the signature to `&secrecy::SecretString` and call `expose_secret()` inside, then drop the `ExposeSecret` import and the two `.expose_secret()` calls from the generated snippets in `rest_contract.rs`.

libs/toolkit/src/context.rs
```diff
@@ -118,11 +130,34 @@ impl GearCtx {
...
+    }
+
+    /// The process-wide platform-plane credential source, if any.
+    ///
+    /// Returned by clone so `#[toolkit::provides]` can move it onto a generated
+    /// client's [`ClientConfig`](toolkit_contract::runtime::config::ClientConfig).
+    /// `None` (Profile 1 / in-process) means platform-plane calls attach no
+    /// credential.
+    #[must_use]
+    pub fn internal_token_provider(&self) -> Option<InternalTokenProvider> {
```
MikeFalcon77, line 176: This is a fully public, non-hidden accessor that hands any gear or plugin holding a `GearCtx` a clone of the process-wide credential source, and `InternalTokenProvider::current()` returns the raw `SecretString`.

The two-plane design says the platform credential is injected below the contract layer and is not available to contract-level code. This getter lets arbitrary co-hosted gear code — including third-party gears — read the rotating credential and forge platform-plane requests. The credential grants access to every system service.

Mark it `#[doc(hidden)]` and document it as wiring-only for `#[toolkit::provides]`, or replace it with a narrow decorator such as `ctx.apply_platform_credential(cfg)` that never yields the `SecretString`.

libs/toolkit-contract/tests/rest_client_platform_secctx.rs
```diff
@@ -0,0 +1,165 @@
...
+    let client = SysApiRestClient::new(cfg).unwrap();
+
+    let first = SysApi::whoami(&client, &platform_ctx(), "a".to_owned())
+        .await
+        .unwrap();
+    let second = SysApi::whoami(&client, &platform_ctx(), "b".to_owned())
+        .await
+        .unwrap();
+
+    assert_eq!(first.internal_token.as_deref(), Some("token-0"));
```
MikeFalcon77: Asserting the exact values `token-0` and `token-1` encodes the assumption that the provider closure is invoked exactly once per request. An added retry, redirect, reconnect, or a second `current()` call in the request builder shifts every observed value and fails this test even though rotation still works.

Make the rotation explicit rather than counting invocations: back the provider with an `Arc<Mutex<String>>`, mutate it between the two calls, and assert the first request saw the pre-rotation value and the second the post-rotation one.

libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -686,3 +749,139 @@ fn generate_projection_impl(model: &GrpcContractModel) -> TokenStream {
...
+    fn platform_plane_method_attaches_internal_token_not_bearer() {
+        let out = expand(quote! {
+            pub trait FooApiGrpc: FooApi {
+                async fn get_thing(&self, ctx: PlatformSecurityContext, id: String) -> Result<Resp, Err>;
+            }
+        });
+        assert!(out.contains("attach_internal_token"), "got:\n{out}");
+        assert!(!out.contains("attach_bearer"), "got:\n{out}");
+        // The off-wire marker is consumed, not forwarded as a credential.
+        assert!(out.contains("let _ = & ctx"), "got:\n{out}");
```
MikeFalcon77: `out.contains("let _ = & ctx")` asserts on `TokenStream::to_string()` whitespace. The spacing (`& ctx` vs `&ctx`) is a proc-macro2 pretty-printing detail and `ctx` is an arbitrary param name, so a proc-macro2 bump or a rename breaks this with no behavior change.

The surrounding `contains("attach_internal_token")` assertions have the same weakness in the other direction — they prove an identifier appears somewhere in the expansion, not that the call is wired to `self.config.internal_token_provider` or `?`-propagated. Changing the emitted `attach_internal_token(...)?;` to `let _ = attach_internal_token(...);` keeps them green.

Parse the expansion with `syn::parse2::<syn::File>` and assert on the located statement, or drop these in favour of the runtime metadata check the gRPC test should be doing.

libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -686,3 +749,139 @@ fn generate_projection_impl(model: &GrpcContractModel) -> TokenStream {
...
+        let item: syn::ItemTrait = syn::parse2(tokens).unwrap();
+        grpc_contract_parse::parse(attr, item).unwrap()
+    }
+
+    fn expand(tokens: TokenStream) -> String {
+        generate(&build_model(tokens)).to_string()
+    }
+
+    #[test]
+    fn tenant_plane_method_attaches_bearer_not_internal_token() {
```
MikeFalcon77, line 846: Every test in this module uses a single-method trait, so nothing covers a trait that mixes a tenant `SecurityContext` method and a platform `PlatformSecurityContext` method.

The mutually exclusive `contains` / `!contains` style only works on single-method traits, so cross-method plane leakage — plane classification derived from the wrong method, the most likely real defect here — cannot be detected.

Add a test whose trait declares both, parse the expansion, locate each generated body by ident, and assert the tenant body contains only `attach_bearer` and the platform body only `attach_internal_token`.

libs/toolkit-http/src/request.rs
```diff
@@ -201,6 +201,38 @@ impl RequestBuilder {
...
+                    http::header::HeaderName::from_static(
+                        toolkit_security::constants::INTERNAL_TOKEN_HEADER,
+                    ),
+                    value,
+                ));
+            }
+            Err(e) => {
+                self.error = Some(HttpError::InvalidHeaderValue(e));
+            }
+        }
```
MikeFalcon77, line 231: `internal_token_auth` has no test — not the header name and value, not the `set_sensitive(true)` guarantee, and not this error branch.

`bearer_auth` has `bearer_auth_sets_sensitive_authorization_header`, which explicitly asserts `value.is_sensitive()`. Dropping that call here would leak the platform credential into header-logging layers with no failing test. This branch matters too: a malformed token must fail the request rather than send it unauthenticated.

Mirror the `bearer_auth` test for `internal_token_auth`, and add a case with a token containing `\n` asserting the builder records `HttpError::InvalidHeaderValue` and the request is not sent.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -745,6 +756,38 @@ async fn build_internal_authenticator(
...
+/// `DirectoryService` calls, but yields a transport-agnostic
+/// [`InternalTokenProvider`](toolkit_contract::runtime::config::InternalTokenProvider):
+///
+/// - `shared_secret` → a static provider yielding the configured secret.
+/// - `kube` with a `token_path` → a rotating provider backed by a
+///   [`ServiceAccountTokenReader`](toolkit_transport_grpc::ServiceAccountTokenReader)
+///   (re-reads the projected token on its refresh cadence).
+/// - `kube` without a `token_path` → `None` (inbound-only: this participant
+///   validates but attaches no outbound credential).
+async fn build_internal_token_provider(
```
MikeFalcon77: No tests for any of this function's four outcomes: shared secret, kube with a token path, kube without one, and an unreadable token file.

It decides whether an out-of-process gear attaches any platform credential at all, and every wrong answer is a silent `Ok(None)` — outbound platform-plane calls go out uncredentialed with nothing failing.

Add unit tests over `InternalAuthConfig`: shared secret yields a provider whose `current()` returns the secret; `kube` pointed at a temp file yields one returning the file contents; `kube` without `token_path` yields `None`; a missing file returns the contextualized `Err`.

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -790,15 +791,28 @@ fn generate_client_method(
...
+                        .as_ref()
+                        .and_then(#support::runtime::config::InternalTokenProvider::current)
+                    {
+                        use ::secrecy::ExposeSecret as _;
+                        __builder = __builder.internal_token_auth(__t.expose_secret());
```
MikeFalcon77: The per-attempt freshness guarantee documented right here — `InternalTokenProvider::current` re-resolved inside `__attempt` on every retry — has no test. Hoisting the resolution above the retry loop would silently lose it.

Add an integration test with a `#[retryable] #[idempotency(Idempotent)]` platform-plane method, a handler returning 503 then 200, and a rotating provider, asserting the two requests carried different tokens.

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -934,10 +948,22 @@ fn generate_streaming_method_body(
...
+                // started) so a rotated/refreshed token is used on reconnects.
                 if let Some(ref __t) = __bearer {
                     __builder = __builder.bearer_auth(__t);
                 }
+                if let Some(__t) = __internal_token_provider
+                    .as_ref()
+                    .and_then(#support::runtime::config::InternalTokenProvider::current)
+                {
+                    use ::secrecy::ExposeSecret as _;
+                    __builder = __builder.internal_token_auth(__t.expose_secret());
```
MikeFalcon77: The SSE reconnect branch has no test at any level — no streaming platform-plane method is exercised in this PR. It is a separate code path from the unary attempt closure, so an unauthenticated or stale-token reconnect would be caught by nothing.

Add an SSE test with a `#[streaming]` platform-plane method, non-zero `sse_reconnect.max_attempts`, and a handler recording the `X-ToolKit-Internal-Token` of both the initial request and the reconnect.

libs/toolkit-contract-macros/src/grpc_contract_parse.rs
```diff
@@ -370,3 +390,144 @@ impl GrpcIdempotency {
...
+
+    #[test]
+    fn accepts_platform_security_context_param() {
+        let model = parse_trait(quote! {
+            pub trait FooApiGrpc: FooApi {
+                async fn do_it(&self, ctx: PlatformSecurityContext, id: String) -> Result<Resp, Err>;
+            }
+        })
+        .expect("a single platform security-context param is valid");
+        assert_eq!(model.methods[0].params.len(), 2);
```
MikeFalcon77, line 459: `accepts_platform_security_context_param` differs from `accepts_single_security_context_param` only in the context type, yet makes the identical assertion `params.len() == 2`. Counting params cannot tell a `PlatformSecurityContext` recognized as a plane marker from one treated as an ordinary wire param, so this passes even if platform detection is broken.

Assert the classification instead: `assert!(crate::projection::is_platform_security_context_type(&model.methods[0].params[0].ty))`, and that `body_param_ident` selects `id`.

libs/toolkit-contract-macros/src/grpc_contract_parse.rs
mattgarmon: Fixed. Now delegates to `is_security_context_type`

libs/toolkit-contract-macros/src/rest_contract_parse.rs
mattgarmon: Fixed. `generate_route_registration` picks the tenant auth axis from the plane marker

libs/toolkit-contract-macros/src/grpc_contract.rs
mattgarmon: Fixed the silent gap: `require_tls` is now honored in gRPC `connect`

libs/toolkit-contract-macros/src/provides.rs
mattgarmon: Fixed. The provider is threaded through

libs/toolkit/src/bootstrap/run.rs
mattgarmon: Addressed with a comment + no provider: The in-process host resolves gear->gear deps as local trait objects (no transport), so no outbound platform header is emitted and no credential is needed

libs/toolkit-contract-macros/src/rest_contract.rs
mattgarmon: Fixed. The platform branch now emits a compile-time by-value type-equality assertion against `toolkit_security::PlatformSecurityContext`, mirroring the gRPC `SecurityContextMarker` guard

libs/toolkit-http/src/request.rs
mattgarmon: Fixed

libs/toolkit/src/context.rs
mattgarmon: Fixed. `internal_token_provider()` is now `#[doc(hidden)]` and documented as wiring-only for `#[toolkit::provides]`

libs/toolkit-contract-macros/src/provides.rs
mattgarmon: Fixed, `require_tls` is now used

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Fixed

libs/toolkit-contract/tests/grpc_client_platform_secctx.rs
mattgarmon: Fixed in test

libs/toolkit-contract/tests/rest_client_platform_secctx.rs
mattgarmon: Fixed

libs/toolkit-contract/src/grpc/mod.rs
mattgarmon: Added

libs/toolkit-http/src/request.rs
mattgarmon: Fixed. Added `INTERNAL_TOKEN_HEADER` to the redirect layer's `SENSITIVE_HEADERS`

libs/toolkit-contract/src/grpc/mod.rs
mattgarmon: Fixed via the three-state `CredentialState`

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Fixed. `build_platform_credentials` matches `InternalAuthConfig` exhaustively

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Fixed. Added `ServiceAccountTokenReader::with_cancellation`

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Added three-state `CredentialState`

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Documented

libs/toolkit-contract-macros/src/grpc_contract.rs
mattgarmon: Fixed. Added enum

libs/toolkit-contract-macros/src/rest_contract.rs
mattgarmon: Fixed. Added `toolkit_contract::runtime::http::attach_internal_token`

libs/toolkit/src/runtime/runner.rs
mattgarmon: Fixed

libs/toolkit-contract/tests/rest_client_platform_secctx.rs
mattgarmon: Done

libs/toolkit-contract-macros/src/grpc_contract.rs
mattgarmon: Fixed

libs/toolkit-contract-macros/src/grpc_contract.rs
mattgarmon: Added

libs/toolkit-http/src/request.rs
mattgarmon: Added

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Added tests for all four outcomes

libs/toolkit-contract-macros/src/rest_contract.rs
mattgarmon: Added test

libs/toolkit-contract-macros/src/rest_contract.rs
mattgarmon: Added test

libs/toolkit-contract-macros/src/grpc_contract_parse.rs
mattgarmon: Fixed

libs/toolkit-contract/tests/rest_client_platform_secctx.rs
```diff
@@ -0,0 +1,466 @@
...
+        ctx: &PlatformSecurityContext,
+        id: String,
+    ) -> Result<HeaderEcho, DemoError>;
+
+    /// Tenant-plane sibling: must forward the caller's bearer token on
+    /// `Authorization` and must NOT attach the platform internal token, even
+    /// when a provider is configured on the client.
+    #[idempotency(SafeRead)]
+    async fn whoami_tenant(&self, ctx: &SecurityContext, id: String)
+    -> Result<HeaderEcho, DemoError>;
```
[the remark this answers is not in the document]
mattgarmon: Done

libs/toolkit/src/discovery.rs
```diff
@@ -178,10 +178,21 @@ pub struct ConsumerRegistration {
...
 
+/// The wiring function pointer emitted by `#[toolkit::consumes]`. Registers the
+/// client into the `ClientHub`, given the endpoint resolver and the process's
+/// platform-plane credential source (threaded onto a directory-resolving remote
+/// client). Aliased to keep the fn-pointer type readable.
+pub type WireFn = fn(
+    &ClientHub,
+    Arc<dyn EndpointResolver>,
+    Option<&toolkit_contract::runtime::config::InternalTokenProvider>,
+) -> anyhow::Result<WireOutcome>;
```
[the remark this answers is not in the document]
mattgarmon: Updated doc

libs/toolkit/src/bootstrap/oop_tests.rs
```diff
@@ -879,3 +879,93 @@ mod advertise_uri {
...
+            "inbound-only kube participant must attach no outbound credential"
+        );
+    }
+
+    #[tokio::test]
+    async fn kube_with_missing_token_file_is_an_error() {
+        let cfg = InternalAuthConfig::Kube {
+            audiences: vec!["toolkit-internal".to_owned()],
+            token_path: Some(std::path::PathBuf::from("/nonexistent/toolkit/token")),
+        };
```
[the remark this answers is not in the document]
mattgarmon: Fixed

libs/toolkit-contract-macros/src/grpc_contract.rs
MikeFalcon77: The `require_tls` check added here does not produce a TLS connection — it only compares the URL string, and the channel is still built without TLS.

`toolkit-contract` depends on `tonic = { features = ["transport"] }`. In tonic 0.14 all TLS support sits behind the internal `_tls-any` feature, enabled only by `tls-ring` / `tls-aws-lc` / the `tls-*-roots` features. Nothing in the workspace turns any of those on — `cargo tree -e features` resolves tonic to exactly `channel`, `codegen`, `default`, `router`, `server`, `transport`.

With `_tls-any` off, both the scheme handling and the TLS wrapping are compiled out:

```rust
// tonic-0.14.5/src/transport/channel/service/connector.rs
#[cfg(feature = "_tls-any")]
let is_https = uri.scheme_str() == Some("https");
...
    #[cfg(feature = "_tls-any")]
    if is_https { /* wrap in TLS */ }
```

and the auto-TLS shortcut in `Endpoint::new` is gated the same way. So `require_tls = true` with an `https://` endpoint passes this check and then does a plaintext TCP connect and speaks h2c. The internal token still goes out in cleartext, but now behind a setting that says it won't — which is worse than the previous state, where at least nothing claimed otherwise.

Enable a TLS feature on the tonic dependency (`tls-ring` plus a roots feature) and apply `ClientTlsConfig` on the endpoint when `require_tls` is set. If pulling TLS into this crate isn't wanted right now, make the check fail closed instead: return an error when `require_tls` is set at all, so the setting can't be silently satisfied by a connection that isn't encrypted.

The scheme string match is also loose next to `Endpoint::from_shared`, which parses the URI — `HTTPS://host` and a URI with leading whitespace parse as https but fail `starts_with("https://")`. Parse once and compare `uri.scheme()`.

PR #4605, opened 2026-08-19
docs(secure-orm): ADR-0002 safe property-graph (SQL/PGQ) policy

MikeFalcon77:
## Summary
- Extends ADR-0001's invariant ("every access to a table passes through a scope condition") to PostgreSQL 19's `GRAPH_TABLE`/`MATCH`, since a graph query is evaluated independently of the outer `WHERE` just like a CTE body.
- Two-layer split: a security-agnostic `toolkit-sea-orm-pgq` crate for the AST/DDL/renderer, `toolkit-db::secure::pgq` as the security boundary. Entry only from `SecureSelect<E, Scoped>::with_graph()`, mirroring `with_ctes()`.
- New policies beyond the CTE precedent: no label-only matching (a PG label can span multiple element tables, so it has no unique security mapping); every element table must resolve a scope property (today's `no_tenant` closure/membership tables are fail-closed to `WHERE false`, which is safe but silently empty — not eligible graph elements); DDL and query AST derive from one Rust declaration.
- Records an open feasibility question (whether an element-pattern `WHERE` may contain a subquery, needed for `InTenantSubtree`/`InGroup*` scope shapes) to settle before the API is pinned, and states plainly that PG19's fixed-depth-only patterns do not replace `recursive_cte`.
- No code changes — policy + test contract for the future implementation PR.

## Summary by CodeRabbit

* **Documentation**
  * Added an accepted architecture decision documenting secure property-graph query policies.
  * Defined scope enforcement, supported graph queries, execution safeguards, database compatibility requirements, and migration behavior.
  * Documented testing expectations, known limitations, fallback approaches, and restrictions on direct graph-query SQL.

### DIFF
--- a/docs/arch/secure-orm/ADR/0002-secure-property-graph-policy.md
+++ b/docs/arch/secure-orm/ADR/0002-secure-property-graph-policy.md
@@ -1,0 +1,730 @@
+---
+status: accepted
+date: 2026-08-18
+decision-makers: Constructor Fabric steering committee
+---
+
+# ADR-0002: Safe Property-Graph Queries (SQL/PGQ) in the Secure ORM
+
+**ID**: `cpt-cf-adr-secure-property-graph-policy`
+
+## Context and Problem Statement
+
+[ADR-0001](./0001-secure-cte-policy.md) settled the Secure ORM's load-bearing invariant —
+**every access to a table passes through a scope condition** — and, for constructs that are
+evaluated *independently* of the outer query, the rule that follows from it: **do not apply
+scope around the construct, embed it inside**. That is why
+[cte.rs](../../../../libs/toolkit-db/src/secure/cte.rs) scopes every CTE body rather than
+the outer `WHERE`, and why the mechanism is structural — `SecureCteSelect` is reachable only
+from `SecureSelect<E, Scoped>`, captures that query's `Arc<AccessScope>`, and re-scopes every
+body from it
+([cte.rs:229-249](../../../../libs/toolkit-db/src/secure/cte.rs#L229-L249)), so
+a mixed-scope query is unrepresentable rather than rejected at runtime.
+
+PostgreSQL 19 adds SQL/PGQ (ISO/IEC 9075-16): `CREATE PROPERTY GRAPH` declares a graph as
+"a kind of read-only view over relational tables", and `GRAPH_TABLE (g MATCH … COLUMNS (…))`
+queries it. The documentation is explicit that `GRAPH_TABLE` "acts like a table function in
+that it produces a computed table as output", usable "anywhere you would use a subquery or
+table", joinable, filterable and orderable from outside.
+
+**This is the ADR-0001 hazard class exactly.** The outer query's scope predicate does not
+reach inside `MATCH`:
+
+```sql
+-- UNSAFE: the traversal saw every tenant's rows; the outer WHERE only trims the output.
+SELECT * FROM GRAPH_TABLE (
+    mygraph MATCH (a IS resource)-[e IS depends_on]->(b IS resource)
+            COLUMNS (a.id AS src, b.id AS dst)
+) x
+WHERE x.tenant_id = $1;
+```
+
+```sql
+-- SAFE: every element pattern carries the scope predicate, so no element ever
+-- held another tenant's rows.
+SELECT * FROM GRAPH_TABLE (
+    mygraph MATCH (a IS resource   WHERE a.tenant_id = $1)
+                 -[e IS depends_on WHERE e.tenant_id = $1]->
+                  (b IS resource   WHERE b.tenant_id = $1)
+            COLUMNS (a.id AS src, b.id AS dst)
+);
+```
+
+PostgreSQL permits `WHERE` inside an element pattern — `(a IS account WHERE a.type =
+'savings')` — and that is the mechanism which makes the safe form expressible at all. It is
+the direct analogue of embedding scope in a CTE body.
+
+**Scope of this ADR.** As with ADR-0001, this policy governs **standard user gears** — the
+handlers / services / repos that query their own domain tables. It does not govern
+`toolkit-db` itself, nor migration infrastructure, where sea_query and hand-written SQL are
+the implementation substrate.
+
+**Why now, before there is a consumer.** Nothing in the repository uses SQL/PGQ today; a
+repo-wide search for `GRAPH_TABLE` / `pgq` / property-graph vocabulary returns only
+considered-and-rejected mentions of graph *databases*
+([authorization/DESIGN.md:156](../../authorization/DESIGN.md#L156),
+`gears/chat-engine/docs/ADR/0001-message-tree-structure.md:27`). But there is one tracked,
+first-class requirement: IRM's `cpt-cf-infrastructure-resource-manager-fr-graph-query`
+(`p2`, Planned —
+[DESIGN.md:1345](../../../../gears/infrastructure-resource-manager/docs/DESIGN.md#L1345),
+[PRD.md:843](../../../../gears/infrastructure-resource-manager/docs/PRD.md#L843)) over a
+typed-edge table `resource_relationships`
+([DESIGN.md:1107](../../../../gears/infrastructure-resource-manager/docs/DESIGN.md#L1107))
+and a `resource_closure` read model, at a stated scale of 1M+ nodes and 5M+ edges. IRM is
+still documentation only — zero `.rs` files — which is precisely the moment to fix the
+policy, for the same reason ADR-0001 was written before ad-hoc CTEs accreted in gears.
+
+**PostgreSQL 19 is pre-GA at the date of this ADR.** Beta 1 was released 2026-06-04, Beta 2
+on 2026-07-16 with the feature set frozen, and GA is expected September/October 2026. This
+ADR is therefore accepted as policy and as the target API, with the implementation gated
+behind an opt-in feature flag (see "Backend gating") and required to be re-validated at GA.
+
+## Decision Drivers
+
+- **Preserve the invariant.** A graph query must be as impossible to construct unscoped as
+  an `Unscoped` select is impossible to execute
+  ([select.rs:151-188](../../../../libs/toolkit-db/src/secure/select.rs#L151-L188)).
+- **No raw SQL in user-gear code** — the rule in
+  [11_database_patterns.md:9](../../../toolkit_unified_system/11_database_patterns.md#L9)
+  applies verbatim to `GRAPH_TABLE` and `CREATE PROPERTY GRAPH` strings.
+- **One scope compiler, not two.** `build_scope_condition`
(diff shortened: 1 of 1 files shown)


vasylcf:
Your open feasibility question — whether an element-pattern WHERE may contain a subquery — has an answer, and a third fallback that beats both of the two you list. Measured on PG19 beta2 against a property graph over real tables, not reasoned about.
 
  No, it may not. Both shapes are rejected outright:
 
  ERROR:  subqueries within GRAPH_TABLE reference are not supported
 
  That covers col IN (SELECT …) and = ANY(ARRAY(SELECT …)). CROSS JOIN LATERAL GRAPH_TABLE (…) is also refused, with a bare syntax error — LATERAL is not accepted before the construct. So InGroup, InGroupSubtree and InTenantSubtree cannot be injected into a pattern predicate in any form.
 
  But the closure does not have to be pre-resolved or rejected. A comma join with a correlated reference — an implicit lateral — is accepted, and reaches the pattern:
 
  WITH closure(descendant_id) AS (SELECT descendant_id FROM tenant_closure WHERE ancestor_id = $1)
  SELECT g.n FROM closure c,
    GRAPH_TABLE (kb MATCH (a IS node)-[e IS edge]->(b IS node)
      WHERE a.id = $2 AND a.tenant_id = c.descendant_id AND b.tenant_id = c.descendant_id
      COLUMNS (b.id AS n)) g;
 
  The closure query stays in the same statement, so there is no extra round trip and no unbounded IN list — your fallback 2 without its two costs. Plan is nested loops over index scans, 0.337 ms on a 199k-node fixture. With a three-tenant closure the semantics are right per tenant and rows do not multiply: each match pairs with exactly one closure row because the tenant predicate pins it.
 
  One caveat, since it will bite later. Correlating turns a semi-join into a join, so if the sibling can carry duplicate keys the result multiplies — a resource authorized through two groups appears twice in a membership list. The sibling needs to be distinct on the correlated column. For tenant_closure descendants that holds alreadip it does not.
 
  This has a consequence for the scope-addressing design. w a resolved column is addressed — for_table() versusfor_graph_element(v) — and that is exactly right for Eq and In. The subquery-producing shapes need something the diagram does not have: the
  compiler must be able to emit a FROM item plus a correlaERE predicate. Worth deciding before the API is pinned,because it changes what build_scope_condition returns rather than only how it renders.
 
  Separately, on PathBuilder. .vertex().edge().to() does not appear to carry direction. On our schema the undirected shorthand (a)-[e]-(b) plans
  as a parallel sequential scan of the edge table — 734.9 wo directed patterns unioned, same ten rows. If the AST can express an undirected element, it ships that. We ended up with no undirected variant at all and build an undirected hop as two patterns.
 
  Context for where these numbers come from: we have a working SQL/PGQ traversal on a dev stand, reaching GRAPH_TABLE through TableRef::FunctionCall + Func::Custom + Expr::cust_with_er than a design, and strictly worse than the typed AST you are proposing. Offering it only as a test oracle: it emits statements PG19 accepts today, so your renderer can be diffed against it. Branch is vasylcf/gears-rust:feature/graph-storage-pgq.

vasylcf:
Probed the risky parts against PG19 beta2 — branch vasylcf/gears-rust:feature/pgq-adr0002-probe, results and scripts under gears/graph-storage/dev/adr0002-probe/. Four things that bear on the API before it is pinned.
 
  The addressing change works, but its signature needs a Result. I implemented ColumnAddress::{Table, GraphElement} in cond.rs: table addressing reproduces the current compiler's SQL exactly, and the graph-addressed predicate executed verbatim inside an element pattern — right rows for the owner, none for a foreign tenant. But the three subquery arms have to fail in graph mode, and fail loudly: dropping a filter is fail-closed in the letter, yet the constraint vanishes and the scope becomes WHERE false — the silent empty traversal Policy 2 exists to prevent. So -> Result<Condition, _>, which changes the shape of the function rather than its internals.
 
  Policy 2 belongs where you put it. An entity resolving no property compiles identically under for_tenant(t) and under deny_all() — the failure is inside resolve_property, not in a filter arm, so the compiler cannot tell them apart. Eligibility has to be checked before compiling.
 
  Direction in PathBuilder, and a correction to what we told you. We said multi-hop path patterns explode, citing our spike. On this schema that is wrong for directed patterns — no penalty against chained one-hop at depth 2 or 3. What explodes is undirected elements, and it compounds:
 
  | pattern                | rows | time     |
  |------------------------|------|----------|
  | one undirected element | 10   | 735 ms   |
  | two undirected         | 83   | 7 967 ms |
  | the same two, directed | 83   | ~1.5 ms  |
 
  The sibling gap decides whether subtree scopes are servable. A pattern can correlate against a plain table in the same FROM, so with_graph() could let a pattern reference E's own columns — seeding a traversal from a scoped entity query with no new construct. But an arbitrary sibling has no room, and that is what a tenant-subtree scope needs: the closure is neither E nor a graph element, and cannot be the outer entity because a scoped select over a closure table is WHERE false by the point above. Since a pattern cannot hold a subquery, whether the API can hold a sibling is what decides between servable and fallback 1. Worth settling with the open question, not after it.
 
  One non-finding, in case it is useful: embedding scope in every element plans identically to the top-level form and runs in the same time, so there is no performance objection to the stricter rule.

vasylcf (approved):
Reviewed against a live PG19 beta2 property graph — probe branch vasylcf/gears-rust:feature/pgq-adr0002-probe, scripts and results under gears/graph-storage/dev/adr0002-probe/ — and against a working gear built on this shape.

Settles it for us, no objections:

- Option A and the "embed scope into every element, vertices and edges alike" principle. This is what we implemented; it holds.
- Policy 2, including where you put the eligibility check. Confirmed: an entity that resolves no property compiles identically under for_tenant(t) and under deny_all(), so the compiler cannot tell them apart after the fact — the check has to be before compiling, as you have it.
- Policy 3, and it is stronger than the ADR states: a column absent from the DDL's PROPERTIES list is invisible to MATCH — not an error, just silently unfilterable. Generating both sides from one declaration is the only thing that stays correct.
- Level C binding immediately, and pgq implying pg.

Three edits needed before the API is pinned:

1. Result on the addressing function. In graph mode the three subquery arms must fail loudly, not drop. A dropped constraint leaves deny_all() → WHERE false → exactly the silent empty traversal Policy 2 exists to prevent. This changes the function's shape, not its internals, so it cannot be left to the implementation.
2. Room for a sibling FROM item in with_graph(). Your open question is answered: no subquery is accepted inside a pattern in any form — IN (SELECT …), = ANY(ARRAY(SELECT …)), and LATERAL before GRAPH_TABLE are all rejected. But fallback 1 is not forced. A comma join with a correlated reference does reach the pattern and works (0.337 ms against a derived table). Without a place for that second FROM item, subtree scopes are permanently rejected rather than served.
3. Always emit an explicit arrow — do not offer an undirected form. Measured: undirected shorthand is 734.9 ms for one element and 7967 ms for two, against ~1.5 ms directed. Worth a line in "Feasibility constraints" so the cost is not rediscovered.

Approving the policy — the three items are text edits to this ADR, not objections to the decision.

Artifizer (approved):

MikeFalcon77:
@vasylcf thanks — the probe answered the open question and all three edits were right. Pushed in dd27f1ac.

**1. `Result` on the addressing function.** Agreed, and for the reason you give: dropping a filter is only fail-closed in the letter. The remaining constraints collapse to `deny_all()` and the traversal returns nothing, which is the same silent empty result Policy 2 exists to prevent, just reached a different way. Signature is now pinned as `-> Result<ScopePredicate, ScopeError>`, where `ScopePredicate` is a `Condition` plus the optional `FROM` item it references. Added the test that table addressing stays byte-identical to today's SQL, since that is the thing that could quietly break the existing select path.

**2. Sibling `FROM` item.** This was the useful part. The open question is now recorded as settled — no subquery in a pattern in any form, including `= ANY(ARRAY(SELECT …))` and `LATERAL` before `GRAPH_TABLE` — and the correlated comma join is recorded as the shape that makes subtree scopes servable rather than rejected. Your framing that this changes what the compiler *returns*, not how it renders, is why it is in the ADR text and not left to the implementation. `with_graph()` now has an explicit requirement to leave room for that second `FROM` item.

Your duplicate caveat is in too, as an obligation with a test. Worth noting it bites us in one place already: `tenant_closure` descendants are distinct per ancestor, but `resource_group_membership` resource ids are not, so the `InGroup` / `InGroupSubtree` siblings need an explicit `DISTINCT`. Good catch — that one would have shown up as a duplicated row in a listing, not as an error.

**3. Explicit direction only.** In, with your numbers. `PathBuilder` has no undirected variant at all; an undirected hop is two directed patterns. The correction on multi-hop is recorded as well, so nobody re-derives the wrong conclusion from the earlier note.

Also folded in two smaller things from your review: Policy 3 now says a column left out of the `PROPERTIES` list is silently unfilterable rather than an error, and the "embedding scope in every element costs nothing measurable" result is recorded so the stricter rule does not get re-argued on performance grounds later.

One thing I did not take: the measurements are all beta 2, so they are attributed to your probe branch and flagged for re-confirmation at GA, and fallbacks 1 and 2 stay in the text as what we do if the sibling shape regresses.

On the test oracle — yes, that is useful, and more useful than a comparison implementation would be. When the renderer lands I will diff its output against what your branch emits, since those statements are known to be accepted by a real server. Will ping you on the implementation PR.

MikeFalcon77:
Second pass over your comments — `dc9d2108`. Four things from them that the first commit missed:

**The two sibling shapes are now distinguished.** Your point that a pattern can correlate against `E`'s own columns is a different capability from the arbitrary-sibling one, and I had collapsed them. It is in as its own item, along with the argument you gave for why the closure cannot simply be promoted to the anchor — a scoped select over `tenant_closure` is `WHERE false` by Policy 2, so it is neither `E` nor a graph element and has nowhere else to go. That argument is what makes the sibling requirement non-negotiable rather than a convenience, so it belongs in the text. It also partly redeems Option A's forced anchor, which is now noted in its cons list.

**Policy 2: the check runs before compiling.** Your observation that a no-property entity compiles identically to a legitimate `deny_all()` forecloses the obvious implementation shortcut, so I wrote it in explicitly — a post-hoc "is this deny-all?" check either misses the bug or rejects correct deny-all queries. The ADR previously said only "not silent", which left that shortcut open.

**The `Func::Custom` route is recorded as Level C.** Worth having in writing that `GRAPH_TABLE` is reachable through sea_query today via `TableRef::FunctionCall` + `Func::Custom` + `Expr::cust_*` — not as a competing design, but because it means the dylint rule has to match the construct inside `cust`-family arguments, not just obvious raw-SQL sinks. I would not have thought to scope the rule that way.

**Your traversal is in the test contract as a renderer oracle.** Diffing against statements a real server already accepts separates a renderer bug from a policy bug, which are indistinguishable from the outside. I also noted in Confirmation which parts your probe already validated — graph-addressed predicate executing verbatim, table addressing reproducing current SQL, the correlated sibling — while keeping the section's "not done" framing, since that covers the mechanism and not the guarantee.

That should be everything from the three comments. If I have still dropped something, say which and I will fold it in.

PR #4608, opened 2026-08-19
feat(toolkit): canonical-error-aware Json/Query/Path extractors + middleware fallback

### LINKED ISSUE #4596: [toolkit][api] axum Json<T> extraction failures bypass CanonicalError/RFC 7807 entirely
## Problem Statement

`toolkit::api::canonical_prelude::Json` (`libs/toolkit/src/api/mod.rs:51`) is a bare
re-export of `axum::Json`. When it fails to extract (malformed JSON, an unknown field
under `#[serde(deny_unknown_fields)]`, an invalid enum variant, etc.), axum's own
`JsonRejection` handles it **before any handler code runs**, returning its default
plain-text body - never a `CanonicalError`/RFC 7807 `Problem` response.

`canonical_error_middleware` (`libs/toolkit/src/api/canonical_error_layer.rs`) is wired
into every gear's router that uses it, but its own doc is explicit: "Non-problem
responses pass through unchanged... handlers are responsible for typing their errors as
`CanonicalError`." That assumption doesn't hold here - a `Json<T>` extraction failure
happens at the axum extraction layer, before any handler runs, so there is no handler in
the loop to type the error at all.

Confirmed via repo-wide search: no `JsonRejection` handling, custom `FromRequest` impl,
or `Json` wrapper exists anywhere in `libs/toolkit*` or any gear. Every gear using
`axum::Json<T>` as a handler parameter has this gap, not just one.

Concrete evidence: `gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/{events,producers}_tests.rs`'s
`publish_event_readonly_field_rejected_via_deny_unknown_fields` and
`register_producer_invalid_mode_returns_422` both assert a raw plain-text 422 body
instead of a `Problem`, annotated `#[temporary]` referencing this issue.

## Design / Proposed Solution(s) (for discussion)

- A custom extractor (e.g. `CanonicalJson<T>`) in `libs/toolkit` wrapping `axum::Json<T>`,
  with a `FromRequest` impl that catches `JsonRejection` and converts it to
  `CanonicalError::invalid_argument()` (matching the wire shape every other validation
  error already uses), then gears migrate their `Json<T>` extractor usages to it.
- Alternative: a response-mapping layer/middleware that recognizes axum's default
  rejection response shape (`Content-Type: text/plain`, specific body prefix) and
  rewrites it into a `Problem` - more invasive to detect reliably, avoided in favor of
  the wrapper-extractor approach unless a strong reason emerges.
- Open question: whether this belongs in `toolkit::api::canonical_prelude` directly (so
  existing `use toolkit::api::canonical_prelude::*; ... Json<T>` call sites migrate with
  a type alias swap) or requires an explicit new import per gear.

## Definition of Done

- [ ] A canonical-error-aware JSON extractor exists in `libs/toolkit`
- [ ] Malformed JSON, unknown-field, and invalid-enum-variant failures all return
      `application/problem+json` with the correct `400`/`422` status
- [ ] At least one gear (event-broker) migrated as the reference adoption
- [ ] `libs/toolkit/src/api/canonical_error_layer.rs`'s doc comment updated to no longer
      claim this is out of scope

## Validation

- Unit test: malformed JSON body → `Problem` response, not plain text
- Unit test: unknown field under `deny_unknown_fields` → `Problem` response
- Unit test: invalid enum variant → `Problem` response
- Regression: event-broker's two `#[temporary]`-annotated tests updated to assert the
  new `Problem` shape and their `#[temporary]` annotation removed

## References

- `libs/toolkit/src/api/mod.rs:51`
- `libs/toolkit/src/api/canonical_error_layer.rs`
- `gears/system/event-broker/event-broker/src/api/rest/handlers/ingest/events_tests.rs`,
  `producers_tests.rs`

striped-zebra-dev:
## Summary
- `Json<T>`/`Query<T>`/`Path<T>` extraction failures (malformed JSON, `deny_unknown_fields`, invalid enum variants, missing `Content-Type`, oversized bodies, malformed query strings, non-deserializable path segments) resolve axum's own rejection types before any handler runs, bypassing `canonical_error_middleware` entirely and returning plain text instead of an RFC 9457 `Problem` body.
- Adds `toolkit::api::rest::extract::{Json, Query, Path}`, drop-in replacements for axum's own extractors (mirrors axum's own module naming) whose `Rejection` is `CanonicalError`, preserving axum's own per-failure-kind status via a new `GenericResourceError` marker for framework-level errors with no single domain resource in scope. A route/type-definition bug (e.g. `Path`'s `MissingPathParams`/`WrongNumberOfParameters`, which axum itself reports at `.status() == 500`) maps to a real `internal` category rather than being mislabeled as client-fault `invalid_argument` - this classification is centralized in one place (`rejection_to_canonical`) shared by all three extractors.
- Widens `canonical_error_middleware` itself: any error-status response that isn't already a `Problem` (a tower-layer short-circuit, an unmatched route, or any other untyped rejection) is now wrapped into a minimal, valid `Problem` - a foreign 4xx uses RFC 9457 §4.2.1's `"about:blank"` convention, a foreign 5xx maps to the real `internal` category. Preserves CORS/rate-limit/`Set-Cookie`/`WWW-Authenticate` headers (including multi-value ones) from the original response. A new `ForeignPassthrough` marker lets reverse-proxy/passthrough code (`toolkit-gateway::Forwarder`, `oagw`'s proxy handler) exempt a relayed upstream response from this wrapping entirely, since the original response there must reach the client byte-for-byte.
- Fixes `Problem`'s `Deserialize` to make `detail`/`context` genuinely optional per RFC 9457 §3.1 - it previously required both present on the wire, rejecting spec-compliant minimal Problem bodies from well-behaved external peers and silently relabeling their real error category as a generic fallback.
- Migrates every users-info example handler off axum's bare `Json`/`Path` onto the new extractors, with a full users-info e2e suite exercising both the golden path and malformed-input/error-representation cases end-to-end.
- Scope: `libs/toolkit`, `libs/toolkit-canonical-errors`, `libs/toolkit-contract`, `libs/toolkit-gateway`, plus the users-info example and `oagw`'s proxy handler (both migration-only). `gears/bss/ledger`'s own private `CanonicalJson` (which independently confirmed this same gap in production) is left as-is.

Closes #4596

## Test plan
- [x] `cargo test -p cf-gears-toolkit --lib` - 347 passed, 0 failed
- [x] `cargo test -p cf-gears-toolkit-canonical-errors --lib` - 4 passed, 0 failed
- [x] `cargo test -p cf-gears-toolkit-contract --lib` - 78 passed, 0 failed
- [x] `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --check` clean across all touched crates
- [x] `make openapi` regenerated (`docs/api/api.json`) - additive-only diff (413/415/422 response blocks)
- [x] Full users-info e2e suite (`testing/e2e/gears/users_info/`) run live against a built server - golden path and malformed json/query/path error-representation cases
- [x] `oagw`'s full e2e suite (81 tests: proxy round-trip, streaming, rate-limiting, SSE, WebSocket) re-run live - zero regressions from the middleware's new fallback-wrap logic, including the upstream-passthrough case
- [x] New tests cover: valid extraction and every rejection kind across all three extractors, machine-readable field-violation reason codes, route/type-definition bugs mapping to `internal` not `invalid_argument`, foreign 4xx/5xx wrapping (including minimal spec-compliant Problem bodies), `ForeignPassthrough` exemption, multi-value header preservation (`Set-Cookie`, `WWW-Authenticate`), rate-limit header preservation, empty-body fallback to reason phrase, `instance`/`trace_id` fill-in, 2xx/3xx pass-through


libs/toolkit/src/api/canonical_json.rs
```diff
@@ -0,0 +1,269 @@
...
+        .with_field_violation("body", message, code)
+        .with_override(Http::status_code(status))
+        .create()
+}
+
+/// Drop-in replacement for `axum::Json<T>` as a handler parameter. Extraction
+/// success is identical to `Json<T>`; extraction failure produces a
+/// `CanonicalError` (rendered as `Problem` by its existing `IntoResponse`
+/// impl) instead of `JsonRejection`'s plain-text body.
+pub struct CanonicalJson<T>(pub T);
```
striped-zebra-dev: Word canonical is overused already and confuses reader.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -20,21 +42,46 @@ use toolkit_canonical_errors::{CanonicalError, Problem};
...
     let response = next.run(request).await;
 
-    if !is_problem_response(&response) {
-        return response;
+    if is_problem_response(&response) {
+        return enrich_problem_response(response, &uri_path, &request_headers).await;
+    }
+
+    let status = response.status();
+    if status.is_client_error() || status.is_server_error() {
```
MikeFalcon77: **HIGH**

This branch rewrites every non-`Problem` 4xx/5xx response on a gear's entire router, which includes api-gateway's reverse-proxy fallback: `gears/system/api-gateway/src/gear.rs:496` layers this middleware over a router whose fallback is `toolkit_gateway::Forwarder` (mounted at `gear.rs:608` / `gear.rs:1069`, before `apply_middleware_stack`).

`upstream_to_response` in `libs/toolkit-gateway/src/forward.rs:153-168` deliberately copies the upstream status and headers and *streams* the upstream body via `into_limited_body()`; after this change every proxied upstream error response is fully buffered by `wrap_as_generic_problem` and its payload is discarded and replaced with a generic `about:blank` Problem, so gateway clients lose the upstream's own error body (and its streaming) on every 4xx/5xx.

Exclude responses the gear did not produce: have `Forwarder` tag proxied responses with a response extension and check for it in `canonical_error_middleware` before calling `wrap_as_generic_problem`.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+                );
+            }
+        }
+        Err(e) => {
+            tracing::warn!(error = %e, "canonical error middleware: failed to read foreign response body while wrapping");
+        }
+    }
+
+    let problem = Problem {
+        problem_type: "about:blank".to_owned(),
```
MikeFalcon77, line 479: **HIGH**

`problem_type: "about:blank"` contradicts two statements in `docs/arch/errors/DESIGN.md` that this PR does not update: §2.1 Fail-Safe Fallback ("Any error that does not match a canonical category is mapped to `internal` with a trace ID. No error escapes the system without a canonical category") and the `p1` constraint §2.2 RFC 9457 Compliance ("The `type` field carries the GTS URI for the error category", restated as a **Contract** row in §RFC 9457 Problem Wire Format, line 474).

DESIGN.md §2.1's own note says full enforcement of that principle "depends on the error middleware catch-all capability" — this PR *is* that capability, and it implements the opposite mapping, so a documented `p1` wire contract that clients are told to rely on is now violated by the framework itself while the design doc still asserts it.

Either build the wrapped Problem from `CanonicalError::internal(...)` so `type` stays a GTS URI, or amend DESIGN.md §2.1 and §2.2 in this PR to sanction `about:blank` as a second wire form (PRD.md alone is not enough).

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+    let reason = status.canonical_reason().unwrap_or("Error");
+
+    // Never surface a foreign body's own content in the client-visible
+    // `detail` - this response was never typed as `CanonicalError`, so
+    // nothing here has vetted it for the "no internal details on the wire"
+    // guarantee the rest of this error system enforces (Internal/Unknown's
+    // `description` is `#[serde(skip)]` for the same reason). A foreign 5xx
+    // in particular could carry a stack trace, an internal hostname, or a
+    // credential. Diagnostics, if any, are logged server-side only, below.
+    match to_bytes(body, MAX_FALLBACK_DETAIL_BYTES).await {
```
MikeFalcon77: **HIGH**

`to_bytes` bounds the foreign body by size but not by time, and this middleware wraps every gear's whole router.

A slow, stalled, or proxied streaming error body (e.g. an upstream 502 forwarded by `toolkit-gateway`) makes the middleware await until the upstream finishes or the connection dies before it can emit the replacement `Problem`, converting a fast error into a hung request — and the body is fully buffered only to be thrown away.

Wrap the read in `tokio::time::timeout` with a short budget and fall through to the reason-phrase `Problem` on elapse, or skip the read entirely and log only `status` and `Body::size_hint`.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+        Ok(bytes) => {
+            let text = String::from_utf8_lossy(&bytes);
+            let text = text.trim();
+            if !text.is_empty() {
+                tracing::warn!(
```
MikeFalcon77: **HIGH**

The entire foreign response body is logged verbatim at `warn!` in field `body` — the body the comment five lines above describes as something that "could carry a stack trace, an internal hostname, or a credential".

That reasoning is why the body is kept off the wire, but it applies just as much to the log pipeline: logs are shipped to central storage with a different and usually broader access boundary than an HTTP response, so this moves the disclosure instead of preventing it.

Drop the `body` field from this event by default and log `status`, `instance`, and the byte length only; put the body text behind `debug!` (or an explicit opt-in flag) and truncate it to a few hundred bytes.

libs/toolkit/src/api/extractors/error.rs
```diff
@@ -0,0 +1,55 @@
...
+/// usual panic-isolation in this codebase, e.g. `CatchPanicLayer`) forces
+/// this match to be updated the first time a future axum upgrade actually
+/// introduces one, instead of quietly mislabeling it forever.
+pub fn json_rejection_code(rejection: &JsonRejection) -> &'static str {
+    match rejection {
+        JsonRejection::JsonSyntaxError(_) => "json_syntax_error",
+        JsonRejection::JsonDataError(_) => "invalid_json_body",
+        JsonRejection::MissingJsonContentType(_) => "missing_json_content_type",
+        JsonRejection::BytesRejection(_) => "json_body_read_error",
+        _ => panic!("unhandled JsonRejection variant, update json_rejection_code: {rejection}"),
```
MikeFalcon77: **HIGH**

`panic!` on the wildcard arm of the `#[non_exhaustive]` `JsonRejection` puts a panic in the request-serving path of a library crate that every gear routes through.

A semver-compatible axum bump that adds one rejection variant turns a client-fault 4xx into a panic — a 500 where panic isolation exists, a dropped connection where it does not — and the workspace lint set (`Cargo.toml` denies `unwrap_used` and `expect_used`) has no `clippy::panic` entry to catch it at compile time.

Return a conservative fallback code (`"invalid_json_body"`) plus a `tracing::error!` naming the unhandled variant, so a new variant degrades to a logged, valid Problem instead of unwinding.

libs/toolkit/src/api/extractors/path.rs
```diff
@@ -0,0 +1,178 @@
...
+            let status = rejection.status().as_u16();
+            if status >= 500 {
+                return CanonicalError::internal(message).create();
+            }
+            GenericResourceError::invalid_argument()
+                .with_field_violation("path", message, "invalid_path_params")
+                .with_override(Http::status_code(status))
+                .create()
+        }
+        _ => panic!(
```
MikeFalcon77: **HIGH**

`panic!` on the wildcard arm of the `#[non_exhaustive]` `PathRejection` aborts the request path, even though this function already contains a variant-independent classifier that would handle an unknown variant correctly.

The `status >= 500 ? internal : invalid_argument` logic on lines 59-66 needs nothing variant-specific, so the panic buys no safety — it only converts a future axum upgrade into a request-path abort in library code.

Route unknown variants through that same `status >= 500 ? internal : invalid_argument` fallback and log the unhandled variant at `error!` instead of panicking.

libs/toolkit/src/api/extractors/query.rs
```diff
@@ -0,0 +1,165 @@
...
+    match rejection {
+        QueryRejection::FailedToDeserializeQueryString(_) => {
+            let status = rejection.status().as_u16();
+            let message = rejection.body_text();
+            GenericResourceError::invalid_argument()
+                .with_field_violation("query", message, "invalid_query_string")
+                .with_override(Http::status_code(status))
+                .create()
+        }
+        _ => panic!(
```
MikeFalcon77: **HIGH**

`panic!` on the wildcard arm of the `#[non_exhaustive]` `QueryRejection` aborts the request path, although the sole known arm's logic is entirely variant-agnostic.

`rejection.status()`, `rejection.body_text()`, and the fixed `"invalid_query_string"` code are inherent on every `QueryRejection`, so a new variant would be handled correctly by the existing arm — the match exists only to panic.

Delete the match and map every `QueryRejection` through the existing status/`body_text()` path, logging at `error!` when the variant is not one this function was written against.

libs/toolkit/src/api/mod.rs
```diff
@@ -45,6 +46,7 @@ pub mod canonical_prelude {
     pub type ApiResult<T = ()> = std::result::Result<T, CanonicalError>;
 
     // Same response sugar / OData / axum re-exports as the legacy prelude
+    pub use super::extractors;
```
MikeFalcon77: **HIGH**

`canonical_prelude` re-exports the `extractors` module here but still glob-re-exports `axum::Json` on line 53, so `use canonical_prelude::*` binds the bare name `Json` to axum's extractor.

A handler writing the idiomatic `Json(body): Json<T>` therefore keeps the plain-text-rejection behavior this PR exists to fix, and the canonical extractor is reachable only via the longer `extractors::Json` path — the defective extractor is the default and the fix is opt-in.

Add `pub use super::extractors::{Json, Path, Query};` and remove `Json` from the `pub use axum::{...}` list on line 53 (re-export axum's as `AxumJson` if any caller still needs it).

libs/toolkit/src/api/rest/extract/json.rs
```diff
@@ -0,0 +1,307 @@
...
+use axum::extract::{FromRequest, Request};
+use toolkit_canonical_errors::CanonicalError;
+
+use super::error::json_rejection_to_canonical;
+
+/// Drop-in replacement for `axum::Json<T>` as a handler parameter. Extraction
+/// success is identical to `axum::Json<T>`; extraction failure produces a
+/// `CanonicalError` (rendered as `Problem` by its existing `IntoResponse`
+/// impl) instead of `JsonRejection`'s plain-text body.
+pub struct Json<T>(pub T);
```
MikeFalcon77, line 32: **HIGH**

`Json<T>` is documented as a "drop-in `axum::Json<T>` replacement" (module doc line 1, this doc comment line 19) but implements neither `IntoResponse` nor `From<T>`, unlike `axum::Json`.

Any handler that both extracts and returns `Json(...)` — the common case — fails to compile on an import swap, which pushes the whole module back to `axum::Json` and reintroduces the plain-text rejection this type exists to prevent.

Add `impl<T: Serialize> IntoResponse for Json<T>` delegating to `axum::Json`, plus `impl<T> From<T> for Json<T>`; if response position is deliberately out of scope, say so explicitly in the doc comment instead of calling it a drop-in replacement.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -55,15 +102,23 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
-            return Response::from_parts(parts, Body::from(bytes));
+            // A response claiming `application/problem+json` that isn't
+            // actually valid `Problem` JSON still needs to become one - the
+            // platform-wide guarantee is "every error response is a valid
+            // RFC 9457 Problem", not "unless it lied about its own
+            // Content-Type". Route it through the same safe fallback as any
+            // other untyped rejection, from the bytes already read (no
+            // second body read).
+            let foreign = Response::from_parts(parts, Body::from(bytes));
+            return wrap_as_generic_problem(foreign, uri_path, request_headers).await;
```
MikeFalcon77: **MEDIUM**

This fallback call has no status guard, but `is_problem_response` is checked before the status check in `canonical_error_middleware`, so a 2xx/3xx response that carries `Content-Type: application/problem+json` with a body that is not valid `Problem` JSON reaches here.

The result is `{"type":"about:blank","title":"OK","status":200,"detail":"OK"}` — an RFC 9457 violation (a Problem with a success status) that also destroys the original success body and strips its `ETag`/`Content-Encoding`, while `log_problem` records nothing because the status falls outside `400..600`.

Guard this path on `status.is_client_error() || status.is_server_error()`; for a non-error status, log the mislabeled `Content-Type` and return the already-read bytes unchanged.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+        status: status.as_u16(),
+        detail: reason.to_owned(),
+        instance: Some(uri_path.to_owned()),
+        trace_id: extract_trace_id(request_headers),
+        context: serde_json::Value::Object(serde_json::Map::new()),
+        error_code: None,
+        error_domain: None,
+    };
+
+    log_problem(&problem, None);
```
MikeFalcon77, line 490: **MEDIUM**

`log_problem(&problem, None)` hardcodes `None`, so when `enrich_problem_response` falls through to this function it drops the `CanonicalError` it recovered from the response extensions on line 91.

For an `Internal`/`Unknown` 5xx that took the malformed-body path, the diagnostic is `#[serde(skip)]` on the wire and now also absent from the log — the root cause has no record anywhere, defeating the DESIGN §3.6 correlation this middleware is responsible for.

Add an `Option<&CanonicalError>` parameter to `wrap_as_generic_problem`, pass the value recovered in `enrich_problem_response`, and forward it to `log_problem`.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -20,21 +42,46 @@ use toolkit_canonical_errors::{CanonicalError, Problem};
 
 const PROBLEM_JSON: &str = "application/problem+json";
 
+/// Cap on how much of a foreign (non-`Problem`) error response body the
+/// generic-wrap fallback will read into `detail`. Every rejection body seen
+/// in this codebase today (axum's own `JsonRejection` messages, tower-http's
+/// body-limit text) is well under a kilobyte; this bounds the worst case for
+/// a response this middleware does not control the size of.
+const MAX_FALLBACK_DETAIL_BYTES: usize = 8 * 1024;
```
MikeFalcon77: **MEDIUM**

The name `MAX_FALLBACK_DETAIL_BYTES` and its doc comment ("how much of a foreign ... body the generic-wrap fallback will read into `detail`") describe behavior the code does not have: `wrap_as_generic_problem` always sets `detail` to the status reason phrase and never puts the body there.

The constant actually bounds only the server-side log read, so the next reader will look for a `detail` truncation path that does not exist, and may "fix" the cap in the wrong direction.

Rename it to `MAX_FOREIGN_BODY_LOG_BYTES` and state in the doc comment that it caps the body read for logging only, since the foreign body is deliberately never placed in the client-visible `detail`.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+    // Never surface a foreign body's own content in the client-visible
+    // `detail` - this response was never typed as `CanonicalError`, so
+    // nothing here has vetted it for the "no internal details on the wire"
+    // guarantee the rest of this error system enforces (Internal/Unknown's
+    // `description` is `#[serde(skip)]` for the same reason). A foreign 5xx
+    // in particular could carry a stack trace, an internal hostname, or a
+    // credential. Diagnostics, if any, are logged server-side only, below.
+    match to_bytes(body, MAX_FALLBACK_DETAIL_BYTES).await {
+        Ok(bytes) => {
+            let text = String::from_utf8_lossy(&bytes);
```
MikeFalcon77, line 346: **MEDIUM**

The foreign body is passed to `tracing::warn!` as raw `String::from_utf8_lossy(...).trim()` with no control-character handling.

axum and tower rejection texts embed client-supplied values, so a caller can inject newlines to forge additional log records, or ANSI escapes into any non-JSON log sink — the classic log-forging vector, and it applies to unauthenticated 4xx paths.

Escape the value before logging (`text.escape_debug()`, or a control-character-stripping helper) and hard-truncate it to a fixed short length.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+    };
+
+    let mut response = Response::from_parts(parts, Body::from(bytes.clone()));
+    let headers = response.headers_mut();
+    // The body just changed from the foreign response's original bytes to a
+    // freshly-serialized `Problem` - any body-representation headers describing
+    // that original body (e.g. `Content-Encoding: gzip`) would otherwise
+    // survive `Response::from_parts` untouched and make a client try to
+    // decode the new plaintext JSON as if it were still compressed.
+    headers.remove(header::CONTENT_ENCODING);
```
MikeFalcon77: **MEDIUM**

The rewrite reuses the foreign response's `parts` and denylists only three headers, so every other header describing the now-discarded body survives onto the synthesized Problem: `Content-Digest`, `Digest`, `Content-MD5`, `Content-Language`, `Content-Location`, `Last-Modified`, plus any fingerprinting header the foreign layer set (`Server`, `X-Powered-By`, internal `X-*`).

A denylist is the wrong default here — the body is fully replaced, so the correct assumption is that no body-describing header from the original response is still true, and each new header a dependency starts emitting silently becomes a new leak or a new client-visible inconsistency.

Build a fresh `HeaderMap` containing an allowlist (tracing/correlation, CORS, `WWW-Authenticate`) plus the Problem's own `Content-Type`/`Content-Length`, instead of removing three names from the original.

libs/toolkit/src/api/extractors/error.rs
```diff
@@ -0,0 +1,55 @@
...
+/// vocabulary already shipped in `gears/bss/ledger`'s own `CanonicalJson`.
+///
+/// # Panics
+///
+/// `JsonRejection` is `#[non_exhaustive]`, so this match needs some arm for
+/// a variant it doesn't know about - deliberately a panic, not a fallback
+/// code. Guessing a code for a variant we haven't looked at would silently
+/// ship a claim we can't stand behind; a panic (caught per-request by the
+/// usual panic-isolation in this codebase, e.g. `CatchPanicLayer`) forces
+/// this match to be updated the first time a future axum upgrade actually
```
MikeFalcon77: **MEDIUM**

The stated justification for the panic — "caught per-request by the usual panic-isolation in this codebase, e.g. `CatchPanicLayer`" — is not a property this library can rely on.

Panic isolation exists in exactly two places: `libs/toolkit/src/runtime/oop_serve.rs`'s `drain_guard_middleware` `catch_unwind`, and `gears/system/api-gateway/src/gear.rs`'s `CatchPanicLayer`. Any router that adopts these extractors outside those two stacks — including the `Router::new()` test apps in this very PR — unwinds into hyper and drops the connection, so the doc comment promises a safety net that is not universally present.

Remove the panic-based arms in favour of logged fallbacks, and add `panic = "deny"` to `[workspace.lints.clippy]` so future panics in request-handling code fail the build.

libs/toolkit/src/api/extractors/error.rs
```diff
@@ -0,0 +1,55 @@
...
+/// explicitly instead of relying on `?`'s automatic conversion.
+pub fn json_rejection_to_canonical(rejection: &JsonRejection) -> CanonicalError {
+    let code = json_rejection_code(rejection);
+    // `.status()` / `.body_text()` are inherent on every axum rejection
+    // type (incl. nested composites, e.g. `BytesRejection`'s own
+    // `LengthLimitError`/`UnknownBodyError`) - resolves to axum's own
+    // intended status per failure kind with no variant matching needed
+    // for the status itself.
+    let status = rejection.status().as_u16();
+    let message = rejection.body_text();
```
MikeFalcon77: **MEDIUM**

`rejection.body_text()` is copied verbatim and unbounded into the client-visible `field_violations[].description` (same at `path.rs:55` and `query.rs:50`), echoing serde/axum internals to unauthenticated callers: target struct field names, the full set of expected enum variants, internal Rust type names, and the caller's own input reflected back.

This PR's own tests document the disclosure — `"unknown field `extra`, expected `name`"`, `"Cannot parse `id` with value `not-a-number` to a `u32`"` — and the sibling middleware in the same PR caps this exact class of content at 8 KiB precisely because it is unvetted; here there is no cap and no vetting at all.

Map each rejection kind to a fixed, vetted client-facing message (the `code` already carries the machine-readable reason), log the raw `body_text()` server-side only, and truncate whatever does reach the wire.

libs/toolkit/src/api/extractors/error.rs
```diff
@@ -0,0 +1,55 @@
...
+        JsonRejection::BytesRejection(_) => "json_body_read_error",
+        _ => panic!("unhandled JsonRejection variant, update json_rejection_code: {rejection}"),
+    }
+}
+
+/// Maps a `JsonRejection` to a `CanonicalError`. Not a `From` impl - both
+/// `JsonRejection` and `CanonicalError` are foreign to this crate, so the
+/// orphan rule forbids it; `extractors::Json::from_request` calls this
+/// explicitly instead of relying on `?`'s automatic conversion.
+pub fn json_rejection_to_canonical(rejection: &JsonRejection) -> CanonicalError {
```
MikeFalcon77: **MEDIUM**

Only the JSON mapper lives in this shared `error` module; `path_rejection_to_canonical` and `query_rejection_to_canonical` are private copies in their own modules, each repeating the same `status()` / `body_text()` / `invalid_argument()` / `with_field_violation` / `with_override` sequence with only the field name and reason code differing.

Three copies of the same five-call sequence means a change to the canonical shape (say, adding the truncation above, or a redacted message) has to be made three times, and one of them will be missed.

Move a single `rejection_to_canonical(field: &str, code: &str, status: u16, message: String) -> CanonicalError` into this module and have all three extractors call it, keeping only their variant-to-code classification local.

libs/toolkit/src/api/extractors/error.rs
```diff
@@ -0,0 +1,55 @@
...
+/// specific to JSON, reusable by future extractor-level fixes
+/// (`extractors::Query`, `extractors::Path`). Mirrors
+/// `toolkit_odata::errors::OdataError`. `error` is a private module (see
+/// `extractors/mod.rs`), so this `pub` item is still crate-internal in
+/// effect - nothing outside `toolkit` constructs it directly.
+#[resource_error(gts_id!("cf.core.http.request.v1~"))]
+pub struct GenericResourceError;
+
+/// Machine-readable reason code per `JsonRejection` kind, matching the
+/// vocabulary already shipped in `gears/bss/ledger`'s own `CanonicalJson`.
```
MikeFalcon77: **MEDIUM**

This PR ships a toolkit-wide replacement for `gears/bss/ledger`'s private `CanonicalJson` but leaves that gear-local extractor in place, and the two now disagree on the wire for identical client mistakes: ledger maps every `JsonRejection` to a fixed 400 with its own message text, while this one propagates axum's status (400/413/415/422) and axum's `body_text()`.

The same malformed body therefore produces a different status and a different `description` depending on which gear received it — the exact cross-gear inconsistency the canonical error system exists to remove.

Migrate the ledger REST handlers to `toolkit::api::extractors::Json` and delete `gears/bss/ledger/ledger/src/api/rest/canonical_json.rs` in this PR, or record in the ADR which status mapping is authoritative and mark the other deprecated.

libs/toolkit/src/api/extractors/json.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// Drop-in replacement for `axum::Json<T>` as a handler parameter. Extraction
+/// success is identical to `axum::Json<T>`; extraction failure produces a
+/// `CanonicalError` (rendered as `Problem` by its existing `IntoResponse`
+/// impl) instead of `JsonRejection`'s plain-text body.
+pub struct Json<T>(pub T);
+
+impl<T, S> FromRequest<S> for Json<T>
+where
+    AxumJson<T>: FromRequest<S, Rejection = JsonRejection>,
```
MikeFalcon77: **MEDIUM**

The public `FromRequest` impl is bounded on `AxumJson<T>: FromRequest<S, Rejection = JsonRejection>`, which leaks the delegate type into toolkit's public API surface.

A caller who forgets `#[derive(Deserialize)]` gets a rustc diagnostic about `axum::Json` not implementing `FromRequest` rather than about their own type's missing bound, and toolkit is now committed to axum's rejection type in its public signature.

Bound the impl on `T: DeserializeOwned` instead — as `gears/bss/ledger`'s existing `CanonicalJson` already does — so `axum::Json` stays an implementation detail.

libs/toolkit/src/api/rest/extract/json.rs
```diff
@@ -0,0 +1,307 @@
...
+
+impl<T, S> FromRequest<S> for Json<T>
+where
+    AxumJson<T>: FromRequest<S, Rejection = JsonRejection>,
+    S: Send + Sync,
+{
+    type Rejection = CanonicalError;
```
MikeFalcon77, line 39: **MEDIUM**

This extractor returns 415, 422, and 413 (asserted by its own tests below), but `OperationBuilder::standard_errors` registers none of them: `libs/toolkit/src/api/operation_builder.rs:1653-1660` omits 413 and 415 entirely, and its comment states 422 is "intentionally absent ... no canonical-handler path produces a 422 response" — which this extractor makes false.

Every operation that adopts `extractors::Json` therefore serves three statuses absent from its published OpenAPI responses, so generated clients have no Problem schema for them.

Register 413/415/422 Problem responses for operations using this extractor (an `OperationBuilder` helper, or an extension of `standard_errors`) and update the now-incorrect 422 rationale comment in `operation_builder.rs`.

libs/toolkit/src/api/rest/extract/json.rs
MikeFalcon77, line 32: **MEDIUM**

The three public extractor newtypes (`Json` here, `Path`, `Query`) derive nothing, while their axum counterparts derive `Debug, Clone, Copy, Default` and impl `Deref`/`DerefMut`.

A public library type with no `Debug` cannot appear in a `dbg!`, a `#[derive(Debug)]` struct that holds it, or a tracing field — `missing_debug_implementations` flags exactly this — and the missing `Deref` is a second silent break for code migrating off axum's extractors.

Add `#[derive(Debug, Clone, Copy, Default)]` (bounds permitting) to all three, and state in the doc comment that `Deref`/`DerefMut` are omitted deliberately so the divergence from axum is a decision rather than an oversight.

libs/toolkit/src/api/rest/extract/json.rs
```diff
@@ -0,0 +1,307 @@
...
+            json!({
+                "type": INVALID_ARGUMENT_TYPE,
+                "title": "Invalid Argument",
+                "status": 400,
+                "detail": "Request validation failed",
+                "context": {
+                    "resource_type": RESOURCE_TYPE,
+                    "field_violations": [{
+                        "field": "body",
+                        "description": "Failed to parse the request body as JSON: key must be a string at line 1 column 2",
```
MikeFalcon77, line 165: **MEDIUM**

This whole-body `assert_eq!` pins axum's and serde_json's exact internal prose (`"Failed to parse the request body as JSON: key must be a string at line 1 column 2"`, and `"unknown field `extra`, expected `name` at line 1 column 19"` on line 142).

Those strings are upstream `Display` output with no stability guarantee, so a patch-level axum or serde bump reddens the suite while proving nothing about this extractor — the test is asserting a third party's wording, not this code's behavior.

Keep exact assertions on the fields this code owns (`type`, `status`, `field: "body"`, `reason`) and reduce `description` to a substring check on a stable fragment.

libs/toolkit/src/api/rest/extract/path.rs
```diff
@@ -0,0 +1,178 @@
...
+            json!({
+                "type": INVALID_ARGUMENT_TYPE,
+                "title": "Invalid Argument",
+                "status": 400,
+                "detail": "Request validation failed",
+                "context": {
+                    "resource_type": RESOURCE_TYPE,
+                    "field_violations": [{
+                        "field": "path",
+                        "description": "Invalid URL: Cannot parse `id` with value `not-a-number` to a `u32`",
```
MikeFalcon77, line 147: **MEDIUM**

This assertion pins axum's verbatim rejection prose `"Invalid URL: Cannot parse `id` with value `not-a-number` to a `u32`"`.

That text is an upstream `Display` impl carrying no cross-release guarantee, so the test breaks on an axum patch bump without any behavior of this module having changed.

Assert `status: 400`, `field: "path"`, and `reason: "invalid_path_params"` exactly, and check `description` with a substring such as `not-a-number`.

libs/toolkit/src/api/rest/extract/query.rs
```diff
@@ -0,0 +1,165 @@
...
+            json!({
+                "type": INVALID_ARGUMENT_TYPE,
+                "title": "Invalid Argument",
+                "status": 400,
+                "detail": "Request validation failed",
+                "context": {
+                    "resource_type": RESOURCE_TYPE,
+                    "field_violations": [{
+                        "field": "query",
+                        "description": "Failed to deserialize query string: page: invalid digit found in string",
```
MikeFalcon77, line 134: **MEDIUM**

Both query tests pin serde's exact wording — `"Failed to deserialize query string: page: invalid digit found in string"` here and `"missing field `page`"` on line 158.

This is upstream prose, not behavior this module defines, so a serde or `serde_urlencoded` patch bump breaks the suite while the extractor is unchanged.

Keep exact assertions on `type`, `status`, `field`, and `reason`, and switch the `description` assertions to substring matches on the field name.

libs/toolkit/src/api/rest/extract/path.rs
```diff
@@ -0,0 +1,178 @@
...
+                        "reason": "invalid_path_params",
+                    }],
+                },
+            })
+        );
+    }
+
+    #[test]
+    fn missing_path_params_maps_to_internal_not_invalid_argument() {
```
MikeFalcon77, line 188: **MEDIUM**

Only the `MissingPathParams` internal mapping is tested; the `status >= 500` branch inside `FailedToDeserializePathParams` (lines 60-62) has no test.

That branch is the one the doc comment justifies at length — axum's `WrongNumberOfParameters` / `UnsupportedType` kinds, and the `.with_override` same-status-class invariant it exists to protect — yet deleting it would leave every test green, so the invariant is documented but unguarded.

Add a test that routes `Path<(u32, u32)>` against the single-parameter route `/items/{id}` and assert the response is the 500 `internal` Problem, not a 4xx `invalid_argument`.

libs/toolkit/src/api/rest/extract/json.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    async fn body_json(response: axum::response::Response) -> Value {
+        let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        serde_json::from_slice(&bytes).expect("response body is valid JSON")
+    }
+
+    #[tokio::test]
+    async fn valid_body_extracts_normally() {
```
MikeFalcon77, line 109: **MEDIUM**

`valid_body_extracts_normally` — and its `path.rs` / `query.rs` twins — asserts only the status code, while the handler discards the extracted value via `Json(_w)`.

The entire success-path contract of this type is "extraction success is identical to `axum::Json<T>`", and a `Json<T>` that yielded a wrong, empty, or defaulted payload would still return `201` and still pass.

Have the handler echo the deserialized value back and assert it, in the success test of all three extractors.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -462,6 +638,367 @@ mod tests {
...
+                get(move || {
+                    let responder = responder.clone();
+                    async move { responder() }
+                }),
+            )
+            .layer(from_fn(canonical_error_middleware))
+    }
+
+    #[tokio::test]
+    async fn wraps_a_foreign_4xx_plain_text_response_as_a_problem() {
```
MikeFalcon77: **MEDIUM**

No foreign-wrap test carries `#[tracing_test::traced_test]`, so the security-relevant side effect `wrap_as_generic_problem` documents — "body logged, not sent to client" — is never asserted.

The positive half (the body is absent from the response) is covered, but deleting the entire `tracing::warn!` block would keep every test passing, which means the server-side diagnostic that justifies suppressing the body on the wire is unprotected.

Add `#[tracing_test::traced_test]` to this test and assert `logs_contain("length limit exceeded")` alongside the existing exact-JSON check.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -462,6 +638,367 @@ mod tests {
         assert_eq!(v["trace_id"].as_str(), Some("abc123"));
     }
 
+    fn foreign_response(status: StatusCode, content_type: &str, body: &'static str) -> Response {
```
MikeFalcon77: **MEDIUM**

No test drives a foreign body larger than `MAX_FALLBACK_DETAIL_BYTES`, so the `to_bytes` `Err` arm on line 183 is uncovered — neither its warning nor the requirement that the response is still wrapped as a valid Problem when the read fails.

The cap itself is equally unguarded: the constant could be set to `0` and the suite would stay green, even though that would send every wrapped error down the error arm.

Add a test whose foreign responder returns a body over 8 KiB and assert the response is still the `about:blank` Problem with the correct status while the over-limit read is logged.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -462,6 +638,367 @@ mod tests {
...
+                "status": 413,
+                "detail": "Payload Too Large",
+                "instance": "/upload",
+                "context": {},
+            })
+        );
+    }
+
+    #[tokio::test]
+    async fn wrapping_strips_stale_content_encoding_header() {
```
MikeFalcon77: **MEDIUM**

This test asserts only the `Content-Encoding` removal; the `Content-Range` and `ETag` removals and the `Content-Length` rewrite on lines 221-224 have no coverage.

`Content-Length` is the one that matters: the tests read bodies with `to_bytes`, which ignores `Content-Length`, so a stale value inherited from the foreign response would pass every test here and then truncate or hang real clients.

Extend this test so the foreign response also carries `Content-Range`, `ETag`, and a `Content-Length` matching its original body, then assert the first two are gone and `Content-Length` equals the serialized Problem's byte length.

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -86,12 +141,107 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+        Err(e) => {
+            tracing::error!(
+                error = %e,
+                "canonical error middleware: failed to serialize generic-wrap problem body"
+            );
+            return Response::from_parts(parts, Body::empty());
+        }
+    };
+
+    let mut response = Response::from_parts(parts, Body::from(bytes.clone()));
```
MikeFalcon77: **LOW**

`Body::from(bytes.clone())` deep-copies the whole serialized Problem buffer solely so `bytes.len()` can be read on line 224 (the same shape as `new_bytes.clone()` in `enrich_problem_response`).

`clippy::redundant_clone` cannot flag it because `bytes` is genuinely used after the clone, so it will not be caught by lints.

Bind `let len = bytes.len();` before constructing the body, then move `bytes` into `Body::from` and use `len` for the `Content-Length` header.

libs/toolkit/src/api/mod.rs
striped-zebra-dev: Not doing this — it would silently change error-handling behavior across 27 gear crates (~130 call sites) in one PR, and it contradicts our own ADR's explicit opt-in migration. Re-pointing a common name via a glob-imported prelude is the wrong pattern regardless of blast radius. Keeping `extractors::Json` explicit-path-only; see `examples/toolkit/users-info`'s `create_user` for the intended usage. Will sharpen the prelude's doc comment instead.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Confirmed real: `Forwarder`'s fallback is layered inside `canonical_error_middleware`, so every proxied non-Problem 4xx/5xx now gets buffered and replaced, losing streaming and the upstream body. Filing a follow-up ticket to tag proxied responses on `Forwarder` so the middleware skips them, rather than fixing in this PR.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Agreed. Split the fallback: a foreign 5xx now builds a real `CanonicalError::internal(...)` (status preserved via `.with_override` when not already 500), so `type` stays a GTS URI as DESIGN.md §2.2 requires; a foreign 4xx still uses `about:blank` since a bare 4xx genuinely doesn't determine one category. Amended DESIGN.md §2.1, §2.2, the §3.3 wire-format table, and the §3.2 Error Middleware component doc to state this split explicitly instead of the old blanket "mapped to internal."

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Added a 2s `tokio::time::timeout` around the foreign-body read; on elapse, wrapping proceeds without the diagnostic log instead of blocking. Read stays capped at 8 KiB, for logging only, never the client response.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Split into two log lines: `warn!` now carries only `status`/`instance`, no body. The foreign body, if any, is `debug!`-only and truncated to 256 chars.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Fixed: a non-error status (2xx/3xx) with an invalid `problem+json` body now logs a warning and passes the original bytes through unchanged, instead of being wrapped into a `Problem` claiming a success status.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Renamed to `MAX_FOREIGN_BODY_LOG_BYTES` with a doc comment stating it bounds the log read only, never `detail`.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Body text is now `.escape_debug()`-escaped and hard-truncated (256 chars) before reaching any log line, on top of the existing byte cap and the move to `debug!`-only.

libs/toolkit/src/api/extractors/error.rs
striped-zebra-dev: Agreed this is a real inconsistency. Not migrating `gears/bss/ledger` in this PR - filing a follow-up ticket to move it to `toolkit::api::extractors::Json` and remove the gear-local duplicate.

libs/toolkit/src/api/rest/extract/json.rs
striped-zebra-dev: Added `impl<T: Serialize> IntoResponse for Json<T>` (delegates to `axum::Json`) and `impl<T> From<T> for Json<T>`, plus a new test (`usable_as_a_response_body`) proving the `Json(body): Json<Req> -> ... -> Json(resp)` shape compiles and round-trips. Doc comment updated; `examples/toolkit/users-info`'s `create_user` handler now uses `extractors::Json` as a live demonstration.

libs/toolkit/src/api/rest/extract/json.rs
striped-zebra-dev: Respectfully, we disagree and are keeping the full-body assertion. This PR's test philosophy (deliberately established for this change) is to assert the complete response body, specifically so drift has zero blind spots - a substring-only check can keep passing while the rest of the description silently degrades or ends up describing the wrong failure entirely, which is exactly the kind of regression a substring match would miss. We accept the (rare, one-line) cost of updating a literal string if axum/serde reword their `Display` output on a future patch bump, in exchange for never having a blind spot in what these tests actually verify. If `3812353948` (redacting `description` to our own fixed vetted string) lands, this becomes moot anyway - the text becomes ours, not upstream's.

libs/toolkit/src/api/rest/extract/path.rs
striped-zebra-dev: Respectfully, we disagree and are keeping the full-body assertion. This PR's test philosophy (deliberately established for this change) is to assert the complete response body, specifically so drift has zero blind spots - a substring-only check can keep passing while the rest of the description silently degrades or ends up describing the wrong failure entirely, which is exactly the kind of regression a substring match would miss. We accept the (rare, one-line) cost of updating a literal string if axum/serde reword their `Display` output on a future patch bump, in exchange for never having a blind spot in what these tests actually verify. If `3812353948` (redacting `description` to our own fixed vetted string) lands, this becomes moot anyway - the text becomes ours, not upstream's.

libs/toolkit/src/api/rest/extract/query.rs
striped-zebra-dev: Respectfully, we disagree and are keeping the full-body assertion. This PR's test philosophy (deliberately established for this change) is to assert the complete response body, specifically so drift has zero blind spots - a substring-only check can keep passing while the rest of the description silently degrades or ends up describing the wrong failure entirely, which is exactly the kind of regression a substring match would miss. We accept the (rare, one-line) cost of updating a literal string if axum/serde reword their `Display` output on a future patch bump, in exchange for never having a blind spot in what these tests actually verify. If `3812353948` (redacting `description` to our own fixed vetted string) lands, this becomes moot anyway - the text becomes ours, not upstream's.

libs/toolkit/src/api/extractors/error.rs
striped-zebra-dev: We checked every `description` string this code actually produces (see the extractor test suites) and none of them expose anything beyond the caller's own submitted input or the request schema's own public field/enum names - already necessarily public via the OpenAPI spec, since a client needs them to build a valid request. This exact text was also already unconditionally on the wire before this PR: axum's own default `JsonRejection` response put it there as unstructured plain text (that's literally #4596). So we're not applying a blanket redaction to a fixed generic string - it would make the API strictly worse for legitimate clients (losing "you sent `not-a-number`, we expected a `u32`") for no confidentiality benefit, since nothing here is sensitive.

What we did add: a 500-character cap that truncates the message (with a "... (truncated)" marker) before it reaches `field_violations[].description`. Normal, well-formed traffic never gets near it - every message this code produces today is under 200 characters - but it stops a client from inflating the response by submitting a pathologically long value that axum/serde would otherwise echo back verbatim (e.g. an extremely long unknown field name under `deny_unknown_fields`). Bandwidth concern, not a confidentiality one.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Fixed properly, not just the incidental improvement from the about:blank/internal split. `wrap_foreign_response` now takes the `CanonicalError` recovered from the original response's extensions and, when present, uses it directly (`wrap_recovered_canonical_error`) instead of ever synthesizing a generic one - the real category, status, and diagnostic were already known, so there was no reason to discard them for a class-based guess. This applies uniformly to both the malformed-`problem+json` path (`enrich_problem_response`) and the fully-foreign path, and covers any category, not just `internal` - a corrupted `service_unavailable` body now correctly wraps as `service_unavailable`, not `internal`. New test: `malformed_body_with_recovered_extension_uses_the_real_category`.

diffora:
**HIGH**

`Problem` requires `detail` and `context` on deserialize, but RFC 9457 makes both optional — so a valid `problem+json` body from anything that isn't this crate fails to parse in `enrich_problem_response` and falls through to `wrap_foreign_response`, which throws the original error away. Missing either field is enough to trigger it.

This hits the proxy paths hardest: `toolkit_gateway::Forwarder` is mounted inside this middleware (`gears/system/api-gateway/src/gear.rs:496`), and oagw's `proxy_handler` sits under it too.

Send this back from a route under the middleware:

```http
HTTP/1.1 409 Conflict
Content-Type: application/problem+json

{"type":"https://example.com/probs/out-of-credit","title":"You do not have enough credit.","status":409}
```

You get:

```json
{"type":"about:blank","title":"Conflict","status":409,"detail":"Conflict","instance":"/upload","context":{}}
```

The 5xx side is worse, because the new `wrap_as_internal_problem` path relabels the fault. Upstream sends:

```http
HTTP/1.1 502 Bad Gateway
Content-Type: application/problem+json

{"type":"https://example.com/probs/upstream-down","title":"Upstream is down.","status":502}
```

Client gets:

```json
{"type":"gts://gts.cf.core.errors.err.v1~cf.core.err.internal.v1~","title":"Internal","status":502,"detail":"An internal error occurred. Please retry later.","instance":"/upload","context":{}}
```

So a proxied upstream failure is now reported as *our* internal error.

Fix: `#[serde(default)]` on `detail` / `context`, or skip the rewrite when the body already parses as a valid RFC 9457 problem.

libs/toolkit/src/api/extractors/error.rs
striped-zebra-dev: Kept the panic here - Json's four known variants each map to a genuinely different code, so an unknown one really is unclassified - but split it by build: debug/test still panics via `debug_assert!` (forcing an update the moment axum adds a variant), release logs and degrades to an honest `unclassified_json_rejection` code instead of unwinding. Mirrors this crate's own status-class-invariant `debug_assert!` pattern.

libs/toolkit/src/api/extractors/path.rs
striped-zebra-dev: Agreed, removed the panic. The `>= 500` classification is already fully status-driven, not variant-specific, so an unknown variant is handled correctly by that same logic with nothing invented. Now just logs at `error!` and falls through to it.

libs/toolkit/src/api/extractors/query.rs
striped-zebra-dev: Agreed, removed the match/panic entirely. The sole known arm was already fully generic (`.status()`/`.body_text()`, one fixed code), so every variant - known or not - goes through the same path now, logging at `error!` only when it isn't the one this was written against.

libs/toolkit/src/api/extractors/error.rs
striped-zebra-dev: Agreed, the `CatchPanicLayer` claim was inaccurate. Doc comment rewritten to state what's actually true - panic isolation exists in exactly 2 call sites, not universally - and the debug-panic/release-degrade split above makes the isolation claim moot regardless.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Fixed - `finish_wrapped_response` now builds headers from an explicit allowlist (`WWW-Authenticate`, `Retry-After`, CORS headers, `Vary`) instead of denylisting three names.

libs/toolkit/src/api/extractors/error.rs
striped-zebra-dev: Fixed - moved a single `rejection_to_canonical(field, code, status, message)` into `error.rs`; json/path/query now keep only their own variant classification, no duplicated construction logic.

libs/toolkit/src/api/extractors/json.rs
striped-zebra-dev: Fixed - bound is now `T: DeserializeOwned` instead of referencing `AxumJson`'s own bound, so the delegate type no longer leaks into the public signature.

libs/toolkit/src/api/rest/extract/json.rs
striped-zebra-dev: Fixed - added `error_413` alongside the existing `error_415`/`error_422`, and wired all three into every users-info operation whose handler takes `extract::Json<T>`.

libs/toolkit/src/api/rest/extract/json.rs
striped-zebra-dev: Added - `Debug, Clone, Copy, Default` on all three newtypes, matching axum's own API surface.

libs/toolkit/src/api/rest/extract/path.rs
striped-zebra-dev: Added - `wrong_number_of_path_params_maps_to_internal_not_invalid_argument`, driving `Path<(u32,u32)>` against a single-param route to hit the `>= 500` branch directly.

libs/toolkit/src/api/rest/extract/json.rs
striped-zebra-dev: Fixed - the three success tests now echo the extracted value back and assert it, not just the status.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Added - `foreign_body_is_logged_at_debug_but_never_sent_to_client`, asserting `logs_contain` on the `debug!` line alongside the existing response-body check.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Added - `oversized_foreign_body_still_wraps_correctly`, driving a body over the cap and asserting the response still wraps correctly while the read failure is logged.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Checked first - not a live bug: `insert` (not `append`) always overwrites any stale value. Extended the test anyway (`wrapping_strips_stale_body_representation_headers`) to assert `Content-Range`/`ETag` removal and the recomputed `Content-Length` explicitly, as regression insurance.

libs/toolkit/src/api/canonical_error_layer.rs
striped-zebra-dev: Fixed - both instances (here and the twin in `enrich_problem_response`) now capture the length before moving the buffer, instead of cloning it.

libs/toolkit/src/api/canonical_error_layer.rs
MikeFalcon77: Looks like this can't be a follow-up, it already fails CI on this branch but green on main

libs/toolkit/src/api/canonical_error_layer.rs
MikeFalcon77: The allowlist is the right shape, but three headers are missing and one of them is a MUST. axum sets Allow on its own 405 responses (set_allow_header in axum/src/routing/route.rs), 405 is a foreign error status, so it goes through the wrap and PRESERVED_FOREIGN_HEADERS drops it. Also missing: Set-Cookie (a 401/403 that clears a session cookie loses it) and Access-Control-Allow-Methods / -Allow-Headers / -Max-Age.

libs/toolkit/src/api/rest/extract/json.rs
MikeFalcon77: The builder change is good. Please do not forget to regenerate docs/api/api.json (make openapi)

libs/toolkit/src/api/mod.rs
MikeFalcon77: Looks like the doc change you promised isn't in the branch. Please add.

libs/toolkit/src/api/extractors/error.rs
MikeFalcon77: ok. please fill a ticket

libs/toolkit/src/api/extractors/error.rs
MikeFalcon77: Ok split works for me. One thing to fix before merge: please udpate ADR 0006 because it still describes the old behaviour. Its tables say the Json mapper "panics", and that Query and Path panic on unknown #[non_exhaustive] variants — Query and Path don't panic at all now, and Json only debug_assert!.

libs/toolkit/src/api/rest/extract/json.rs
MikeFalcon77: OK, but for the record I didn't ask for a substring-only check. I asked for exact assertions on every field this code owns (type, status, field, reason) plus a substring on description only, so there's no blind spot either way. 

The stronger argument for your side is one you didn't make: description is client-visible wire text, so upstream rewording is a real contract change worth catching. Please record that trade-off in the ADR or in separate issue.

libs/toolkit/src/api/rest/extract/path.rs
MikeFalcon77: The same answer is above

libs/toolkit/src/api/rest/extract/query.rs
MikeFalcon77: OK. already mentioned by me

libs/toolkit/src/api/extractors/error.rs
MikeFalcon77: agreed

libs/toolkit/src/api/canonical_error_layer.rs
MikeFalcon77: you are right. ok

striped-zebra-dev:
> **HIGH**
> 
> `Problem` requires `detail` and `context` on deserialize, but RFC 9457 makes both optional — so a valid `problem+json` body from anything that isn't this crate fails to parse in `enrich_problem_response` and falls through to `wrap_foreign_response`, which throws the original error away. Missing either field is enough to trigger it.
>  ...
> 

@diffora 
Thank you, fixed.

docs/arch/errors/ADR/0006-cpt-cf-adr-error-middleware-catchall.md
```diff
@@ -0,0 +1,243 @@
...
+  Router/extractor machinery runs at all
+* Avoid guessing a specific canonical category (`invalid_argument` vs.
+  `failed_precondition` vs. `out_of_range`, etc.) from information that
+  doesn't uniquely determine it - a bare HTTP status code is ambiguous
+  across several categories
+* Reuse already-shipped primitives (`.with_override`, gears-rust#4466;
+  `#[resource_error]`) rather than introducing new machinery
+* Zero required changes to any gear - the fix must apply through
+  `libs/toolkit` alone, given every gear already receives
+  `canonical_error_middleware` as its outermost layer via the OoP bootstrap
```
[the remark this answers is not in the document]
striped-zebra-dev: Out of scope for this PR - we'll file follow-up tickets to update the PR description's stated scope and track anything further. The underlying complaint no longer fully applies anyway: users-info is now fully migrated with a real e2e suite, so "no gear touched, no regression test" is already stale in our favor, not against it.

docs/arch/errors/ADR/0006-cpt-cf-adr-error-middleware-catchall.md
```diff
@@ -0,0 +1,243 @@
...
+  it with a client- or server-error status that is not already
+  `application/problem+json` is rewritten into a minimal `Problem` using RFC
+  9457 §4.2.1's `about:blank` convention (`type: "about:blank"`, `title`
+  *and* `detail` both the status's standard reason phrase - the foreign
+  body's own content is never placed on the wire, only logged server-side,
+  since nothing has vetted it against this platform's "no internal details
+  on the wire" guarantee). A response that claims
+  `application/problem+json` but fails to deserialize as a valid `Problem`
+  is routed through this same branch rather than passed through malformed -
+  the "every error is a valid Problem" guarantee has no exception for a
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - updated the ADR's main bullet and mapping tables to describe the actual split: foreign 5xx maps to a real `internal` CanonicalError, foreign 4xx uses `about:blank`, and a recovered CanonicalError (from surviving extensions) wins over either guess. Also documented the new `ForeignPassthrough` escape hatch and updated the "Confirmation" section to list the tests that actually exist now (5xx-vs-4xx split, extensions-recovered path, `ForeignPassthrough`, and `oagw`'s own e2e coverage).

docs/arch/errors/ADR/0006-cpt-cf-adr-error-middleware-catchall.md
```diff
@@ -0,0 +1,243 @@
...
+
+`PathRejection` -> `CanonicalError` mapping (`extract::Path`) - not
+purely a client-fault mapping, unlike `Json`/`Query`:
+
+| `PathRejection` variant | status | maps to |
+|---|---|---|
+| `FailedToDeserializePathParams` (most `ErrorKind`s) | 400 | `invalid_argument`, code `invalid_path_params` |
+| `FailedToDeserializePathParams` (`WrongNumberOfParameters`/`UnsupportedType`) | 500 | `internal` - a route/type definition bug, not the client's |
+| `MissingPathParams` | 500 | `internal` - the extractor used outside a matched route |
+| any future (`#[non_exhaustive]`) variant | n/a | n/a - panics, same rationale as `Json` |
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - the PathRejection mapping table's wildcard row no longer says "panics." It now describes the real behavior: an unknown variant falls through to the same `>= 500`-driven classification the known variant already uses, logged at `error!` for visibility, never unwinding.

docs/arch/errors/DESIGN.md
```diff
@@ -471,7 +469,7 @@ Every REST error response follows this structure:
 
 | Field | Source | Part | Description |
 |-------|--------|------|-------------|
-| `type` | GTS URI from category | **Contract** | Error type URI (e.g., `gts.cf.core.errors.err.v1~cf.core.err.not_found.v1~`) |
+| `type` | GTS URI from category, or `"about:blank"` for an unattributable foreign 4xx | **Contract** | Error type URI (e.g., `gts.cf.core.errors.err.v1~cf.core.err.not_found.v1~`); `"about:blank"` per RFC 9457 §4.2.1 when the error-middleware catch-all fallback could not attribute a canonical category |
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - the Base Error Schema's `type` property description now says "GTS type identifier for the error category, or \"about:blank\" (RFC 9457 §4.2.1) for an error the platform could not attribute to any canonical category."

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -5,36 +5,114 @@
...
+
 /// Tower middleware function that fills `trace_id` / `instance` on canonical
-/// Problem responses and logs at `warn!` (4xx) / `error!` (5xx).
+/// Problem responses and logs at `warn!` (4xx) / `error!` (5xx); wraps any
+/// other error-status response into a minimal RFC 9457 `Problem`.
 ///
-/// Non-problem responses pass through unchanged. Malformed Problem bodies
-/// are logged at `error!` and returned to the client as-is.
+/// Success/redirect responses pass through unchanged. Malformed Problem
+/// bodies are logged at `error!` and returned to the client as-is.
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed with your suggested diff - now reads "Malformed Problem bodies are logged at `error!`; on an error status they are re-wrapped through the same fallback as any other untyped rejection, and on a non-error status they pass through unchanged."

libs/toolkit/src/api/canonical_error_layer.rs
```diff
@@ -79,19 +181,252 @@ pub async fn canonical_error_middleware(request: Request, next: Next) -> Respons
...
+/// CORS headers and `Vary` are needed for a wrapped error to still pass a
+/// browser's CORS check.
+const PRESERVED_FOREIGN_HEADERS: &[axum::http::HeaderName] = &[
+    header::WWW_AUTHENTICATE,
+    header::RETRY_AFTER,
+    header::ACCESS_CONTROL_ALLOW_ORIGIN,
+    header::ACCESS_CONTROL_ALLOW_CREDENTIALS,
+    header::ACCESS_CONTROL_EXPOSE_HEADERS,
+    header::VARY,
+];
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - `Set-Cookie` is now preserved via `get_all`/`append` (not `get`/`insert`) so multiple cookies survive. Also ran a broader sweep for what else might be missing in the same allowlist: found `WWW-Authenticate` has the identical multi-value latent risk (fixed alongside Set-Cookie, though no current call site actually appends more than one), and real rate-limit/quota headers (`RateLimit-Policy`, `RateLimit-Limit`, `X-RateLimit-Limit/-Remaining/-Reset`) that api-gateway's rate-limit middleware and oagw's `error_response` attach to `CanonicalError`/`Problem` responses - these are reachable via the same malformed-body recovery path this allowlist already exists for, so added them too. All three additions have new tests.

libs/toolkit/src/api/rest/extract/error.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+    tracing::debug!(
+        original_len = message.len(),
+        "extractor rejection message exceeded MAX_FIELD_VIOLATION_DESCRIPTION_CHARS, truncating before sending to client"
+    );
+    let mut truncated: String = message
+        .chars()
+        .take(MAX_FIELD_VIOLATION_DESCRIPTION_CHARS)
+        .collect();
+    truncated.push_str("... (truncated)");
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - reserved space for the truncation suffix before the char take, and updated the test to assert the final length is at most MAX_FIELD_VIOLATION_DESCRIPTION_CHARS.

docs/arch/errors/DESIGN.md
```diff
@@ -163,9 +163,11 @@ See [§ 4. Category Reference](#4-category-reference) for full definitions inclu
...
+There is no alternative path for returning errors. Every REST error response *this platform produces* is built from a `CanonicalError` via `From<CanonicalError> for Problem`. This eliminates inconsistent error formats across gears.
 
-**ADRs**: `cpt-cf-errors-adr-typed-enum-impl`
+This is a producer guarantee, not a claim about what's accepted on ingestion: a response received from a genuinely foreign peer (proxied through `toolkit-gateway::Forwarder`, or returned by a contract-generated RPC client) was never constructed via `CanonicalError` and is not required to be - `canonical_error_middleware`'s fallback (`cpt-cf-errors-adr-error-middleware-catchall`) and `toolkit-contract`'s transport-error mapping both parse it as a plain RFC 9457 `Problem` instead. `Problem`'s own `Deserialize` accordingly follows RFC 9457 §3.1's actual optionality (`detail`/`context`/`title`/`type` all default rather than requiring the fields this platform always happens to emit) - the Base Error Schema below documents this platform's own emission contract, not a requirement placed on foreign peers.
+
+**ADRs**: `cpt-cf-errors-adr-typed-enum-impl`, `cpt-cf-errors-adr-error-middleware-catchall`
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - scoped the universal-Problem guarantee language in DESIGN.md (Single Error Gateway, Fail-Safe Fallback, RFC 9457 Compliance, error-middleware component doc, wire-format table intro, base schema description) and the ADR to exclude ForeignPassthrough-tagged and non-error-status responses, which bypass the middleware's rewriting entirely.

docs/arch/errors/ADR/0006-cpt-cf-adr-error-middleware-catchall.md
```diff
@@ -0,0 +1,271 @@
...
+  is unambiguously this platform's own fault), a foreign 4xx uses RFC 9457
+  §4.2.1's `about:blank` convention instead (`type: "about:blank"`, `title`
+  *and* `detail` both the status's standard reason phrase - a bare 4xx
+  doesn't determine one canonical category over another). If a
+  `CanonicalError` can be recovered from the original response's
+  extensions (e.g. a `Problem` body that failed to deserialize but whose
+  extension survived), that real category is used directly instead of
+  either class-based guess. In every case, the foreign body's own content
+  is never placed on the wire, only logged server-side (capped, timeboxed,
+  escaped), since nothing has vetted it against this platform's "no
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - enrich_problem_response's initial body read is now bounded (64 KiB via a new MAX_PROBLEM_BODY_BYTES, distinct from the 8 KiB diagnostic-log cap) and time-limited (reusing FOREIGN_BODY_READ_TIMEOUT). On limit-exceeded or timeout it now routes through the same wrap_foreign_response fallback as a deserialize failure, rather than the prior empty-body dead end (which was itself not a valid Problem). New test: oversized_problem_json_body_is_wrapped_not_left_empty.

libs/toolkit-canonical-errors/src/problem.rs
```diff
@@ -151,17 +151,47 @@ impl ProblemCategory {
...
     pub title: String,
     pub status: u16,
+    #[serde(default)]
     pub detail: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub instance: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub trace_id: Option<String>,
+    #[serde(default = "default_context")]
     pub context: serde_json::Value,
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed - `status` is now `Option<u16>` rather than a defaulted `u16` (a synthetic 0 would be an actively wrong status, not an honest absence, matching `default_context`'s own reasoning). Every deserialize-then-consume call site with a real status to fall back on normalizes `None` from it: `enrich_problem_response` (middleware) and `map_http_error` (HTTP client) both fill it from the real response status. SSE has no such status and preserves the absence. `TryFrom<Problem>` treats a missing status as "no override recoverable" and falls back to the category's own canonical default rather than erroring. New tests: status_omitted_on_the_wire_deserializes_to_none, try_from_with_no_status_keeps_the_category_default, a_problem_omitting_status_is_normalized_from_the_real_response_status.

Issue #4389, opened 2026-07-31
resource-group: a group is always created in the caller's own tenant


parviz-93:
## Problem Statement

`POST /resource-group/v1/groups` always creates the group in the tenant of the caller. A caller cannot name a target tenant.

The handler derives the tenant from the security context and hands it to the service (`src/api/rest/handlers/groups.rs:95-96`):

```rust
// Derive tenant_id from SecurityContext
let tenant_id = ctx.subject_tenant_id();
```

`create_group` takes that value as a separate parameter (`src/domain/group_service.rs:102-107`), so the request body never reaches it.

There is no other way to state the target:

- `CreateGroupDto` has no `tenant_id` field. It carries `id`, `type`, `name`, `parent_id` and `metadata` (`src/api/rest/dto.rs:173`).
- The group routes carry no tenant path segment and no tenant header (`src/api/rest/routes/groups.rs`).
- `parent_id` does not help. A tenant is not a resource group, so there is no group to point at, and the `am.user_group` type allows only `am.user_group` as a parent.

Who this blocks: a platform administrator that manages several tenants. The administrator selects a child tenant in a console and creates a user group. The group lands in the platform root tenant instead. The console filters its list by the selected tenant, so the new group disappears from view at once. It is not lost. It belongs to the wrong tenant. The console already holds the selected tenant id at the point of the call, and it has nowhere to put it.

## Design / Proposed Solution(s)

*(for discussion)*

- **An optional `tenant_id` on `CreateGroupDto`.** When a caller omits it, the behavior stays exactly as today, so nothing breaks for current callers. When a caller sets it, the gear validates it against the scope of the caller and refuses a tenant outside that scope.
- **Model tenant scopes as resource groups** and allow them in `allowed_parent_types`. Then `parent_id` states the target. This is a larger change, and it alters what a tenant means inside this gear.
- **An acting-scope header** across the write surface of the gear. This keeps the body unchanged but adds a cross-cutting concept.

In every option a child group still inherits its tenant from its parent. A `tenant_id` that conflicts with the tenant of the parent is a 400.

## Definition of Done

- [ ] A caller with sufficient visibility creates a group owned by a stated tenant
- [ ] The read projection reports that tenant in `hierarchy.tenant_id`
- [ ] A target tenant outside the visibility of the caller is refused, and the response does not reveal whether that tenant exists
- [ ] A caller that omits the target keeps today's behavior
[...]

PR #4613, opened 2026-08-19
perf(resource-group,toolkit-db): shorten the transaction, then stop paying for the isolation it held (ABC-2344)

ffedoroff:
Where #4444 removed statements, this removes the time a transaction holds them open, and stops paying for an isolation level the work does not need. Two things grew on top of that while it was in review and are described below: a guard table that makes "one tenant per resource" an invariant the schema enforces, and authorization on the type-registry surface.

---

## The retry loop was dead code

Every repository in this gear maps its errors through `DomainError::database(e.to_string())`, which wraps into `DbErr::Custom`. `is_retryable_contention` matched only `DbErr::Exec`/`Query` and returned `false` for everything else — so a serialization failure raised inside *any* repository call arrived in a shape the classifier refused to recognize. `transaction_with_retry` never retried a single write path here (RG-15).

That also means the jittered backoff added in #4444 could not run, and neither could the give-up logging it feeds. The classifier now checks `DbErr::Custom` against the same message patterns, with five regression tests.

Worth naming for anyone adding metrics later: a retry counter added *before* this fix would have honestly reported zero retries, and zero retries looks like a healthy system.

## Transaction window

**GTS metadata validation moved out of the transaction (RG-09).** Resolving the chained schema is a cross-gear network round-trip, and it is then compiled; both ran with a `SERIALIZABLE` transaction open, and every retry paid for them again. Neither reads database state. The fix is structural: `create_group_inner` and `update_group_inner` no longer take a `TypesRegistryClient`, so the call cannot drift back inside.

**`delete_type` had no transaction at all (RG-02)** — resolve, count and delete on a bare connection, so a concurrent `create_group` of that type could land between the count and the delete.

## Isolation, chosen per operation

| Operation | Level | Why |
|---|---|---|
| create, move, parent-changing update, force delete | `SERIALIZABLE` | a predicate over rows the write does not lock — closure rewrite across a subtree, or `create`'s depth/width check |
| rename/metadata-only update | default | one row by primary key |
| non-force delete | default | one row, with `SELECT ... FOR UPDATE` on the target |
| create type | default | `UNIQUE(schema_id)` holds the invariant at any level |
| delete type | default | `ON DELETE RESTRICT` holds it |
| update type | `SERIALIZABLE` | the removed-parent and removed-membership-type checks are predicates over rows it does not lock |
| add / remove membership | `SERIALIZABLE` | see the guard table below — the level is load-bearing on **both**, or SSI pairs neither |

`update_group` cannot know before `BEGIN` whether it is a move, so it takes a hint from a read it already does and escalates: the authoritative in-transaction read returns `NeedsSerializable` **before any write**, and the operation reruns at `SERIALIZABLE`. Overshooting is harmless; undershooting cannot commit.

These operations do not stop being correct under `SERIALIZABLE` — they simply do not require it. What changes is where contention surfaces: on PostgreSQL's default they wait on a row lock instead of aborting with `40001`. Retry stays in place either way.

**What actually serializes the non-force delete against a concurrent create is the foreign key, and the code says so.** A plain `SELECT` does not wait on `FOR UPDATE`; what blocks is the `FOR KEY SHARE` the RI trigger takes on the parent row when inserting a child, which conflicts with the `FOR UPDATE` the delete holds. Whichever arrives first, the other waits — the delete's lock makes the ordering decidable rather than doing the blocking. Every foreign key into `resource_group` is `ON DELETE RESTRICT`, so no orphan is reachable; this was review-challenged (one engine called it Critical) and survived four independent attempts to construct one.

**And the race the lowered level opens gets a real answer.** Under `SERIALIZABLE` on both sides, a create that lost to a concurrent delete of its parent surfaced as `40001`, retried, and re-read a clean `group_not_found`. With the delete at the backend default, SSI has no second serializable party — the foreign key answers instead, and unmapped that was a 500. `GroupRepository::insert` now classifies the FK violation, and only for `fk_resource_group_parent` (matched by constraint name): the table's other FK, `fk_rg_gts_type`, fails on a concurrent `delete_type` and must not be reported as a missing parent.

**Two constraint classifications came along**, but only where a constraint was already enforcing the invariant: a duplicate type code was a 500 until `SERIALIZABLE` retried into the in-transaction check, and a type still in use answered from `ON DELETE RESTRICT` with no mapping. `is_foreign_key_violation` joins `is_unique_violation` in `toolkit-db`, with tests that they are not confused for each other.

## One tenant per resource, enforced by the schema

`add_membership` checked tenant compatibility by scanning existing membership rows, then inserted — on a pooled connection, not a transaction. The membership PK includes `group_id`, so two concurrent first-memberships from different tenants both passed their own check and both committed (RG-01). `remove_membership` had the same check-then-write shape (RG-14).

A new table, `resource_membership_tenant`, keyed `(gts_type_id, resource_id)`, carries the tenant that owns the resource. Its PK is what serializes the first membership: the claim and the membership insert share one transaction, so a failed insert rolls the claim back rather than pinning a resource with nothing to justify it. The claim uses `ON CONFLICT DO NOTHING` and re-reads unconditionally — a deliberate unique violation would abort the transaction it now shares (SQLSTATE `25P02`) and take the recovery read with it.

**The guard is released on every path that can orphan it.** `remove_membership` re-counts and deletes it when the last membership goes. `force_delete_subtree` deletes a whole subtree's memberships in one statement, so it releases the rows that delete orphaned — checked against the table rather than inferred from the subtree, so a pair still referenced from outside it keeps its row. An orphaned row would not only pin the resource: it holds an `ON DELETE RESTRICT` reference to `gts_type`, so a later `delete_type` failed on the foreign key and reported "group(s) or membership(s) of this type exist" when neither existed.

**All three paths run `SERIALIZABLE`, and that is one invariant, not three choices.** PostgreSQL's SSI only tracks conflicts among transactions that are all at that level, so lowering any one of them lets the other two through — a `READ COMMITTED` add re-adding the tenant a removal is vacating is invisible to that removal's count as anything but a phantom. Both commit, and the resource ends up with a live membership and no guard row. This is written up as TX-04 in the audit report.

## Migration and API change

**There is a migration**, and it does more than create a table: `m20260819_000001_membership_tenant_guard` backfills one guard row per existing `(gts_type_id, resource_id)` pair from the owning group's tenant. Without it the table starts empty on a live database, `add_membership` no longer scans memberships, and the first caller after deploy claims resources it may not own.

A pair that already spans several tenants — never actually forbidden before this table existed — does not fail the migration. It takes the earliest membership's tenant, ties broken by `tenant_id`, and reports every such pair through `tracing::warn!` with a bounded sample so an operator can reconcile the rest. The two backends get separate rankings (`DISTINCT ON` and `ROW_NUMBER()`), and both are exercised against their own engine.

**One additive API change:** `POST /resource-group/v1/groups` accepts an optional `tenant_id` to place a group in another tenant, authorized against the caller's scope. `docs/api/api.json` is regenerated; nothing is removed or renamed.

## Authorization on the type registry, and the bootstrap surface it needed

Type-registry CRUD now goes through the `PolicyEnforcer` like the rest of the gear. That broke the server's boot, which is worth stating plainly rather than leaving in the commit log: account-management registers the RG type schemas it owns from its own `Gear::init`, and there is no phase of gear initialization at which a gated call there can both find a plugin and be admitted by it. `init` is the toolkit's config phase, where types-registry keeps plugin registrations in a staging buffer until `post_init` — `list_instances` returns zero even for a plugin that registered earlier in the same pass, so the PDP fails closed with `PluginNotFound`. Deferring to `post_init` does not help either: the only actor a gear-init path has is platform-scoped, and the static AuthZ plugin rejects nil-tenant callers.

So the SDK gains `ResourceGroupTypeBootstrap`: a narrow, deliberately un-gated trait registered in `ClientHub` beside the existing clients, backed by `TypeService`'s own unscoped methods — the same ones RG's `seed_types` uses at its own init for the identical reason. It takes the same posture as `ResourceGroupReadHierarchy`, and must never be exposed through REST. Its doc records that the carve-out exists only because the GTS type registry lives inside this gear; if the registry is split out, it should not survive the split as-is.

## Measurement (P0)

`resource-group` was the only system gear without `infra/metrics.rs`; nine others have one, so this wires into an existing platform rather than building one. Five instruments: operation duration by operation and outcome, subtree size, closure rows written (the database's own count — the rows never reach the process), isolation escalations, and metadata-validation duration.

Not injected through the constructor: it has ~150 call sites, nearly all tests. The service defaults to a no-op port and `gear.rs` installs the real one, so the domain never mentions OpenTelemetry.

**Not here:** the retry loop's own counters — attempts, serialization failures, exhausted budgets. Those live in `toolkit-db`, which carries no instruments at all today; adding the first is a platform decision that deserves its own review.

## Behaviour changes

None in the API contract beyond the additive field above.

1. **Isolation levels** as tabled above.
2. **Error precedence on create.** Metadata is validated before the type is resolved, so a request with *both* an unregistered type and invalid metadata reports the metadata error.
3. **A create losing the delete race answers 404**, not 500.
4. **A membership whose resource is already claimed by another tenant is rejected** with a `failed_precondition` that deliberately does not name the owning tenant.
5. **Type-registry operations are authorized.** In-process gear-init registration goes through the un-gated bootstrap trait instead.
6. **`TenantNotFound` maps to 404** with the tenant as the reported resource.

## What guards it

Four source-scan rules cover what is not observable as SQL: retry on every transaction in **both** service files, the external call kept out of the transaction, and the non-force delete's row lock — invisible on SQLite, which has no row locks. Each rule is scoped to the function bodies it is about, through a shared `fn_body` helper that panics rather than scanning nothing: an earlier revision scanned whole files, and one rule passed on `update_group`'s literals regardless of what `delete_group` did — verified by hard-coding the level back and watching the rule stay green. It fails now.

**Isolation level was the recorder's documented blind spot, and is not any more.** `SET TRANSACTION ISOLATION LEVEL` bypasses the metric callback, so no statement-count rule could see a downgrade — and lowering one of the guard row's three paths left the whole suite green, including the PostgreSQL tests written for that exact race, because they open their own transactions. The recorder now carries the level each transaction was *asked* to open at, in the task-local that already carried transaction identity, and the audit suite asserts it on all three paths plus the other direction: a non-force delete must stay at the backend default, so "raise everything to `SERIALIZABLE`" cannot satisfy the first three while undoing the saving this branch is about.

Where a test pins a security check or a defect class, it was validated by breaking the thing it guards and confirming it fails. That found a coverage gap: the cross-tenant check on `update_group`'s path had **no** test — deleting it left the whole suite green. The same method later found that the `update_group` isolation escalation and the "removed membership type still in use" rejection ran in no test at all.

## Verification

- `cargo test -p cf-gears-resource-group -p cf-gears-toolkit-db` — **1214 passed, 0 failed**.
- `cargo test -p cf-gears-resource-group --features integration` — **409 passed, 0 failed**, against real PostgreSQL via `testcontainers`.
- `cargo fmt --check`, `clippy --all-targets` and `xtask check-test-container-pins` clean.
- CDSL markers unchanged in every file touched; `make cfs-validate` reports 0 errors.

**PostgreSQL coverage is in this branch**, unlike the earlier revision of this description: `pg_concurrency_test.rs` (11 barrier-synchronized races, each ending in a post-state invariant check against the tables), `pg_smoke_test.rs`, and the backfill's conflict case. Two of those races are worth naming — one drives an add against a force-delete cascade at `SERIALIZABLE` and observes the `40001` pivot cancellation; its negative control runs the same barrier at the backend default and observes the corruption, so the pairing cannot be relaxed without a test failing.

The three remaining `testcontainers` suites from the audit's full apparatus (`pg_membership_filter_test.rs`, `pg_group_filter_test.rs`, `secure_group_scope_postgres.rs`) live on the branch carrying the tenant-scope fixes they exercise.

## Summary by CodeRabbit

* **New Features**
  * Added operational metrics for resource-group lifecycle operations.
  * Added optional tenant targeting when creating groups, with authorization and validation checks.
  * Added authorization enforcement for resource-group type management.
  * Added atomic membership tenant consistency safeguards.
  * Added a dedicated type-bootstrap interface for deployment-time setup.

* **Improvements**
  * Improved transaction isolation, retries, locking, hierarchy validation, and error handling.
  * Added clearer duplicate and foreign-key conflict reporting.

* **Documentation**
  * Clarified transaction isolation, retry behavior, audit coverage, and database configuration options.


gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
```diff
@@ -240,4 +240,144 @@ impl MembershipRepositoryTrait for MembershipRepository {
...
+
+        // @cpt-begin:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-3
+        // Whether the insert above won the claim or lost it to a concurrent
+        // one, this read returns whichever tenant is now on record.
+        GuardEntity::find()
+            .filter(guard_entity::Column::GtsTypeId.eq(gts_type_id))
+            .filter(guard_entity::Column::ResourceId.eq(resource_id))
+            .secure()
+            .scope_with(&sc)
+            .one(db)
```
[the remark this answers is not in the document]
ffedoroff: Not adding the lock protocol, and the reason is testable rather than rhetorical.

Every path that touches `resource_membership_tenant` runs `SERIALIZABLE`: the claim and the release in `membership_service`, and the cascading release inside `force_delete_subtree`, which `delete_group_inner` reaches only under `force` — exactly when `delete_group` selects that level. With both sides there, the interleaving you describe is a two-transaction cycle of rw-antidependencies: the cleanup reads the membership predicate the add's insert falls under, and the add reads the guard row the cleanup deletes. PostgreSQL cancels one side as a pivot and the bounded retry already on both paths reruns it.

`pg_concurrency_test.rs` demonstrates it on a barrier that holds the add uncommitted while the cleanup runs its "still referenced?" check. At `SERIALIZABLE` the run raises `40001`, "Canceled on identification as a pivot", and the post-state invariant holds; `forced_overlap_at_read_committed_orphans_the_guard` runs the same barrier with the cascade at the backend default and observes both sides committing and the guard row orphaned — the corruption this pairing prevents, kept as a permanent negative control.

Explicit `FOR UPDATE` would be redundant under SSI, inert on SQLite, and would add lock-ordering risk: the cascade locks a data-dependent key set while the add locks one key.

Since this comment the invariant is also no longer held by review alone. Lowering `delete_group`'s force branch used to leave the whole suite green — both tests above open their own transactions and never saw the service layer's choice. The query recorder now captures the isolation level each transaction was asked to open at, and the audit suite asserts it on all three paths, plus the other direction so raising everything to `SERIALIZABLE` cannot satisfy them while undoing the saving. Each assertion was checked by lowering the path it guards.

On the third bullet: `delete_membership_guard` is not unconditional at its call site — `remove_membership` deletes the membership, re-counts, and only then releases the guard, all in one transaction. The cascade does the equivalent check against the table after its bulk delete rather than inferring it from the subtree, so a pair still referenced from outside the deleted subtree keeps its row.

gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
```diff
@@ -240,4 +240,63 @@ impl MembershipRepositoryTrait for MembershipRepository {
...
+                .await
+                .map_err(|e| DomainError::database(e.to_string()))?
+                .ok_or_else(|| {
+                    DomainError::database("Guard row disappeared after UNIQUE violation".to_owned())
+                })?;
+            Ok(winner.tenant_id)
+        } else {
+            Err(DomainError::database(err.to_string()))
+        }
+    }
```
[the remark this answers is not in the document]
ffedoroff: Addressed since this comment. The claim and the membership insert now share one transaction, so a failed insert rolls the claim back instead of leaving it behind. The guard is released on both paths that can orphan it: `remove_membership` re-counts and deletes it when the last membership goes, and `force_delete_subtree` releases the rows its bulk membership delete orphaned — checked against the table, so a pair still referenced from outside the subtree keeps its row. Existing databases are handled by a backfill in the migration, which seeds one row per existing `(gts_type_id, resource_id)` pair from the owning group's tenant and reports pairs already spanning several tenants instead of failing the deploy.

Covered by `concurrent_remove_membership_releases_the_guard_exactly_once`, `group_force_delete_releases_orphaned_guard_and_unblocks_type_delete`, `group_force_delete_preserves_guard_for_membership_outside_subtree`, and the backfill cases in `membership_tenant_guard_backfill_test.rs` — including one against real PostgreSQL, since the two engines get separate rankings.

gears/system/resource-group/resource-group/tests/api_rest_test.rs
```diff
@@ -2209,3 +2230,163 @@ async fn rest_route_smoke_all_endpoints_registered() {
...
+// `group_service_test.rs` (TC-GRP-06/07, via `move_group_unscoped`) and
+// error-mapping coverage exists in `domain_unit_test.rs`
+// (`domain_to_problem_cycle_detected_is_400`), but neither one drives the
+// actual HTTP path. These two tests close that gap.
+//
+// They now drive `POST /resource-group/v1/groups/{id}/move` rather than
+// `PUT /resource-group/v1/groups/{id}`: re-parenting was split out of the
+// update payload, so a `parent_id` in a PUT body is a rejected unknown field
+// (see `rest_update_group_rejects_parent_id_field`) and can no longer reach
+// cycle detection at all.
```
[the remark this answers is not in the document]
ffedoroff: There is no move endpoint to send to: `POST /resource-group/v1/groups/{id}/move` does not exist in this gear's REST surface — there is no `/move` route in `src/api/rest/`. Re-parenting is part of the `PUT` payload (`UpdateGroupDto.parent_id`), and `update_group_inner` delegates to the move logic when it changes, which is what these tests exercise. They assert 400 with a `hierarchy` precondition violation, so the `CycleDetected` mapping is covered through the HTTP path. The comment block above the tests states this.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -868,15 +1239,27 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
         // `DomainError::invalid_parent_type` when the parent's type is not in
         // the moved subtree's `allowed_parent_types`).
         let parent_changed = existing.parent_id != req.parent_id;
+        // The fresh read is the authoritative one. If it says this request
+        // moves the group but the transaction was opened below SERIALIZABLE
+        // on a stale hint, stop here -- before the move branch and before the
+        // update below, so nothing has been written -- and let the caller
+        // rerun the whole operation at the right level.
+        if parent_changed && !serializable {
```
parviz-93: Blocking. This attempt runs at the backend default (`TxConfig::default()` at line 549), but the read at line 1151 takes no row lock (`group_repo.rs:492-505` has no `.lock`). `group_repo.update` at line 1276 always writes `parent_id`: the predicate is `Id` only (`group_repo.rs:824`) and `ParentId` is always assigned (`:826`).

If a `SERIALIZABLE` `move_group` commits between line 1151 and line 1276, this UPDATE reverts its `parent_id`. The closure table keeps the ancestry of the move (`rebuild_subtree_closure` at line 1683). The two tables then disagree, and cycle detection and hierarchy reads consume the wrong data. A READ COMMITTED UPDATE by primary key never raises `40001`, and SSI tracks only `SERIALIZABLE` peers, so nothing detects this.

At BASE `update_group` always ran `SERIALIZABLE` (`bdf9987d2:263`), so the loser got `40001` and the retry read the moved parent. Every PUT sends `parent_id` (`resource-group-sdk/src/models.rs:273-277`, "an omitted key is rejected"), so a plain rename takes this path.

Fix: use `find_model_by_id_for_update` at line 1151, as the non-force delete does at line 1496. Under READ COMMITTED, `FOR UPDATE` waits for the in-flight move and returns the fresh parent, so line 1247 escalates. Or add `parent_id` to the UPDATE predicate and return `NeedsSerializable` when 0 rows change.

No test covers this. `pg_concurrency_test.rs` has no rename race. `group_service_test.rs:3697` uses `ParentFlippingGroupRepo`, which flips the parent in the pre-transaction read only (`:3408-3430`, fires once).

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -258,15 +427,134 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
...
+        //
+        // The same row also carries the current `parent_id`, which decides
+        // the transaction's isolation level below. `conn` is scoped to this
+        // block so the pooled connection is back before the transaction asks
+        // for its own.
+        let existing = {
+            let conn = db
+                .conn()
+                .map_err(|e| DomainError::database(e.to_string()))?;
+            let existing = group_repo
```
parviz-93, line 492: High. `find_model_by_id` builds `system_scope()` (`group_repo.rs:497`, `:39-41`), so this read sees every tenant. The first scoped read is at line 1146, inside the transaction. Metadata validation at lines 451-465 runs before it and returns on error.

So a PUT with invalid metadata returns 400 (`DomainError::Validation`, `api/rest/error.rs:37`) for a group in another tenant, and 404 (`GroupNotFound`, `:47-51`) for an unknown id. This tells the caller that the foreign group exists. The 400 text is schema-shaped (`src/domain/validation.rs:188-191`), so it also tells the caller which schema the group is validated against. The oracle works whenever the type schema declares a `metadata` property (`src/domain/validation.rs:175-178`).

At BASE the scoped `find_by_id(tx, scope, group_id)` was the first statement of `update_group_inner` and validation came after it (`bdf9987d2:782-826`). This file states the rule at lines 728-731: a cross-tenant id must look the same as a non-existent id. `docs/toolkit_unified_system/06_authn_authz_secure_orm.md:13` allows an `allow_all` prefetch only with the compensating checks applied.

Fix: call `find_by_id(&conn, &scope, group_id)` (`src/domain/repo.rs:22`) before this line. Then validate the metadata.

`tenant_filtering_db_test.rs:294` does not catch this. It sends `metadata: None` and asserts only `is_err()` and an unchanged row, never which error came back.

gears/system/resource-group/resource-group/src/domain/type_service.rs
```diff
@@ -190,33 +288,68 @@ impl<TR: TypeRepositoryTrait> TypeService<TR> {
...
 
-    /// List GTS type definitions with `OData` filtering and pagination.
+    /// List GTS type definitions with `OData` filtering and pagination
+    /// (`AuthZ`-gated: `read` on [`RG_TYPE_RESOURCE`]).
     pub async fn list_types(
+        &self,
+        ctx: &SecurityContext,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupType>, DomainError> {
+        self.gate(ctx, "read").await?;
```
parviz-93: Guideline (SEM-CODE-007, `docs/checklists/CODING.md:87`: API contracts are honored). `resource-group/docs/DESIGN.md:1279` maps `listTypes` to action `list`. `resource-group/docs/DESIGN.md:1297` defines the vocabulary: `list` for a collection, `read` for a single resource. This gate uses `read`. The sibling collection endpoints use `list` (`group_service.rs:374`, `membership_service.rs:370`). The string `"list"` appears nowhere in this file, and no test pins the action.

If a policy grants `read` on types but not `list`, this gate exposes the full type catalog. The whole gate is new in this PR (BASE had no `enforcer` in this file), so this is a new mismatch.

Fix: use `"list"` here. Update the rustdoc at lines 311-312, which also says `read`.

gears/system/resource-group/resource-group-sdk/src/api.rs
```diff
@@ -263,3 +263,98 @@ pub trait ResourceGroupReadHierarchy: Send + Sync {
...
+/// a carve-out on `resource-group`'s.
+///
+/// # Error envelope
+///
+/// Like [`ResourceGroupClient`], every fallible method returns
+/// `Result<_, CanonicalError>` per [ADR 0005].
+///
+/// [ADR 0005]: https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/errors/ADR/0005-cpt-cf-adr-sdk-canonical-projection.md
+#[async_trait]
+pub trait ResourceGroupTypeBootstrap: Send + Sync {
```
parviz-93, line 345: Guideline (`docs/checklists/CODING.md:94` and `:99`). This trait bypasses `PolicyEnforcer`, and every in-process gear can resolve it from `ClientHub` (`libs/toolkit/src/client_hub.rs:182` has no caller restriction). `docs/checklists/CODING.md:94` asks for a review record for privilege-boundary changes: the guardrail, the rationale, the owner, and the validation. `docs/toolkit_unified_system/16_defect_class_to_control_map.md:137-139` states the rule this deviates from: a bypass granted to a shared client is granted to every consumer of it.

The rationale lives only in this rustdoc block. `resource-group/docs/DESIGN.md:1290` and `:1298` record the `ResourceGroupReadHierarchy` bypass in the AuthZ matrix. This trait has no such entry.

Fix: add an `_(AuthZ bypassed)_` row and a note for this trait next to the `ResourceGroupReadHierarchy` entry in `resource-group/docs/DESIGN.md`. Name the one permitted caller (`register_user_group_types`), the guardrail (never exposed through REST, pinned by `bootstrap_bypasses_policy_enforcer_while_gated_path_denies`), and the sunset condition (type registry split). Also add a source-scan test: no file under `src/api/rest/` names `ResourceGroupTypeBootstrap`. Keep `update_type`: `account-management/src/domain/user_groups/registration.rs:185-193` explains why the two-step patch is necessary.

gears/system/resource-group/resource-group/src/domain/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+//!
+//! The second needs the retry loop's own counters (attempts, serialization
+//! failures, exhausted budgets), and that loop lives in `toolkit-db`, which
+//! carries no instruments at all today. Adding them there is a platform
+//! change touching every gear and belongs in its own review, not smuggled
+//! into a resource-group PR.
+
+/// The write operation a measurement belongs to.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Operation {
```
parviz-93, line 25: Guideline (`docs/toolkit_unified_system/02_gear_layout_and_sdk_pattern.md:273`, MUST). All `struct` and `enum` types in `domain/` that are visible beyond their module must have `#[domain_model]`. `Operation` (line 24), `Outcome` (line 51), `NoopMetrics` (line 109), and `ParentSnapshot` / `MoveOutcome` in `group_service.rs:123,130` do not.

The other 11 domain types in this crate use `#[allow(unknown_lints, de0309_must_have_domain_model)]`. The backing lint `de0309` is `Deny`, but the CI job `cargo gears lint --dylint` is green, so the lint does not fire on this layout. The rule holds by review only.

Fix: add the same allow attribute (or `#[domain_model]`) to the five types.

gears/system/resource-group/resource-group/src/infra/storage/group_repo.rs
```diff
@@ -733,6 +750,40 @@ impl GroupRepositoryTrait for GroupRepository {
...
+                    // and inserting. Answering *that* with "group not found,
+                    // id = parent" would name the wrong resource and the wrong
+                    // cause, so only the parent constraint maps. PostgreSQL
+                    // puts the constraint name in the message; SQLite says
+                    // only "FOREIGN KEY constraint failed", so there the
+                    // answer stays a database error -- unhelpful, but not a
+                    // confident lie, and the race needs concurrent writers
+                    // SQLite does not have.
+                    let msg = e.to_string();
+                    if msg.contains("fk_resource_group_parent")
```
parviz-93: Guideline (RUST-TEST-001, `docs/pr-review/toolkit-rust-review.md:278` and `:290`). No test asserts this mapping. `concurrent_non_force_delete_and_create_child` (`pg_concurrency_test.rs:433`) reaches this arm on Postgres, but its Err/Err branch asserts on `format!("{e1}{e2}")` with `contains("conflict") || contains("not found") || contains("precondition")` (`:506-519`). The Ok/Err branch is the same. `GroupNotFound` is never asserted in that file. If this arm returns `DomainError::database`, the test still passes.

On SQLite the message is `FOREIGN KEY constraint failed` with no constraint name, so this arm returns a 500 there. The constraint name is also a bare literal here and in the migration DDL (`m20260306_000001_initial.rs:56` and `:135`), with nothing that ties them together.

Fix: move the mapping into `fn map_insert_error(e: &ScopeError, id: Uuid, parent_id: Option<Uuid>) -> DomainError`. Unit-test it with `fk_resource_group_parent` (expect `GroupNotFound`) and `fk_rg_gts_type` (expect a database error). In the race test, assert each error of the Err/Err arm on its own. Share the constraint name as a `const`.

gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
```diff
@@ -184,60 +183,153 @@ impl MembershipRepositoryTrait for MembershipRepository {
...
+        // @cpt-begin:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-1
+        // Optimistic: try to read first; if none exists, insert.
+        let existing = GuardEntity::find()
+            .filter(guard_entity::Column::GtsTypeId.eq(gts_type_id))
+            .filter(guard_entity::Column::ResourceId.eq(resource_id))
+            .secure()
+            .scope_with(&sc)
+            .one(db)
+            .await
+            .map_err(|e| DomainError::database(e.to_string()))?;
```
parviz-93: Guideline (RUST-ERR-001, `docs/pr-review/toolkit-rust-review.md:131` and `:144`). This PR adds 11 new `DomainError::database(e.to_string())` sites (type_repo 25 to 28, group_repo 28 to 32, membership_repo 5 to 9). Five are on this guard path: lines 208, 242, 269, 307, 332.

`DomainError::database` wraps the text in `DbErr::Custom` (`src/domain/error.rs:218-219`). The typed `DbErr` is lost, so `sql_err()` cannot classify the error later. `is_retryable_contention` now accepts `DbErr::Custom`, so the retry itself survives. The error class in the give-up diagnostics does not, and `is_unique_violation` / `is_foreign_key_violation` must fall back to message text.

Fix: add `impl From<ScopeError> for DomainError` that maps `ScopeError::Db(db)` to `DomainError::Database(db)` and stringifies only the other variants. `group_repo.rs:913-918` already does this match by hand. Then use `?` at the new sites. The ~50 inherited sites can wait for a follow-up.

gears/system/resource-group/resource-group/src/infra/storage/migrations/m20260819_000001_membership_tenant_guard.rs
```diff
@@ -0,0 +1,221 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let backend = manager.get_database_backend();
+        let conn = manager.get_connection();
+
+        match backend {
+            sea_orm::DatabaseBackend::Postgres => {
+                let sql = r"
+CREATE TABLE IF NOT EXISTS resource_membership_tenant (
```
parviz-93: Guideline (SEM-CODE-007, `docs/checklists/CODING.md:97`: database schemas align with the design). This migration adds a table that the design does not know. `resource-group/docs/DESIGN.md:1312` lists six owned tables. The `#### Table:` sections cover six tables. `resource-group/docs/DESIGN.md:1154` (and `:468`) says: "Tenant scope is not stored on membership rows." This table stores it per `(gts_type_id, resource_id)`. `docs/migration.sql`, which `resource-group/docs/DESIGN.md:247` calls the reference DDL, ends at `resource_group_membership` (line 130).

Fix: add `resource_membership_tenant` to the data-ownership list, add a `#### Table:` section (PK, `tenant_id`, FK to `gts_type` with `ON DELETE RESTRICT`, claim and release lifecycle), correct line 1154, and append the `CREATE TABLE` to `docs/migration.sql`.

gears/system/resource-group/resource-group/src/infra/storage/migrations/m20260819_000001_membership_tenant_guard.rs
```diff
@@ -0,0 +1,221 @@
+// Created: 2026-08-19 by Constructor Tech
+// Updated: 2026-08-24 by Constructor Tech
+// @cpt-dod:cpt-cf-resource-group-dod-sdk-foundation-persistence:p1
+/// Migration: add `resource_membership_tenant` guard table for
```
parviz-93: Recommendation (`CONTRIBUTING.md:389`, PR checklist: code is documented with rustdoc). This 24-line `///` block sits directly above `use sea_orm::{...}` at line 28. A `///` comment attaches to the next item, so this text documents a private `use` statement. rustdoc does not render it. The block holds the backfill rationale and the tie-break rule, which is the most important text in this file.

Fix: change `///` to `//!` on lines 4-27.

gears/system/resource-group/resource-group/src/infra/storage/entity/resource_membership_tenant.rs
```diff
@@ -0,0 +1,22 @@
+// Created: 2026-08-19 by Constructor Tech
```
parviz-93: The other seven files in `entity/` start with a `// @cpt-dod:` traceability marker after the header. This file has none. `cfs validate` still passes, but the new table is not traceable to a Definition-of-Done item.

Fix: add the `// @cpt-dod:` marker for the persistence DoD, the same as the sibling entities. Keep `#[secure(no_tenant, ...)]` as it is: this table is cross-tenant by design, and `no_tenant` is the documented declaration for that (`libs/toolkit-db/src/secure/db_ops.rs:829-831`).

gears/system/resource-group/resource-group/src/domain/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+    Create,
+    Update,
+    Move,
+    Delete,
+    ForceDelete,
+}
+
+impl Operation {
+    #[must_use]
+    pub const fn as_str(self) -> &'static str {
```
parviz-93, line 38: Recommendation (`CONTRIBUTING.md:389`, PR checklist: code is documented with rustdoc). `Operation::as_str` here and `Outcome::as_str` at line 58 have `#[must_use]` and no doc. These strings are the metric label values. Dashboards and alerts depend on them, so a rename is a breaking change for operators. The doc is the place to say that.

Fix: add one `///` line to each: the value is the stable label for the `operation` (or `outcome`) attribute, and a change breaks existing dashboards.

libs/toolkit-db/src/contention.rs
```diff
@@ -82,21 +82,35 @@ const SQLITE_LOCKED_MSG: &str = "database is locked";
...
-                _ => false,
-            }
+            is_contention_message(backend, &msg)
         }
+        DbErr::Custom(msg) => is_contention_message(backend, msg),
+        _ => false,
+    }
+}
+
+fn is_contention_message(backend: DbBackend, msg: &str) -> bool {
```
parviz-93, line 121: Recommendation (`CONTRIBUTING.md:389`, PR checklist: code is documented with rustdoc). `is_contention_message` is new in this PR and has no doc. Every other function in this file has one. The function is the single point where backend dispatch happens, and the `_ => false` arm for unknown backends is a decision worth one sentence.

Fix: add a `///` comment: matches the message against the backend's contention signatures, and returns `false` for a backend with no known signatures.

gears/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -322,6 +389,13 @@ pub trait TypeRepositoryTrait: Send + Sync + 'static {
         type_id: i16,
     ) -> Result<Vec<(Uuid, String)>, DomainError>;
 
+    async fn find_groups_violating_removed_membership_types<C: DBRunner>(
```
parviz-93, line 371: Recommendation (`CONTRIBUTING.md:389`, PR checklist: code is documented with rustdoc). This new trait method has no doc. The sibling `find_groups_violating_removed_parents` at lines 359-378 has a full doc that explains the batch shape and the scope decision. This method makes the same decisions (system scope, subquery, chunked binds) and says nothing.

Fix: add a `///` doc with the same content as the sibling: what it returns (one tuple per violating membership), why it uses the system scope, and the bind-parameter chunking.

parviz-93 (changes requested):
One blocking defect, one high, seven guideline items, four rustdoc recommendations, and cleanup. Details are in the inline comments. The tiers are: Blocking = must fix before merge. High = real defect with a plausible trigger. Guideline = a written project rule is not met, with the rule cited. Recommendation = no hard rule, our suggestion. Cleanup and Minor = quality and hardening.

Verified locally on b6553dfcd: `cargo fmt --all --check` is clean. `cargo test -p cf-gears-toolkit-db -p cf-gears-resource-group` gives 1214 passed, 0 failed, 85 ignored. `cfs validate` gives 0 errors and 20 warnings. CI runs the PostgreSQL suites through `make test-rg-pg` (`.github/workflows/ci.yml:375`).

Items with no diff line to attach to:

- `group_repo.rs:798-801`: the doc on `update` says that both callers read the row inside the same `SERIALIZABLE` transaction. After this PR the rename caller runs at the backend default. That is what makes `rows_affected == 0` reachable. Update the doc.
- `group_repo.rs:915-917` and `:1462-1464`: these comments say that `is_retryable_contention` does not recognize `DbErr::Custom`. `contention.rs:101` in this PR makes that false. Update both comments.
- `membership_repo.rs:117-129`: `insert` maps only the unique violation. An FK violation on `fk_rgm_group_id` returns a 500. This happens when a membership insert loses to a concurrent group delete. Map it to `group_not_found`, the same as `GroupRepository::insert` does now at `group_repo.rs:753`.
- `group_service.rs:1146-1178`: a plain rename reads the group at line 448, again at line 1146, and again at line 1151. It then resolves the type path and loads the full type with both junction tables. `rg_type` is used only at line 1263, inside the parent-changed branch. With metadata, `resolve_type_path_from_id` runs twice (lines 453 and 1176). One locked read inside the transaction removes most of this. It is also the fix for the blocking comment.
- `resource-group/docs/DESIGN.md:67`: the SDK trait list names `ResourceGroupClient` and `ResourceGroupReadHierarchy` only. Add `ResourceGroupTypeBootstrap` with one sentence on why it bypasses `PolicyEnforcer`.
- Commit title: 111 characters. `CONTRIBUTING.md:346` asks for 50. More important: under squash-merge this one commit carries a perf change, a security fix (RG-01, RG-14), a schema migration, an additive API field, and a new authorization surface. List all five in the description, so the history stays readable.

Pre-existing, for a follow-up ticket, not for this PR: `membership_service.rs:94-98` (add) and `:274-278` (remove) compute an `AccessScope` into `_scope` and do not use it. The group is then read with `system_scope()` (`group_repo.rs:497`). A caller in tenant B can add or remove memberships on a group in tenant A. The code is the same at BASE on every relevant line. The simple fix does not fit: `find_by_id` returns the SDK `ResourceGroup`, which has no `gts_type_id` (needed at line 159). Thread `Option<&AccessScope>` into `add_membership_inner`, with `None` for the seeding caller, and reject unless `scope.is_unconstrained() || scope.contains_uuid(OWNER_TENANT_ID, group_model.tenant_id)`.

## Cleanup and hardening (not inline)

**14. `src/domain/type_service.rs:55` — Recommendation**

`RG_TYPE_RESOURCE` is new and `pub`, but it is used only in this file. The one other mention (`resource-group-sdk/src/api.rs:273`) is prose in a doc comment, not an import. Recommendation, no project rule behind it. The sibling constants `RG_GROUP_RESOURCE` and `RG_MEMBERSHIP_RESOURCE` are `pub` because handlers and tests import them.

Fix: make it `pub(crate)`. If a handler or test needs it later, widen it then.

**15. `libs/toolkit-db/src/secure/db.rs:297` — Cleanup**

This doc says that the function sets `IN_TX = true` for the closure. The body now also allocates a fresh id from `NEXT_TX_ID` and scopes `TX_ID` and `TX_ISOLATION` (lines 304-310). Both are read only under `test-support`, but the atomic and the task-locals are compiled into every production transaction. A reader of this doc does not learn that.

Fix: add two sentences: the function assigns a process-unique transaction id and records the requested isolation level, and both are for the test recorder only.

**16. `src/domain/group_service.rs:85` — Cleanup**

This doc says that `metrics` is "bound to the process-global meter, not injected". The constructor sets `NoopMetrics` at line 154, and `gear.rs` injects the OTel recorder through `with_metrics`. So the field is injected, and the default is a no-op struct, not the global meter. `with_metrics` at lines 181-187 repeats the 150-call-site rationale.

Fix: reword this doc: the field defaults to `NoopMetrics`, and the composition root installs the real recorder with `with_metrics`. Keep the rationale in one place.

**17. `libs/toolkit-db/src/secure/error.rs:87` — Cleanup**

This doc lists SQLite extended codes 787 and 1811 as foreign-key codes. 1811 is `SQLITE_CONSTRAINT_TRIGGER` (19 + 7<<8). 787 is `SQLITE_CONSTRAINT_FOREIGNKEY` (19 + 3<<8). sea-orm 2.0.2 (`DbErr::sql_err`) maps only 787 to `ForeignKeyConstraintViolation`. A reader who trusts this doc expects a match that never happens.

Fix: remove 1811.

**18. `src/infra/storage/type_repo.rs:801` — Cleanup**

This query returns one row per membership, and line 812 extends `violating` with one `(type_id, group_id)` tuple per row. A group with five memberships of a removed type gives five equal tuples. `type_service.rs:678-692` then joins the names without dedup, so the rejection message names the group five times. The read itself is bounded by the violating memberships, so this is a message defect, not an unbounded read.

Fix: collect into a set keyed by `(gts_type_id, group_id)` before you resolve names, or `sort_unstable` and `dedup` the tuples.

**19. `src/domain/type_service.rs:196` — Cleanup**

This read is now redundant. `TypeRepository::insert` at `type_repo.rs:393-394` (new in this PR) maps the UNIQUE violation to the same `DomainError::type_already_exists(schema_id)`. The comment at lines 190-196 already says that the read does not close the race and exists for the message. The message is now the same on both paths. So the read is one extra round-trip inside the transaction, in a PR that exists to shorten transactions.

Fix: remove lines 200-203. Keep the `inst-create-type-8` CFS markers (lines 189 and 204) around the insert call, so `cfs validate` stays at 0 errors.

**20. `src/domain/group_service.rs:850` — Cleanup**

`create_group` (lines 208-229) and `create_group_unscoped` (lines 850-872) repeat the same prefix: `validate_type_code`, `validate_name`, the `is_tenant` derivation, `reject_tenant_id_on_tenant_type`, and `validate_metadata_via_gts`. This PR made them diverge. Only `create_group` records `metadata_validation_duration` (lines 218-229) and `record_op` (line 330). The unscoped path records nothing. The divergence comes from moving validation out of `create_group_inner` into both call sites.

Fix: move the shared prefix into one private helper that both paths call. Target-tenant resolution and the PDP gate (lines 236-280) stay in `create_group`.

**21. `src/domain/group_service.rs:544` — Cleanup**

Eight of the ten `transaction_with_retry` call sites pass a literal `TxConfig`. This one and `delete_group` (line 683) derive it from a `bool` (`serializable`, `force`). Line 1247 then reads as `parent_changed && !serializable`, a negated flag. The isolation policy is the central decision of this PR, and it is spelled three ways.

Fix: one small `enum WriteIsolation { RowLocked, Serializable }` with `fn tx_config(self) -> TxConfig`, or a comparison on the existing `TxConfig.isolation` (`libs/toolkit-db/src/secure/tx_config.rs:117`). Then every write names its policy the same way, and line 1247 becomes a level comparison.

**22. `tests/db_behavior_audit_test.rs:1565` — Cleanup**

This rule matches source text: `delete_group.contains("if force {")` and the `TxConfig` literals. A comment that contains `if force {` satisfies it. The exact count `retried == 5` at line 1500 breaks on any behavior-preserving split of the file. The recorder now records `tx_isolation` for every statement, so this one property is observable at run time.

The other text rules cannot move yet: retry wrapping, the `FOR UPDATE` on SQLite (sea-query omits the clause), and the `types_registry` signature check have no runtime signal. `resource-group/docs/db-behavior-audit.md:237-238` already calls them interim.

Fix: replace the `TxConfig` literal check with an assertion on the recorder's `tx_isolation` for the non-force and force delete traces. Keep the other rules, labeled as interim.

**23. `src/domain/membership_service.rs:334` — Cleanup**

The rule "release the guard when nothing references the pair" exists twice. Here it is count-then-delete (`count_memberships` then `delete_membership_guard`). In `group_repo.rs:1107-1116` it is a `NOT IN` bulk delete (`delete_orphaned_membership_guards`). The guard-table operations are split across `MembershipRepositoryTrait` and `GroupRepositoryTrait`. Both are correct today. One invariant with two implementations in two traits is how the next divergence starts.

Fix: keep one implementation. `remove_membership` can call `delete_orphaned_membership_guards(tx, &[(gts_type_id, resource_id)])`. Then `count_memberships` and `delete_membership_guard` can go, and all guard-table code lives in one repository.

**24. `src/domain/group_service.rs:219` — Minor**

This block runs the GTS round-trip and the JSON-schema compile before the `access_scope_with` gate at line 257. A caller that the gate then denies still gets the network call, the compile, and a schema-derived 400 text instead of `permission_denied`. The cost is one round-trip per request, and type schemas are platform-global, so this is hardening, not a leak.

Fix: move lines 218-229 after the gate. The block stays outside the transaction, so the PR goal is not affected.

**25. `tests/pg_concurrency_test.rs:856` — Minor**

Both sides of this test call `transaction_ref_mapped`, which opens with a bare `begin()` (`libs/toolkit-db/src/secure/db.rs:518`), so the test runs at the backend default. Production runs this claim at `SERIALIZABLE` (`membership_service.rs:200`). At `SERIALIZABLE` the lost claim ends in `40001`, then retry, then the pre-read at `membership_repo.rs:201` finds the winner. The `RecordNotInserted` fall-through that this test certifies is not the production outcome on Postgres.

Fix: add a `SERIALIZABLE` variant of this test, or state in the comment that this test covers the default-level path only.

**26. `tests/pg_concurrency_test.rs:896` — Minor**

This `sleep(200ms)` is the one timed wait in the suite. The PR description calls the suite barrier-synchronized. The comment above explains why the sleep exists: the loser must reach its INSERT before the winner is released. A slow CI runner can miss that window, and the test then passes without the conflict it wants to show.

Fix: replace the sleep with a two-way handshake, as the tests at lines 1239 and 1464 do. If the loser cannot signal from inside the driver, keep the sleep and add an assertion that the conflict branch was actually taken.

**27. `src/infra/storage/migrations/m20260819_000001_membership_tenant_guard.rs:190` — Minor**

On SQLite `created_at` is TEXT (`m20260306_000001_initial.rs:165`). sqlx writes `OffsetDateTime` as RFC 3339 through `time`, which removes trailing subsecond zeros. `ORDER BY m.created_at` is then a text sort. `...00.12Z` sorts after `...00.123Z`, but 0.12 s is earlier than 0.123 s. So the "earliest" winner on SQLite can differ from the Postgres `TIMESTAMPTZ` winner. Rows written by `datetime('now')` have no fraction and sort correctly.

Fix: order by `julianday(m.created_at)`. Then the two backends rank the same rows first.

**28. `src/domain/group_service.rs:1733` — Minor**

Both release paths keep or delete the guard row. They never re-point it to another tenant. `count_memberships` counts across all tenants. On a pair that the backfill collapsed from two tenants to one, the guard can stay on the tenant whose last membership was removed, while the other tenant still holds one. That tenant can then no longer add memberships for the resource. The migration warns about such pairs, but it does not say this consequence.

Fix: add one sentence to the migration warning and to `resource-group/docs/features/0004-membership.md`: reconcile a multi-tenant pair before any membership on it is removed.

**29. `src/infra/storage/group_repo.rs:1107` — Minor**

This subquery selects every membership of the type, with no filter on the resource ids of the chunk. The read cost and the `SERIALIZABLE` predicate footprint then follow the popularity of the type, not the size of the subtree. A force delete of one small group under a busy type reads all memberships of that type.

Fix: add `resource_id IN (chunk)` to the subquery, or use a correlated `NOT EXISTS` on `(gts_type_id, resource_id)`.

**30. `libs/toolkit-db/src/contention.rs:101` — Minor**

`is_pg_contention` (line 131) matches the bare substring `40001`. Before this PR, only driver-formatted messages reached it. This arm now also feeds it caller-composed strings, and some of them embed identifiers: `group_repo.rs:1493-1496` interpolates a group UUID, and the `"Type ID {type_id} not found"` sites interpolate an id. A UUID whose hex contains `40001` (about 2.6e-5 per UUID) is then retried as contention. The result is a few wasted re-runs, not a wrong answer.

Fix: match `SQLSTATE 40001` or `(40001)` for the numeric form, and keep the message patterns as the primary signal.

**31. `src/infra/storage/type_repo.rs:542` — Minor**

This message shows the internal SMALLINT `type_id`. The caller has the code, and the count-based message in `delete_type_unscoped` (`type_service.rs:518`) names the code. The two messages for the same conflict then look different, and one of them shows a surrogate key that the API never exposes.

Fix: drop `(type_id {type_id})`, or pass the code into `delete_by_id` and name it here.

**32. `src/domain/membership_service.rs:199` — Minor**

The `allowed_membership_types` check at lines 163-173 reads the type on the pooled connection from line 129, outside this transaction. `update_type` runs `SERIALIZABLE` (`type_service.rs:379`), but SSI can pair only reads inside a `SERIALIZABLE` transaction. So an `update_type` that removes this membership type can commit between line 163 and the insert at line 242, and the insert succeeds. This is the same shape as TX-04 in `resource-group/docs/db-behavior-audit.md`. The read was outside a transaction at BASE too.

Fix: move `load_full_type_by_id` and the check inside this transaction, or add one line to the audit document that names this race as accepted.

**33. `resource-group-sdk/src/models.rs:254` — Minor**

`CreateGroupRequest` is a plain `pub struct` with all-`pub` fields, no `#[non_exhaustive]`, and no constructor. A new field breaks struct-literal construction in every consumer. `src/domain/seeding.rs:157-164` in this repo already had to add `tenant_id: None`. The SDK is published at 0.2.x, so this is a 0.3.0-category change under pre-1.0 SemVer. The PR description says "one additive API change" and does not mention the SDK.

Fix: state the SemVer category in the description and the changelog. Or mark the struct `#[non_exhaustive]` and add a constructor, so the next field is not a break.

**34. `src/domain/type_service.rs:123` — Minor**

At BASE this file had three `@cpt-flow` anchors for type management: create-type (`:40`), update-type (`:211`), delete-type (`:344`). At HEAD only the update-type anchor survives, at line 343 above `update_type_unscoped`. `cfs validate` still reports 0 errors, so this is traceability drift, not a broken gate. The flow markers inside the bodies (`inst-create-type-*`, `inst-delete-type-*`) have no function-level anchor to hang from.

Fix: restore `// @cpt-flow:cpt-cf-resource-group-flow-type-mgmt-create-type:p1` above this function, and the delete-type anchor above `delete_type_unscoped` at line 490.

gears/system/resource-group/resource-group/src/api/rest/dto.rs
```diff
@@ -181,6 +181,9 @@ pub struct CreateGroupDto {
     pub name: String,
     /// Parent group ID (null for root groups).
     pub parent_id: Option<Uuid>,
+    /// Optional target tenant for the created group.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tenant_id: Option<Uuid>,
```
parviz-93, line 186: This field implements #4389 (the first option in that issue): a caller names the target tenant, the gear checks it against the caller's scope, and a tenant outside the scope gets a 404 that does not reveal whether the tenant exists. `docs/api/api.json` has the new field.

One item of that issue's Definition of Done is still open: `gears/system/resource-group/docs/openapi.yaml` does not list `tenant_id` in the `CreateGroup` schema (line 1238, `properties:`). Please add it there too, so the two API documents agree.

After merge, #4389 can be closed with a reference to this PR.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Fixed in e0206426a. `update_group_inner`'s authoritative read is now `find_model_by_id_for_update`, so under READ COMMITTED it waits for the in-flight move and `parent_changed` sees the committed parent -- the escalation happens before anything is written.

Two guards: `group_rename_does_not_revert_a_concurrent_reparent` (group_service_test.rs) and the source rule `static_rule_update_group_locks_the_row_it_may_rewrite`. The decorator needed extending for the first one -- it now serves a stale parent from unlocked reads and performs a real reparent including `rebuild_subtree_closure`, otherwise the two halves of the race cancel out. Verified it fails without the fix, on the closure/parent_id mismatch.

One correction to the report: there is no REST `move` endpoint, so the concurrent writer is not `move_group`. The same race is reachable from two concurrent PUTs -- one changing `parent_id` (escalates to SERIALIZABLE), one renaming. Conclusion unchanged.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Fixed. The scoped `find_by_id(&conn, &scope, group_id)` now runs before metadata validation, so a group in another tenant and an unknown id both answer `GroupNotFound`.

New test: `tenant_isolation_update_cross_tenant_is_not_an_existence_oracle`. Verified it fails without the fix -- `Validation failed: Failed to resolve GTS type '...' for metadata validation`. The shared stub registry gained a mode where schema lookups fail with `Internal` instead of `NotFound`, since `NotFound` makes `validate_metadata_via_gts` skip validation entirely.

gears/system/resource-group/resource-group/src/domain/type_service.rs
ffedoroff: Fixed -- `list_types` gates on `list`, and the rustdoc at the call site now cites the action vocabulary and the sibling collection endpoints rather than repeating `read`.

gears/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: Partly. `DESIGN.md` now carries the `_(AuthZ bypassed)_` row for `ResourceGroupTypeBootstrap` plus a note naming the one permitted caller (`register_user_group_types`), why gear init leaves no gated path available, the guardrail (never exposed through REST, pinned by `bootstrap_bypasses_policy_enforcer_while_gated_path_denies`) and the sunset condition. The SDK trait list at the top of the document lists it too.

Not done yet: the source-scan test asserting that no file under `src/api/rest/` names the trait.

gears/system/resource-group/resource-group/src/domain/metrics.rs
ffedoroff: Fixed -- `#[allow(unknown_lints, de0309_must_have_domain_model)]` added to `Operation`, `Outcome`, `NoopMetrics` and to `ParentSnapshot` / `MoveOutcome`, so all five match the other 11 domain types in the crate.

gears/system/resource-group/resource-group/src/infra/storage/membership_repo.rs
ffedoroff: Addressed by removal rather than by the `From<ScopeError>` impl.

The guard table is gone in e0206426a -- tenant ownership is derived from the membership rows again -- so all five guard-path sites went with it. `membership_repo.rs` is now at 6 `DomainError::database` sites against 5 at BASE, i.e. one net new (the FK classification on `insert`). The typed-`DbErr` refactor you describe is still worth doing and is not in this branch.

gears/system/resource-group/resource-group/src/infra/storage/migrations/m20260819_000001_membership_tenant_guard.rs
ffedoroff: Resolved by dropping the table rather than by documenting it.

`resource_membership_tenant`, its migration and the backfill are gone. The "one tenant per resource" check is back to a read derived from the membership rows -- `has_membership_in_other_tenant`, one statement with `LIMIT 1` -- sharing one SERIALIZABLE transaction with the insert it gates, which is what RG-01 actually needed: a predicate read followed by a write into the same predicate is the write skew SSI cancels.

So DESIGN.md's six owned tables and the "tenant scope is not stored on membership rows" line stand as written, and `docs/migration.sql` is unchanged. New PostgreSQL coverage instead: `forced_overlap_at_serializable_keeps_one_tenant` drives both transactions through the overlap window and observes the `40001`, with `forced_overlap_at_read_committed_splits_the_resource` as a permanent negative control.

gears/system/resource-group/resource-group/src/infra/storage/migrations/m20260819_000001_membership_tenant_guard.rs
ffedoroff: Moot now -- the migration file is gone along with the guard table (see the reply on the table's own thread).

gears/system/resource-group/resource-group/src/infra/storage/entity/resource_membership_tenant.rs
ffedoroff: Moot now -- the entity is gone along with the guard table. The marker was added first, then the file removed.

gears/system/resource-group/resource-group/src/domain/metrics.rs
ffedoroff: Fixed -- both `as_str` methods now document that the returned string is the stable label value for the `operation` / `outcome` attribute and that renaming it breaks existing dashboards and alerts.

libs/toolkit-db/src/contention.rs
ffedoroff: Fixed -- `is_contention_message` now has a doc covering what it matches and the `_ => false` arm for a backend with no known signatures.

gears/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Fixed -- the method now documents what it returns (one tuple per violating pair, deduplicated), the system-scope decision and the bind-parameter chunking, matching its sibling. The dedup is new too: the query returns one row per membership, so a group with several memberships of a removed type used to be named once per membership in the rejection message.

gears/system/resource-group/resource-group/src/api/rest/dto.rs
ffedoroff: Fixed -- `tenant_id` is now in the `CreateGroup` schema in `gears/system/resource-group/docs/openapi.yaml`, describing the scope check and the 404-rather-than-403 behaviour, so it agrees with the generated `docs/api/api.json`.

parviz-93 (approved):
**Approve.** Re-checked `e0206426a` against review 5045110500.

Fixed, with tests:
- Blocking: `update_group_inner` now takes `find_model_by_id_for_update` before it decides `parent_changed`. `group_rename_does_not_revert_a_concurrent_reparent` performs a real reparent and checks the closure table.
- High: `update_group` does the scoped read before metadata validation. `tenant_isolation_update_cross_tenant_is_not_an_existence_oracle` covers it.
- `list_types` gates on `list`. `de0309` on all five domain types. `ResourceGroupTypeBootstrap` recorded in the DESIGN.md AuthZ matrix. `openapi.yaml` has `tenant_id`, which closes the last item of #4389.

The design reversal is the better fix. `has_membership_in_other_tenant` is one `LIMIT 1` read inside the `SERIALIZABLE` insert transaction, so two first memberships from different tenants are the write skew SSI cancels. It uses `idx_rgm_gts_type_resource`, so the predicate lock is narrow. The two `forced_overlap_*` tests pin the `40001` and the negative control.

Verified locally on `e0206426a`: `cargo fmt --all --check` clean, `cargo test -p cf-gears-toolkit-db -p cf-gears-resource-group` 1212 passed, `cfs validate` 0 errors.

Not blocking, for a follow-up:
- No source-scan test that `src/api/rest/` never names `ResourceGroupTypeBootstrap`.
- No unit test for the `fk_resource_group_parent` mapping; the race test still asserts on both errors concatenated.
- `ScopeError::Db` is still stringified at the new sites; a `From<ScopeError> for DomainError` would keep it typed.
- `create_group` validates metadata before the `access_scope_with` gate.

gears/system/resource-group/resource-group-sdk/src/api.rs
MikeFalcon77, line 345: **HIGH**

The doc comment on `ResourceGroupTypeBootstrap` states it must never be exposed through REST, but nothing enforces that beyond the comment itself.

Any code holding a `ClientHub` reference — not only another gear's `init` — can resolve `Arc<dyn ResourceGroupTypeBootstrap>` at any lifecycle phase and call its PolicyEnforcer-bypassing methods, silently reopening the authorization gap this PR closes.

Make the init-only restriction structural: pass the bootstrap capability explicitly into dependent gears' `Gear::init` instead of registering it in the globally resolvable `ClientHub`, or block resolution of this trait once the gear passes `post_init`.

gears/system/resource-group/resource-group/src/gear.rs
```diff
@@ -136,8 +152,17 @@ impl Gear for ResourceGroup {
...
             .register::<dyn ResourceGroupReadHierarchy>(read_client);
 
+        // TEMPORARY: bootstrap-only, un-gated surface for another gear's
+        // `init` to register RG types before AuthZ is reachable — see
+        // `ResourceGroupTypeBootstrap`'s doc comment ("Temporary — revisit
+        // when the type registry becomes its own gear").
+        let type_bootstrap_client: Arc<dyn ResourceGroupTypeBootstrap> =
+            Arc::new(RgTypeBootstrapService::new(type_svc));
+        ctx.client_hub()
+            .register::<dyn ResourceGroupTypeBootstrap>(type_bootstrap_client);
```
MikeFalcon77, line 176: **MEDIUM**

`ResourceGroupTypeBootstrap` is registered in `ClientHub` the same way as `ResourceGroupClient`, with no scoping that limits resolution to the init phase.

The 'init-only, never from REST' contract for this trait exists only as a doc comment; there is no runtime or compile-time check preventing a later handler or background task from resolving it and skipping authorization entirely.

Restrict resolution to the init lifecycle — pass the bootstrap client explicitly to dependent gears' `init` rather than registering it in the globally resolvable `ClientHub`, or gate lookup of this type once `post_init` completes.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -258,15 +428,145 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait> GroupService<GR, TR> {
...
+    #[allow(clippy::too_many_arguments)]
+    async fn attempt_update_group(
+        db: &Db,
+        group_repo: &Arc<GR>,
+        type_repo: &Arc<TR>,
+        scope: &toolkit_security::AccessScope,
+        group_id: Uuid,
+        req: &UpdateGroupRequest,
+        profile: &QueryProfile,
+        serializable: bool,
```
MikeFalcon77: **MEDIUM**

`attempt_update_group` selects the transaction isolation level via a bare `bool` parameter (`serializable`) instead of a dedicated isolation-level type.

A boolean carries no meaning at call sites and is easy to pass wrong, especially since this PR's entire point is choosing the correct isolation level per operation.

Replace the `serializable: bool` parameter with `TxIsolationLevel` (or an equivalent enum) so every call site is self-describing.

gears/system/resource-group/resource-group/tests/pg_concurrency_test.rs
```diff
@@ -0,0 +1,1037 @@
...
+        .expect("create group B");
+
+    let resource_id = "res-1".to_owned();
+
+    // Both tenants race to be the first membership on the same resource.
+    // The check and the insert share one `SERIALIZABLE` transaction (RG-01),
+    // so the loser is either cancelled on `40001` and retried into the
+    // rejection, or reads the winner's committed row outright. Either way it
+    // must surface as a domain error, not a bare database one.
+    let (r1, r2) = tokio::join!(
```
MikeFalcon77, line 618: **MEDIUM**

`concurrent_add_membership_from_two_tenants_claims_exactly_one` races two `add_membership` calls with plain `tokio::join!` and no barrier, so the two transactions are not guaranteed to actually overlap.

This file's own comment on the neighboring scenario admits that interleaving is not guaranteed to land inside the RG-01 contention window, so this test can pass by exercising the ordinary sequential rejection path instead of the race it claims to cover.

Force the overlap deterministically with the oneshot-channel barrier pattern already used by the adjacent scenarios in this file, or drop the RG-01 claim from this test's assertions.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
```diff
@@ -1292,75 +1416,244 @@ async fn negative_control_read_paths_produce_no_write_statements() {
...
+    // `delete_group`.
+    assert!(
+        delete_inner.contains("find_model_by_id_for_update"),
+        "the non-force delete must lock the target row before checking its \
+         children: without it, lowering the isolation level below SERIALIZABLE \
+         opens the window it was closing"
+    );
+
+    // And the level must still be chosen, not hard-coded back.
+    assert!(
```
MikeFalcon77, line 1737: **MEDIUM**

`static_rule_non_force_delete_locks_before_it_decides` only checks that `"if force {"`, `"TxConfig::serializable()"`, and `"TxConfig::default()"` all appear somewhere in `delete_group`'s source text, without checking which branch of the `if` each `TxConfig` call is actually inside.

A refactor that swaps which branch uses which `TxConfig` would still pass this text-based check, defeating its purpose as a guard against exactly that regression.

Assert the branch-to-config association directly — scan the true/false arms of the `if` separately, or check the isolation level actually opened via the query-recorder mechanism already used elsewhere in this file instead of a source-text scan.

gears/system/resource-group/resource-group/src/domain/seeding.rs
```diff
@@ -43,7 +43,7 @@ pub async fn seed_types<TR: TypeRepositoryTrait>(
     // @cpt-begin:cpt-cf-resource-group-algo-type-mgmt-seed-types:p1:inst-seed-2
     for seed in seeds {
         // @cpt-begin:cpt-cf-resource-group-algo-type-mgmt-seed-types:p1:inst-seed-2a
-        match type_service.get_type(&seed.code).await {
+        match type_service.get_type_unscoped(&seed.code).await {
```
MikeFalcon77, line 46: **MEDIUM**

`seed_types` now calls the unscoped `get_type_unscoped`/`update_type_unscoped`/`create_type_unscoped` methods to avoid needing AuthZ at gear init, but every existing test in `seeding_test.rs` wires `TypeService` with an allow-all `PolicyEnforcer`.

A regression back to the gated calls would still pass every existing seeding test, so there is no test protecting the fix this PR makes here.

Add a seeding regression test that wires `TypeService` with a deny-all `PolicyEnforcer` and asserts `seed_types` still succeeds, mirroring the equivalent test already present in `type_service_test.rs`.

gears/system/resource-group/resource-group/src/domain/membership_service.rs
```diff
@@ -172,52 +173,94 @@ impl<GR: GroupRepositoryTrait, TR: TypeRepositoryTrait, MR: MembershipRepository
...
+                    // @cpt-end:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-3
+                    // @cpt-end:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-2
+                    // @cpt-end:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-1
+
+                    // @cpt-begin:cpt-cf-resource-group-flow-membership-add:p1:inst-add-memb-9
+                    // @cpt-begin:cpt-cf-resource-group-flow-membership-add:p1:inst-add-memb-10
+                    // @cpt-begin:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-4
+                    // @cpt-begin:cpt-cf-resource-group-algo-membership-check-tenant-compat:p1:inst-tenant-check-5
+                    if owned_elsewhere {
+                        debug!(
```
MikeFalcon77, line 248: **LOW**

The tenant-incompatibility `debug!` log in `add_membership` dropped the `group_id` field that the prior version logged, leaving only `resource_type`/`resource_id`.

Without `group_id`, a rejected `add_membership` call cannot be correlated to the specific group involved when debugging from logs alone.

Add `group_id` back as a structured field on the `debug!` call alongside `resource_type` and `resource_id`.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Fixed in 2ca1788a6. `serializable: bool` on `attempt_update_group`/`update_group_inner` is now `WriteIsolation` (`RowLocked` / `Serializable`) with a `tx_config()` method. Both call sites in `update_group` are self-describing now (`WriteIsolation::Serializable` instead of a bare `true`), and the `!serializable` check became `!isolation.is_serializable()`.

gears/system/resource-group/resource-group/tests/pg_concurrency_test.rs
ffedoroff: Reworded rather than barriered. `add_membership` is called through the public API here, unlike the raw-repository scenarios in this file that use the oneshot-handshake pattern -- there's no injection point inside the service call to force the exact overlap. Reworked the comment and dropped the RG-01 attribution from the panic message: this test is a black-box check that the outcome holds under either interleaving, not a proof of the write-skew abort itself -- that proof is deterministic and already covered by `forced_overlap_at_serializable_keeps_one_tenant` / `..._splits_the_resource`.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
ffedoroff: Fixed. Added `split_if_else`/`brace_close` to isolate each arm of `if force { ... } else { ... }` by brace depth and assert on it independently, instead of checking that both `TxConfig` literals appear somewhere in `delete_group`. Verified the fix actually catches the regression: swapped the two `TxConfig` calls locally and confirmed the rule fails before reverting.

gears/system/resource-group/resource-group/src/domain/seeding.rs
ffedoroff: Fixed -- added `seed_types_bypasses_policy_enforcer` in `seeding_test.rs`, wiring `TypeService` with the deny-all enforcer and asserting all three branches (`created`/`unchanged`/`updated`) still succeed, mirroring `bootstrap_bypasses_policy_enforcer_while_gated_path_denies`.

gears/system/resource-group/resource-group/src/domain/membership_service.rs
ffedoroff: Fixed -- `group_id` is back on the `debug!` call alongside `resource_type`/`resource_id`.

gears/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: Fixed in bad05ab4c. `RgTypeBootstrapService` now seals itself (`AtomicBool`, checked first in all three methods) once `register_rest` runs -- the toolkit runtime only calls that after every gear's `init`/`post_init`, in both runtime profiles, so the one sanctioned caller (account-management's `init`) has always already run by then. `register_rest` also removes the `ClientHub` registration itself.

This is stronger than gating a future lookup alone: `seal()` lives on the instance, so it also revokes an `Arc` a caller already resolved and held from inside the window -- the case your suggested remediation (block resolution after `post_init`) wouldn't have covered.

Scoped to this crate, no `ClientHub`/`Gear` trait change. `ResourceGroupReadHierarchy` has the identical shape and predates this PR; left alone as a separate concern.

One honest gap: in the out-of-process profile, `register_rest` runs after `run_start_phase`, so a hypothetical background task started during another gear's `start()` could theoretically resolve this before the seal. No gear does that today -- only `init` ever touches this trait -- but it's not an absolute guarantee for the future.

New test: `bootstrap_fails_closed_once_sealed` (`type_service_test.rs`).

gears/system/resource-group/resource-group/src/gear.rs
ffedoroff: Same fix as the sibling comment on `api.rs` -- see that reply for the mechanism and its one open caveat (OoP `start()` ordering). Short version: `register_rest` now removes the `ClientHub` registration and calls `seal()` on the kept concrete instance, closing both a future lookup and an already-held `Arc`.

gears/system/resource-group/resource-group-sdk/src/api.rs
MikeFalcon77: Checked. seal() plus ClientHub::remove in register_rest, and register_rest runs after every gear's init/post_init in both run profiles. This closes it, including the already-held-Arc case. Good fix.

gears/system/resource-group/resource-group/src/gear.rs
MikeFalcon77: Same fix as the api.rs comment. Checked, confirmed fixed.

gears/system/resource-group/resource-group/src/domain/group_service.rs
MikeFalcon77: Checked. serializable: bool is now WriteIsolation. Fixed.

gears/system/resource-group/resource-group/tests/pg_concurrency_test.rs
MikeFalcon77: Checked. You dropped the RG-01 claim instead of adding a barrier, and the real barrier-based proof is in forced_overlap_at_serializable_keeps_one_tenant / forced_overlap_at_read_committed_splits_the_resource. That's a valid fix, not just a rename. Confirmed.

gears/system/resource-group/resource-group/tests/db_behavior_audit_test.rs
MikeFalcon77: Checked. split_if_else checks each branch on its own now, not just that both strings exist in the file. Fixed.

gears/system/resource-group/resource-group/src/domain/seeding.rs
MikeFalcon77: Checked. seed_types_bypasses_policy_enforcer test is there and wires a deny-all enforcer. Fixed.

gears/system/resource-group/resource-group/src/domain/membership_service.rs
MikeFalcon77: Checked. group_id is back on the debug! call. Fixed.

MikeFalcon77 (approved):

PR #4618, opened 2026-08-20
fix(authz): type native group predicates

### LINKED ISSUE #4644: authz: make native group predicates type-safe and PostgreSQL-compatible
## Summary

Native `InGroup` and `InGroupSubtree` authorization predicates do not currently carry the resource-group member-handle type needed to evaluate membership safely. Their SecureORM SQL also compares native entity identifiers with the opaque `TEXT` identifiers stored by resource-group membership in a way that is not executable for UUID-backed entities on PostgreSQL.

The native path must be type-qualified, PostgreSQL-compatible, capability-negotiated, tenant-bounded, and fail closed when the required metadata or projections are unavailable.

## Current behavior

- Group scope filters do not identify which external GTS member-handle type their `resource_id` values belong to.
- Membership rows for different resource types can therefore be conflated when they contain the same opaque external ID.
- PostgreSQL cannot directly compare a UUID entity column with `resource_group_membership.resource_id TEXT`.
- Casting membership IDs to UUID is not safe because RG identifiers are intentionally opaque and valid member types may use non-UUID IDs.
- A resource can advertise or receive a native group predicate without a trusted membership-type mapping.
- Group predicates constructed outside the high-level PEP compiler can bypass property and tenant-pairing assumptions unless SecureORM enforces the same invariants.
- Empty or unsupported group selectors in bundled PDP implementations must not silently widen to tenant-only access.
- Debug-query assertions do not prove that the generated SQL executes on PostgreSQL.

## Expected behavior

- A PEP resource descriptor explicitly associates an AuthZ resource with its RG member-handle GTS schema ID.
- Native group capabilities are advertised only when the resource has that mapping and the querying service has the required local tables.
- The response compiler rejects unsolicited native predicates and missing/invalid hierarchy metadata.
- Native group predicates target the resource `id` property and are AND-paired with an `owner_tenant_id` predicate.
- SecureORM independently enforces the same property, type, and tenant invariants for directly constructed scopes.
- Membership rows are qualified by resolving the external type through `gts_type.schema_id` to the RG-local `gts_type_id`.
- SQL casts the entity identifier to text and never casts opaque membership IDs to UUID.
- Missing mappings, malformed UUID hierarchy keys, empty selectors, unknown types, and unavailable capabilities fail closed.
- Services unable to execute native predicates receive PDP-expanded explicit `In` predicates or a deny; the PEP does not silently degrade or widen predicates.

## Acceptance criteria

- [ ] Add an explicit RG membership-type mapping to `ResourceType` and propagate it into typed scope filters.
- [ ] Preserve source compatibility for existing constructors and low-level compilation APIs while making untyped native filters deny safely.
- [ ] Enforce negotiated `GroupMembership` and `GroupHierarchy` capabilities.
- [ ] Restrict native group predicates to `id` and require `owner_tenant_id` in the same AND constraint.
- [ ] Compile membership SQL using `CAST(entity_id AS TEXT)` and a `gts_type.schema_id` lookup.
- [ ] Ensure direct SecureORM scope construction cannot bypass the type/property/tenant invariants.
- [ ] Make bundled PDP plugins deny empty or unsupported group selectors rather than widening scope.
- [ ] Add real PostgreSQL execution coverage for UUID/text matching, opaque IDs, duplicate IDs across member types, unknown types, tenant isolation, direct groups, and group subtrees.
- [ ] Document projection prerequisites, canonical identifier representation, capability degradation, and consumer-specific expression-index guidance.

## Non-goals

- Enabling native group pushdown in every downstream service. Each consumer must advertise capabilities only after its required RG tables are co-located or projected.
- Creating entity-side expression indexes from the generic toolkit. Consuming gears own their entity tables and migrations.
- Changing RG's opaque `resource_id` storage contract.

## Implementation

Draft PR: #4618

Downstream tracking: VHP-2344

fdlockgraf:
## Description
Fix native `InGroup` and `InGroupSubtree` scope compilation so resource-group membership is both type-safe and PostgreSQL-compatible.

- carry an explicit RG member-handle GTS type from `ResourceType` into compiled scope filters
- suppress group capabilities for resource descriptors without that mapping and reject native predicates that were not negotiated
- require UUID hierarchy keys, `id` as the group-scoped property, and `owner_tenant_id` in the same AND constraint
- enforce the same tenant/type invariants for direct SecureORM scope construction
- deny empty or unsupported group selectors instead of widening them to tenant-only scope
- document canonical opaque-ID encoding and the consuming-gear expression-index pattern
- qualify `resource_group_membership` through `gts_type.schema_id` / `gts_type_id`
- cast the querying entity ID to `TEXT` instead of casting RG's opaque `resource_id` to UUID
- retain the existing low-level compiler and group-filter constructors for source compatibility; legacy untyped paths now deny safely
- make the bundled TR PDP honor capability negotiation and configure RG to advertise its canonical native capabilities
- add PostgreSQL execution tests for UUID/text matching, duplicate IDs across member types, opaque non-UUID IDs, type and tenant isolation, unknown types, and descendant groups
- document the mapping and projection requirements

## Type of Change
- [x] Documentation update

## Testing
- [x] Unit tests pass (`cf-gears-toolkit-security`, `cf-gears-authz-resolver-sdk`, `cf-gears-toolkit-db`, `cf-gears-tr-authz-plugin`, `cf-gears-resource-group`)
- [x] Integration tests pass (`make test-pg`: 538/538)
- [x] New PostgreSQL regression tests added
- [x] Architecture lints pass (`make dylint`)
- [x] GTS documentation validation passes (`make gts-docs`: 798 files)
- [x] Strict Clippy passes for all modified crates and targets

The full workspace `make clippy` currently stops on an unrelated pre-existing `collapsible_if` warning in `libs/rustls-corecrypto-provider/tests/handshake_smoke.rs` under local Rust 1.97.0. The modified crates pass with `-D warnings -D clippy::perf`.

## Documentation
- [x] Authorization and SecureORM documentation updated
- [ ] API documentation updated (not applicable)

## Checklist
- [x] Self-review completed
- [x] Tests pass
- [x] Commit includes DCO sign-off

## Compatibility / Versioning
This is a non-breaking security/correctness fix. New typed APIs are additive; existing public constructors and `compile_to_access_scope` remain source-compatible. Untyped native group predicates now fail closed rather than querying ambiguous membership rows.

## Related Issues
Fixes #4644

Downstream tracking: VHP-2344

## Summary by CodeRabbit

- **New Features**
  - Added type-aware resource-group membership and subtree authorization.
  - Added tenant-scoped membership filtering and capability negotiation.
  - Group membership now supports UUID and non-UUID resource identifiers.

- **Bug Fixes**
  - Prevented overly broad access when mappings, capabilities, or tenant constraints are missing.
  - Unsupported or empty group selectors now deny access instead of being ignored.
  - Membership filters exclude unrelated resource types and groups.
  - Authorization failures now return appropriate access-denied responses without exposing sensitive details.

- **Documentation**
  - Expanded guidance and examples for membership configuration, projections, and capability behavior.


gears/system/authz-resolver/authz-resolver-sdk/src/constraints.rs
```diff
@@ -92,7 +92,10 @@ impl InPredicate {
 
 /// Group membership predicate: resource is visible if it belongs to any of the listed groups.
 ///
-/// Compiles to: `property IN (SELECT resource_id FROM resource_group_membership WHERE group_id IN (group_ids))`
+/// The PEP compiler combines this wire-level predicate with the current
+/// [`crate::pep::ResourceType`]'s RG member-handle type. `SecureORM` compiles the
+/// result to a type-qualified membership subquery and casts the entity property
+/// to text to match RG's opaque `resource_id` storage.
```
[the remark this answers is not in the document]
fdlockgraf: Addressed in 5af78da6. Both predicate descriptions now document that a missing `ResourceType` membership mapping fails native compilation closed and that degraded PDP responses must use explicit `InPredicate` values.

docs/arch/authorization/RESOURCE_GROUP_MODEL.md
```diff
@@ -55,14 +55,26 @@ Whether and which tables to project depends on the deployment topology and acces
...
 >
-> **Architecture guidance:** default to consuming degraded `in` predicates from PDP. The `in_group` and `in_group_subtree` predicates are natively executable within the RG gear; domain services that choose not to project the membership table rely on PDP capability degradation.
-
-PEP within the RG gear compiles `in_group`/`in_group_subtree` predicates into SQL subqueries using the membership table. Domain services without the membership projection receive degraded `in` predicates and do not need group-related projection tables for authorization filtering.
+> **Architecture guidance:** default to consuming degraded `in` predicates from PDP. The `in_group` and `in_group_subtree` predicates are natively executable wherever the membership/type tables (and closure table for subtree predicates) are co-located or projected. Domain services that choose not to project the membership table rely on PDP capability degradation.
```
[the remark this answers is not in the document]
fdlockgraf: Addressed in 5af78da6. The architecture guide now has an explicit predicate/capability/table matrix: `in_group` requires membership + `gts_type`, while `in_group_subtree` requires those plus closure and both group capabilities. It also states that capabilities must be omitted when projections are unavailable, that the PEP does not degrade automatically, and that a PDP unable to expand to explicit `in` predicates must deny. The resolver README mirrors the contract.

fdlockgraf:
Addressed the CodeRabbit follow-up in `3a9b121f` and `5af78da6`:

- centralized the shared typed-membership SQL and external GTS type resolution for direct/subtree predicates;
- added an enforcer-level test proving an untyped resource suppresses configured group capabilities and rejects an unsolicited native predicate;
- extended the real PostgreSQL subtree test with the canonical closure self-row and direct ancestor membership (wrong-type isolation was already covered and remains covered);
- corrected the `ScopeFilter` SQL documentation to describe `gts_type.schema_id` resolution;
- documented projection prerequisites, capability omission, PDP-owned explicit-`in` degradation, and missing-mapping fail-closed behavior.

Performance disposition: the entity-side cast direction remains intentional because RG member IDs are opaque. `ca800385` documents the consumer-specific composite expression index. A synthetic PostgreSQL plan over 200,000 tenant rows changed from a sequential/hash-semi plan (~25.2 ms) to nested expression-index probes (~0.19 ms). No generic migration was added because toolkit-db does not know consumer tables. No extra membership index is needed for this SQL shape: RG already has `PRIMARY KEY (group_id, gts_type_id, resource_id)` plus `idx_rgm_gts_type_resource (gts_type_id, resource_id)`.

Validation completed: strict Clippy for affected crates, SDK/security/toolkit-db tests, the two real PostgreSQL group-scope tests, `make dylint`, `make gts-docs`, formatting, and diff checks.

gears/system/resource-group/resource-group/src/domain/group_service.rs
```diff
@@ -35,7 +35,8 @@ use crate::domain::validation;
 pub const RG_GROUP_RESOURCE: ResourceType = ResourceType::from_static(
     GROUP_RESOURCE_TYPE,
     &[pep_properties::OWNER_TENANT_ID, pep_properties::RESOURCE_ID],
-);
+)
+.with_group_membership_type(GROUP_RESOURCE_TYPE);
```
ffedoroff: **HIGH — `RG_GROUP_RESOURCE`'s new membership-type mapping can never resolve to a real row**

`RG_GROUP_RESOURCE` is declared here with `.with_group_membership_type(GROUP_RESOURCE_TYPE)`, where `GROUP_RESOURCE_TYPE` (`resource_group_sdk::gts::GROUP_RESOURCE_TYPE`) is the literal string `"gts.cf.core.rg.group.v1~"`.

`validate_type_code` (`domain/validation.rs`) requires every registered `gts_type.schema_id` to start with `RG_TYPE_PREFIX = "gts.cf.core.rg.type.v1~"` — note `rg.type` vs `rg.group`, two different strings. That validation runs on **every** path that can insert a `gts_type` row: the gated `create_type`, `seeding::seed_types`, and the bootstrap surface `RgTypeBootstrapService` all funnel through `TypeService::create_type_unscoped`, which calls `validate_type_code` before insertion. There is no migration or seed SQL that inserts `gts_type` rows directly.

I verified this by calling `create_type_unscoped` with `code: GROUP_RESOURCE_TYPE.to_owned()` directly: it is rejected with
```
Validation failed: Type code must start with prefix 'gts.cf.core.rg.type.v1~', got: 'gts.cf.core.rg.group.v1~'
```

Since `cond.rs`'s `membership_type_subquery` resolves the membership type via `SELECT id FROM gts_type WHERE schema_id = '<membership type>'`, and no supported path can ever create that row, the generated subquery for `RG_GROUP_RESOURCE` is:
```sql
SELECT id FROM gts_type WHERE schema_id = 'gts.cf.core.rg.group.v1~'
```
which can never match. Every `InGroup`/`InGroupSubtree` predicate compiled against `RG_GROUP_RESOURCE` therefore always ANDs in a never-true condition and returns zero rows against a real database — contradicting the new comment in `gear.rs` ("RG owns the canonical membership and closure tables, so it can execute both native group predicate shapes").

**Not caught by any test in this PR.** The PostgreSQL tests in `libs/toolkit-db/tests/group_scope_postgres.rs` use a synthetic member-type string with the correct `rg.type.v1~` prefix, not the real `GROUP_RESOURCE_TYPE`. The only test that uses the real `RG_GROUP_RESOURCE` with `InGroup` (`tenant_filtering_db_test.rs::group_based_in_group_predicate_produces_combined_scope`) only asserts the compiled `AccessScope` structure and never executes SQL against a database.

To be fair about severity: this is not currently exploitable or user-visible — no production PDP in this repo currently emits `InGroup`/`InGroupSubtree` for RG's own resource requests, and on `main` (before this PR) the same predicate against `RG_GROUP_RESOURCE` compiled to a `uuid = text` comparison that PostgreSQL rejects outright, so this PR didn't break a working flow — it replaced a loud SQL error with a silent empty result and an inaccurate capability claim.

There's also a broader inconsistency this exposes: `validation.rs` explicitly documents that a *membership* resource type (as opposed to a registered *group* type) need not carry the `rg.type` prefix — external gears are expected to register their own types under their own paths. But nothing in this repo currently exercises that path with a real, executable native predicate either. Worth a follow-up: does any registered `gts_type` schema_id currently exist that this mapping could correctly point at, or does `RG_GROUP_RESOURCE` need a different mapping entirely (or none)?

**Suggested fix:** either remove `.with_group_membership_type(GROUP_RESOURCE_TYPE)` from `RG_GROUP_RESOURCE` until there's a real, registrable type it should point at, or add a PostgreSQL execution test (not just a structural `AccessScope` test) that proves `InGroup`/`InGroupSubtree` against `RG_GROUP_RESOURCE` actually returns rows for a real membership row.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/compiler.rs
```diff
@@ -151,46 +250,64 @@ fn compile_constraint(
...
+                    "InGroupSubtree",
+                    &p.property,
+                )?;
                 (
                     p.property.as_str(),
-                    ScopeFilter::in_group_subtree(&p.property, ancestor_ids),
+                    ScopeFilter::in_group_subtree_typed(&p.property, membership_type, ancestor_ids),
                 )
             }
             Predicate::InTenantSubtree(p) => {
```
ffedoroff, line 416: **MEDIUM — the new `TenantHierarchy` capability requirement for `InTenantSubtree` breaks an already-merged, unrelated test (`create_group_target_tenant_test.rs::in_tenant_subtree_scope_is_denied_fail_closed` in `resource-group`)**

This branch now requires `Capability::TenantHierarchy` to have been negotiated before an `InTenantSubtree` predicate is allowed to compile (via `require_negotiated_capabilities`, ~line 381). `resource-group`'s own `PolicyEnforcer` (built in `gear.rs`) only advertises `Capability::GroupMembership` and `Capability::GroupHierarchy` — never `TenantHierarchy`.

There is a pre-existing test (from an earlier, already-merged feature, VHP-2162 — `git log` shows it predates and is untouched by this branch) that builds a `PolicyEnforcer` with no capabilities and a mock `AuthZResolverClient` that returns an `InTenantSubtree`-only constraint, to pin that `GroupService::create_group`'s cross-tenant check fails closed via its own `scope.contains_uuid(...)` fallback, landing on a clean `DomainError::TenantNotFound`.

I reproduced both sides directly:
- On current `main` (without this branch's compiler.rs changes): `cargo test -p cf-gears-resource-group --test create_group_target_tenant_test in_tenant_subtree_scope_is_denied_fail_closed` → **passes**.
- With this branch rebased on top of current `main`: the same test → **fails**:
  ```
  panicked at .../create_group_target_tenant_test.rs:321: expected DomainError::TenantNotFound, got: InternalError
  ```

Mechanism: the capability check now fails compilation (`EnforcerError::CompileFailed`) *before* `create_group` ever reaches its own `contains_uuid` fallback, and `resource-group`'s `From<EnforcerError> for DomainError` maps `CompileFailed` to `DomainError::InternalError` — so the same request now surfaces as a 500-shaped internal error instead of the previous clean 404-shaped domain error. It's the only failure across the full `resource-group` test suite; everything else is green.

To be clear, I don't think the fix here is to *remove* the capability gate — `resource-group` genuinely has no `tenant_closure` projection, so a nonconforming PDP that got this predicate past compilation could go on to hit a real `relation does not exist` SQL error, which the gate correctly prevents. The gate is defensible on its own merits, distinct from the uuid/text motivation behind gating `InGroup`/`InGroupSubtree`. But this branch needs to either update `create_group_target_tenant_test.rs` to expect the new error shape, or give `create_group`'s cross-tenant path a way to distinguish "capability negotiation failed" from "generic internal error" so it can still answer with a clean domain error instead of a 500. Since this branch is currently `CONFLICTING` against `main`, this collision likely wasn't visible in the PR's own CI run against its original (older) base.

libs/toolkit-security/src/access_scope.rs
```diff
@@ -450,18 +525,53 @@ impl ScopeFilter {
         ))
     }
 
-    /// Create a group membership filter.
+    /// Create an untyped group membership filter that compiles to deny-all.
+    ///
+    /// Retained for source compatibility. New callers should use
+    /// [`Self::in_group_typed`] so membership rows can be safely qualified.
     #[must_use]
     pub fn in_group(property: impl Into<String>, group_ids: Vec<ScopeValue>) -> Self {
```
ffedoroff, line 545: **MEDIUM — the "source-compatible" legacy constructors silently change behavior for any existing caller**

`ScopeFilter::in_group`/`in_group_subtree` (and the corresponding `InGroupScopeFilter::new`/`InGroupSubtreeScopeFilter::new`) are kept for source compatibility, per the PR description ("retain the existing low-level compiler and group-filter constructors for source compatibility; legacy untyped paths now deny safely").

That's true at the type level, but it understates the behavior change: any pre-existing caller of these constructors compiled and ran real membership SQL before this PR. After this PR, the identical call compiles fine but silently produces a deny-all condition (`membership_resource_type` defaults to `""`, and `cond.rs`'s `typed_membership_condition` returns `None` for an empty type, which the caller ANDs in as `Expr::value(false)`). No compile error, no panic, no log line at the call site — the query that used to return real rows now silently returns zero.

This is fail-closed, so it's not a security regression, but it's a debugging trap: a caller migrating to this PR's changes with no code changes on their end gets a quiet, hard-to-diagnose "always empty" result instead of a signal that they need to switch to `in_group_typed`/`in_group_subtree_typed`. Consider either a `#[deprecated]` annotation on the untyped constructors so callers get a compile-time nudge, or a `tracing::warn!` in the deny-all path in `cond.rs` naming which untyped filter triggered it, so this doesn't look identical to "the caller's actual group membership is empty" in production logs.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer.rs
```diff
@@ -312,6 +342,24 @@ impl PolicyEnforcer {
 
         let bearer_token = ctx.bearer_token().cloned();
 
+        // Native group predicates are only executable when the resource
+        // descriptor supplies the RG member-handle type needed to qualify
+        // `resource_group_membership.gts_type_id`. Suppress group capabilities
+        // for untyped resources so the PDP takes its degraded expansion path
+        // instead of returning a predicate that must fail compilation.
+        let capabilities = self
```
ffedoroff: **LOW/MEDIUM — capability advertisement isn't validated for internal consistency, so a `PolicyEnforcer` can advertise a capability set that can never actually compile**

This filter suppresses `GroupMembership`/`GroupHierarchy` when the resource has no `group_membership_type`, which is good. But it doesn't check two other invalid combinations before sending them to the PDP:

- A `PolicyEnforcer` can be configured with `Capability::GroupHierarchy` alone, without `Capability::GroupMembership`. `compiler.rs`'s `require_negotiated_capabilities` for `InGroupSubtree` requires *both* (`&[Capability::GroupMembership, Capability::GroupHierarchy]`), so a PDP that takes the advertised `GroupHierarchy` at face value and returns an `InGroupSubtree` predicate will have it rejected at compile time even though the capability was technically "advertised."
- A resource can declare `group_membership_type` on a property other than `id` (native group predicates are hard-restricted to the `id`/`RESOURCE_ID` property in `compiler.rs`'s `require_resource_id_group_property`), and nothing here catches that mismatch before the capability is advertised to the PDP.

In both cases, the PEP ends up advertising a capability it cannot actually execute, and the PDP has no way to know until its predicate is rejected — the negotiation looks successful but isn't. I didn't find a production misconfiguration that hits this today, but it seems worth validating/normalizing the capability set (or panicking early in a debug assertion) at `ResourceType`/`PolicyEnforcer` construction time rather than relying on every future caller getting the combination right by convention.

docs/toolkit_unified_system/06_authn_authz_secure_orm.md
```diff
@@ -139,30 +139,50 @@ Each resource type declares which properties the PEP can compile from PDP constr
...
+    &[pep_properties::OWNER_TENANT_ID, pep_properties::RESOURCE_ID],
+)
+.with_group_membership_type("gts.cf.core.rg.type.v1~example.core.mygear.user.v1~");
+```
+
+This value is not inferred from the AuthZ resource name: gears may use different
+GTS paths for policy matching and RG membership storage. `PolicyEnforcer`
+suppresses `GroupMembership` and `GroupHierarchy` capabilities for resource
+descriptors without this mapping, causing the PDP to use its degraded explicit
+`In` path. The querying database must also contain or project
```
ffedoroff: **LOW — this doc overstates `InGroup`'s prerequisites relative to `InGroupSubtree`**

This section states that a resource capable of native `InGroup`/`InGroupSubtree` predicates needs the RG member-handle mapping, and the surrounding text a few lines below ("the querying database must also contain or project `resource_group_membership`, `resource_group_closure`, and RG's `gts_type` registry") reads as if both predicate shapes need all three tables.

`InGroup` (direct membership, no ancestor traversal) never touches `resource_group_closure` — only `InGroupSubtree` does, because it needs the closure table to expand ancestor groups into the full group set before joining against membership. `docs/arch/authorization/RESOURCE_GROUP_MODEL.md`'s table (added by this same PR) gets this right and lists the required tables per predicate separately:

| Native predicate | Required local tables |
|---|---|
| `in_group` | `resource_group_membership`, `gts_type` |
| `in_group_subtree` | `resource_group_membership`, `gts_type`, `resource_group_closure` |

Worth aligning this file's wording with that table so a reader doesn't come away thinking `in_group` alone requires projecting the (much larger) closure table.

gears/system/resource-group/resource-group/src/domain/group_service.rs
fdlockgraf: Fixed in 5a363845a.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/compiler.rs
fdlockgraf: Fixed in 5a363845a.

libs/toolkit-security/src/access_scope.rs
fdlockgraf: Fixed in 5a363845a.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer.rs
fdlockgraf: Fixed in 5a363845a.

docs/toolkit_unified_system/06_authn_authz_secure_orm.md
fdlockgraf: Fixed in 5a363845a.

gears/system/resource-group/resource-group/src/domain/group_service.rs
ffedoroff: Verified independently against `5a363845a`: `GROUP_RESOURCE_TYPE` (`resource-group-sdk/src/gts.rs:74`) resolves to `gts.cf.core.rg.group.v1~`, which never satisfies `validate_type_code`'s required `gts.cf.core.rg.type.v1~` prefix (`domain/validation.rs:9,30`) — confirms the original silent-zero-rows diagnosis. The mapping is removed in `group_service.rs:59-62` with a doc comment explaining why, and `gear.rs:77-86` no longer advertises the capability in production wiring. LGTM.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/compiler.rs
ffedoroff: Verified independently against `5a363845a`: the capability check in `compile_constraint` (`compiler.rs:296-320`) now runs before shape validation (`:323+`), the new `UnadvertisedCapabilities` variant (`compiler.rs:49,76-81`) maps to `AccessDenied`/403 (`resource-group/domain/error.rs:253-283`), and REST returns a generic 403 with details only in debug logs (`api/rest/error.rs:156-160`). Both the new fail-closed scenario and the original `TenantNotFound` case are covered by separate tests (`create_group_target_tenant_test.rs:266-289` and `:295-335`), and `api_rest_test.rs:2392-2434` confirms no predicate/capability names leak in the response body. LGTM.

libs/toolkit-security/src/access_scope.rs
ffedoroff: Verified independently against `5a363845a`: all four legacy constructors (`access_scope.rs:289-292,356-359,540-543,569-572`) carry `#[deprecated(since = "0.8.2")]` pointing at the typed replacements, and the structured `warn_group_filter_denied` (`toolkit-db/src/secure/cond.rs:69-76`, called at `:238,256,265,278,303`) logs predicate/property/cause only — no filter values. LGTM.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer.rs
ffedoroff: Verified independently against `5a363845a`: `with_capabilities` warns when `GroupHierarchy` is configured without `GroupMembership` (`enforcer.rs:359-367`), per-request executability requires both a membership-type mapping and a resolvable `id` property (`enforcer.rs:441-444`), and unmet conditions suppress the corresponding capability before advertisement (`enforcer.rs:461-472`). Covered by `enforcer_tests.rs:315-384`. LGTM.

docs/toolkit_unified_system/06_authn_authz_secure_orm.md
ffedoroff: Verified independently against `5a363845a`: `06_authn_authz_secure_orm.md:166-188` and `RESOURCE_GROUP_MODEL.md:65-72` now correctly separate `in_group` (needs `resource_group_membership` + `gts_type`) from `in_group_subtree` (additionally needs `resource_group_closure`), and are consistent with each other. LGTM.

libs/toolkit-db/src/secure/cond.rs
```diff
@@ -106,42 +252,63 @@ where
...
-                let group_values = scope_values_to_sea_values(gf.group_ids());
-                let subquery = Query::select()
-                    .column(Alias::new(rg_tables::MEMBERSHIP_RESOURCE_ID))
-                    .from(Alias::new(rg_tables::MEMBERSHIP_TABLE))
-                    .and_where(
-                        Expr::col(Alias::new(rg_tables::MEMBERSHIP_GROUP_ID)).is_in(group_values),
-                    )
-                    .to_owned();
-                and_cond = and_cond.add(col.into_expr().in_subquery(subquery));
+                let Some(group_values) = uuid_scope_values_to_sea_values(gf.group_ids()) else {
```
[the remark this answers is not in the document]
ffedoroff: Re-reviewed this independently, specifically the question of what an empty `IN` list actually does at the query-builder level rather than assuming the worst case. A few points worth separating:

1. **No privilege-escalation vector.** `and_where`/`is_in` never silently drops the condition — it always emits a `SimpleExpr`. The only two possible renders are an early-`false` collapse (deny) or a syntactically empty `IN ()` that errors out as `DbErr` (500). Neither widens access; worst case is a query error, not an authorization bypass.
2. **Pre-existing, not introduced by this PR.** The same missing `is_empty()` guard existed at this call site before this PR — the plain `In` filter arm handles selector lists the same way and isn't touched here either. This PR added typing, the non-UUID guard, and `warn_group_filter_denied`, but didn't touch the empty-list case on any of these arms.
3. **Broader than group selectors.** The gap isn't specific to `InGroup`/`InGroupSubtree` — the plain `In` arm has the same shape.
4. **`docs/arch/authorization/DESIGN.md:718` (PEP requirement #11) does call for fail-closed handling here** and explicitly names the `IN ()` risk, so closing this gap is legitimate — just as a `cond.rs`-level hardening item (explicit `is_empty()` → `warn_group_filter_denied` + `Expr::value(false)`, plus a regression test covering an empty group/ancestor list), not something specific to the native-group work in this PR.

Given (1)–(3) I'd classify this MEDIUM rather than the reported Major/bypass severity, and track it as a follow-up hardening ticket on `cond.rs` (covering the `In` arm too) rather than block this PR on it — unless someone has concrete evidence that the pinned sea-query/sea-orm versions render an empty `IN` some other way than described above.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/compiler.rs
```diff
@@ -127,70 +292,123 @@ pub fn compile_to_access_scope(
...
     supported_properties: &[&str],
-) -> Result<ScopeConstraint, String> {
+    group_membership_type: Option<&str>,
+    negotiated_capabilities: Option<&[Capability]>,
+) -> Result<ScopeConstraint, ConstraintFailure> {
+    // Capability negotiation is checked for every predicate BEFORE any shape
+    // validation, so a predicate that is both unadvertised and malformed
+    // still classifies as the typed negotiation violation — the highest-
+    // priority failure class, which enforcing services map to a denial.
+    for predicate in &constraint.predicates {
```
ffedoroff, line 314: **HIGH — a `Constraint` with empty `predicates: []` inside a non-empty `constraints` list compiles to `allow_all()`, even under `require_constraints(true)`**

Not a regression introduced by this PR specifically, but it's directly in the code path this PR modifies, and I think it should block merge as a fail-open defect.

- `Constraint.predicates` (`constraints.rs:28`) is a plain `Vec<Predicate>` with no minimum-length validation, and `EvaluationResponseContext.constraints` deserializes straight off the wire from the PDP response — nothing rejects `Constraint { predicates: [] }` at that boundary.
- The capability-negotiation loop starting here (this line, `for predicate in &constraint.predicates`) simply doesn't iterate when `predicates` is empty, and neither does the second loop further down that builds `filters` — so `compile_constraint` returns `Ok(ScopeConstraint::new(vec![]))` for an empty-predicates constraint: a *successful*, empty constraint, not an error.
- Back in `compile_to_access_scope_impl`, `constraints.iter().all(ScopeConstraint::is_empty)` (line 281) only checks whether every *compiled* constraint ended up empty — it does not consult `require_constraints`. The `require_constraints` gate (line 225) only fires when the *outer* `constraints` list itself is empty.

Net effect: a PDP response body like `{"decision":true,"context":{"constraints":[{"predicates":[]}]}}` — a non-empty outer list containing one degenerate constraint — compiles to `AccessScope::allow_all()` even when the caller asked for `require_constraints(true)`.

This is a concrete violation of `docs/arch/authorization/DESIGN.md:714` (PEP requirement #7): *"If a constraint has empty `predicates: []` ... treat constraint as malformed -> deny all."*

It also has a live consumer: `resource-group/src/domain/group_service.rs:336-337` treats `scope.is_unconstrained()` as sufficient to permit `create_group` across tenants — so a degenerate empty-predicate permit from the PDP would grant unrestricted cross-tenant group creation. `usage-collector/src/domain/authz.rs:517-530,802-813` independently discovered and defends against exactly this case (its own comment: *"allow_all only arises from a degenerate empty-predicate permit"*) — `resource-group`'s `create_group` has no equivalent guard.

Suggested fix: reject `Constraint { predicates: [] }` as a compile error (fail-closed) inside `compile_constraint`, consistent with `DESIGN.md` §7, rather than letting it fold into the `all(is_empty)` → `allow_all()` branch here. No existing test covers this exact shape — `compiler_tests.rs`'s `require_constraints`-empty test only covers a fully empty *outer* `constraints` list, not a non-empty list containing a degenerate constraint.

fdlockgraf: Fixed in e5c2f82ec: compile_constraint now rejects empty-predicates constraints as malformed (fail-closed, DESIGN.md PEP requirement #7), and the all-constraints-empty -> allow_all() fold is removed as unreachable. A valid sibling constraint still survives (OR semantics). Both require_constraints modes and the mixed case are pinned in compiler_tests.rs.

libs/toolkit-db/src/secure/cond.rs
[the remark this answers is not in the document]
fdlockgraf: Fixed in 54caca6d6: both group arms now deny and warn on an empty selector before building the membership subquery, mirroring the malformed-UUID handling, with direct and subtree regression tests.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/compiler.rs
```diff
@@ -34,13 +34,19 @@
...
+use crate::models::{BarrierMode, Capability, EvaluationResponse};
 
 /// Error during constraint compilation.
+///
+/// Marked `#[non_exhaustive]`: variants have been added before (most recently
+/// `UnadvertisedCapabilities`), and downstream crates consume this enum
+/// through a pinned toolkit, so exhaustive matches there must not break on a
+/// cargo-compatible patch release.
 #[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
fdlockgraf: Fixed in cf53ec6b5: cf-gears-authz-resolver-sdk bumped to 0.4.0 (leftmost non-zero component), so the new ConstraintCompileError variant and #[non_exhaustive] ship as a breaking release rather than a caret-compatible 0.3.x update. All in-workspace dependents updated.

ffedoroff (approved):
Approved from my side

entropyshift (approved):

fdlockgraf:
Superseded by #4744 — same head commit (75613a206); the head branch was renamed to fix/type-native-group-predicates, which a cross-repo PR cannot follow. All review threads here are resolved and their fixes are in the new PR's commit.

Issue #4616, opened 2026-08-20
Pin database container versions in one place: PostgreSQL 18 floor, PostgreSQL 19 for graph-storage, MySQL 9.7


MikeFalcon77:
## Problem

Nearly every integration-test fixture in the workspace calls `Postgres::default()` or `Mysql::default()` from `testcontainers-modules`. Those defaults are hardcoded in the crate:

- `postgres:11-alpine` — PostgreSQL 11, EOL since November 2023
- `mysql:8.1` — EOL

Consequences:

- We test against EOL database versions.
- The version is pinned only transitively, through `Cargo.lock`. Bumping `testcontainers-modules` silently changes the database version under every test in the repo.
- Pins have already started to diverge. `resource-group` sets `with_tag("16-alpine")` locally (`gears/system/resource-group/resource-group/tests/pg_smoke_test.rs:114`, with a comment that PG 11 lacks built-in `gen_random_uuid()`), and the usage-collector plugin pins `timescale/timescaledb:2.17.2-pg16`. That is three different Postgres versions in one repository.
- ADR-0002 already records this as a blocker: `docs/arch/secure-orm/ADR/0002-secure-property-graph-policy.md:384-387`.

## Goal

A single place that defines the database images used by the whole workspace:

- PostgreSQL 18 as the floor, everywhere.
- PostgreSQL 19 for the future graph-storage gear, which needs SQL/PGQ (`GRAPH_TABLE`).
- MySQL 9.7 everywhere.

## Tags

Verified against Docker Hub on 2026-08-20:

| Constant | Value | Note |
| --- | --- | --- |
| `POSTGRES_TAG` | `18-alpine` | exists |
| `POSTGRES_GRAPH_TAG` | `19beta3-alpine` | exists; beta3 is the latest. There is no GA `19-alpine` tag yet |
| `MYSQL_TAG` | `9.7` | exists; latest patch is 9.7.2 |
| `TIMESCALEDB_TAG` | `2.29.2-pg18` | latest pg18 build, non-oss variant to match the current pin |
| `MARIADB_TAG` | `11.8` | currently a floating `lts` tag in the outbox bench; confirm the exact pin during implementation |

## Proposed change

Add a crate at `libs/test-containers`, package name `cf-gears-test-containers`, with `publish = false`.

[...]

PR #4639, opened 2026-08-20
feat(secure-orm): implement ADR-0002 — safe property-graph (SQL/PGQ) queries

vasylcf:
Implements `docs/arch/secure-orm/ADR/0002` (merged in #4605), and #4616 as its
prerequisite. Seven commits, each independently verifiable — worth reading in
order rather than as one diff, since the 62 ledger test files in the stat are a
mechanical `sed` from the first commit.

## What ships

**`libs/toolkit-sea-orm-pgq`** — a typed AST and renderer for
`GRAPH_TABLE … MATCH … COLUMNS` and for the `CREATE`/`DROP PROPERTY GRAPH` DDL.
`sea_query` models none of it. Security-agnostic by construction: no
`AccessScope`, no `ScopableEntity`, no runner, and no dependency that could bring
one. The crate documentation says outright that using it from a gear is unsafe,
because nothing in it embeds scope.

**`toolkit-db::secure::pgq`**, behind a `pgq` feature that implies `pg` —
`SecureGraphSelect`, reachable only from `SecureSelect<E, Scoped>`, with the
scope injected into every element pattern, vertices and edges alike.

**The scope compiler is now addressable and fallible.**
`build_scope_predicate::<E>(scope, address)` generalises `build_scope_condition`:
`ColumnAddress::Table` renders as before, `GraphElement { var, .. }` renders
`dst.tenant_id`. One compiler serves selects, CTE bodies and graph elements.

**`libs/test-containers`** — database image pins in one place, and the workspace
floor raised from `postgres:11-alpine` (EOL since 2023) to `18-alpine`.

## The three policies

Policy 1, 2 and 3 are each enforced where they can still be enforced usefully:

- **Elements are addressed by entity type**, never by label. `PathBuilder` takes
  `J: VertexOf<G>` / `J: EdgeOf<G>`, so an entity outside the graph — or an edge
  where a vertex belongs — is a type error. One label may span several element
  tables, so a label-addressed element would carry several security mappings.
- **Policy 2 is checked before any condition is compiled.** An entity that
  resolves no scope column compiles to exactly the same predicate under a real
  scope as a legitimate deny-all does, so afterwards the two are
  indistinguishable. The platform's closure and membership tables are correctly
  shaped for what they are and consequently ineligible.
- **One declaration is the source of both the DDL and the labels.** The
  `PROPERTIES` list is derived from the entity, not repeated by the caller,
  because a column absent from it is invisible to `MATCH` — not an error, just
  silently unfilterable.

## Subtree scopes are servable

The open question in the ADR is settled the way the probe measured it. No
subquery is accepted inside a pattern predicate in any form, so `InGroup`,
`InGroupSubtree` and `InTenantSubtree` compile to a correlated sibling `FROM`
item instead — placed once, referenced by every element:

```sql
SELECT "cf_graph"."neighbour" FROM "graph_node",
  (SELECT "descendant_id" FROM "tenant_closure" WHERE …) AS "__cf_scope_0_0",
  GRAPH_TABLE("kb_pgq" MATCH
    ("a" IS "node" WHERE "a"."tenant_id" = "__cf_scope_0_0"."descendant_id")
    -["e" IS "edge" WHERE "e"."tenant_id" = "__cf_scope_0_0"."descendant_id"]->
    ("b" IS "node" WHERE "b"."tenant_id" = "__cf_scope_0_0"."descendant_id")
    COLUMNS ("b"."id" AS "neighbour")) AS "cf_graph"
WHERE "graph_node"."tenant_id" IN (SELECT …)
```

A caller with nowhere to put that second `FROM` item declares
`SiblingSupport::Rejected` and gets an error rather than a dropped filter — which
is also the ADR's documented v1 fallback if the shape does not survive GA.

Two details there are load-bearing rather than incidental. `DISTINCT` belongs to
the correlated form only: correlating turns a semi-join into a join, so a
membership relation with duplicate keys would return a resource authorized
through two groups twice — while inlined as `IN (subquery)` it cannot, and adding
one there would move SQL this change is required not to move. And sibling aliases
are derived from the filter's position in the scope, so the same scope compiled
for two elements names the same relation and the two references share one `FROM`
item instead of multiplying rows.

## One change to `ScopableEntity` worth reviewing deliberately

The trait gains `scope_columns()`. `resolve_property` answers "which column does
this property mean" but cannot be enumerated, so nothing could ask which columns
are scope columns at all — and both the `PROPERTIES` list and the Policy 2 check
need exactly that. The default covers the four dimension columns, so every
hand-written impl keeps working; the `Scopable` derive overrides it to include
`pep_prop` columns, which it knows and the default cannot.

Without this, the ADR's promise that the generator exposes
"`tenant_col`, `resource_col`, `owner_col`, `type_col`, `pep_prop(...)`" is not
implementable. Flagging it because it is a trait change, not an internal one.

## Tests

Assertions are on rendered SQL, and **per element** — counting a scope column
across the statement passes even when one element body is unscoped, because the
column also appears in `COLUMNS` and in the outer `WHERE`.

Live tests in `tests/pg/` against PostgreSQL 19beta3, using
`POSTGRES_GRAPH_TAG`. The fixture is a trap: two tenants and one edge crossing
between them, with `the_trap_edge_exists` asserting that edge is really there so
a fixture that loses it fails loudly instead of turning the guards into vacuous
passes. And `the_unscoped_pattern_does_reach_across_tenants` runs the same
pattern with no scope anywhere and asserts it *does* return the foreign row —
without that control, "the scoped query returned the right rows" cannot tell a
working guard from an absent one.

Compile-fail guard in `tests/ui/fail_pgq/graph_from_unscoped.rs`, in its own
directory because its expected output only exists when the feature is on.

Every guarantee was verified by mutation. Three of those mutations found weak
tests rather than weak code, and all three are noted in their commit messages:

| Mutation | Before | After |
|---|---|---|
| scope dropped for non-head elements | 6 tests fail | — |
| `Rejected` drops the filter instead of erroring | 2 tests fail | — |
| `DISTINCT` leaks into the table path | 1 test fails | — |
| scope columns omitted from `PROPERTIES` | **all green** | 1 test fails |
| sibling dedup removed at one of two sites | **all green** | (both sites) 1 fails |
| DDL clause order wrong | **unit tests green** | live migration fails, + ordering test |

That last row is the one worth knowing about: SQL/PGQ orders an edge's clauses
`KEY`, `SOURCE`, `DESTINATION`, `LABEL`, and the generator emitted `LABEL` first.
The unit test passed because it asserted the clauses were *present* — a statement
with every clause in the wrong order still contains all of them. The live
migration rejected it with a syntax error at `LABEL`.

## Verified

- `cargo check --workspace --all-targets` clean.
- clippy clean on `pg`, `pgq`, and `pgq` with every other backend.
- `cargo doc --all-features` adds no warnings — required, since docs.rs builds
  with all features on.
- The whole `toolkit-db` suite passes on the new pins, including the Docker-backed
  lanes; plus `coord`'s postgres fence, one ledger Postgres suite and
  `resource-group`'s smoke test.
- The PG19 lane skips with a reason when the image is unavailable and fails
  naming the tag under `GEARS_TEST_PG_GRAPH_REQUIRED=1` — checked both ways
  against a bad tag, so a lane meant to cover PG19 cannot pass by skipping.

## Deliberately not in scope

- **The Level C dylint rule.** The ADR asks to extend the rule ADR-0001
  specifies so it also matches `GRAPH_TABLE` in `cust`-family arguments. There is
  nothing to extend here: the lints live in the external `cargo-gears`
  (`cargo gears lint --dylint`), and `tools/dylint_lints/` is not in this
  checkout — only CI path filters and comments reference it. Separate task, other
  repository.
- **The `pg_catalog` drift check.** The ADR records it as a follow-up and
  CodeRabbit asked to promote it to a startup precondition. That is a scope call
  for @MikeFalcon77 rather than something to decide inside this PR.
- **Adoption in IRM.** No consumer yet — IRM is documentation only.

## Summary by CodeRabbit

* **New Features**
  * Added secure PostgreSQL 19 property-graph queries with scoped traversal and graph definition management.
  * Added SQL/PGQ statement generation with validation and safe parameter handling.
  * Added explicit scope-column support for secure entities.

* **Bug Fixes**
  * Improved handling of unrecognized scope errors by reporting them as internal errors.

* **Tests**
  * Added unit, compile-time, and PostgreSQL 19 integration coverage for graph queries and scope enforcement.

* **Documentation**
  * Updated testing and secure ORM documentation for PGQ support and CI behavior.


vasylcf:
@MikeFalcon77 ADR-0002 is implemented in #4639 (#4616 as prerequisite). Details are in the description; two things there are your call rather than mine:

1. ScopableEntity gains scope_columns() — a trait change, not an internal one. Policy 3 is not implementable without it.
2. The pg_catalog drift check. The ADR parks it as a follow-up. I left it out, but the open part is whose layer it is: toolkit-db is a library and has no startup — healthcheck belongs to a gear. Happy to add a comparison function here that a consumer calls, if you'd rather it live in the ORM.

vasylcf:
Third one, and it does not block this PR.
 
ADR-0004 (errors) sends sea_orm::DbErr to Internal through the blanket From, and delegates anything finer to each gear — "Gears needing finer-grained mapping implement their own From instead of relying on the blanket." #4645 is what that delegation cost: PostgreSQL 18 reports an ON DELETE RESTRICT refusal as 23001 instead of 23503, the usage-collector's own SQLSTATE table went stale, and a documented 409 answered 500 until #4639.
 
What I can't decide is whether the delegation still stands for driver-code detection, or whether the shared piece gets an owner. The seam is awkward either way: toolkit-canonical-errors owns the mapping direction but has no sqlx dependency, so it cannot read a SQLSTATE; toolkit-db can read one but does not own categories. Neither closes it alone, which is presumably how it ended up per-gear.
 
If it is worth an ADR, it looks like a follow-up to errors/0004 rather than anything under arch/database.

libs/toolkit-sea-orm-pgq/src/render.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Render the construct as an aliased `FROM` item.
+pub fn table_ref(table: &GraphTable, alias: &str) -> Result<TableRef, PgqError> {
+    require_named(alias, "a GRAPH_TABLE alias")?;
+    let (sql, values) = body(table)?;
+
+    // `Func::cust` renders its name raw, which is what a keyword needs, and the
+    // single argument becomes the parenthesised body. The values ride along as
+    // bound parameters: `cust_with_values` substitutes each `$n` with a
+    // placeholder rather than with the value's text.
+    let call = Func::cust(Alias::new(GRAPH_TABLE)).arg(Expr::cust_with_values(sql, values));
```
MikeFalcon77: **CRITICAL**

Placeholders inside an edge element are never renumbered, so the edge's scope predicate binds the wrong value.

`Expr::cust_with_values` re-tokenises the pre-rendered body, and sea-query's `Tokenizer` treats `[` / `]` as a string-delimiter pair (`token.rs:288`, MSSQL bracket quoting). The whole `["e" IS "edge" WHERE ... $n]` element comes out as one `Token::Quoted` and is re-emitted verbatim: its `$n` is left pointing at the inner writer's numbering, and its value is dropped from the outer bound-value list.

Built against this branch, a two-hop pattern whose elements compile different numbers of values:

```
MATCH ("a" ... "a"."tenant_id" = $1 OR "a"."id" = $2)
     -["e" IS "redge" WHERE "e"."id" = $3]->
      ("b" ... "b"."tenant_id" = $3 OR "b"."id" = $4)
$1=AAAA(tenant) $2=BBBB(resource) $3=AAAA $4=BBBB $5=AAAA $6=BBBB
```

Seven values were compiled, six are bound. The edge must compare against the resource UUID but gets the tenant UUID, because its `$3` now aliases element `b`'s first parameter. When the edge's inner index exceeds the outer list length the server fails with `there is no parameter $n`.

Every existing test passes because all elements compile the same scope and therefore the same values; the bug only shows when two elements resolve different property subsets.

Do not round-trip the rendered body through `cust_with_values`. Write the element brackets outside the custom-expression text, or hand element predicates to the outer builder as `Expr` values so it does the numbering. Add a test asserting the exact bound-value list for a pattern whose elements bind different values.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+            // correlation into a cross join and multiply rows.
+            if !self.siblings.iter().any(|p| p.alias == sibling.alias) {
+                self.siblings.push(SiblingPlacement {
+                    alias: sibling.alias,
+                    query: sibling.query,
+                });
+            }
+        }
+
+        Ok(toolkit_sea_orm_pgq::Element::new(variable, label).and_where(condition))
```
MikeFalcon77: **HIGH**

An empty condition is accepted as a scoped element, so the element renders with no `WHERE` at all.

A scope built from one `ScopeConstraint` with no filters reports `is_unconstrained() == false` and `is_deny_all() == false`, but `build_constraint_condition` returns an empty `Condition::all()` for it, and `Element::and_where` drops empty conditions. Run against this branch:

```
MATCH ("a" IS "node")-["e" IS "edge"]->("b" IS "node")   -- outer: WHERE TRUE
```

Every element traverses every tenant's rows under a scope that is not allow-all. The select path has the same shape today, but the graph path is where the ADR requires every element to carry a predicate.

Reject an empty `condition` here with `ScopeError::Invalid` unless `self.scope.is_unconstrained()`, so "no predicate" is reachable only from an explicitly unconstrained scope.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+    fn scoped_element<J>(
+        &mut self,
+        variable: &'static str,
+        label: &'static str,
+    ) -> Result<toolkit_sea_orm_pgq::Element, ScopeError>
+    where
+        J: ScopableEntity + EntityTrait,
+        J::Column: sea_orm::ColumnTrait + Copy,
+    {
+        if J::scope_columns().is_empty() {
```
MikeFalcon77, line 568: **HIGH**

This checks that the entity declares *some* scope column, not that the query's scope resolves against it, so Policy 2 still lets a silent deny-all through.

An entity with `#[secure(no_tenant, resource_col = "id", no_owner, no_type)]` has a non-empty `scope_columns()` and passes here, but under `AccessScope::for_tenant(...)` `resolve_property("owner_tenant_id")` returns `None`, the constraint is dropped, and `build_scope_predicate` falls to `deny_all()`. Built against this branch:

```
MATCH ("v" IS "thing" WHERE $1)    values: [Bool(false), ...]
```

The build succeeds. That is exactly the silent empty traversal the ADR says must be refused loudly.

Check resolvability against the live scope before compiling: for each constraint in `self.scope.constraints()`, resolve every filter's property through `J::resolve_property`; if no constraint survives intact, return `ScopeError::Invalid` naming the element and the unresolved property.

libs/toolkit-db-macros/src/scopable.rs
```diff
@@ -156,6 +162,42 @@ fn generate_col_impl(
...
+/// property-graph declaration derives its `PROPERTIES` list from this, and a
+/// scope column left out of that list is silently unfilterable.
+fn generate_scope_columns(config: &SecureConfig, span: Span) -> TokenStream {
+    let mut columns = Vec::new();
+
+    for col_name in [
+        config.tenant_col.as_ref().map(|(name, _)| name),
+        config.resource_col.as_ref().map(|(name, _)| name),
+        config.owner_col.as_ref().map(|(name, _)| name),
+        config.type_col.as_ref().map(|(name, _)| name),
```
MikeFalcon77: **HIGH**

`scope_columns()` includes `type_col`, but `generate_resolve_property` emits no arm for it, so the two sets disagree and an entity that can never be scoped passes Policy 2.

`generate_resolve_property` (line 200) maps `tenant_col`, `resource_col`, `owner_col` and `pep_prop` entries — `type_col` is not among them. For `#[secure(no_tenant, no_resource, no_owner, type_col = "kind")]`, built against this branch:

```
scope_columns() = [Kind]        (non-empty -> both Policy 2 gates pass)
resolve_property(x)  = None     for every property, including "kind" and "type"
declaration() ok? true
=== BUILD SUCCEEDED ===  MATCH ("v" IS "t" WHERE $1)   values: [Bool(false), ...]
```

This contradicts the doc comment above, which says the two "cannot disagree about what counts as a scope column".

Either exclude `type_col` from `generate_scope_columns`, or emit a `resolve_property` arm for it under a well-known property name.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+        // Siblings first, then the pattern: a comma join, which PostgreSQL
+        // treats as an implicit lateral, so the pattern's correlated references
+        // resolve. `LATERAL` itself is refused before `GRAPH_TABLE`.
+        for sibling in self.siblings {
+            self.outer.from(TableRef::SubQuery(
+                Box::new(sibling.query),
+                Alias::new(sibling.alias.as_str()).into_iden(),
+            ));
+        }
+        self.outer.from(
```
MikeFalcon77: **HIGH**

The anchor table is comma-joined with `GRAPH_TABLE` and nothing ever correlates them, so every graph query is a cross join.

`SelectStatement::from()` appends, so the anchor, the siblings and the construct all land in one `FROM` with no join predicate. Built from `two_hop()` on this branch:

```sql
SELECT "cf_graph"."neighbour" FROM "graph_node", GRAPH_TABLE(...) AS "cf_graph"
WHERE "graph_node"."tenant_id" IN ($1)
```

`column()` calls `clear_selects()`, so the anchor contributes no output column either — it is a pure row multiplier. On the PG fixture tenant A has two nodes and one match, so the query returns the same row twice; `a_scoped_hop_does_not_leave_the_tenant` only passes because it calls `ids.dedup()`.

Either drop the anchor from the `FROM` when no element references it, or add the element-predicate API that lets a pattern correlate against `E`'s columns and require at least one such correlation before the anchor stays.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+    state: PathState,
+    siblings: Vec<SiblingPlacement>,
+    error: Option<ScopeError>,
+    _marker: PhantomData<G>,
+}
+
+impl<G: PropertyGraph> PathBuilder<G> {
+    /// Start the pattern at a vertex.
+    #[must_use]
+    pub fn vertex<J>(mut self, variable: &'static str) -> Self
```
MikeFalcon77, line 420: **HIGH**

`PathBuilder` has no way to attach a caller predicate to an element, so two things the ADR pins are not implemented.

`vertex`, `edge_to`, `edge_from` and `to` all take only a variable name. That means the ADR's "scope is applied after the caller's predicate, so a predicate cannot filter it back off" has nothing to order against, and a pattern cannot reference `E`'s own columns — which is the only mechanism for the "start from these resources, walk out one hop" shape the ADR calls the most common one, and the only thing that would stop the anchor being a cross join.

Add a predicate parameter (`vertex::<J>(variable, Condition)`, or a `.where_(Condition)` on the last added element) and attach it with `Element::and_where` **before** the scope condition, so scope is AND-ed on top.

libs/toolkit-db/tests/pg/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Live `PostgreSQL` 19 tests.
+//!
+//! SQL/PGQ exists on no other engine and on no earlier major, so these cannot
+//! live beside the `SQLite` suite. The image comes from `cf-gears-test-containers`
+//! (`POSTGRES_GRAPH_TAG`); while `PostgreSQL` 19 is pre-GA an unavailable image is
+//! a skip, and `GEARS_TEST_PG_GRAPH_REQUIRED=1` turns it into a failure so a CI
+//! lane meant to cover PG19 cannot pass vacuously.
+
+#![cfg(feature = "pgq")]
```
MikeFalcon77, line 9: **HIGH**

Nothing in the repo enables the `pgq` feature, so none of this PR's tests are compiled or run.

`grep -rn pgq .github/ Makefile` returns nothing. `make test-no-macros` (the `cargo test` CI job) runs `cargo nextest run --workspace` with toolkit-db's `default = []`, and the DB targets are `test-sqlite`, `test-pg`, `test-mysql` — there is no `test-pgq`. So this file, `src/secure/pgq_tests.rs` (`#[cfg(all(test, feature = "pgq"))]`) and `tests/ui/fail_pgq/*` never build in CI, and neither does the `pgq` code itself.

Add a `test-pgq` target running `cargo nextest run -p cf-gears-toolkit-db --features pgq,integration` and wire it into the `integration` job in `.github/workflows/ci.yml`.

libs/toolkit-db/src/secure/pgq_tests.rs
```diff
@@ -0,0 +1,784 @@
...
+            body.contains(&format!(r#""{variable}"."tenant_id""#)),
+            "{body}"
+        );
+    }
+}
+
+/// A caller predicate narrows an element; it cannot remove what the library put
+/// there. Both predicates must survive in the same body.
+#[test]
+fn a_caller_predicate_cannot_remove_element_scope() {
```
MikeFalcon77, line 647: **HIGH**

This test adds no caller predicate, so it does not test what it is named for.

The chain is byte-for-byte the same as the `two_hop()` helper, and the only assertion is that element `a` contains `"a"."tenant_id"` — which `every_graph_element_embeds_scope` already covers. It cannot fail for the reason the name gives, because `PathBuilder` exposes no way to attach a caller predicate at all.

This is one of the ADR's required test-contract items, so it currently reads as green coverage of a guarantee nothing checks. Once the element-predicate API exists, pass a predicate that would exclude the scoped rows and assert both it and the scope predicate appear AND-ed in the same element body; until then delete the test.

libs/toolkit-db/tests/pg/secure_graph.rs
```diff
@@ -0,0 +1,587 @@
...
+                .to::<node::Entity>("b")
+        })
+        .column("b", "id", "neighbour")
+        .all_as(&conn)
+        .await
+        .expect("graph query");
+
+    let mut ids: Vec<i64> = rows.into_iter().map(|r| r.neighbour).collect();
+    ids.sort_unstable();
+    ids.dedup();
```
MikeFalcon77: **HIGH**

`dedup()` here discards exactly the signal the ADR asks these tests to pin.

The query really does return two identical rows, because the anchor is cross-joined with `GRAPH_TABLE` (tenant A has two nodes, the pattern has one match). Sorting and deduplicating makes the suite structurally unable to observe row multiplication — including the ADR item "a correlated-sibling scope does not multiply rows: a resource reachable through two group rows appears exactly once". `each_tenant_sees_only_its_own_graph:471` normalises the same way.

Pin the anchor to one row with `.filter(node::Column::Id.eq(1))` and assert on the raw `Vec<i64>` including duplicates.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+
+        let mut table = toolkit_sea_orm_pgq::GraphTable::new(G::GRAPH_NAME, pattern);
+        for column in self.columns {
+            table = table.column(column);
+        }
+
+        // Siblings first, then the pattern: a comma join, which PostgreSQL
+        // treats as an implicit lateral, so the pattern's correlated references
+        // resolve. `LATERAL` itself is refused before `GRAPH_TABLE`.
+        for sibling in self.siblings {
```
MikeFalcon77, line 868: **MEDIUM**

Siblings from different OR-ed constraints are inner-joined, so one empty alternative zeroes the whole result.

Aliases are `__cf_scope_{constraint}_{filter}`, so a scope with two subquery-producing constraints places two relations. The element predicate is `(x = __cf_scope_0_0.c) OR (y = __cf_scope_1_0.c)`, but the `FROM` comma-joins both unconditionally: if either relation is empty the query returns nothing even when the other branch matches, and if both are non-empty each match is multiplied by the other relation's cardinality.

Either reject a scope that produces siblings from more than one constraint with `ScopeError::Invalid`, or emit each alternative's sibling as `LEFT JOIN ... ON true`.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+                "a graph query needs a head vertex; call vertex()",
+            ));
+        };
+
+        let mut pattern = toolkit_sea_orm_pgq::GraphPattern::new(head);
+        for (edge, direction, target) in state.hops {
+            pattern = pattern.hop(edge, direction, target);
+        }
+
+        let mut table = toolkit_sea_orm_pgq::GraphTable::new(G::GRAPH_NAME, pattern);
```
MikeFalcon77, line 853: **MEDIUM**

`build()` never calls `G::declaration()`, so Policy 3's single-source guarantee does not hold on the query path.

The pattern is assembled from `G::GRAPH_NAME` and per-element `VertexOf`/`EdgeOf::LABEL`; `declaration()` is only ever called by tests and migrations. Nothing ties a `VertexOf<G>` impl to the declaration, so an entity can be patterned into a graph it was never registered in. Built against this branch, with a declaration registering only `node`:

```
declared labels: ["node"]
=== BUILD SUCCEEDED ===  MATCH ("g" IS "ghost" WHERE "g"."tenant_id" IN ($1))
```

The graph has no `ghost` label, so this fails at the server instead of at build time. The duplicate-label check in `element()` is bypassed the same way.

Call `G::declaration()` in `build()` and verify each element label is declared and each projected property is in `properties_of(label)`.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+        );
+
+        Ok(self.outer)
+    }
+
+    /// Render the statement without executing it.
+    ///
+    /// # Errors
+    /// Returns [`ScopeError::Invalid`] for a pattern that cannot be built.
+    pub fn build_statement(
```
MikeFalcon77: **MEDIUM**

`build_statement` is `pub`, which puts `sea_orm::Statement` and `sea_orm::DbBackend` into toolkit-db's public API.

The CTE equivalent is deliberately gated — `cte.rs:594` is `#[cfg(test)] pub(crate) fn build_statement`, with the private `into_statement` used on the execution path. The ADR lists "no `Statement`/`ConnectionTrait` escape from `toolkit-db`" as a property of the chosen design. The only callers here are `all_as` and the in-crate unit tests.

Mirror `cte.rs`: make the execution-path renderer private, and expose the SQL-assertion helper as `#[cfg(test)] pub(crate)`.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+        for sibling in self.siblings {
+            self.outer.from(TableRef::SubQuery(
+                Box::new(sibling.query),
+                Alias::new(sibling.alias.as_str()).into_iden(),
+            ));
+        }
+        self.outer.from(
+            table
+                .into_table_ref(self.graph_alias)
+                .map_err(|_| ScopeError::Invalid("the graph pattern could not be rendered"))?,
```
MikeFalcon77: **MEDIUM**

`map_err(|_| ...)` collapses three distinct caller mistakes into one message.

`into_table_ref` returns `NoColumns` (no `.column()` was projected), `DuplicateVariable(var)` (a pattern variable bound twice) and `EmptyIdentifier { what }` — all caller-reachable, and `DuplicateVariable` even names the offending variable. All three arrive as "the graph pattern could not be rendered".

Add a `Pgq(#[from] toolkit_sea_orm_pgq::PgqError)` variant to `ScopeError` behind `#[cfg(feature = "pgq")]` and propagate with `?`.

libs/toolkit-db/src/secure/entity_traits.rs
```diff
@@ -152,4 +152,37 @@ pub trait ScopableEntity: EntityTrait {
...
+    /// rejected up front when it resolves no scope column at all, rather than
+    /// compiling to a deny-all traversal that looks like missing data
+    /// (Policy 2).
+    ///
+    /// The default covers the four dimension columns, so every existing
+    /// hand-written implementation keeps working. `#[derive(Scopable)]`
+    /// overrides it to include `pep_prop(...)` columns as well, which it knows
+    /// and this default cannot.
+    #[must_use]
+    fn scope_columns() -> Vec<Self::Column> {
```
MikeFalcon77: **MEDIUM**

This default enumerates `type_col`, which `resolve_property` has no property name for, and omits any custom property a hand-written impl maps.

Both directions are wrong for the two callers that depend on this method. An entity whose only dimension is `type_col` gets a non-empty `scope_columns()` and passes the Policy 2 gates while resolving nothing. Conversely, a hand-written `ScopableEntity` that maps extra properties in `resolve_property` (the manual equivalent of `pep_prop`) inherits a `scope_columns()` that omits them, so the generated `PROPERTIES` list is missing exactly the column the element predicate will address — and a column absent from `PROPERTIES` is invisible to `MATCH`.

Drop `Self::type_col()` from the default, and make `scope_columns()` a required method so manual implementors must enumerate the same set their `resolve_property` can return.

libs/toolkit-db/src/secure/cond.rs
```diff
@@ -1,11 +1,109 @@
...
+/// a variable, and needs `dst.tenant_id`. Parameterising the addressing is what
+/// lets one compiler serve ordinary selects, CTE bodies and graph elements
+/// alike; a second, PGQ-specific compiler would double the number of places
+/// tenant isolation could be wrong (`docs/arch/secure-orm/ADR/0002`).
+// The graph-addressed half of this module is the API `docs/arch/secure-orm/ADR/0002`
+// pins, and `secure::pgq` is what will consume it. Until that lands nothing in a
+// non-test build names it, which is what these allows say — the shape is fixed
+// here on purpose, separately from the builder that uses it.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[allow(dead_code)]
```
MikeFalcon77: **MEDIUM**

The blanket `#[allow(dead_code)]` and the `pub` here are both wrong now that `secure::pgq` lands in the same PR.

The comment above says `secure::pgq` "will" consume these and that "until that lands nothing in a non-test build names it" — it lands here. The allow is repeated on `SiblingSource`, `ScopePredicate` and its impl block, and it suppresses unused-item detection for every variant, field and method of those types, including any that really do become dead. `pub` is also misleading: `cond` is a private module and none of these are re-exported from `secure::mod`.

Make them `pub(crate)`, replace the blanket allows with `#[cfg_attr(not(feature = "pgq"), allow(dead_code))]`, and update the comment.

libs/toolkit-db/src/secure/cond.rs
```diff
@@ -79,31 +227,137 @@ where
...
+    for (filter_index, filter) in constraint.filters().iter().enumerate() {
+        let Some(col) = E::resolve_property(filter.property()) else {
+            return Ok(None);
+        };
+        let column = addressed::<E>(col, address);
+        // Derived from the filter's position in the scope, so the same scope
+        // compiled for two pattern elements produces the same alias — the two
+        // references then share one `FROM` item instead of duplicating the
+        // relation, which is what keeps a correlated join from multiplying rows.
+        let alias = format!("__cf_scope_{constraint_index}_{filter_index}");
```
MikeFalcon77: **MEDIUM**

This `format!` runs for every filter of every scoped query in the platform, and is discarded on all but the graph path.

`alias` is read only by the `ColumnAddress::GraphElement` branch of `attach_set_membership`; the `Eq` and `In` arms never touch it, and the `Table` branch ignores `arm.alias` entirely. `build_constraint_condition` is on the path of every secure select, which previously allocated nothing here.

Pass `constraint_index`/`filter_index` through `SetMembership` and build the alias inside the `GraphElement` arm.

libs/toolkit-db/src/secure/cond.rs
```diff
@@ -194,6 +495,285 @@ mod tests {
...
+    }
+
+    /// Generalising the addressing must not move the select path. These are the
+    /// exact strings the compiler emitted before the parameter existed, one per
+    /// `ScopeFilter` arm.
+    #[test]
+    fn table_addressing_is_unchanged_for_every_arm() {
+        use custom_prop_entity::Entity as E;
+
+        let cases: Vec<(AccessScope, &str)> = vec![
```
MikeFalcon77, line 615: **MEDIUM**

The golden table is missing the `Eq` arm, so "byte-identical for every `ScopeFilter` arm" is not actually pinned for it.

`AccessScope::for_tenant` builds `in_uuids`, so the four cases cover `In`, `InGroup`, `InGroupSubtree` and `InTenantSubtree`. `ScopeFilter::Eq` now routes through the new `addressed::<E>()` indirection with no golden string guarding its output.

Add a fifth case built from `ScopeFilter::eq(...)` with its exact pre-change rendering.

libs/toolkit-db/tests/pg/secure_graph.rs
```diff
@@ -0,0 +1,587 @@
...
+        .filter(
+            sea_orm::Condition::all()
+                .add(edge::Column::TenantId.eq(TENANT_A))
+                .add(edge::Column::DstId.eq(11_i64)),
+        )
+        .all(&conn)
+        .await
+        .expect("query");
+
+    assert_eq!(
```
MikeFalcon77, line 664: **MEDIUM**

The trap edge cannot be traversed, so this precondition guards nothing.

Edge `(TENANT_A, 300, src 1, dst 11)` has destination key `(tenant_id, dst_id)` = `(A, 11)`, and the declaration references `pgq_node (tenant_id, id)`. No node row satisfies `(A, 11)`, so the edge yields no match in any query in this file, scoped or unscoped — as the `seed()` comment itself says. Deleting the row would not change a single assertion below.

Seed a node `(TENANT_A, 11)` so the crossing edge actually resolves and the trap can spring when a guard is removed.

libs/toolkit-db/tests/pg/secure_graph.rs
```diff
@@ -0,0 +1,587 @@
...
+    let raw = stand.raw().await;
+    let rows = Neighbour::find_by_statement(stmt)
+        .all(&raw)
+        .await
+        .expect("unscoped query");
+
+    let mut ids: Vec<i64> = rows.into_iter().map(|r| r.neighbour).collect();
+    ids.sort_unstable();
+    ids.dedup();
+    assert_eq!(
```
MikeFalcon77, line 850: **MEDIUM**

The control does not show a traversal crossing a tenant boundary, so it does not establish that the suite can see the hole.

The doc comment says "tenant A's crossing edge is resolvable, so the walk reaches a node A does not own", but `[2, 11]` is the union of two intra-tenant walks (A: 1->2, B: 10->11) — the crossing edge is unresolvable because of its composite endpoint key. So this shows only that an unscoped query returns both tenants' rows, not that an unscoped element body returns foreign rows.

Build the control with scope on both vertices and none on the edge body, over a fixture whose crossing edge resolves, and assert that variant returns the foreign neighbour while the fully scoped query does not.

libs/toolkit-sea-orm-pgq/src/render.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    /// Render the construct's SQL and bound values, without executing it.
+    ///
+    /// Exposed for tests: assertions belong on rendered SQL, because a predicate
+    /// that never reaches the database would satisfy a `Debug`-form assertion
+    /// and still leak.
+    ///
+    /// # Errors
+    /// As [`Self::into_table_ref`].
+    pub fn render_for_test(&self) -> Result<(String, Vec<Value>), PgqError> {
```
MikeFalcon77: **MEDIUM**

`render_for_test` is a `pub` method on a published crate whose only callers are this crate's own `#[cfg(test)] mod tests`.

It ships a test-only entry point to every downstream consumer of `cf-gears-toolkit-sea-orm-pgq`.

Mark it `#[cfg(test)] pub(crate)`, or put it behind an opt-in `test-support` feature.

libs/toolkit-sea-orm-pgq/src/ddl.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub element: VertexTable,
+    /// Source endpoint.
+    pub source: EndpointRef,
+    /// Destination endpoint.
+    pub destination: EndpointRef,
+}
+
+/// A whole property-graph declaration.
+#[derive(Clone, Debug)]
+pub struct PropertyGraph {
```
MikeFalcon77, line 188: **MEDIUM**

The DDL types expose every field as `pub`, which is the thing `ast.rs` documents as unsafe two modules away.

`ast.rs:13-17` explains that its fields stay `pub(crate)` because `pub` "would become part of the public API and let a caller build a `GraphTable` around the builders whose invariants the renderer checks". `PropertyGraph`, `VertexTable`, `EndpointRef` and `EdgeTable` do exactly that — e.g. the source/destination arity invariant is enforced in `toolkit_db::secure::pgq::into_endpoint`, in a different crate, so a direct user of this crate bypasses it and gets `SOURCE KEY (a, b) REFERENCES t (c)` rendered without error.

Make the fields `pub(crate)` behind validating constructors, matching `GraphPattern`/`GraphTable`/`ProjectedColumn`.

libs/toolkit-sea-orm-pgq/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "cf-gears-toolkit-sea-orm-pgq"
```
MikeFalcon77, line 2: **MEDIUM**

This crate is published (no `publish = false`, and it has a `release-plz.toml` entry) but is missing the manifest fields every sibling in `libs/` carries.

Absent: `repository.workspace`, `rust-version.workspace`, `readme = "README.md"` and the README file itself, `keywords`, `categories`, `metadata.docs.rs.all-features = true`, and `[lib] name = "toolkit_sea_orm_pgq"`. Compare `libs/toolkit-gts/Cargo.toml` and `libs/toolkit-db/Cargo.toml`, which have all of them. The ADR's registration checklist names `[lib] name`, `README.md`, `description`, keywords/categories and `[lints] workspace = true` explicitly.

The lib target is currently `cf_gears_toolkit_sea_orm_pgq`; it only imports as `toolkit_sea_orm_pgq` because the `[workspace.dependencies]` alias renames it, so the convention is silently unmet for anyone outside this workspace.

Add the missing fields and create `libs/toolkit-sea-orm-pgq/README.md`.

libs/toolkit-sea-orm-pgq/src/error.rs
```diff
@@ -0,0 +1,23 @@
+//! The one error type the AST and the DDL builders share.
+//!
+//! Its own module so that `ast` and `ddl` stay independent of each other: both
+//! need this type, neither needs the other.
+
+/// Something the AST refuses to render.
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum PgqError {
```
MikeFalcon77, line 13: **LOW**

`PgqError` is the public error enum of a newly published crate and is not `#[non_exhaustive]`, so adding a variant later is a breaking change for every downstream `match`.

Also, `EmptyIdentifier { what: &'static str }` carries four structurally different refusals — empty vertex-table list (`ddl.rs:105`), empty element key (`ddl.rs:158`), empty property list (`ddl.rs:162`), empty endpoint key (`ddl.rs:200`) — none of which is an identifier, so callers can only tell them apart by string-matching prose.

Add `#[non_exhaustive]`, and split the non-identifier cases into their own variants.

libs/toolkit-db/src/secure/pgq.rs
```diff
@@ -0,0 +1,657 @@
...
+            ));
+            return self;
+        }
+
+        for sibling in builder.siblings {
+            if !self.siblings.iter().any(|p| p.alias == sibling.alias) {
+                self.siblings.push(sibling);
+            }
+        }
+        self.pattern = Some(builder.state);
```
MikeFalcon77, line 707: **LOW**

A second `match_path` call silently replaces the first pattern but keeps the siblings it contributed.

Line 520 merges the closure's siblings into `self.siblings` before this assignment overwrites `self.pattern`. If the discarded pattern was the only thing referencing one of those relations, it stays in the `FROM` as an uncorrelated comma join.

Call `self.fail(ScopeError::Invalid(...))` when `self.pattern.is_some()` on entry.

libs/toolkit-db/src/secure/cond.rs
```diff
@@ -52,24 +150,74 @@ fn deny_all() -> Condition {
...
 /// | multiple constraints | OR of ANDed filter groups |
 pub fn build_scope_condition<E>(scope: &AccessScope) -> Condition
+where
+    E: ScopableEntity + EntityTrait,
+    E::Column: ColumnTrait + Copy,
+{
+    // Table addressing renders every arm, so the `Result` cannot be `Err` here
+    // and the sibling list is always empty. Kept as one code path so the select
+    // path cannot drift from the graph path.
+    build_scope_predicate::<E>(scope, ColumnAddress::Table)
```
MikeFalcon77: **LOW**

The `Err` arm is swallowed with no assertion and no log.

`map_or_else(|_| deny_all(), ...)` is correct today because table addressing renders every arm, but if a future `ScopeFilter` variant is fallible under `ColumnAddress::Table` the select path degrades to `WHERE false` platform-wide with nothing recording why — the silent empty result this module's own doc comment calls unacceptable.

Add a `debug_assert!` and a `tracing::error!` on the `Err` arm before falling back.

libs/test-containers/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn the_pins_are_not_the_eol_defaults() {
+        assert_ne!(POSTGRES_TAG, Postgres::default().tag());
+        assert_ne!(MYSQL_TAG, Mysql::default().tag());
+    }
+
+    /// An exported-but-empty variable must not produce `postgres:`.
+    #[test]
+    fn an_empty_override_falls_back_to_the_pin() {
+        assert_eq!(
+            tag_from_env("GEARS_TEST_ABSENT_VAR_XYZ", "fallback"),
```
MikeFalcon77: **LOW**

This test passes an unset variable, so it exercises the `Err` arm of `env::var`, not the guard the doc comment describes.

`GEARS_TEST_ABSENT_VAR_XYZ` is absent, so `Ok(v) if !v.trim().is_empty()` is never evaluated and the "exported-but-empty variable must not produce `postgres:`" case is untested.

`temp-env` is already a workspace dependency — use `temp_env::with_var` to set the variable to `""` and to `"   "` and assert both fall back to the pin.

docs/TESTING.md
```diff
@@ -136,11 +136,79 @@ make test-db               # all three
...
+`Cargo.lock` and a dependency bump changed the database under every test in the
+repository silently.
+
+```rust
+// in a fixture
+let request = cf_gears_test_containers::postgres()
+    .with_env_var("POSTGRES_PASSWORD", "pass");
+
+// when the image itself needs configuring (with_db_name, with_user, ...)
+let request = cf_gears_test_containers::postgres_from(
```
MikeFalcon77: **LOW**

`cf_gears_test_containers::postgres_from` does not exist.

The crate exposes `postgres()`, `postgres_with_db(db_name: &str)` and `postgres_graph()`. This example also calls `Postgres::default()` in fixture code, which the `db-image-pins` job added in this same PR rejects outright.

Change the example to `cf_gears_test_containers::postgres_with_db("my_test")`, or rename the function to match the documented signature.

.github/workflows/ci.yml
```diff
@@ -355,6 +355,13 @@ jobs:
       - name: Test toolkit-db (pg integration)
         run: make test-pg
 
+      - name: Test toolkit-db (pgq unit + PG19 integration)
+        # Compiles and runs everything behind the `pgq` feature — the secure
+        # graph builder, its unit suites and the trybuild guards — and the
+        # PostgreSQL 19 integration suite, which skips itself while the pre-GA
```
diffora: Test comment

libs/toolkit-sea-orm-pgq/src/render.rs
vasylcf: Confirmed — Tokenizer::is_string_delimiter_start treats [ as a quote opener, so the whole edge element became one Token::Quoted re-emitted verbatim, and its values never reached the outer bound list. Fixed in 5068b48b7: the body is now a single Expr::cust_with_exprs whose template holds only $n markers, never brackets — raw syntax rides in Expr::cust fragments (rendered verbatim, no tokenisation) and each element predicate is handed to the enclosing builder as a Condition-derived expression, so the outer statement does the numbering itself. render_for_test renders the same expression tree, so assertions can't drift from the execution path. Added the requested tests: edge_values_survive_into_the_final_statement (syntax crate — final statement, elements binding different value counts, exact value list) and the_exact_value_list_is_bound_when_elements_differ (toolkit-db — exact bound list through build_statement).

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51: scoped_element refuses an empty compiled condition unless scope.is_unconstrained(), so "no predicate" is reachable only from an explicit allow-all. Test: an_empty_constraint_scope_is_refused_for_elements.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51: require_scope_resolves resolves every filter of every constraint through J::resolve_property before compiling; if no constraint survives intact the build fails with a new ScopeError::UnresolvedScopeProperty { element, property } — a new variant because Invalid(&'static str) can't carry the names you asked for. 
Test: a_scope_no_constraint_of_which_resolves_is_refused.

libs/toolkit-db-macros/src/scopable.rs
vasylcf: Went with the first option in 9046ffb51: type_col is out of generate_scope_columns — resolve_property has no property name for it, so no scope constraint can address it and listing it only let unscopable entities through the gates. Together with making scope_columns() required on the trait (your entity_traits.rs thread), both methods now come from one configuration and can't disagree.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51, via your second option with one honest caveat. The element-predicate API now exists (where_, next thread), and build() keeps the anchor only when the pattern carries at least one caller predicate — the only construct that can reference the anchor — so the default no-predicate path drops the anchor from the FROM entirely (its scope isn't lost; every element embeds the same scope). The caveat: we don't introspect the predicate to prove it actually names the anchor's table, so a non-correlating caller predicate still keeps the anchor. The multiplication your comment demonstrates is gone from the default path, and the anchored path is pinned by a live-server test (an_anchored_walk_starts_from_the_anchor_rows) that narrows the anchor and watches the result change — i.e. the comma-join correlation into GRAPH_TABLE is real, not a cross join.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Added in 9046ffb51: PathBuilder::where_(Condition) narrows the last added element, vertex or edge. Elements are held as (caller predicates, scope condition) until finalization, so the rendered body always reads caller predicate first with scope AND-ed after it — and since Element::and_where only narrows, the caller can't filter scope back off. An empty condition passed to where_ is refused loudly rather than silently dropped.

libs/toolkit-db/tests/pg/mod.rs
vasylcf: Added in e13ad9c5a: make test-pgq runs cargo nextest run -p cf-gears-toolkit-db --features pgq,integration; it's part of test-db and a step in the CI integration job. The PG19 suite still skips itself while the pre-GA image is unavailable (GEARS_TEST_PG_GRAPH_REQUIRED=1 upgrades that to a failure for a lane that's expected to be green). Ran locally against 19beta3-alpine: 581/581.

libs/toolkit-db/src/secure/pgq_tests.rs
vasylcf: Rewritten in 9046ffb51 now that the API exists: the test passes a predicate that would exclude the scoped rows and asserts both predicates survive in the same element body, AND-ed, caller before scope (a_caller_predicate_cannot_remove_element_scope).

libs/toolkit-db/tests/pg/secure_graph.rs
vasylcf: Fixed in cea549871: every assertion is on the raw, sorted, non-deduplicated Vec<i64>. With the uncorrelated anchor dropped (FROM thread) the scoped queries return each hop exactly once, and that's asserted literally now. The .filter(id = 1) idea became the dedicated anchored-walk test instead, where the anchor legitimately stays.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Went with the refusal in 9046ffb51: build_scope_predicate rejects a scope whose siblings come from more than one OR-ed constraint, while table addressing still inlines both subqueries and stays servable. LEFT JOIN ... ON true is doable if a real scope ever needs it, but I'd rather not grow the correlated-join surface before the PG19 GA re-validation. Test: siblings_from_two_or_ed_constraints_are_refused.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51: build() calls G::declaration() and verifies every pattern label is declared and every projected column names a bound variable whose element exposes the property. Tests: a_label_outside_the_declaration_is_refused, a_property_outside_the_declaration_is_refused.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51, mirroring cte.rs: #[cfg(test)] pub(crate) build_statement plus a private into_statement on the execution path.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Done in 9046ffb51 with one deliberate deviation: ScopeError::Pgq(#[from] PgqError) exists and propagates via ?, but it is not behind #[cfg(feature = "pgq")]. Six gears match ScopeError exhaustively, and CI clippy runs --workspace --all-features: under feature unification a gated variant changes the enum's shape for every crate in the build the moment anything enables pgq, and downstream can't name the variant conditionally (their cfg(feature) tests their own features, not toolkit-db's). So the variant is unconditional, toolkit-sea-orm-pgq became a regular dependency (tiny, security-agnostic, zero new transitive deps), and the pgq feature still gates the builder itself. Happy to revisit if you see a way to keep the cfg without breaking --all-features builds.

libs/toolkit-db/src/secure/entity_traits.rs
vasylcf: Both directions fixed in 9046ffb51: the default is gone — scope_columns() is required, documented as "exactly the set resolve_property can return", and type_col is explicitly not in it. All manual impls in the repo (~30) now enumerate their sets.

libs/toolkit-db/src/secure/cond.rs
vasylcf: The allows and the stale comment are fixed in 9046ffb51: blanket #[allow(dead_code)] became targeted cfg_attr(not(feature = "pgq"), allow(dead_code)) (and cfg_attr(not(test), ...) for the accessors only tests read), so genuinely dead items are visible again. On pub vs pub(crate): cond is a private module, so the two are equivalent there, and the workspace denies clippy::redundant_pub_crate, which flags pub(crate) in exactly this position — the same situation ast.rs documents for Hop. I've extended the comment to say that explicitly, so the pub no longer reads as an intended re-export.

libs/toolkit-db/src/secure/cond.rs
vasylcf: Fixed in 9046ffb51: SetMembership carries constraint_index/filter_index and the alias is built inside the GraphElement arm only.

libs/toolkit-db/src/secure/cond.rs
vasylcf: Added in 9046ffb51, with the exact pre-change rendering as the fifth case.

libs/toolkit-db/tests/pg/secure_graph.rs
vasylcf: Fixed in cea549871: node (A, 11) is seeded, so edge 300 resolves and tenant A's scoped result is [2, 11] — both tenants now own an 11, so a leaked B-walk surfaces as an extra 11 that the non-dedup assertions can see. The unresolvable shape didn't go to waste: a new edge (A, 301, 2 -> 10) is asserted absent from even the unscoped control, pinning that the composite endpoint key refuses cross-tenant joins structurally. The preconditions moved to the_fixture_traps_exist.

libs/toolkit-db/tests/pg/secure_graph.rs
vasylcf: Fixed in cea549871, with a fixture change your comment implied: on the composite-key graph a single unscoped element body can't return foreign rows — the endpoint key blocks the join before any predicate — so the per-element control lives on a second graph (pgq_open) keyed on id alone, where crossing edges really resolve. There, with scope on both vertices and none on the edge body, B's edge between A's two nodes carries a second walk ([2, 2]) that the fully scoped query doesn't have ([2]) — an_unscoped_edge_element_admits_foreign_rows. The whole-pattern control stays on the composite graph, asserting [2, 11, 11] and no 10.

libs/toolkit-sea-orm-pgq/src/render.rs
vasylcf: Fixed in 5068b48b7: #[cfg(test)] pub(crate).

libs/toolkit-sea-orm-pgq/src/ddl.rs
vasylcf: Fixed in 5068b48b7: fields are pub(crate) behind validating constructors — EndpointRef::new owns the arity/emptiness invariants (moved out of toolkit-db), VertexTable::new refuses empty keys and property lists — and toolkit-db reads labels/properties through accessors. ElementKey left the public API.

libs/toolkit-sea-orm-pgq/Cargo.toml
vasylcf: Fixed in 5068b48b7: all the listed fields, README.md, and [lib] name = "toolkit_sea_orm_pgq" so the import name no longer depends on the workspace dependency rename.

libs/toolkit-sea-orm-pgq/src/error.rs
vasylcf: Fixed in 5068b48b7: #[non_exhaustive], and the four non-identifier refusals get their own variants (NoVertexTables, EmptyElementKey, EmptyProperties, EmptyEndpointKey) plus MismatchedEndpointArity { key, references }.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 9046ffb51: a second match_path fails before the closure runs, so the first pattern's siblings can't be orphaned. Test: a_second_match_path_is_refused.

libs/toolkit-db/src/secure/cond.rs
vasylcf: Fixed in 9046ffb51: debug_assert! plus tracing::error! (with the entity's table name) before the deny_all() fallback — still fail-closed, no longer silent.

libs/test-containers/src/lib.rs
vasylcf: Fixed in 802259e2d with temp_env: "" and "   " both fall back to the pin, the absent case keeps covering the Err arm, and a set, non-empty value is pinned to win over the fallback.

docs/TESTING.md
vasylcf: Fixed in e13ad9c5a: the example uses postgres_with_db("my_test"), and the doc now says why there's deliberately no configure-the-image-yourself hook.

libs/toolkit-db/src/secure/pgq.rs
MikeFalcon77: The unification argument is right, but it forces the wrong fix. `#[cfg]` on the variant breaks downstream matches because `ScopeError` is exhaustive — that is the thing to change, not the dependency.

Two separate problems.

**The dependency is now unconditional.** `toolkit-sea-orm-pgq = { workspace = true }` in `[dependencies]` plus `pgq = ["pg"]` means every crate linking toolkit-db links the SQL/PGQ AST crate. `gears/simple-user-settings` builds toolkit-db with `features = ["sqlite"]` and no `pg`, and it is one of the six sites that had to add the arm. ADR-0002 "Backend gating" says the opposite: "a gear cannot even name the types on a build without a Postgres backend."

**The six arms are dead code in every build that has them.** Both variants are constructed only inside `secure::pgq`, which is `#[cfg(feature = "pgq")]`. No gear enables `pgq`, so in those six files neither variant is constructible.

Two independent fixes:

1. Don't carry the foreign type. `PgqError` isn't re-exported from toolkit-db, so no gear can inspect it anyway — the variant is unavoidable and unusable at the same time. Make the payload toolkit-db's own (a `String`, or a small local `GraphSyntaxError` if you want `pgq_tests` to keep matching on it), and `toolkit-sea-orm-pgq` goes back to `optional = true` under `pgq = ["pg", "dep:toolkit-sea-orm-pgq"]`. The variant stays unconditional, so the unification problem never arises and the six gears don't move again. Cost is one `map_err` in the gated module instead of `?`.
2. Add `#[non_exhaustive]` to `ScopeError`. You applied exactly this reasoning to `PgqError` in this PR — "adding a variant must not be a breaking change for every downstream `match`". `ScopeError` is matched by six sites in five gears and is where it actually costs something. With it, this round's churn is the last one.

libs/toolkit-db/src/secure/pgq.rs
MikeFalcon77: The guard counts constraints that produced a sibling, so it misses the mixed case, which has the same symptom.

`C1 OR C2` where only `C1` needs a correlated relation — say `C1 = in_group(resource_id, …)` and `C2 = eq(owner_tenant_id, X)` — gives `constraints_with_siblings == 1` and passes. The predicate is `(a.id = __cf_scope_0_0.resource_id) OR (a.tenant_id = X)`, but the sibling is still placed as an unconditional comma join. An empty relation zeroes the `C2` branch too; a relation with k rows returns every `C2` match k times.

The condition to reject is "a sibling exists and the scope has more than one constraint", not "more than one constraint has a sibling". `siblings_from_two_or_ed_constraints_are_refused` passes without covering this.

libs/toolkit-db/tests/pg/mod.rs
MikeFalcon77: The lane runs, but `GEARS_TEST_PG_GRAPH_REQUIRED` is set nowhere in CI, so the PG19 half can still pass by skipping.

`grep -rn GEARS_TEST .github/workflows/` returns nothing. The new step has no `env:` block, and the `integration` job's only job-level env is `RUSTFLAGS`. Every test in `tests/pg/secure_graph.rs` opens with `let Some(stand) = stand().await else { return };`, so if `postgres:19beta3-alpine` can't start the whole isolation suite is green and empty — the vacuous pass `graph_lane_required()` exists to prevent. The Makefile comment says to set it; nothing does.

One `env: GEARS_TEST_PG_GRAPH_REQUIRED: 1` on the new step closes it.

Separately: `pgq = ["pg"]`, so `make test-pgq` re-runs the whole pg integration suite that `make test-pg` ran in the previous step.

libs/toolkit-db/src/secure/pgq.rs
MikeFalcon77: Agreed on the default path. On the caveat you name: `has_caller_predicate` is set by any `where_` call, so a predicate that doesn't mention the anchor keeps it in the `FROM` with no correlation. `where_(Expr::col(("a", "id")).eq(5))` reproduces the original cross join, and with `clear_selects()` and no `DISTINCT` a multi-row anchor still multiplies matches.

`an_uncorrelated_anchor_is_dropped_from_the_from_clause` and `an_anchored_walk_starts_from_the_anchor_rows` cover the two ends. The middle case — anchor kept, predicate doesn't correlate — is the one that regresses, and nothing tests it.

If introspecting the predicate isn't worth it, an explicit opt-in reads better than an inference: a `correlate_with_anchor()` on the builder, so keeping the anchor is a decision rather than a side effect of having called `where_`.

libs/toolkit-db/src/secure/pgq.rs
MikeFalcon77: `check_declaration` matches on `LABEL`, not on the entity, so membership still isn't what Policy 3 asks for.

Two entities can both `impl VertexOf<G>` with the same `LABEL`. If only one is registered, the other passes `properties_of(label).is_some()`. The declaration's own uniqueness check (`self.labels.insert`) runs inside `declaration()` and can't see marker impls that were never registered.

The failure isn't just a bad error message: scope is compiled from the unregistered entity's `resolve_property` while the label resolves to the registered entity's table, so when the two share scope column names the query silently traverses the wrong table under a predicate built for a different one.

Comparing the element's table name against the declared element's would close it — `GraphDeclaration::element` already records `J::default().table_name()`.

libs/toolkit-db/src/secure/pgq.rs
MikeFalcon77: Graph path is right. Noting that the same shape is still live on the ordinary select path, since it's the same defect class: `cond.rs` still returns `Ok(Some(Condition::all()))` for an empty constraint and `build_scope_condition` hands that back, so a scope of one empty `ScopeConstraint` — `is_unconstrained() == false`, `is_deny_all() == false` — compiles to `WHERE TRUE` on a plain secure select.

Pre-existing and not this PR's job, but the guard you added here is the one the select path also wants. Please create a separate issue about it

libs/toolkit-sea-orm-pgq/src/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+    }
+    Ok(out)
+}
+
+/// A hostile name must render as exactly one identifier — it must not be able
+/// to close its own quoting and become syntax.
+#[test]
+fn a_hostile_identifier_renders_as_one_escaped_identifier() {
+    let hostile = r#"a" IS "node") COLUMNS ("x"."y" AS "z"); DROP TABLE t; --"#;
+    let table = GraphTable::new("kb", GraphPattern::new(Element::new(hostile, "node")))
```
MikeFalcon77: **MEDIUM**

The hostile-identifier test covers two of the four positions the ADR names.

`hostile` is passed as the element variable and the projected-column variable; the graph name (`"kb"`), the label (`"node"`), the property (`"id"`) and the alias (`"n"`) stay benign. The ADR contract is "hostile graph name / label / element variable renders as exactly one quoted, escaped identifier, asserted structurally".

`ddl.rs` has its own `write_ident` path and no hostile input reaches it — `graph_ddl()` uses `"kb"`, `"graph_node"`, `"node"`, `"tenant_id"` throughout, so table names, labels and `PROPERTIES` entries in `CREATE PROPERTY GRAPH` are unpinned.

Parameterise `outside_identifiers` over each position and add a DDL case.

gears/simple-user-settings/simple-user-settings/src/infra/storage/sea_orm_repo.rs
```diff
@@ -36,6 +36,11 @@ fn map_scope_error(e: ScopeError) -> DomainError {
         ScopeError::TenantNotInScope { tenant_id } => {
             DomainError::forbidden(format!("tenant {tenant_id} not in scope"))
         }
+        // Graph-query refusals; this gear issues no graph queries, so reaching
+        // either is a programmer error, like `Invalid`.
+        err @ (ScopeError::UnresolvedScopeProperty { .. } | ScopeError::Pgq(_)) => {
```
MikeFalcon77: **MEDIUM**

This gear builds toolkit-db with `features = ["sqlite"]` and no `pg`, and it now links a PostgreSQL-19 SQL/PGQ crate to write this arm.

`ScopeError::Pgq` wraps `toolkit_sea_orm_pgq::PgqError`, so `toolkit-sea-orm-pgq` had to become an unconditional dependency of toolkit-db (`pgq = ["pg"]`, and the crate is in plain `[dependencies]`). ADR-0002 "Backend gating" states the opposite requirement: "a gear cannot even name the types on a build without a Postgres backend."

The arm is also unreachable here. Both variants are constructed only inside `secure::pgq`, which is `#[cfg(feature = "pgq")]`, and this gear does not enable `pgq` — so neither variant is constructible in this build.

Fix belongs in toolkit-db, not here: give the variant a toolkit-db-owned payload instead of the foreign error type, and `toolkit-sea-orm-pgq` goes back to `optional = true` under `pgq = ["pg", "dep:toolkit-sea-orm-pgq"]`. The variant can stay unconditional, so nothing about feature unification changes and this file does not move again. Detail in the `ScopeError` thread on `libs/toolkit-db/src/secure/pgq.rs`.

gears/chat-engine/chat-engine/src/infra/db/error_map.rs
```diff
@@ -66,6 +66,14 @@ impl From<ScopeError> for ChatEngineError {
                 reason: format!("invalid scope: {msg}"),
                 source: None,
             },
+            // Graph-query refusals; chat-engine issues no graph queries, so
+            // reaching either is a programmer error, like `Invalid`.
+            err @ (ScopeError::UnresolvedScopeProperty { .. } | ScopeError::Pgq(_)) => {
```
MikeFalcon77: **MEDIUM**

Six sites across five gears had to add this same dead arm, and none of them needed to.

The arms exist only because `ScopeError` is exhaustive. Add `#[non_exhaustive]` to it and a downstream `match` needs one wildcard instead of a new arm per ORM error variant — this round's churn becomes the last one, and the `Pgq` variant can then also be `#[cfg]`-gated without breaking anyone under feature unification.

You already applied exactly this reasoning to `PgqError` in this PR: "adding a variant must not be a breaking change for every downstream `match`." `ScopeError` is the enum where that actually costs something — it is matched in `chat-engine`, `credstore`, `mini-chat`, `simple-user-settings` and `account-management` (two sites).

Trade-off worth stating: gears lose exhaustiveness checking on `ScopeError`, so a future variant they should handle specifically falls into the wildcard instead of failing the build. That is the standard cost for a library error enum, and the alternative is repeating these six edits every time the ORM grows an error.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: You're right on both counts. Fixed in 3388c4c5d, both independent parts:
 1. The payload is toolkit-db's own: ScopeError::GraphSyntax(String), filled by a plain map_err adapter in the gated module (not a From impl — .to_string() inside From is what the DE1302 dylint refuses). toolkit-sea-orm-pgq is optional = true again under pgq = ["pg", "dep:toolkit-sea-orm-pgq"], so a SQLite-only gear links nothing from PostgreSQL 19. pgq_tests matches on the message text for the arity case, which is enough there.
 2. ScopeError is #[non_exhaustive], with the trade-off you state written into its doc. The seven downstream matches  now keep one wildcard arm.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 86601db07: the condition is now "a sibling exists and more than one alternative survives" (!siblings.is_empty() && compiled.len() > 1), which covers in_group OR eq as well. Counted over surviving alternatives rather than declared ones, since a constraint dropped for an unresolvable property contributes no branch. a_sibling_beside_a_plain_alternative_is_refused_too pins the mixed case and that a lone sibling-needing constraint stays servable.

libs/toolkit-db/tests/pg/mod.rs
vasylcf: Set in 69a269931: env: GEARS_TEST_PG_GRAPH_REQUIRED: "1" on the pgq step. It earned its keep immediately — locally Docker had pruned the image and testcontainers' pull hit a 401 on stale hub credentials; with the flag the lane failed instead of passing green and empty. On the duplication: e4ed9c075 narrows make test-pgq to -E 'kind(lib) | binary(mod) | binary(ui)' — the unit suites under the pgq cfg, the PG19 binary and the trybuild guards — so it no longer re-runs the PG18 suites test-pg just ran.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Agreed that an inference here is worse than a decision. 3388c4c5d adds PathBuilder::correlate_with_anchor() as the explicit opt-in; where_() no longer influences the FROM at all. The middle case you name is now a test (a_predicate_that_does_not_correlate_leaves_the_anchor_out: where_(a.id = 5) and no opt-in → no anchor), the anchored-walk PG19 test opts in explicitly, and the exact-bound-values test lost the anchor's value accordingly.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Fixed in 3388c4c5d: PendingElement carries the entity's table, GraphDeclaration::table_of(label) exposes what was registered (via a new VertexTable::table() accessor in the syntax crate, 3cd951166), and check_declaration refuses a mismatch by name. Test: a_label_bound_to_an_unregistered_entity_is_refused — two entities under one LABEL, only one registered.

libs/toolkit-db/src/secure/pgq.rs
vasylcf: Filed as #4697, with a unit-level reproducer and the same guard proposed for build_scope_condition. Left out of this PR as you suggest.

libs/toolkit-sea-orm-pgq/src/tests.rs
vasylcf: Fixed in 3cd951166. One parametrised case now exercises each pattern position alone — graph name, vertex label, edge label, element variable, projected property, projected alias — each with the structural check, so a position that bypassed write_ident fails on its own assertion. A DDL case feeds the same payload to the graph name, a table, a label, a key column, a property and both endpoint references, and asserts that only the keyword frame survives outside the identifiers (for DROP the surviving syntax is exactly DROP PROPERTY GRAPH IF EXISTS ).

gears/simple-user-settings/simple-user-settings/src/infra/storage/sea_orm_repo.rs
vasylcf: Fixed in toolkit-db, as you say (3388c4c5d): the variant's payload is toolkit-db's own and the syntax crate is optional again, so this gear links nothing SQL/PGQ-related. The arm here became a wildcard (ScopeError is #[non_exhaustive] now), so this file does not move again when the ORM grows an error.

gears/chat-engine/chat-engine/src/infra/db/error_map.rs
vasylcf: Done in 3388c4c5d — ScopeError is #[non_exhaustive], the trade-off you state is recorded in its doc comment, and all seven sites (this one, credstore, mini-chat, simple-user-settings, both account-management sites, and github-mirror, which had already broken on the merge with main) keep a single wildcard arm mapping to their internal error.

libs/toolkit-db/src/secure/entity_traits.rs
```diff
@@ -152,4 +156,36 @@ pub trait ScopableEntity: EntityTrait {
     /// Manual implementors must provide all property arms explicitly.
     #[must_use]
     fn resolve_property(property: &str) -> Option<Self::Column>;
+
```
striped-zebra-dev, line 159: Please rework this AI-slope comment, in its current form it makes no sense.

libs/toolkit-db/src/secure/entity_traits.rs
```diff
@@ -152,4 +156,36 @@ pub trait ScopableEntity: EntityTrait {
...
+    /// entity then passes the Policy 2 gates while resolving nothing — the
+    /// silent deny-all those gates exist to refuse). Manual implementors
+    /// enumerate the same set their `resolve_property` can return;
+    /// `#[derive(Scopable)]` generates both from one configuration so they
+    /// cannot disagree.
+    ///
+    /// Note `type_col` is **not** a scope column: `resolve_property` has no
+    /// well-known property name for it, so no scope constraint can address it.
+    #[must_use]
+    fn scope_columns() -> Vec<Self::Column>;
```
striped-zebra-dev, line 176: We already have tenant column, why do we need this in the same interface?

striped-zebra-dev (changes requested):

libs/toolkit-db/src/secure/entity_traits.rs
vasylcf: Rewritten in 0982bf940

libs/toolkit-db/src/secure/entity_traits.rs
vasylcf: The 3 getters are not the full set. `resolve_property` can also map custom `pep_prop(...)` columns, such as `department_id`. Every scope-addressable column must be included in the graph element’s `PROPERTIES` list; otherwise, it cannot be filtered inside `MATCH`.

So `scope_columns` is the enumerable inverse of `resolve_property`, not a duplicate of the three getters. A getter-based default was proposed earlier but rejected because custom properties would be silently omitted.

The duplication is real, though. A clean follow-up would introduce a single `SCOPE_PROPERTIES` table, with both methods generated from it by the derive. I’d prefer to keep that outside this PR ...

docs/TESTING.md
```diff
@@ -140,6 +140,7 @@ behind the `integration` Cargo feature so that `cargo test --workspace` (without
 |---------|----------|---------|
 | `cf-gears-toolkit-db` | `sqlite,integration` | SQLite (in-process) |
 | `cf-gears-toolkit-db` | `pg,integration` | PostgreSQL (requires running instance) |
+| `cf-gears-toolkit-db` | `pgq,integration` | PostgreSQL 19 SQL/PGQ (testcontainers; `make test-pgq`) |
```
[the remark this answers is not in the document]
vasylcf: Fixed in 965ffe8f1: make test-pgq is listed in the quick-reference block and in 4.2, 4.3 names the PGQ step (and that it runs with GEARS_TEST_PG_GRAPH_REQUIRED=1), and test-db is documented as including it.

libs/toolkit-db/src/secure/entity_traits.rs
MikeFalcon77: It's hard to understand an answer but anyway please fill an issue for it

vasylcf: "Scope_columns" is the list of columns a scope may touch. 
"resolve_property" can't give a list (it's a lookup), and the graph DDL needs the list — a column missing from PROPERTIES can't be filtered in MATCH. The three getters are not enough because custom pep_prop columns aren't behind any getter.

You're right that having two methods that must agree is a weak spot. Filed #4726 with the details and a proposal to replace both with one table.

MikeFalcon77 (approved):

PR #4640, opened 2026-08-20
chore(testing): pin database container versions in one place

### LINKED ISSUE #4616: Pin database container versions in one place: PostgreSQL 18 floor, PostgreSQL 19 for graph-storage, MySQL 9.7

MikeFalcon77:
- Add libs/test-containers (cf-gears-test-containers, not published) as the single place these versions are pinned, with env var overrides (GEARS_TEST_PG_TAG etc.) for running a version matrix in CI. Route every fixture through it: ~240 call sites across ledger, toolkit-db, account-management, coord, resource-group, cluster, and the timescaledb plugin.

- Raise the pins: PostgreSQL 11 -> 18, MySQL 8.1 -> 9.7, TimescaleDB 2.17.2-pg16 -> 2.29.2-pg18, MariaDB lts -> 11.8.
PostgreSQL 19 is reserved for the future graph-storage gear (needs SQL/PGQ) but unused for now.

- The version bump surfaced one real bug: PostgreSQL 18 reports ON DELETE RESTRICT as SQLSTATE 23001 (restrict_violation) instead of the old 23503 (foreign_key_violation). Fixed to accept both codes, with a test.

- Add a CI check that fails if anything outside libs/test-containers calls the default constructors again.

Closes #4616

## Summary by CodeRabbit

* **New Features**
  * Centralized pinned database containers for PostgreSQL, graph PostgreSQL, MySQL, MariaDB, and TimescaleDB integration tests.
  * Added environment-based image tag overrides and optional graph-database test requirements.
  * Updated TimescaleDB testing to use a PostgreSQL 18 image.

* **Bug Fixes**
  * Improved classification of PostgreSQL foreign-key constraint errors.

* **Documentation**
  * Documented image pinning, overrides, and graph test behavior.

* **Tests**
  * Standardized container setup and added CI validation against unpinned images.
  * Added coverage for an additional foreign-key error code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,11 @@ jobs:
               # Input to `make check-release-config` in the test job: a config-only
               # edit must still get the release configuration validated.
               - 'release-plz.toml'
+              # Input to `e2e_sidecar_pins_the_same_timescaledb_image` in
+              # libs/test-containers: a Python-only edit to the E2E sidecar must
+              # still get the cross-lane TimescaleDB pin checked, or the guard
+              # silently never runs on exactly the diff it exists to catch.
+              - 'testing/e2e/lib/sidecars.py'
             fips:
               - 'libs/toolkit/**'
               - 'libs/toolkit-http/**'
@@ -166,6 +171,20 @@ jobs:
           else
             echo "sccache failed, falling back to plain rustc"
           fi
+
+      # Database image tags are pinned in libs/test-containers (docs/TESTING.md
+      # 4.4). A fixture that constructs a container any other way re-introduces
+      # the transitive pin through Cargo.lock that issue #4616 removed: a
+      # dependency bump then moves the database under every test with nothing in
+      # the diff to show it.
+      #
+      # This parses the workspace with `syn` rather than grepping it, so import
+      # renames, type aliases, non-literal image names and creative formatting
+      # cannot slip past. Placed here, after the cache and sccache setup, because
+      # it needs a toolchain -- but still ahead of clippy, which is the expensive
+      # step in this job.
+      - name: No unpinned database container images
+        run: cargo xtask check-test-container-pins
 
       - name: cargo clippy
         run: make clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ members = [
     "libs/toolkit",
     "libs/toolkit-macros",
     "libs/toolkit-macros-tests",
+    "libs/test-containers",
     "libs/toolkit-db",
     "libs/toolkit-db-macros",
     "libs/toolkit-auth",
@@ -675,6 +676,11 @@ heck = "0.5"
 proc-macro-crate = "3"
 
 # Testing utilities
+# Internal, unpublished: the single place every database container tag is
+# pinned (see libs/test-containers). Lives here rather than with the other
+# `libs/**` entries because it is only ever a dev-dependency, alongside the
+# `testcontainers` pair it wraps.
+test-containers = { package = "cf-gears-test-containers", path = "libs/test-containers" }
 trybuild = "1"
 serial_test = { version = "3", features = ["file_locks"] }
 criterion = { version = "0.8", default-features = false }
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -50,6 +50,9 @@ make fuzz                  # fuzz — 30 s smoke per target
 make check                 # full quality gate (fmt + clippy + test + security)
 make all                   # full pipeline (build + check + test-sqlite + e2e-local)
 
+# CI gate: no database container built outside libs/test-containers (4.4)
+cargo xtask check-test-container-pins
+
 make build GEAR=file-parser      # one gear package plus SDK package
 make test GEAR=file-parser       # one gear package plus SDK package
 make run GEAR=file-parser        # server with only this gear feature set
@@ -154,6 +157,71 @@ make test-users-info-pg    # users-info Postgres integration
 
 The `integration` job in `ci.yml` runs SQLite, Postgres, and MySQL integration tests
 plus macro UI tests on every PR (Ubuntu only).
+
+### 4.4 Database container images
+
+Every fixture that starts a database container goes through
+[`libs/test-containers`](../libs/test-containers/src/lib.rs) (crate
+`cf-gears-test-containers`, imported as `test_containers`). It is the single place image
+versions are pinned, so a version change is one reviewed diff instead of a grep across the
+workspace.
+
+**Do not call `Postgres::default()` or `Mysql::default()` in a test**, and do not call
+`GenericImage::new()` outside that crate. Those constructors take their tag from
+`testcontainers-modules`, which pins it transitively through `Cargo.lock` — a dependency bump
+then changes the database under every test with nothing in the diff to show it.
+
+`cargo xtask check-test-container-pins` enforces this in CI (the `clippy` job). It parses every
+tracked `.rs` file with `syn`, so an import rename (`Postgres as Pg`), a type alias, a
+non-literal image name or an oddly wrapped call is caught the same as the plain spelling.
+
+```rust
+// Add to [dev-dependencies]:  test-containers = { workspace = true }
+use testcontainers::runners::AsyncRunner;
+
+let container = test_containers::postgres().start().await?;
+
+// Non-default database name (POSTGRES_DB is applied at image level):
+let container = test_containers::postgres_named("cluster_test").start().await?;
+
+// A suite pinned to a floor of its own, above or below the workspace pin —
+// "16-alpine" is just an illustration value here. GEARS_TEST_PG_TAG still
+// wins over it, so the suite stays in the CI version matrix; chaining
+// `.with_tag(...)` onto `postgres()` would silently opt out of the override.
+let container = test_containers::postgres_tagged("16-alpine").start().await?;
+```
+
+Helpers: `postgres()`, `postgres_named()`, `postgres_tagged()`, `postgres_graph()`, `mysql()`,
+`timescaledb()`, `mariadb()`.
+
+#### Version-matrix overrides
+
+Each tag can be overridden from the environment, so CI can run a matrix without touching code.
+An unset *or empty* variable means "use the pinned constant".
+
+| Variable | Overrides |
+|---|---|
+| `GEARS_TEST_PG_TAG` | `POSTGRES_TAG` |
+| `GEARS_TEST_PG_GRAPH_TAG` | `POSTGRES_GRAPH_TAG` |
+| `GEARS_TEST_MYSQL_TAG` | `MYSQL_TAG` |
+| `GEARS_TEST_TIMESCALEDB_TAG` | `TIMESCALEDB_TAG` |
+| `GEARS_TEST_MARIADB_TAG` | `MARIADB_TAG` |
+
+```bash
+GEARS_TEST_PG_TAG=16-alpine cargo nextest run -p cf-gears-toolkit-db --features pg,integration
+```
+
+`GEARS_TEST_TIMESCALEDB_TAG` is read by both lanes: the Rust plugin fixtures via
+`test_containers::timescaledb()`, and the Python E2E sidecar via `timescaledb_tag()` in
+[`testing/e2e/lib/sidecars.py`](../testing/e2e/lib/sidecars.py). A matrix run therefore keeps
+migrations and E2E on the same image instead of silently testing two different ones.
+
+`GEARS_TEST_PG_GRAPH_REQUIRED=1` turns an unavailable PostgreSQL 19 image into a failure rather
+than a graceful skip; it is off by default while that tag is pre-GA. Unset, empty, `0`, `false`,
+`off` and `no` all mean off.
+
+Only concrete tags belong in the constants — a floating alias such as `lts` or `latest`
+re-points under CI with nothing in the diff, which is the drift this crate exists to prevent.
 
 ---
 
--- a/libs/test-containers/src/lib.rs
+++ b/libs/test-containers/src/lib.rs
@@ -1,0 +1,557 @@
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+//! Single source of truth for the database container images this workspace's
+//! integration tests start.
+//!
+//! Every fixture that brings up a database goes through this crate, so a
+//! version change is one edit here instead of a grep across ~240 call sites
+//! (<https://github.com/constructorfabric/gears-rust/issues/4616>).
+//!
+//! # Why not `Postgres::default()`
+//!
+//! `testcontainers-modules` hardcodes its own tags — `postgres:11-alpine` and
+//! `mysql:8.1`, both long past end of life. Calling the default constructor
+//! pins the database version transitively, through `Cargo.lock`: bumping
+//! `testcontainers-modules` then changes the database under every test in the
+//! repository with nothing in the diff to show it. [`postgres()`] and
+//! [`mysql()`] delegate to those same images but override the tag with the
+//! constants below, so the version is stated in one place and reviewed like
+//! any other change.
+//!
+//! # Overrides
+//!
+//! Each tag can be overridden by an environment variable, so CI can run a
+//! version matrix without touching code:
+//!
+//! | Variable | Overrides |
+//! |---|---|
+//! | `GEARS_TEST_PG_TAG` | [`POSTGRES_TAG`] |
+//! | `GEARS_TEST_PG_GRAPH_TAG` | [`POSTGRES_GRAPH_TAG`] |
+//! | `GEARS_TEST_MYSQL_TAG` | [`MYSQL_TAG`] |
+//! | `GEARS_TEST_TIMESCALEDB_TAG` | [`TIMESCALEDB_TAG`] |
+//! | `GEARS_TEST_MARIADB_TAG` | [`MARIADB_TAG`] |
+//!
+//! An unset *or empty* variable means "use the constant".
+//!
+//! Only concrete tags belong in the constants below. A floating alias such as
+//! `lts` or `latest` re-points under CI with nothing in the diff to show it,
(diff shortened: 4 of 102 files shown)


.github/workflows/ci.yml
```diff
@@ -123,6 +123,22 @@ jobs:
...
+        run: |
+          if hits=$(grep -rnE \
+              'Postgres::default\(\)|Mysql::default\(\)|GenericImage::new\("(mariadb|timescale/timescaledb)"' \
+              --include='*.rs' . \
+              | grep -v '/target/' \
+              | grep -v 'libs/test-containers/'); then
+            echo "::error::Database images must come from libs/test-containers (see docs/TESTING.md 4.4)."
+            echo "$hits"
+            exit 1
+          fi
```
MikeFalcon77: fixed

Cargo.toml
```diff
@@ -56,6 +56,7 @@ members = [
     "libs/toolkit",
     "libs/toolkit-macros",
     "libs/toolkit-macros-tests",
+    "libs/test-containers",
```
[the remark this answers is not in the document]
MikeFalcon77: non-sense

docs/TESTING.md
```diff
@@ -141,6 +141,60 @@ make test-users-info-pg    # users-info Postgres integration
...
+
+let container = test_containers::postgres().start().await?;
+
+// Non-default database name (POSTGRES_DB is applied at image level):
+let container = test_containers::postgres_named("cluster_test").start().await?;
+
+// A suite whose migrations need a higher floor than the workspace pin.
+// GEARS_TEST_PG_TAG still wins, so the suite stays in the CI version matrix —
+// chaining `.with_tag(...)` onto `postgres()` would silently opt out of it.
+let container = test_containers::postgres_tagged("16-alpine").start().await?;
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

libs/test-containers/src/lib.rs
```diff
@@ -0,0 +1,512 @@
...
+            POSTGRES_GRAPH_TAG,
+            MYSQL_TAG,
+            TIMESCALEDB_TAG,
+            MARIADB_TAG,
+        ] {
+            assert!(
+                !matches!(tag, "latest" | "lts" | "stable" | "edge"),
+                "{tag} is a floating alias, not a pin"
+            );
+        }
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

.github/workflows/ci.yml
```diff
@@ -123,6 +123,22 @@ jobs:
...
+        run: |
+          if hits=$(grep -rnE \
+              'Postgres[[:space:]]*::[[:space:]]*default[[:space:]]*\([[:space:]]*\)|Mysql[[:space:]]*::[[:space:]]*default[[:space:]]*\([[:space:]]*\)|GenericImage[[:space:]]*::[[:space:]]*new[[:space:]]*\([[:space:]]*"(mariadb|timescale/timescaledb)"' \
+              --include='*.rs' . \
+              | grep -v '/target/' \
+              | grep -v 'libs/test-containers/'); then
+            echo "::error::Database images must come from libs/test-containers (see docs/TESTING.md 4.4)."
+            echo "$hits"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
MikeFalcon77: fixed

testing/e2e/lib/sidecars.py
```diff
@@ -69,13 +69,15 @@ def skip_without_docker() -> None:
...
+    libs/test-containers/src/lib.rs, which is where the Rust side pins every
+    database image. A skew means the plugin's migrations are validated against
+    a different PostgreSQL major than E2E runs; the Rust test
+    `e2e_sidecar_pins_the_same_timescaledb_image` fails the build if this
+    literal drifts from the constants.
     """
 
-    IMAGE = "timescale/timescaledb:2.17.2-pg16"
+    IMAGE = "timescale/timescaledb:2.29.2-pg18"
     LABEL = "cf-gears-e2e=usage-collector"
```
fdlockgraf, line 103: ### TimescaleDB override does not reach Python E2E

`testing/e2e/lib/sidecars.py` hardcodes:

```python
IMAGE = "timescale/timescaledb:2.29.2-pg18"
```

Therefore `GEARS_TEST_TIMESCALEDB_TAG` affects Rust plugin tests but not E2E. A version-matrix run can silently test Rust migrations against one TimescaleDB version and E2E against another.

The cross-file Rust test only verifies the default literal; it does not solve effective-tag skew. The Python sidecar should honor the same environment variable.

.github/workflows/ci.yml
```diff
@@ -126,6 +126,37 @@ jobs:
...
```
fdlockgraf: ### Suggestion: replace the grep policy with an AST-based `xtask`

Add an `xtask` command:

```bash
cargo xtask check-test-container-pins
```

Using `syn`, it should:

1. Scan tracked `*.rs` files, excluding `libs/test-containers`.
2. Parse Rust syntax, so comments and formatting are handled correctly.
3. Reject:
   - `Postgres::default()`
   - `Mysql::default()`
   - Aliased forms such as `Pg::default()`
   - Type-alias forms where practical
   - Any `GenericImage::new(...)` outside the centralized crate
4. Report the path and source location.
5. Include regression tests for multiline calls, aliases, and comments.

Banning every external `GenericImage::new` is simpler and safer than trying to evaluate its first argument. Currently, all legitimate database generic images should come from `test_containers`.

CI then becomes:

```yaml
- name: No unpinned database container images
  run: cargo xtask check-test-container-pins
```

Example visitor shape:

```rust
impl<'ast> syn::visit::Visit<'ast> for PolicyVisitor {
    fn visit_expr_call(&mut self, call: &'ast syn::ExprCall) {
        if is_banned_default_constructor(call, &self.aliases)
            || is_generic_image_constructor(call, &self.aliases)
        {
            self.report(call);
        }

        syn::visit::visit_expr_call(self, call);
    }
}
```

This would make the policy formatting-independent and testable, rather than relying on line-oriented regex matching.

fdlockgraf:
apart from these two AI comments - looks good. maybe we can even record a separate issue to make a dylint rule regarding hardcoding the versions?

fdlockgraf (changes requested):

.github/workflows/ci.yml
MikeFalcon77: Filed the dylint follow-up: constructorfabric/cargo-gears#112

The lints live in `cargo-gears-lints`, not here, so that's where it has to go — this repo only carries the config in `dylint.toml`. Suggested `DE1102` in `de11_testing`.

Worth noting the dylint version can be strictly better than the xtask one: as a late pass it has type information, so it can match on "resolves to `testcontainers_modules::postgres::Postgres`" instead of "is called `Postgres`". That covers re-exports and aliases arriving through other modules, which the name-based approach here can't follow. The issue points at `tools/xtask/src/pins.rs` and its 16 regression cases as the starting point, and notes that this gate can be deleted once the lint is enabled.

MikeFalcon77:
> apart from these two AI comments - looks good. maybe we can even record a separate issue to make a dylint rule regarding hardcoding the versions?

I did it: https://github.com/constructorfabric/cargo-gears/issues/112

testing/e2e/lib/sidecars.py
MikeFalcon77: Fixed. `sidecars.py` now has its own `timescaledb_tag()` that reads `GEARS_TEST_TIMESCALEDB_TAG` (unset or empty falls back to the pinned constant), mirroring `test_containers::timescaledb_tag()` on the Rust side. `IMAGE` is built from `TIMESCALEDB_IMAGE:timescaledb_tag()` instead of the hardcoded literal.

Also updated the cross-file test in `libs/test-containers/src/lib.rs` (`e2e_sidecar_pins_the_same_timescaledb_image`) to check for `TIMESCALEDB_IMAGE`, `TIMESCALEDB_TAG`, and `ENV_TIMESCALEDB_TAG` in `sidecars.py` — the last one is what actually catches this class of bug, since it fails if the Python side ever stops reading the override.

And added `testing/e2e/lib/sidecars.py` to the `rust` path filter in `ci.yml` so a Python-only edit here still triggers that Rust test.

Pushed in the latest commit.

fdlockgraf (approved):

PR #4641, opened 2026-08-20
feat(types-registry): implement P0-1

### LINKED ISSUE #4620: [Types Registry P0] Phase 0 — gts-rust 0.12.0 upgrade (fail fast)
Parent: #4619

**Delivery:** this phase and Phase 1 land together in the **first PR**, which also adds the P0
spec, plan and task list. Nothing of P0 is on `main` yet.

## Goal

Move the whole workspace onto `gts-rust` 0.12.0 so the tri-state compatibility verdict,
`ContentModel` classification and `compare_documents` become available — and prove the
corrected semantics break nothing already declared. This task's failure mode is *other gears*,
which is why it runs first and alone, before a line of registry code is written.

## Tasks

- [ ] **T1** — upgrade to `gts-rust` 0.12.0 and re-validate every declared identifier.
  `gts`, `gts-id` and `gts-macros` are pinned together at 0.12.0 (a split pin puts the
  identifier crate and the semantics crate on different specifications); `gts-dylint` and
  `gts-macros-cli` must not lag.

## Exit criteria — Checkpoint 0

- [ ] `make ci` green
- [ ] Every declared GTS identifier still admits under 0.12.0; anything that does not is fixed
      or explicitly waived in writing
- [ ] Every difference in `#[gts_type_schema]`-generated schema documents versus 0.11.0 is
      enumerated and accounted for — none silent
- [ ] Human review

## Notes

The API is additive; the semantics are not. Known blast radius includes schemas that live
outside compiled macro literals — YAML configs and e2e fixtures — which a compile-time sweep
does not cover.

## References

- `gears/system/types-registry/docs/p0/SPEC.md` — §7 (the upgrade), D8 (the three crates move together)
- `gears/system/types-registry/docs/p0/plan.md` — Phase 0
- `gears/system/types-registry/docs/p0/todo.md` — T1

### LINKED ISSUE #4621: [Types Registry P0] Phase 1 — one global entity of each kind, persisted, async, end to end
Parent: #4619 · Depends on: Phase 0 (#4620)

**Delivery:** this phase and Phase 0 land together in the **first PR**, which also adds the P0
spec, plan and task list. Nothing of P0 is on `main` yet.

## Goal

Prove the architecture on the narrowest possible slice: register one global Type Schema and
one registered Instance over REST, asynchronously, through the database, end to end —
migration → entity → repository → transient store → acceptance → worker → REST. Everything
after this phase *widens* this path rather than reshaping it.

The new surface is **additive**: the existing v1 routes keep serving the in-memory store, and
the async surface ships under `/types-registry/v2/`. No dual-write, no fallback read — a route
never straddles the two stores. Consumers are untouched.

## Tasks

- [ ] **T2** — one initial SeaORM migration for the 9 P0 tables. Tenant columns and CHECK
      constraints created exactly as specified and never populated with tenant scope;
      `source_claim` and `routing_config` are not created.
- [ ] **T3** — SeaORM entities for the core six, each `#[derive(Scopable)]` +
      `#[secure(unrestricted)]` with a comment recording the P1 switch to `owner_tenant_id`.
- [ ] **T4** — repositories on `DBRunner`: keyed reads, insert-if-absent for `version_family`,
      compare-and-swap on `entity.resource_version`, canonical-order locking helpers.
- [ ] **T5** — a transient `gts-rust` store built from database rows for one admission unit and
      dropped with it. No `ArcSwap`, no snapshot, no process-lifetime store; reads go to the
      database.
- [ ] **T6** — typed configuration: `allow_compatibility_force`, `limits.*` (including
      `activation_write_set`), `registration_policy`, `worker.*`.
- [ ] **T7** — the acceptance path: envelope and batch bounds, canonical identifier check,
      registration policy, Draft-07 dialect gate, request fingerprint, `Idempotency-Key`
      resolution, then one transaction inserting the operation, its items and the outbox message.
- [ ] **T8** — the admission worker as a plain function of `(operation_id, runner)` for one
      dependency-free candidate: evaluate, then commit family, entity, revision and the
      current-state projection with materialized artifacts.
- [ ] **T9** — REST: `POST /entities` (`202` + operation), `GET /operations/{id}`,
      `GET /entities/{key}`.
- [ ] **T9a** — restore the v1 contract verbatim; move the async surface to `/types-registry/v2/`.
- [ ] **T10** — registered Instances, plus the identifier-derived dependency closure (a
      derivation base and an Instance's conforming type are pure functions of the identifier,
      so they need no edge table) and the family *kind* rule.

## Exit criteria — Checkpoint 1 (proves the architecture)

- [ ] A fixture Type Schema registers over REST, the operation reaches `completed`, and the
      entity with its resolved artifacts is readable
- [ ] Entity and artifacts survive a process restart byte-identically
- [ ] An Instance registers against a Type Schema committed by an earlier operation, and a
      derived Type Schema admits against a committed base with the `dependency` table empty
- [ ] Consumers untouched: the old trait is still served from the in-memory repository
- [ ] The new path holds no entity state between admissions
- [ ] Gear tests green on SQLite, PostgreSQL and MySQL
- [ ] `make dylint` green over the full workspace
- [ ] v1 intact, `make e2e-local` green, no e2e file edited for the new surface
- [ ] Human review. Two hazards this gate is expected to settle rather than discover:
      another gear (`resource-group`) registers five routes inside this gear's
      `/types-registry/v1/*` namespace, which T27 / T24a later widen; and several gear groups
      sit outside `config/e2e-features.txt`, so their GTS declarations are proven at compile
      time only, never admitted into a live registry

## References

- `gears/system/types-registry/docs/p0/SPEC.md` — §8.1 write path, §8.2 read path, §9 database, D1–D3
- `gears/system/types-registry/docs/p0/plan.md` — P1 (vertical slices), P2/P3 (self-seeding, worker start), P6 (no store between admissions), P12 (v1 stays, async ships as v2), P13 (Instances in Phase 1)
- `gears/system/types-registry/docs/p0/todo.md` — T2–T10

Already on `main`:

- `gears/system/types-registry/docs/DESIGN.md` §3.7, §3.2
- `gears/system/types-registry/docs/database.sql` — the normative schema target

### LINKED ISSUE #4685: Flaky test: LeaseClock virtual advance can undershoot by 1 ms
## Summary

`cf-gears-cluster-sdk::lease::lease_tests::the_production_clock_ignores_virtual_time_but_the_test_clock_tracks_it` is flaky on Ubuntu CI.

## Evidence

- In [CI run 33408383650, attempt 1](https://github.com/constructorfabric/gears-rust/actions/runs/33408383650/job/99542233340), the test passed.
- In [the same run, attempt 2](https://github.com/constructorfabric/gears-rust/actions/runs/33408383650/job/99627187535), the test failed on the same PR head SHA with:
  ```
  the injected virtual clock must still track `advance`
  (1788208873255 -> 1788212473254)
  ```
  The observed delta is 3,599,999 ms after advancing virtual time by 3,600,000 ms.
- PR #4641 does not change `gears/system/cluster`.

## Likely cause

`LeaseClock::virtual_clock()` records its wall-clock anchor before the test captures `test_before`. If the wall clock crosses a millisecond boundary between those two reads, `test_before` can be 1 ms ahead of the anchor. After `tokio::time::advance(Duration::from_hours(1))`, the hybrid clock resolves from the original anchor, so the exact assertion:

```rust
test_after >= test_before + 3_600_000
```

can fail by 1 ms even though virtual time was tracked correctly.

## Suggested fix

Make the assertion tolerant of the wall-clock sampling offset (or assert against a baseline derived from the virtual clock anchor) while preserving the intent that the injected clock advances by approximately one hour.

aviator5:
closes #4620
closes #4621
closes #4685

## Summary by CodeRabbit

* **New Features**
  * Added durable Types Registry storage with SQLite, PostgreSQL, and MySQL support.
  * Added asynchronous registration, operation polling, entity retrieval, idempotency handling, validation, and configurable policies.
  * Added persistence for schemas, instances, revisions, dependencies, lifecycle states, and operation history.
  * Added quickstart and local end-to-end configurations.
* **Bug Fixes**
  * Standardized tenant schemas around a required `payload` envelope.
  * Improved SQLite connection-string handling.
* **Documentation**
  * Documented the Types Registry v2 API and database-backed architecture.
  * Updated GTS 0.12.0 compatibility guidance.


Cargo.toml
```diff
@@ -694,11 +694,10 @@ fnv = "1.0"
...
 
-# GTS library.
-# Keep in sync with tools/dylint_lints/Cargo.toml.
-gts-id = "0.11.0"
-gts = "0.11.0"
-gts-macros = "0.11.0"
+# GTS library — published on crates.io.
+gts-id = "0.12.0"
+gts = "0.12.0"
+gts-macros = "0.12.0"
```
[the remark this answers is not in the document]
aviator5: We keep the workspace’s existing SemVer convention (the previous GTS declarations were also non-exact); `Cargo.lock` pins all three to 0.12.0. `cf-studio-path` is already configured in `AGENTS.md`/`CLAUDE.md`, where Studio reads it, not in `Cargo.toml`.

gears/system/types-registry/types-registry/src/domain/family.rs
```diff
@@ -0,0 +1,92 @@
...
+            key.push('.');
+            key.push_str(last.type_name());
+        }
+        // A UUID tail is refused at acceptance (§8.1 step 4), so this is
+        // unreachable from the write path. Falling back to the raw form keeps the
+        // function total rather than panicking on an input it cannot see.
+        GtsIdSegment::UuidTail(_) => key.push_str(raw),
+    }
+    FamilyKey(key)
+}
```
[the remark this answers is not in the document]
aviator5: Intentional: `family_key` is kind-agnostic so Type/Instance conflicts serialize on one family. Admission rejects a mismatched kind as `family_kind_conflict`, with tests covering both insertion orders.

gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs
```diff
@@ -0,0 +1,602 @@
...
+
+/// The path of the first `$schema` below the root that does not normalize onto
+/// the same dialect. A nested `$schema` equal to the root's — after
+/// normalization, so `…/schema` and `…/schema#` agree — is not a conflict
+/// (ADR-0014).
+fn conflicting_dialect(value: &Value, path: &str, is_root: bool) -> Option<String> {
+    match value {
+        Value::Object(map) => {
+            if !is_root
+                && let Some(declared) = map.get("$schema").and_then(Value::as_str)
```
Artifizer: **[MAJOR] F-1-MAJOR-codex-gpt-5.6-sol** in `gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs`

The nested-dialect check only examines `$schema` values for which `Value::as_str()` succeeds. A nested declaration such as `"$schema": 7` or `"$schema": null` is therefore silently ignored, even though the acceptance contract says every nested `$schema` must normalize to the supported Draft-07 dialect. Such a request can pass synchronous acceptance, consume its `Idempotency-Key`, and create an operation before later schema evaluation rejects it. This differs materially from the intended `ConflictingDialect` refusal and allows malformed dialect declarations through the validation stage specifically responsible for rejecting them.

**Suggested fix:** Treat every nested object containing `$schema` as a declaration. Return its path when the value is not a non-empty supported string or when the string does not normalize to Draft-07. Add tests covering non-string and empty nested declarations.

docs/api/api.json
```diff
@@ -30474,6 +30756,411 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/OperationAcceptedDto"
+                }
+              }
+            },
+            "description": "Replay of an operation that is already terminal"
+          },
+          "202": {
```
Artifizer, line 43745: **[MAJOR] F-1-MAJOR-codex-default** in `docs/api/api.json`

The new submission operation promises a `Location` header and relies on response headers for polling and replay behavior, but neither the `200` nor `202` response declares any headers. The implementation emits `Location`, emits `Retry-After` for `202`, and conditionally emits `Idempotency-Replayed`, so generated clients built from this contract may discard or fail to expose these protocol signals. In particular, the documented instruction to poll “at the returned Location” is not represented structurally in the API contract. The omission originates in the route definition as well as the generated API document: the route declares only the JSON response bodies and does not describe the runtime response headers.

**Suggested fix:** Declare `Location` for both successful responses, `Retry-After` for `202`, and the optional `Idempotency-Replayed` header wherever a replay can be returned. Extend the route/OpenAPI builder if necessary, add a contract test for these headers, and regenerate `docs/api/api.json` rather than editing the generated artifact directly.

gears/file-storage/file-storage/tests/service_test.rs
```diff
@@ -53,15 +53,7 @@ async fn build_service_with_page_sizes(
...
-    // already starts with '/'.
-    let mut file = path.to_string_lossy().replace('\\', "/");
-    if !file.starts_with('/') {
-        file.insert(0, '/');
-    }
-    let dsn = format!("sqlite://{file}?mode=rwc");
+    let dsn = format!("sqlite://{}?mode=rwc", path.display());
     let opts = ConnectOpts {
         max_conns: Some(1),
         min_conns: Some(1),
```
Artifizer, line 59: **[MAJOR] F-3-MAJOR-codex-default** in `gears/file-storage/file-storage/tests/service_test.rs`

The new DSN construction interpolates `Path::display()` directly into a SQLite URL. On Windows this produces a value such as `sqlite://C:\Users\...\file.db?mode=rwc`, retaining backslashes and omitting the leading slash needed for a valid absolute SQLite URL. The removed code explicitly normalized both conditions. Consequently, these tests can no longer open their temporary databases on Windows. The same regression was copied into all 14 changed file-storage test harnesses in this chunk, so the platform’s file-storage test suite broadly loses Windows portability rather than only one test failing.

**Suggested fix:** Restore conversion to forward slashes and prepend `/` for Windows drive paths before constructing the DSN, or introduce one shared helper that converts a `Path` into a correctly encoded SQLite file URL and use it across all affected tests. The helper should also percent-encode URL-significant path characters.

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    if !mark_running(stores, db, scope, operation_id, now).await? {
+        tracing::warn!(
+            %operation_id,
+            "types_registry operation was already running; continuing with CAS-protected items"
+        );
+    }
+
+    let mut outcomes = Vec::with_capacity(items.len());
+    for item in items {
```
Artifizer, line 106: **[MAJOR] F-4-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/domain/admission/worker.rs`

Items are processed strictly in request order, while `evaluate` builds its transient store from only the current candidate and already-committed database rows. Therefore, a batch containing schema B followed by schema A, where B references A, terminally fails B as unresolvable; reversing the same two items succeeds. The public request permits up to `batch_candidates` items and does not reject references, so admission outcomes incorrectly depend on caller ordering.

**Suggested fix:** Before processing, extract dependencies among all candidates and evaluate them in dependency/SCC order while making the entire candidate set available to the transient store. Until that support exists, synchronously reject batches containing intra-batch references with a clear validation error instead of recording order-dependent outcomes.

gears/system/types-registry/types-registry/src/domain/artifacts.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Digest the three artifacts. Length-prefixed and version-tagged for the same
+/// reasons as the request fingerprint: no two field splits can collide, and a
+/// future change to the inputs cannot read as an unchanged resolution.
+#[must_use]
+pub fn resolution_fingerprint(
+    resolved_schema: &str,
+    effective_traits: &str,
+    effective_traits_schema: &str,
+) -> Vec<u8> {
+    fnv1a_64(&[
```
Artifizer: **[MAJOR] F-5-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/domain/artifacts.rs`

`resolution_fingerprint` uses 64-bit FNV-1a even though its resolved-schema input ultimately comes from client-authored schemas. Contrary to the module’s premise that these inputs are not client-controlled, an authorized schema author can influence the resolved artifacts and deliberately construct FNV collisions. Equality of this fingerprint is intended to decide whether dependency-driven resolution changed and whether propagation stops; a collision can therefore leave stale effective schemas or traits visible to consumers. The same PR correctly explains in `fingerprint.rs` that FNV-1a is invertible and unsuitable when an attacker controls input. That reasoning also applies here because there is no exact artifact comparison described as a fallback for `resolution_fingerprint`.

**Suggested fix:** Use a collision-resistant digest from the approved `aws-lc-rs` provider for `resolution_fingerprint`, with the existing domain tag and length prefixes. Alternatively, treat the hash only as a prefilter and always compare all three canonical artifact strings exactly before declaring the resolution unchanged.

gears/system/types-registry/types-registry/src/gear.rs
```diff
@@ -135,6 +175,41 @@ impl Gear for TypesRegistryGear {
...
+        if let Some(db) = ctx.db() {
+            // Admission runs inline until T21 starts the outbox worker in
+            // `init()`. `NullDispatch` therefore enqueues nothing — the
+            // dispatch is still written inside the acceptance transaction, so
+            // the shape T21 needs is already in place.
+            let dispatch: Arc<dyn OperationDispatch> = Arc::new(NullDispatch);
+            // The domain names its persistence ports and never the repositories;
+            // this is the one place the database-backed adapter is chosen.
+            let stores: Arc<dyn Stores> = Arc::new(Repos);
+            let registry = Arc::new(RegistryService::new(
```
Artifizer, line 193: **[MAJOR] F-7-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/gear.rs`

Inventory and statically configured entities are registered only with the legacy in-memory `TypesRegistryService`. The database-backed `RegistryService` is then constructed from an empty repository without importing the same entities. Consequently, the newly exposed database-backed registry cannot resolve dependencies on the platform's preloaded type schemas. Registrations derived from inventory or configured base types can fail admission even though those types are available through the legacy registry.

**Suggested fix:** Seed or reconcile all inventory and statically configured entities into the database-backed registry before publishing it through the client hub. Alternatively, keep the database-backed routes disabled until a complete migration/cutover mechanism is implemented.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
```diff
@@ -0,0 +1,422 @@
...
+                    .map(|doc| parse_stored(&doc.raw_schema, &row.gts_id))
+                    .transpose()?;
+                match current {
+                    Some(current) => (
+                        content,
+                        Some(parse_stored(&current.resolved_schema, &row.gts_id)?),
+                        Some(parse_stored(&current.effective_traits, &row.gts_id)?),
+                        Some(parse_stored(&current.effective_traits_schema, &row.gts_id)?),
+                    ),
+                    None => (content, None, None, None),
```
Artifizer: **[MAJOR] F-8-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/domain/registry_service.rs`

`entity()` treats a missing current type-schema revision or instance revision as a valid entity and returns an `EntityRecord` with `None` for its content and related state. These rows represent incomplete or corrupt persisted state, not a legitimate form of either entity kind. Returning `Some(EntityRecord)` masks the integrity failure as a successful lookup and can produce an API response with absent schema or instance data. This also makes database corruption difficult for callers and operators to distinguish from an intentionally empty value.

**Suggested fix:** Require the corresponding current-state row and its document or value for every returned entity. If any required row or payload is absent, return a dedicated data-integrity error such as `ServiceError::CorruptDocument` rather than constructing a partially empty `EntityRecord`. Add tests for missing Type Schema and Instance current-state records.

gears/system/types-registry/types-registry/src/api/rest/dto.rs
```diff
@@ -30,7 +34,7 @@ impl From<&GtsIdSegment> for GtsIdSegmentDto {
             package: segment.package().to_owned(),
             namespace: segment.namespace().to_owned(),
             type_name: segment.type_name().to_owned(),
-            ver_major: segment.ver_major(),
+            ver_major: segment.ver_major_opt().unwrap_or(0),
```
Artifizer, line 37: **[MAJOR] F-2-MAJOR-claude-sonnet-5** in `gears/system/types-registry/types-registry/src/api/rest/dto.rs`

`GtsIdSegmentDto::from(&GtsIdSegment)` maps the segment's major version with `segment.ver_major_opt().unwrap_or(0)`, collapsing “no major version present” (`None`) into the same wire value (`0`) as “major version explicitly `0`” (`Some(0)`). This distinction is significant because admission explicitly treats `Some(0)` as a dedicated quarantined state. Public clients consuming `GtsIdSegmentDto` or `GtsEntityDto` therefore cannot distinguish versionless segments from entities that explicitly declared the unstable `v0` major.

**Suggested fix:** Change `GtsIdSegmentDto.ver_major` to `Option<u32>` and map it directly from `segment.ver_major_opt()`, or use another explicitly documented representation that preserves the distinction between no version and major version zero.

```diff
--- a/gears/system/types-registry/types-registry/src/api/rest/dto.rs
+++ b/gears/system/types-registry/types-registry/src/api/rest/dto.rs
@@ -24,7 +24,7 @@ pub struct GtsIdSegmentDto {
     /// Type name component of the segment.
     pub type_name: String,
     /// Major version number.
-    pub ver_major: u32,
+    pub ver_major: Option<u32>,
 }
 
 impl From<&GtsIdSegment> for GtsIdSegmentDto {
@@ -34,7 +34,7 @@ impl From<&GtsIdSegment> for GtsIdSegmentDto {
             namespace: segment.namespace().to_owned(),
             type_name: segment.type_name().to_owned(),
-            ver_major: segment.ver_major_opt().unwrap_or(0),
+            ver_major: segment.ver_major_opt(),
         }
     }
 }
```

gears/system/types-registry/types-registry/src/api/rest/dto.rs
```diff
@@ -428,16 +432,286 @@ mod tests {
...
+#[derive(Debug, Clone)]
+#[toolkit_macros::api_dto(request)]
+pub struct SubmitEntitiesRequest {
+    #[schema(min_items = 1)]
+    pub items: Vec<SubmitEntityDto>,
+    /// Run every check and commit nothing. Part of the request fingerprint, so
+    /// reusing one `Idempotency-Key` for a dry run and then a commit is a
+    /// conflict, not a replay of the dry-run result.
+    #[serde(default)]
+    pub dry_run: Option<bool>,
```
Artifizer, line 487: **[MAJOR] F-3-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/api/rest/dto.rs`

The new request contract advertises revisions and dry runs as functional: a positive `expected_resource_version` is documented as a version to match, and `dry_run: true` promises to run every check without committing. The current admission path instead unconditionally rejects every positive version with `RevisionNotAccepted` and every dry run with `DryRunNotAccepted`. Consequently, generated documentation and clients expose operations that can only return a synchronous error, contrary to the DTO descriptions.

**Suggested fix:** Do not publish these options as supported until their execution paths are implemented. Either complete revision CAS and rollback-only dry-run handling before exposing the fields, or temporarily remove them from this request DTO or explicitly document them as unsupported and always rejected. Update the OpenAPI contract when the corresponding implementation tasks land.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
```diff
@@ -0,0 +1,422 @@
...
+            config,
+            dispatch,
+            admission_mode,
+        }
+    }
+
+    /// See the module header for why this is `allow_all` and why that is honest
+    /// rather than a bypass.
+    fn scope() -> AccessScope {
+        AccessScope::allow_all()
```
Artifizer, line 205: **[MAJOR] F-6-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/domain/registry_service.rs`

Every operation uses `AccessScope::allow_all()` rather than a scope derived from the authenticated caller. Consequently, any caller able to reach this service can submit global registry mutations and read any entity or operation, with no authorization decision or row-level restriction. Keeping the routes internal reduces exposure but does not establish authorization between internal principals. It also bypasses the secure-ORM guideline that sensitive database access be tied to a request security context.

**Suggested fix:** Pass the authenticated `SecurityContext` or platform identity from each transport into `RegistryService`, perform the appropriate permission check before mutation or reads, and derive an `AccessScope` from that context. If authorization cannot be implemented in this release, do not register the mutation routes until the required identity and policy enforcement are available.

gears/system/types-registry/types-registry/src/domain/gts_store.rs
```diff
@@ -0,0 +1,362 @@
...
+pub async fn load_unit_store(
+    stores: &dyn Stores,
+    tx: &DbTx<'_>,
+    scope: &AccessScope,
+    candidates: Vec<UnitDocument>,
+) -> Result<UnitStore, StoreBuildError> {
+    let candidate_ids: Vec<String> = candidates.iter().map(|c| c.gts_id.clone()).collect();
+    let overlay: HashSet<&str> = candidate_ids.iter().map(String::as_str).collect();
+
+    let closure = stores.closure(tx, scope, &candidate_ids).await?;
```
Artifizer, line 303: **[MAJOR] F-7-MAJOR-codex-default** in `gears/system/types-registry/types-registry/src/domain/gts_store.rs`

The transient store’s dependency closure is seeded only with the candidates’ own identifiers. For a newly submitted schema, no stored dependency edges exist yet, so an external `$ref` or `x-gts-ref` appearing in its document cannot cause the referenced, already-registered schema to be loaded. Validation therefore sees the reference target as absent even when it exists in the registry. The same problem affects changed references on future revisions because the closure would follow the old persisted edge set rather than the candidate document’s new references.

**Suggested fix:** Extract direct reference targets from each candidate document before loading the store and include those identifiers as closure roots alongside the candidate identifier chains. Load their transitive stored dependencies in the same snapshot, retain explicit information about genuinely missing targets, and add an integration test admitting a new schema that references an existing schema outside its identifier chain.

MikeFalcon77 (commented):
.

gears/system/types-registry/types-registry/src/api/rest/handlers.rs
```diff
@@ -99,13 +107,197 @@ pub async fn get_entity(
...
+    // Absent and unusable are kept apart. An absent header is the domain's refusal
+    // to make — `validate` takes the empty string and answers "an Idempotency-Key
+    // header is required" — but a header that *was* sent and cannot be decoded is
+    // this layer's own answer, because a `String` cannot carry the distinction any
+    // further and "required" would be a lie about what the caller did.
+    let idempotency_key = match headers.get("idempotency-key") {
+        None => String::new(),
+        Some(value) => value
+            .to_str()
+            .map_err(|_| super::error::idempotency_key_not_utf8())?
```
MikeFalcon77, line 153: **MEDIUM** — `RUST-ERR-001`

map_err(|_| ...) discards the original ToStrError when converting the Idempotency-Key header, losing the invalid bytes that caused the failure.

**Fix:** Use .map_err(|e| super::error::idempotency_key_not_utf8_with_cause(e)) or chain the original error via .context() so the source is preserved for diagnostics.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    /// Accept a submission, and — while admission is inline — admit it.
+    ///
+    /// # Errors
+    /// [`ServiceError::Acceptance`] for every synchronous refusal, including the
+    /// fingerprint conflict; [`ServiceError::Worker`] only for an infrastructure
+    /// failure during inline admission, never for a candidate that was refused on
+    /// its merits.
+    pub async fn submit(
```
MikeFalcon77, line 224: **MEDIUM** — `RUST-ASYNC-001`

The submit async function lacks a cancel-safety annotation. If the future is dropped between accept() and run_operation(), the operation is left in pending state with no outbox to redeliver it.

**Fix:** Add a // cancel-safe or // NOT cancel-safe: <reason> doc comment to submit and every async fn that may run inside select!/timeout, per the Rust async guidelines section 5.

gears/system/types-registry/types-registry/src/gear.rs
```diff
@@ -201,6 +276,35 @@ impl SystemCapability for TypesRegistryGear {
...
+        info!("Providing types-registry database migrations");
+        let mut migrations = crate::infra::storage::Migrator::migrations();
+        let outbox = match toolkit_db::outbox::outbox_migrations_with_prefix(OUTBOX_TABLE_PREFIX) {
+            Ok(outbox) => outbox,
+            // `migrations()` cannot fail in its signature. The prefix is a
+            // compile-time constant that the helper only rejects for an invalid
+            // shape (e.g. a schema-qualified name), so this arm is unreachable
+            // in practice; fail the process rather than booting a gear whose
+            // outbox tables are missing.
+            Err(e) => panic!(
```
MikeFalcon77, line 299: **LOW** — `RUST-PANIC-001`

panic!() in the migrations() method for an invalid outbox prefix crashes the process at boot. While documented as unreachable (the prefix is a compile-time constant), panic! in production code violates the no-panic guideline.

**Fix:** Return Result from migrations() if the trait allows, or use expect() with a descriptive message to make the invariant assertion explicit. If the signature cannot change, add a SAFETY-style comment justifying the panic.

gears/system/types-registry/types-registry/src/gear.rs
```diff
@@ -135,6 +175,41 @@ impl Gear for TypesRegistryGear {
             .set(service.clone())
             .map_err(|_| anyhow::anyhow!("{} gear already initialized", Self::MODULE_NAME))?;
```
MikeFalcon77, line 176: **LOW** — `RUST-ERR-001`

map_err(|_| anyhow::anyhow!(...)) on OnceCell::set discards the original SetError, which carries the value that could not be stored.

**Fix:** Chain the original error: .map_err(|e| anyhow::anyhow!("{} gear already initialized: {e}", Self::MODULE_NAME)) to preserve the source for diagnostics.

gears/system/types-registry/types-registry/src/config.rs
```diff
@@ -32,6 +36,226 @@ pub struct TypesRegistryConfig {
...
+        let split = trimmed
+            .find(|c: char| !c.is_ascii_digit())
+            .unwrap_or(trimmed.len());
+        let (digits, suffix) = trimmed.split_at(split);
+        if digits.is_empty() {
+            return Err(format!("'{trimmed}' does not start with a number"));
+        }
+        let value: usize = digits
+            .parse()
+            .map_err(|_| format!("'{digits}' is not a byte count"))?;
```
MikeFalcon77: **LOW** — `RUST-ERR-001`

map_err(|_| format!(...)) in the ByteSize parser discards the original ParseIntError, losing the specific parse failure reason.

**Fix:** Include the original error: .map_err(|e| format!("'{digits}' is not a byte count: {e}")) to preserve the parse error context.

gears/system/types-registry/types-registry/src/infra/storage/repo/operation_repo.rs
```diff
@@ -0,0 +1,385 @@
...
+/// The multi-row insert budget for `operation_item`, where every row binds 14
+/// columns rather than the one parameter per row that `IN_CHUNK`'s
+/// `SQLITE_MAX_VARIABLE_NUMBER = 999` reasoning budgets for: 70 × 14 = 980.
+const ITEM_INSERT_CHUNK: usize = 70;
+
+/// One stored operation as the domain names it. See `entity_repo::row` for why the
+/// mapper sits beside the repository rather than on the entity.
+fn operation_row(m: operation::Model) -> Result<OperationRow, ScopeError> {
+    let idempotency_scope_hash =
+        ScopeHash::from_stored(m.idempotency_scope_hash).map_err(|_| {
```
MikeFalcon77, line 30: **LOW** — `RUST-ERR-001`

map_err(|_| ScopeError::Invalid(...)) discards the original TryFromSliceError when converting stored bytes to a 32-byte array, losing the slice length that caused the failure.

**Fix:** Include the original error in the message: .map_err(|e| ScopeError::Invalid(format!("stored idempotency_scope_hash is not 32 bytes: {e}"))) to preserve the conversion error.

gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs
```diff
@@ -0,0 +1,602 @@
...
+
+fn normalize_dialect(declared: &str) -> Option<&'static str> {
+    DRAFT_07_SPELLINGS.contains(&declared).then_some(DRAFT_07)
+}
+
+/// The path of the first `$schema` below the root that does not normalize onto
+/// the same dialect. A nested `$schema` equal to the root's — after
+/// normalization, so `…/schema` and `…/schema#` agree — is not a conflict
+/// (ADR-0014).
+fn conflicting_dialect(value: &Value, path: &str, is_root: bool) -> Option<String> {
```
MikeFalcon77: **LOW** — `RUST-TYPE-001`

The is_root: bool parameter in conflicting_dialect is a boolean parameter that should be an enum per the type-safety guidelines. Boolean parameters are unreadable at call sites.

**Fix:** Replace is_root: bool with an enum like enum DialectPosition { Root, Nested } and match on it at call sites for self-documenting code.

gears/system/types-registry/types-registry/src/api/rest/dto.rs
aviator5: The observation is correct in isolation, but UUID-tail segments are structurally impossible for any stored entity: acceptance enforces ADR-0001 and rejects every candidate whose identifier contains an explicit UUID tail (`ExplicitUuidTail` error in acceptance). So `ver_major_opt()` can never return `None` on a segment that reached the response DTO — the `Option<u32>` only added unreachable `null` to the API contract.

docs/api/api.json
aviator5: Response header declarations in OperationBuilder are not yet supported — tracked in #4681. The runtime headers (Location, Retry-After, Idempotency-Replayed) are emitted correctly; the OpenAPI contract will reflect them once that issue lands.

gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs
aviator5: Fixed in bdc5715. `conflicting_dialect` now treats any nested object carrying `$schema` as a declaration and refuses it unless the value is a non-empty supported string — the same test the root check applies. `a_nested_dialect_must_be_a_supported_string` covers null, number and object.

gears/system/types-registry/types-registry/src/domain/artifacts.rs
aviator5: Agreed — the "not client-controlled" premise does not hold for resolved artifacts, and unlike `content_hash` there is no exact-comparison fallback. Fixed in bdc5715: SHA-256 via `aws-lc-rs`, same domain tag and length prefixes. The columns already fit (`bytea` / `BLOB` / `VARBINARY(64)`), and the field is write-only until propagation lands, so no stored digest is invalidated.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
aviator5: Fixed in bdc5715: a missing current-state row or document is now `ServiceError::CorruptDocument`, mapped to an opaque 500 with the detail going to the operator log. Strictness is safe here — `commit_creation` writes the entity and its current state in one transaction, and `mark_deleted` only moves `lifecycle_status`, so a tombstone keeps its row. Test covers both kinds.

gears/file-storage/file-storage/tests/service_test.rs
aviator5: Fixed in 7305559, in the library rather than in the 14 harnesses. `sqlx` does not parse a SQLite DSN as a URL: `SqliteConnectOptions::from_str` trims the scheme, splits at the first `?` and keeps the path byte-for-byte, so `sqlite://C:\Users\...\f.db?mode=rwc` opens fine. The real breakage was toolkit-db round-tripping the DSN through `url::Url`, which turned `C:` into a host. `split_sqlite_dsn` now follows sqlx's grammar and is the single source of truth for the other helpers. `Test Suite (windows-latest)` is green.

gears/system/types-registry/types-registry/src/api/rest/dto.rs
aviator5: Taking the documentation option. bdc5715 marks both fields reserved and synchronously refused — `expected_resource_version` until T11, `dry_run` until T20 — and `docs/api/api.json` is regenerated. They stay in the request shape rather than being removed because `dry_run` is already an input to the request fingerprint; dropping and reinstating it would move idempotency identity twice.

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
aviator5: Correct, and out of P0-1 scope: T19 (Dependency-aware partial admission) owns SCC ordering against the candidate overlay. The interim refusal needs the same reference extractor as the real fix (T13), so it buys nothing. No exposure meanwhile — the v2 routes are internal-only (no `.exposed()`) and the database path has no consumer until T24.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
aviator5: Correct, and tracked as ceiling C6 in the module header: there is no PDP, the managed entities are `#[secure(unrestricted)]`, and no `SecurityContext` reaches this layer yet (`allow_all` is a decision with no row filtering; `AccessScope::default()` is deny-all). Taking your second option — the mutation routes stay internal-only (no `.exposed()`, ceiling C8 in `routes.rs`) until a platform listener can authenticate a platform principal. P1 derives a request-scoped `AccessScope` from that context.

gears/system/types-registry/types-registry/src/gear.rs
aviator5: Accurate. Taking your second option: the database-backed routes stay internal-only and have no consumer until T24, which is where seeding lands. bdc5715 documents both seed sources — types-registry's own inventory and the operator-configured `cfg.entities`, whose identifiers are deployment-specific and cannot be inventory items — in SPEC §8.1 and in T24's acceptance criteria, including that an invalid or oversized combined seed set fails boot.

gears/system/types-registry/types-registry/src/domain/gts_store.rs
aviator5: Correct, including the revision half — the overlay covers documents but not edges. Worth adding that the edge writes T13 already planned do not fix this on their own: rows are replaced at commit, so they do not exist during the read that validates the candidate that authored them. bdc5715 turns this into explicit T13 criteria — run the extractor over the candidate document before the closure read and seed its targets as roots beside `chain_ids()`, take a revision's roots from the document rather than the stored edge set, and report genuinely absent targets apart from `missing_roots`, which holds the candidate itself on every first admission. The integration test you describe is listed too.

gears/system/types-registry/types-registry/src/api/rest/handlers.rs
aviator5: Not changing. `http::header::ToStrError` is a unit struct: it carries no bytes, and its `Display` is the fixed string "failed to convert header to a str" — there is no cause to preserve. The invalid bytes are caller-controlled and this error is client-facing, so they should not be echoed anyway.

gears/system/types-registry/types-registry/src/domain/registry_service.rs
aviator5: Fixed in ac8a918c — annotated `NOT cancel-safe`, naming the commit-then-admit window and the replay-under-the-same-key recovery that covers it until T21's outbox.

gears/system/types-registry/types-registry/src/gear.rs
aviator5: Already satisfied: the justifying comment is at gear.rs:292-297. `expect()` is not an option — `expect_used` is `deny` in the workspace lints.

gears/system/types-registry/types-registry/src/gear.rs
aviator5: Not changing. The field is `std::sync::OnceLock<Arc<TypesRegistryService>>` (gear.rs:55), so `set` returns `Result<(), Arc<TypesRegistryService>>` — there is no `SetError`, and `Arc<TypesRegistryService>` has no `Display`, so `{e}` would not compile. The only possible cause is re-initialization, which the message already states.

gears/system/types-registry/types-registry/src/config.rs
aviator5: Fixed in ac8a918c.

gears/system/types-registry/types-registry/src/infra/storage/repo/operation_repo.rs
aviator5: Not changing. `ScopeHash::from_stored` returns `Result<Self, Vec<u8>>` (fingerprint.rs:71) and `Vec<u8>` has no `Display`; `ScopeError::Invalid` takes `&'static str` (libs/toolkit-db/src/secure/error.rs:12), so `format!` would require changing a shared toolkit type. Reachable only on a corrupted DB, where the length is the only useful datum and belongs in a log line.

gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs
aviator5: Fixed in ac8a918c, but without an enum — the flag is gone entirely. The root exemption is now the recursion's entry point (`conflicting_dialect(root)` → `conflicting_dialect_below` → `conflicting_dialect_at`) instead of a parameter carried through every frame; an enum would have added `::Nested` to three recursive call sites. Guarded by a new test covering array descent and depth ≥ 2.

Issue #4554, opened 2026-08-14
[Directory] Authorize instance registration against authenticated peer identity


mattgarmon:
## Context

Follow-up from PR #4540 review (label-based instance targeting). Raised on `gears/system/gear-orchestrator/src/server.rs` (`DirectoryServiceImpl::register_instance`).

## Problem

`register_instance` / `deregister_instance` / `heartbeat` call `authenticate_peer` but **discard** the resolved `PlatformIdentity`, then take `gear_name` and `instance_id` verbatim from the request body. Any peer holding a valid internal token can therefore:

- **Tamper** with another gear's registration (overwrite its `labels` / `rest_endpoint_uri`), redirecting that gear's tenant traffic; and
- **Inject a rogue instance** under another gear's name (register `gear_name="billing"` with a fresh `instance_id`), inserting an attacker endpoint into that gear's round-robin set.

The RPC validates the *shape* of registrant data (labels/URI checks added in #4540) but never the peer's *authority* over the identity it claims.

## Why it isn't a small fix

Robust authorization needs a per-gear-distinct peer identity, which isn't reliably available yet. `PlatformIdentity::peer_name()` (`libs/toolkit-security/src/internal_auth.rs`) returns:

- the **K8s ServiceAccount name** (Profile 3),
- the **SPIFFE workload/gear name** (mTLS end-state), or
- for **shared-secret** (Profile 2 / dev), a **single configured label shared by every peer** (default `toolkit-internal`).

Under shared-secret all callers present one identity, so per-gear binding is structurally impossible there. A naive `peer.name == gear_name` check today would break shared-secret and any Kube deployment whose SA name differs from the gear name. This work should land **with** the SA-per-gear rollout / SPIFFE, where per-gear identity becomes authoritative.

## Proposed fix

Stop discarding the identity; authorize register/deregister/heartbeat with one predicate correct across all providers:

```rust
fn registration_authorized(
    identity: &PlatformIdentity,
    gear_name: &str,
    trusted_registrars: &HashSet<String>,
) -> bool {
    match identity {
        // Shared secret / bootstrap token = one deliberately-shared trust
        // boundary; per-gear binding is not expressible (documented model).
        PlatformIdentity::Shared { .. } => true,
        // Per-gear identity (SA-per-gear now, SPIFFE next): may act only on its
        // own gear. `Unknown` -> peer_name "<unknown>" -> fails closed.
        _ => {
            let peer = identity.peer_name();
[...]

PR #4642, opened 2026-08-20
feat(grpc-hub): enforce inbound platform-plane auth at the gRPC boundary

### LINKED ISSUE #4576: Enforce platform auth on the server side over gRPC (inbound)
### Summary
Validate the platform credential on incoming gRPC calls and expose the authenticated identity to handlers via `PlatformSecurityContext`. When a client presents a platform credential over gRPC, the server side currently has no way to validate it or reject unauthenticated callers. This is the `InternalAuthMiddleware` specified by [ADR-0006](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md), applied to the gRPC path.

### Problem
The gRPC serving path has no authentication layer, and the platform-plane authenticator is currently wired only into the HTTP path. A platform-plane gRPC method therefore cannot know who called it or refuse an invalid caller. This gap is gRPC-specific — the REST server path already validates the internal token and surfaces the identity to handlers.

### Design (per [ADR-0006](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md) / [ADR-0008](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md))
Add an **async server-side `InternalAuthMiddleware` at the gRPC boundary**: extract the internal token, validate it via the `InternalAuthenticator` trait, and insert `PlatformSecurityContext { identity }` plus `PeerAuthenticated { name }` for the handler. It must be an async server middleware (not a synchronous interceptor), because validation can require an out-of-process call (K8s TokenReview). `PeerAuthenticated` / `PlatformIdentity` are for workload policy only and are never passed to the tenant `PolicyEnforcer`.

- **Bootstrap-selected, not discovered.** The `InternalAuthenticator` implementation MUST be selected at bootstrap by deployment profile/config and wired in-process — never GTS-discovered. Platform auth protects the very calls (Directory registration, heartbeats, global GTS registration) a gear makes before its ClientHub/discovery chain exists, so discovering the validator over the network would be a bootstrapping cycle.
- **TokenReview caching is required.** ADR-0006 specifies TokenReview validation with a cached TTL; the external validation path is currently uncached and too slow for a data plane at scale. A short-lived cache is part of this work (possibly a precursor).

### Open questions
- **Exempt methods.** Which infrastructure endpoints (health checks, reflection, etc.) bypass enforcement, and how they're marked (allowlist vs. marker).
- **Permissive vs. required.** Whether an absent credential is rejected or allowed, configurable per deployment profile.

### Out of scope
REST inbound (already works)

### References
- [ADR-0006 — Platform-Plane Authentication](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0006-cpt-cf-adr-platform-plane-auth.md)
- [ADR-0008 — Two-Plane Authentication Model](https://github.com/constructorfabric/gears-rust/blob/main/docs/arch/toolkit-oop/ADR/0008-cpt-cf-adr-two-plane-auth.md)

mattgarmon:
Closes #4576

Move `x-toolkit-internal-token` (platform-plane) enforcement out of per-service handler checks and into a single transport-level Tower layer applied by `grpc-hub` across every inbound RPC (`cpt-cf-adr-platform-plane-auth`). Previously, enforcement lived in `gear-orchestrator`'s `DirectoryService` and had to be re-implemented per handler; now it is enforced once, at the gRPC boundary, for all services mounted on the hub.
 
- `toolkit-transport-grpc`: new `InternalAuthGrpcLayer` - a Tower layer that validates the internal token on every inbound RPC (with configurable Required/Permissive enforcement and an exempt-method allowlist for health checks and reflection), and populates request extensions (`PlatformSecurityContext`, `PeerAuthenticated`) for downstream handlers.
- `toolkit-security`: new `CachingInternalAuthenticator` - a generic, TTL-bounded cache for any `InternalAuthenticator`, so a remote validation backend (e.g. Kubernetes `TokenReview`) isn't hit on every call.
- `toolkit-k8s-auth`: wires the kube `TokenReview` authenticator through the shared cache.
- `grpc-hub`: applies the layer to all three serve paths (TCP, Unix socket, named pipe); adds `internal_auth`, `internal_auth_enforcement`, `internal_auth_exempt_methods`, and `internal_auth_cache_ttl_secs` config; now owns the `k8s-auth` feature.
- `gear-orchestrator`: drops the old per-handler auth config and plumbing now that enforcement lives at the transport layer; fails `init` with a migration message if a deployment still has the old config set, so an upgrade can't silently drop enforcement.

## Summary by CodeRabbit

* **New Features**
  * Added configurable platform authentication for gRPC Hub connections using shared secrets or Kubernetes tokens.
  * Added required or permissive enforcement, exempt method prefixes, and successful-authentication caching across supported gRPC transports.

* **Bug Fixes**
  * Invalid authentication settings and unsupported providers now produce clear startup errors.

* **Documentation**
  * Added configuration guidance for authentication providers, enforcement, exemptions, and caching.

* **Migration**
  * Legacy orchestrator authentication settings are rejected with guidance to configure authentication through gRPC Hub.

### DIFF
--- a/gears/system/gear-orchestrator/src/gear.rs
+++ b/gears/system/gear-orchestrator/src/gear.rs
@@ -3,7 +3,6 @@
 use anyhow::Result;
 use async_trait::async_trait;
 use std::sync::{Arc, OnceLock};
-use tokio::sync::RwLock;
 
 use toolkit::DirectoryClient;
 use toolkit::context::GearCtx;
@@ -16,21 +15,9 @@ use toolkit::registry::GearRegistry;
 use toolkit::runtime::GearManager;
 
 use cf_system_sdks::directory::DIRECTORY_SERVICE_NAME;
-use toolkit_security::{DynInternalAuthenticator, InternalAuthConfig};
 
 use crate::domain::service::GearsService;
 use crate::server;
-
-/// Configuration for the gear orchestrator
-#[derive(Clone, Debug, Default, serde::Deserialize)]
-pub struct GearOrchestratorConfig {
-    /// Platform-plane (internal) authentication for the `DirectoryService`
-    /// gRPC RPCs. When set, every RPC requires a valid
-    /// `x-toolkit-internal-token` (`cpt-cf-adr-two-plane-auth`); when absent,
-    /// enforcement is disabled (Profile 1 / in-process only).
-    #[serde(default)]
-    pub internal_auth: Option<toolkit_security::InternalAuthConfig>,
-}
 
 /// Gear Orchestrator - system gear for service discovery
 ///
@@ -45,69 +32,19 @@ pub struct GearOrchestratorConfig {
     client = cf_system_sdks::directory::DirectoryClient
 )]
 pub struct GearOrchestrator {
-    config: RwLock<GearOrchestratorConfig>,
     directory_api: OnceLock<Arc<dyn DirectoryClient>>,
     gear_manager: OnceLock<Arc<GearManager>>,
     gears_service: OnceLock<Arc<GearsService>>,
-    /// Platform-plane validator built from `config.internal_auth`, applied to
-    /// the `DirectoryService` gRPC RPCs. `None` disables enforcement.
-    authenticator: OnceLock<Option<DynInternalAuthenticator>>,
 }
 
 impl Default for GearOrchestrator {
     fn default() -> Self {
         Self {
-            config: RwLock::new(GearOrchestratorConfig::default()),
             directory_api: OnceLock::new(),
             gear_manager: OnceLock::new(),
             gears_service: OnceLock::new(),
-            authenticator: OnceLock::new(),
         }
     }
-}
-
-/// Build the platform-plane authenticator from configuration.
-///
-/// The dependency-light shared-secret provider is built here directly. The
-/// Kubernetes `TokenReview` provider requires the `k8s-auth` feature; without
-/// it, configuring `provider: kube` is a hard error rather than a silent
-/// enforcement downgrade.
-#[cfg_attr(not(feature = "k8s-auth"), allow(clippy::unused_async))]
-async fn build_internal_authenticator(
-    cfg: Option<&InternalAuthConfig>,
-) -> Result<Option<DynInternalAuthenticator>> {
-    let Some(cfg) = cfg else {
-        return Ok(None);
-    };
-
-    // Shared-secret (and any future dependency-light provider) builds here.
-    if let Some(auth) = cfg.build_authenticator() {
-        return Ok(Some(auth));
-    }
-
-    #[cfg(feature = "k8s-auth")]
-    {
-        if cfg.is_kube() {
-            let audiences = cfg.kube_audiences().unwrap_or_default().to_vec();
-            let authenticator =
-                toolkit_k8s_auth::K8sTokenReviewAuthenticator::try_default(audiences)
-                    .await
-                    .map_err(|e| {
-                        anyhow::anyhow!("failed to init Kubernetes TokenReview authenticator: {e}")
-                    })?;
-            return Ok(Some(DynInternalAuthenticator::new(authenticator)));
-        }
-    }
-    #[cfg(not(feature = "k8s-auth"))]
-    {
-        if cfg.is_kube() {
-            anyhow::bail!(
-                "gear-orchestrator internal_auth provider=kube requires the `k8s-auth` feature"
-            );
-        }
-    }
-
-    Ok(None)
 }
 
 #[async_trait]
@@ -123,19 +60,27 @@ impl SystemCapability for GearOrchestrator {
 #[async_trait]
 impl toolkit::Gear for GearOrchestrator {
     async fn init(&self, ctx: &GearCtx) -> Result<()> {
-        // Load configuration if present
-        let cfg = ctx.config_or_default::<GearOrchestratorConfig>()?;
-
-        // Build the platform-plane validator (if any) before moving `cfg`.
-        let authenticator = build_internal_authenticator(cfg.internal_auth.as_ref()).await?;
-        if authenticator.is_some() {
-            tracing::info!("DirectoryService platform-plane enforcement enabled");
+        // Migration guard: platform-plane enforcement lives on `grpc-hub` (a
+        // transport-level `InternalAuthGrpcLayer` applied to every inbound
+        // RPC, `cpt-cf-adr-platform-plane-auth`), not here. This gear reads no
+        // config at all, so a leftover `internal_auth` key would silently do
+        // nothing — fail loudly instead of booting the `DirectoryService`
+        // unauthenticated.
+        if let Some(internal_auth) = ctx
+            .config_provider()
+            .get_gear_config(ctx.gear_name())
+            .and_then(|raw| raw.get("config"))
+            .and_then(|config| config.get("internal_auth"))
+        {
+            let provider = internal_auth.get("provider").and_then(|p| p.as_str());
+            let detail = provider.map_or_else(String::new, |p| format!(" (provider = {p:?})"));
+            anyhow::bail!(
+                "gear-orchestrator config still sets `internal_auth`{detail}, which no longer \
+                 has any effect: platform-plane enforcement moved to grpc-hub's own \
+                 `internal_auth` config (cpt-cf-adr-platform-plane-auth). Move this key under \
+                 `gears.grpc-hub.config` instead."
+            );
         }
-        self.authenticator
-            .set(authenticator)
-            .map_err(|_| anyhow::anyhow!("authenticator already set (init called twice?)"))?;
-
-        *self.config.write().await = cfg;
 
         // Use the injected GearManager to create the DirectoryClient
         let manager = self
@@ -199,9 +144,7 @@ impl GrpcServiceCapability for GearOrchestrator {
             .cloned()
             .ok_or_else(|| anyhow::anyhow!("DirectoryClient not initialized"))?;
 
-        // Build DirectoryService with the configured platform-plane validator.
-        let authenticator = self.authenticator.get().cloned().flatten();
-        let directory_svc = server::make_directory_service(api, authenticator);
+        let directory_svc = server::make_directory_service(api);
 
         Ok(vec![RegisterGrpcServiceFn {
             service_name: DIRECTORY_SERVICE_NAME,
@@ -209,5 +152,73 @@ impl GrpcServiceCapability for GearOrchestrator {
                 routes.add_service(directory_svc.clone());
             }),
         }])
+    }
+}
+
+#[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
+mod tests {
+    use super::*;
+    use tokio_util::sync::CancellationToken;
+    use toolkit::client_hub::ClientHub;
+    use toolkit::config::ConfigProvider;
+    use uuid::Uuid;
+
+    struct StaticConfigProvider(serde_json::Value);
+    impl ConfigProvider for StaticConfigProvider {
+        fn get_gear_config(&self, gear_name: &str) -> Option<&serde_json::Value> {
+            (gear_name == "gear-orchestrator").then_some(&self.0)
+        }
+    }
+
+    async fn init_with_config(config: serde_json::Value) -> Result<()> {
+        let mut wrapped = serde_json::Map::new();
+        wrapped.insert("config".to_owned(), config);
+        let ctx = GearCtx::new(
+            "gear-orchestrator",
+            Uuid::new_v4(),
+            Arc::new(StaticConfigProvider(serde_json::Value::Object(wrapped))),
+            Arc::new(ClientHub::default()),
+            CancellationToken::new(),
+        );
+        let gear = GearOrchestrator::default();
+        let gear_manager = Arc::new(GearManager::new());
+        gear.gear_manager
+            .set(gear_manager)
+            .map_err(|_| anyhow::anyhow!("gear_manager already set"))?;
+        toolkit::Gear::init(&gear, &ctx).await
+    }
+
+    #[tokio::test]
+    async fn init_rejects_leftover_internal_auth_config() {
+        let err = init_with_config(serde_json::json!({
+            "internal_auth": { "provider": "shared_secret", "secret": "top-secret-value" }
+        }))
+        .await
+        .unwrap_err();
+        let msg = err.to_string();
+        assert!(
+            msg.contains("internal_auth"),
+            "expected a migration error mentioning internal_auth, got {err}"
+        );
+        assert!(
+            msg.contains("grpc-hub"),
+            "expected the error to point at grpc-hub's config, got {err}"
+        );
+        assert!(
+            msg.contains("shared_secret"),
+            "expected the non-sensitive provider tag to be named, got {err}"
+        );
+        assert!(
+            !msg.contains("top-secret-value"),
+            "the error must never leak the secret value, got {err}"
+        );
+    }
+
+    #[tokio::test]
+    async fn init_succeeds_without_internal_auth_config() {
+        init_with_config(serde_json::json!({}))
+            .await
+            .expect("init without a leftover internal_auth key must succeed");
     }
 }
--- a/gears/system/gear-orchestrator/src/lib.rs
+++ b/gears/system/gear-orchestrator/src/lib.rs
@@ -6,7 +6,7 @@
 
 // === MODULE DEFINITION ===
 pub mod gear;
-pub use gear::{GearOrchestrator, GearOrchestratorConfig};
+pub use gear::GearOrchestrator;
 
 // === INTERNAL MODULES (pub for integration tests) ===
 pub mod api;
--- a/gears/system/gear-orchestrator/src/server.rs
+++ b/gears/system/gear-orchestrator/src/server.rs
@@ -4,13 +4,7 @@
 
 use std::sync::Arc;
 
-use secrecy::ExposeSecret;
-use tonic::metadata::MetadataMap;
 use tonic::{Request, Response, Status};
-use toolkit_security::{
-    DynInternalAuthenticator, InternalAuthNError, InternalAuthenticator, PeerAuthenticated,
-};
-use toolkit_transport_grpc::extract_internal_token_grpc;
 
 use cf_system_sdks::directory::labels::is_valid_label_segment;
 use cf_system_sdks::directory::{
@@ -42,61 +36,25 @@ fn lookup_status(err: &anyhow::Error) -> Status {
     }
 }
 
-/// gRPC service implementation of Directory Service
+/// gRPC service implementation of Directory Service.
+///
+/// Platform-plane (`x-toolkit-internal-token`) enforcement is applied at the
+/// gRPC server boundary by `grpc-hub`'s `InternalAuthGrpcLayer`
+/// (`cpt-cf-adr-platform-plane-auth`): the token is validated and, on success, a
+/// `PlatformSecurityContext` / `PeerAuthenticated` is placed in the request
+/// extensions before the handler runs.
+///
+/// This is the authentication half only. Authorizing a peer against the gear
+/// name it claims when registering, deregistering, or heartbeating is deferred.
 #[derive(Clone)]
 pub struct DirectoryServiceImpl {
     api: Arc<dyn DirectoryClient>,
-    /// Platform-plane validator. When `Some`, every RPC requires a valid
-    /// `x-toolkit-internal-token` (`cpt-cf-adr-two-plane-auth`); when `None`,
-    /// enforcement is disabled (Profile 1 / in-process only).
-    authenticator: Option<DynInternalAuthenticator>,
 }
 
 impl DirectoryServiceImpl {
-    /// Create a `DirectoryService`. When `authenticator` is `Some`, every RPC
-    /// validates the platform-plane internal token; when `None`, enforcement
-    /// is disabled (Profile 1 / in-process only).
-    pub fn with_authenticator(
-        api: Arc<dyn DirectoryClient>,
-        authenticator: Option<DynInternalAuthenticator>,
-    ) -> Self {
-        Self { api, authenticator }
-    }
-
-    /// Enforce the platform plane on an inbound RPC.
-    ///
-    /// When an authenticator is configured, the `x-toolkit-internal-token`
-    /// metadata must be present and valid; the resolved [`PeerAuthenticated`]
-    /// is returned for workload-policy checks (and traced). When no
-    /// authenticator is configured, enforcement is skipped.
-    async fn authenticate_peer(
-        &self,
-        meta: &MetadataMap,
-    ) -> Result<Option<PeerAuthenticated>, Status> {
-        let Some(authenticator) = &self.authenticator else {
-            return Ok(None);
-        };
-
-        let token = extract_internal_token_grpc(meta)?;
-        match authenticator.authenticate(token.expose_secret()).await {
-            Ok(identity) => {
-                let peer = PeerAuthenticated {
-                    name: identity.peer_name().to_owned(),
-                };
-                tracing::debug!(peer = %peer.name, "platform-plane call authenticated");
-                Ok(Some(peer))
-            }
-            Err(InternalAuthNError::InvalidToken) => {
-                Err(Status::unauthenticated("invalid internal token"))
-            }
-            Err(InternalAuthNError::Unavailable) => {
-                Err(Status::unavailable("internal-auth backend unavailable"))
-            }
-            Err(err) => {
-                tracing::warn!(error = %err, "platform-plane authentication failed");
-                Err(Status::internal("internal authentication failure"))
-            }
-        }
+    /// Create a `DirectoryService` backed by `api`.
+    pub fn new(api: Arc<dyn DirectoryClient>) -> Self {
+        Self { api }
     }
 }
 
@@ -106,7 +64,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<ResolveGrpcServiceRequest>,
     ) -> Result<Response<ResolveGrpcServiceResponse>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let service_name = request.into_inner().service_name;
         validate_lookup_name("service_name", &service_name)?;
 
@@ -125,7 +82,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<ResolveRestServiceRequest>,
     ) -> Result<Response<ResolveRestServiceResponse>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let gear_name = request.into_inner().gear_name;
         validate_lookup_name("gear_name", &gear_name)?;
 
@@ -144,7 +100,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<GetOpenApiSpecRequest>,
     ) -> Result<Response<GetOpenApiSpecResponse>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let gear_name = request.into_inner().gear_name;
         validate_lookup_name("gear_name", &gear_name)?;
 
@@ -161,7 +116,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<ListInstancesRequest>,
     ) -> Result<Response<ListInstancesResponse>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let req = request.into_inner();
         let gear_name = req.gear_name;
         validate_lookup_name("gear_name", &gear_name)?;
@@ -205,7 +159,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         _request: Request<ListAllInstancesRequest>,
     ) -> Result<Response<ListAllInstancesResponse>, Status> {
-        self.authenticate_peer(_request.metadata()).await?;
         let instances = self
             .api
             .list_all_instances()
@@ -234,7 +187,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<RegisterInstanceRequest>,
     ) -> Result<Response<()>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let req = request.into_inner();
 
         validate_identity(&req.gear_name, &req.instance_id)?;
@@ -290,7 +242,6 @@ impl DirectoryService for DirectoryServiceImpl {
         &self,
         request: Request<DeregisterInstanceRequest>,
     ) -> Result<Response<()>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let req = request.into_inner();
 
         validate_identity(&req.gear_name, &req.instance_id)?;
@@ -303,7 +254,6 @@ impl DirectoryService for DirectoryServiceImpl {
     }
 
     async fn heartbeat(&self, request: Request<HeartbeatRequest>) -> Result<Response<()>, Status> {
-        self.authenticate_peer(request.metadata()).await?;
         let req = request.into_inner();
 
         validate_identity(&req.gear_name, &req.instance_id)?;
@@ -506,16 +456,14 @@ fn validate_identity(gear_name: &str, instance_id: &str) -> Result<(), Status> {
     Ok(())
 }
 
-/// Create a `DirectoryService` server with the given API implementation and
-/// optional platform-plane `authenticator`.
+/// Create a `DirectoryService` server backed by `api`.
 ///
-/// When `authenticator` is `Some`, every RPC requires a valid
-/// `x-toolkit-internal-token`; when `None`, enforcement is disabled.
+/// Platform-plane enforcement is applied at the gRPC server boundary by
+/// `grpc-hub`'s `InternalAuthGrpcLayer`.
 pub fn make_directory_service(
     api: Arc<dyn DirectoryClient>,
-    authenticator: Option<DynInternalAuthenticator>,
 ) -> DirectoryServiceServer<DirectoryServiceImpl> {
-    DirectoryServiceServer::new(DirectoryServiceImpl::with_authenticator(api, authenticator))
+    DirectoryServiceServer::new(DirectoryServiceImpl::new(api))
 }
 
 #[cfg(test)]
@@ -534,7 +482,7 @@ mod tests {
     fn service() -> DirectoryServiceImpl {
         let manager = Arc::new(GearManager::new());
         let api: Arc<dyn DirectoryClient> = Arc::new(LocalDirectoryClient::new(manager));
-        DirectoryServiceImpl::with_authenticator(api, None)
+        DirectoryServiceImpl::new(api)
     }
 
     #[tokio::test]
@@ -851,7 +799,7 @@ mod tests {
         // Directory service backed by an in-memory GearManager.
         let manager = Arc::new(GearManager::new());
         let api: Arc<dyn DirectoryClient> = Arc::new(LocalDirectoryClient::new(manager));
-        let grpc_service = make_directory_service(api, None);
+        let grpc_service = make_directory_service(api);
 
         // Reserve a free port, then let the tonic server bind it.
         let addr = std::net::TcpListener::bind("127.0.0.1:0")
@@ -935,29 +883,48 @@ mod tests {
         assert!(gears.contains(&"catalog"));
     }
 
-    /// `cpt-cf-adr-platform-plane-auth` acceptance: with a platform-plane
-    /// authenticator installed, the
+    /// `cpt-cf-adr-platform-plane-auth` acceptance: with the platform-plane
+    /// [`InternalAuthGrpcLayer`] on the server (as `grpc-hub` wires it), the
     /// `DirectoryService` gRPC RPCs reject callers lacking a valid internal
-    /// token and accept those that attach the matching shared secret via an
-    /// [`InternalAuthInterceptor`] — the full outbound→inbound loop.
+    /// token and accept those attaching the matching shared secret via an
+    /// [`InternalAuthInterceptor`] — the full outbound→inbound loop. The
+    /// `probe_layer` (a [`tower::util::MapRequestLayer`] mounted just below
+    /// `InternalAuthGrpcLayer`) additionally proves the layer populates the
+    /// `PeerAuthenticated` extension on the request it forwards downstream.
     #[tokio::test]
     async fn grpc_enforces_internal_token_end_to_end() {
         use cf_system_sdks::directory::DirectoryGrpcClient;
         use secrecy::SecretString;
         use tonic::transport::Server;
         use toolkit_security::{DynInternalAuthenticator, SharedSecretInternalAuthenticator};
-        use toolkit_transport_grpc::InternalAuthInterceptor;
+        use toolkit_transport_grpc::{InternalAuthGrpcLayer, InternalAuthInterceptor};
 
         const SECRET: &str = "dev-internal-token";
 
-        // DirectoryService that enforces the platform plane via a shared secret.
         let manager = Arc::new(GearManager::new());
         let api: Arc<dyn DirectoryClient> = Arc::new(LocalDirectoryClient::new(manager));
+        let grpc_service = make_directory_service(api);
+
+        // Required mode: an absent token is rejected.
         let authenticator = DynInternalAuthenticator::new(SharedSecretInternalAuthenticator::new(
             SecretString::from(SECRET),
             "peer".to_owned(),
         ));
-        let grpc_service = make_directory_service(api, Some(authenticator));
+        let auth_layer = InternalAuthGrpcLayer::new(authenticator);
+        let saw_expected_peer = Arc::new(std::sync::atomic::AtomicBool::new(false));
+        let probe_layer = tower::util::MapRequestLayer::new({
+            let saw_expected_peer = Arc::clone(&saw_expected_peer);
+            move |req: http::Request<_>| {
+                if req
+                    .extensions()
+                    .get::<toolkit_security::PeerAuthenticated>()
+                    .is_some_and(|peer| peer.name == "peer")
+                {
+                    saw_expected_peer.store(true, std::sync::atomic::Ordering::SeqCst);
+                }
+                req
+            }
+        });
 
         let addr = std::net::TcpListener::bind("127.0.0.1:0")
             .unwrap()
@@ -967,6 +934,8 @@ mod tests {
         // ephemeral port still bound.
         let server = tokio::spawn(async move {
             Server::builder()
+                .layer(auth_layer)
+                .layer(probe_layer)
                 .add_service(grpc_service)
                 .serve(addr)
                 .await
@@ -1015,6 +984,12 @@ mod tests {
             .expect("authenticated list");
         assert_eq!(all.len(), 1);
         assert_eq!(all[0].gear, "billing");
+
+        assert!(
+            saw_expected_peer.load(std::sync::atomic::Ordering::SeqCst),
+            "InternalAuthGrpcLayer must populate the PeerAuthenticated extension on the request \
+             it forwards to the layers below it"
+        );
 
         server.abort();
     }
@@ -1347,7 +1322,7 @@ mod tests {
 
         let manager = Arc::new(GearManager::new());
         let api: Arc<dyn DirectoryClient> = Arc::new(LocalDirectoryClient::new(manager));
-        let grpc_service = make_directory_service(api, None);
+        let grpc_service = make_directory_service(api);
 
         let addr = std::net::TcpListener::bind("127.0.0.1:0")
             .unwrap()
--- a/gears/system/grpc-hub/src/gear.rs
+++ b/gears/system/grpc-hub/src/gear.rs
@@ -27,11 +27,28 @@ use tokio_stream::wrappers::TcpListenerStream;
 use tokio_util::sync::CancellationToken;
 use tonic::{service::RoutesBuilder, transport::Server};
 
+use toolkit_security::{DynInternalAuthenticator, InternalAuthConfig};
+use toolkit_transport_grpc::{InternalAuthEnforcement, InternalAuthGrpcLayer};
+
 #[cfg(windows)]
 use toolkit_transport_grpc::create_named_pipe_incoming;
 
 const DEFAULT_LISTEN_ADDR: SocketAddr =
     SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 50051));
+
+/// Default seconds a positive platform-plane validation result is cached.
+/// Derived from `toolkit_security::DEFAULT_TOKEN_REVIEW_CACHE_TTL` (not just
+/// mirrored as a literal) so the config surface cannot drift from the shared
+/// cache implementation while still not depending on the optional `k8s-auth`
+/// feature.
+const DEFAULT_INTERNAL_AUTH_CACHE_TTL_SECS: u64 =
+    toolkit_security::DEFAULT_TOKEN_REVIEW_CACHE_TTL.as_secs();
+/// Upper bound on the internal-auth cache TTL. Caps how long a revoked or
+/// expired token can keep validating from cache, so a misconfiguration cannot
+/// widen the revocation window unboundedly. Derived from
+/// `toolkit_security::MAX_TOKEN_REVIEW_CACHE_TTL` for the same reason.
+const MAX_INTERNAL_AUTH_CACHE_TTL_SECS: u64 =
+    toolkit_security::MAX_TOKEN_REVIEW_CACHE_TTL.as_secs();
 
 /// Configuration for the gRPC Hub gear.
 ///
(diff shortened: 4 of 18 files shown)


gears/system/grpc-hub/src/gear.rs
```diff
@@ -94,6 +142,79 @@ fn parse_advertise_addr(advertise_addr: &str) -> anyhow::Result<(String, Option<
...
+        }
+    }
+    #[cfg(not(feature = "k8s-auth"))]
+    {
+        if cfg.is_kube() {
+            anyhow::bail!("grpc-hub internal_auth provider=kube requires the `k8s-auth` feature");
+        }
+    }
+
+    Ok(None)
```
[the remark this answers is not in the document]
mattgarmon: Fixed

gears/system/grpc-hub/src/gear.rs
```diff
@@ -160,6 +287,17 @@ impl GrpcHub {
         *self.bound_endpoint.write() = Some(endpoint);
     }
 
+    /// The platform-plane middleware to install on the server, defaulting to a
+    /// pass-through layer when none was configured.
+    fn effective_auth_layer(&self) -> InternalAuthGrpcLayer {
+        self.auth_layer.get().cloned().unwrap_or_else(|| {
```
MikeFalcon77: HIGH

This still fails open. If `auth_layer` was never set, the hub logs a warning and serves every RPC with a pass-through layer. The earlier review round asked for a log line and got one, but the underlying behaviour is unchanged: a missing security control is downgraded to a warning instead of stopping the process. This layer is now the only enforcement point, because the per-RPC checks were deleted from `DirectoryServiceImpl`.

Fix: return `anyhow::Result<InternalAuthGrpcLayer>` and error when `auth_layer` is unset. All three serve paths (`serve_tcp`, `serve_uds`, `serve_named_pipe`) already return `anyhow::Result<()>`, so this just propagates.

gears/system/gear-orchestrator/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 
 // === MODULE DEFINITION ===
 pub mod gear;
-pub use gear::{GearOrchestrator, GearOrchestratorConfig};
+pub use gear::GearOrchestrator;
```
MikeFalcon77, line 9: HIGH

`GearOrchestratorConfig` is deleted and the gear no longer calls `ctx.config_or_default()` at all. An existing deployment that has `gears.gear-orchestrator.config.internal_auth` set will have that key silently ignored after the upgrade. If the operator does not also add `internal_auth` under `grpc_hub`, enforcement disappears with no error and no log line.

Fix: fail `init` with a migration message when the gear-orchestrator config still carries an `internal_auth` key, so the upgrade breaks loudly instead of running unauthenticated. Also mention the move in the changelog.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    /// Replace the exempt method-path allowlist (default:
+    /// [`DEFAULT_EXEMPT_PREFIXES`]).
+    ///
+    /// Each entry is matched as a prefix of the gRPC method path
+    /// (`/<package>.<Service>/<Method>`). Pass an empty vector to enforce on
+    /// every method.
+    #[must_use]
+    pub fn with_exempt_prefixes(mut self, prefixes: Vec<String>) -> Self {
```
MikeFalcon77, line 186: HIGH

The exempt list is taken as an unvalidated `Vec<String>`, and grpc-hub feeds it straight from `internal_auth_exempt_methods` in user config. A single empty string in that list makes `starts_with` match every path, so one config typo disables inbound auth server-wide and nothing reports it. The neighbouring `internal_auth_cache_ttl_secs` is bounds-checked and fails loud; this is not.

Fix: validate here (or in `init`) and reject entries that are empty or do not start with `/`.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -427,6 +565,7 @@ impl GrpcHub {
 
         let incoming = TcpListenerStream::new(listener);
         Server::builder()
+            .layer(self.effective_auth_layer())
```
MikeFalcon77: HIGH

Nothing tests that the layer is actually installed on the server that gets served. `test_run_with_installers_starts_server` and `test_serve_with_system_context` boot a real listener but never issue an RPC, so deleting `.layer(self.effective_auth_layer())` from all three serve paths would leave the whole suite green while inbound enforcement is off.

Fix: add a test that inits the hub with a `shared_secret` `internal_auth` config, serves on `127.0.0.1:0`, and asserts a token-less RPC gets `Code::Unauthenticated` while a token-carrying one succeeds.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+use tower::{Layer, Service};
+
+/// gRPC method path prefixes exempt from platform-plane enforcement by default.
+///
+/// These are the infrastructure services a client or load balancer probes
+/// before (or independently of) authenticating: the standard health-checking
+/// service and both versions of the reflection service. The `grpc.reflection.`
+/// prefix (no trailing service segment) intentionally covers both
+/// `grpc.reflection.v1` and `grpc.reflection.v1alpha`.
+pub const DEFAULT_EXEMPT_PREFIXES: &[&str] = &["/grpc.health.v1.Health/", "/grpc.reflection."];
```
MikeFalcon77: MEDIUM

Exempt matching is a raw `starts_with`, so a prefix can match past a service or method boundary. `/grpc.reflection.` exempts any service whose package merely begins with that string, and an operator-supplied `/pkg.Svc/List` would also exempt `/pkg.Svc/ListSecrets`. For an auth allowlist that is the wrong default direction.

Fix: require the match to end on a segment boundary (the prefix ends with `/`, or the path continues with `/`), and spell the two reflection services out in full: `/grpc.reflection.v1.ServerReflection/` and `/grpc.reflection.v1alpha.ServerReflection/`. Add a negative test for `/grpc.reflection.evil.Svc/Method`.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -56,13 +68,49 @@ pub struct GrpcHubConfig {
...
+    /// Platform-plane (internal) authentication for **all** inbound gRPC RPCs
+    /// served by this hub. When set, an [`InternalAuthGrpcLayer`] validates the
+    /// `x-toolkit-internal-token` on every non-exempt RPC
+    /// (`cpt-cf-adr-platform-plane-auth`); when absent, enforcement is disabled
+    /// (Profile 1 / in-process only).
+    ///
+    /// Mirrors the REST `oop_http.internal_auth` surface: a single flat
+    /// `InternalAuthConfig` provider, with the gRPC-only enforcement / exempt /
+    /// cache knobs as sibling fields below.
+    pub internal_auth: Option<InternalAuthConfig>,
```
MikeFalcon77, line 86: MEDIUM

`GrpcHubConfig` is declared `#[serde(default)]` with no `deny_unknown_fields`, so a misspelled key such as `internal-auth` deserializes to `None` and enforcement is silently off. The two sibling config types that carry the same field both deny unknown fields: `OopHttpConfig` (libs/toolkit/src/bootstrap/config/mod.rs:147) and `GatewayProxyConfig` (gears/system/api-gateway/src/config.rs:92).

Fix: change the container attribute to `#[serde(deny_unknown_fields, default)]` to match `GatewayProxyConfig`.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+    async fn authenticate(&self, token: &str) -> Result<PlatformIdentity, InternalAuthNError> {
+        let now = Instant::now();
+
+        if let Some(identity) = self.lookup(token, now) {
+            return Ok(identity);
+        }
+
+        // Cache miss: validate against the backend. Only cache success — a
+        // rejection or backend outage must be re-evaluated next time.
+        let identity = self.inner.authenticate(token).await?;
```
MikeFalcon77: MEDIUM

Two things compound on the miss path. There is no single-flight, so N concurrent calls carrying the same token issue N TokenReviews - which is the exact burst the cache exists to collapse. And since failures are never cached, an unauthenticated caller sending a fresh bad token per request turns inbound RPC rate into API-server request rate one-to-one.

Fix: dedupe in-flight lookups per token (shared future or per-token entry lock), put a bounded semaphore over backend calls, and rate-limit or briefly negative-cache `InvalidToken` so a caller with no valid credential cannot drive TokenReview traffic.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// `await`), so the returned future stays `Send`.
+    fn lookup(&self, token: &str, now: Instant) -> Option<PlatformIdentity> {
+        let cache = self.cache.lock();
+        let entry = cache.get(token)?;
+        (entry.expires_at > now).then(|| entry.identity.clone())
+    }
+
+    /// Store a freshly validated `identity` under `token` and drop expired entries.
+    fn store(&self, token: String, identity: PlatformIdentity, now: Instant) {
+        let expires_at = now + self.ttl;
```
MikeFalcon77: MEDIUM

`expires_at` is `now + ttl` and ignores the credential's own lifetime. A projected token with two seconds left is cached for the full TTL and keeps authenticating after it expires, which TokenReview would have rejected. grpc-hub allows a TTL up to 300s, so that window can be five minutes.

Fix: clamp `expires_at` to the token's remaining validity by reading `exp` from the JWT payload (TokenReview has already verified the signature, so a plain base64 decode is enough) and take whichever is sooner.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+        let now = Instant::now();
+
+        if let Some(identity) = self.lookup(token, now) {
+            return Ok(identity);
+        }
+
+        // Cache miss: validate against the backend. Only cache success — a
+        // rejection or backend outage must be re-evaluated next time.
+        let identity = self.inner.authenticate(token).await?;
+        self.store(token.to_owned(), identity.clone(), now);
```
MikeFalcon77: MEDIUM

`now` is sampled at the top of `authenticate`, before the backend round-trip, and then used as the base for `expires_at`. The effective TTL is short by the TokenReview latency. With a short configured TTL and a slow API server the entry is already expired when it is inserted, so the cache never hits and silently stops doing its job.

Fix: re-sample `Instant::now()` after the `.await` and pass that to `store`; keep the pre-await instant for the lookup only.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+        let cache = self.cache.lock();
+        let entry = cache.get(token)?;
+        (entry.expires_at > now).then(|| entry.identity.clone())
+    }
+
+    /// Store a freshly validated `identity` under `token` and drop expired entries.
+    fn store(&self, token: String, identity: PlatformIdentity, now: Instant) {
+        let expires_at = now + self.ttl;
+        let mut cache = self.cache.lock();
+        cache.retain(|_, entry| entry.expires_at > now);
```
MikeFalcon77: MEDIUM

`store` is the only reclamation path - `lookup` shadows an expired entry without removing it. Once traffic goes idle, expired entries and the raw tokens they are keyed by stay in the heap indefinitely. The sweep is also a full O(n) map walk under the mutex on every cache miss.

Fix: remove the expired entry in `lookup` when it is found stale, and replace the per-store full sweep with an amortized one (every Nth insert, or a timer-driven pass) so the lock is never held for O(n) work on the request path.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -94,6 +142,81 @@ fn parse_advertise_addr(advertise_addr: &str) -> anyhow::Result<(String, Option<
...
+/// [`CachingInternalAuthenticator`](toolkit_k8s_auth::CachingInternalAuthenticator)
+/// so a burst of calls carrying the same projected token collapses to a single
+/// API-server round-trip; `0` uses the uncached validator. Without the feature,
+/// configuring `provider: kube` is a hard error rather than a silent enforcement
+/// downgrade.
+#[cfg_attr(
+    not(feature = "k8s-auth"),
+    allow(clippy::unused_async, unused_variables)
+)]
+async fn build_internal_authenticator(
```
MikeFalcon77, line 164: MEDIUM

This is a second copy of provider selection plus `k8s-auth` gating; `libs/toolkit/src/bootstrap/oop.rs:721` has a near-identical `build_internal_authenticator`. The copies already differ in ways that matter: only this one wraps the validator in the positive cache, and only this one fails when the provider cannot be built (oop.rs ends in `Ok(None)`, a silent pass-through). So the HTTP path still pays an uncached TokenReview per request, and the two paths disagree on what a bad provider config means.

Fix: move provider construction plus the optional caching wrapper into one shared helper in toolkit-security/toolkit and call it from both sites.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+///
+/// # async fn wire() -> Result<(), Box<dyn std::error::Error>> {
+/// let validator =
+///     K8sTokenReviewAuthenticator::try_default(vec!["toolkit-internal".to_owned()]).await?;
+/// let cached = CachingInternalAuthenticator::new(validator, Duration::from_secs(30));
+/// # let _ = cached;
+/// # Ok(())
+/// # }
+/// ```
+pub struct CachingInternalAuthenticator<A> {
```
MikeFalcon77: MEDIUM

`CachingInternalAuthenticator<A>` is generic over any `InternalAuthenticator` and touches no `kube` or `k8s-openapi` type, but it ships from the Kubernetes-specific crate. Anyone who wants positive caching for another provider has to depend on the whole Kubernetes implementation and turn on `k8s-auth`.

Fix: move the type and its default-TTL constant into `toolkit-security` next to `InternalAuthenticator` and `DynInternalAuthenticator`, rename the constant off `TOKEN_REVIEW`, and have `toolkit-k8s-auth` consume it.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+                        }
+                        Err(err) => {
+                            tracing::warn!(error = %err, "platform-plane gRPC authentication failed");
+                            Ok(authn_error_to_status(&err).into_http())
+                        }
+                    }
+                }
+                // A malformed / empty credential is rejected in either mode.
+                TokenOutcome::Invalid => {
+                    Ok(Status::unauthenticated("invalid internal token").into_http())
```
MikeFalcon77, line 364: MEDIUM

This rejection and the missing-token rejection below it are completely silent - no log, no metric. Only the authenticator-error path warns. A fleet-wide credential misconfiguration would show up as clients failing with no server-side trace of why.

Fix: emit a `tracing::warn!` in both arms with the gRPC method path and the rejection reason (never the token), and add a rejection counter keyed by reason. The existing warn at line 257 should carry the method path too.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    #[tokio::test]
+    async fn invalid_token_is_rejected() {
+        let resp = call(&authed_layer(), request("/pkg.Svc/Method", Some("nope"))).await;
+        assert_eq!(grpc_status(&resp), Some(16));
+    }
+
+    #[tokio::test]
+    async fn empty_token_is_rejected_even_when_permissive() {
```
MikeFalcon77, line 523: MEDIUM

The module doc's central claim - a present-but-invalid token is rejected regardless of mode - is not tested. This test only covers an empty token, which short-circuits on the `TokenOutcome::Invalid` branch and never reaches the authenticator. A regression that let Permissive fall through on an `authenticate()` error would pass.

Fix: add a Permissive-mode test with token `"nope"` asserting grpc-status 16, and one with `"down"` asserting 14.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    #[tokio::test]
+    async fn errors_are_not_cached() {
+        let cached = CachingInternalAuthenticator::new(CountingAuth::new(), Duration::from_mins(1));
+        cached.inner.fail.store(1, Ordering::SeqCst);
+
+        assert!(cached.authenticate("tok").await.is_err());
+        assert!(cached.authenticate("tok").await.is_err());
+        assert_eq!(cached.inner.calls(), 2, "failures must not be cached");
```
MikeFalcon77: MEDIUM

This only exercises `Unavailable`. The documented guarantee covers rejections too, and `InvalidToken` is the variant that actually matters here - caching a rejection would pin a token to a stale "denied" answer, and caching it as success would be worse. Neither is covered.

Fix: give `CountingAuth` a switch for `InvalidToken` and assert two consecutive rejections both reach the backend, then that the token validates once the backend starts accepting it.

gears/system/gear-orchestrator/src/server.rs
```diff
@@ -506,16 +453,14 @@ fn validate_identity(gear_name: &str, instance_id: &str) -> Result<(), Status> {
...
-/// When `authenticator` is `Some`, every RPC requires a valid
-/// `x-toolkit-internal-token`; when `None`, enforcement is disabled.
+/// Platform-plane enforcement is applied at the gRPC server boundary by
+/// `grpc-hub`'s `InternalAuthGrpcLayer`.
 pub fn make_directory_service(
     api: Arc<dyn DirectoryClient>,
-    authenticator: Option<DynInternalAuthenticator>,
 ) -> DirectoryServiceServer<DirectoryServiceImpl> {
-    DirectoryServiceServer::new(DirectoryServiceImpl::with_authenticator(api, authenticator))
+    DirectoryServiceServer::new(DirectoryServiceImpl::new(api))
```
MikeFalcon77, line 466: MEDIUM

The handlers no longer receive the caller identity at all. The old `authenticate_peer` returned a `PeerAuthenticated` that the doc described as being for workload-policy checks; now nothing reads `PlatformSecurityContext` from the request extensions. Authorization was already missing (the old call discarded its result), but this removes the only path a handler had to get the identity, so any holder of a valid platform-plane token - any ServiceAccount in the audience, or the single shared secret - can register or deregister an arbitrary `gear_name` and hijack another gear's discovery entry.

Fix: read `PlatformSecurityContext` from the extensions in the mutating RPCs (`register_instance`, `deregister_instance`, `heartbeat`) and check the caller identity against `gear_name` before touching the directory.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -600,6 +741,38 @@ impl Gear for GrpcHub {
         // Parse listen_addr into appropriate transport type
         self.apply_listen_config(&cfg.listen_addr)?;
 
+        // Fail loud at init on an out-of-bound cache TTL rather than silently
+        // running with a dangerously wide token-revocation window.
+        if cfg.internal_auth_cache_ttl_secs > MAX_INTERNAL_AUTH_CACHE_TTL_SECS {
```
MikeFalcon77, line 779: MEDIUM

The `MAX_INTERNAL_AUTH_CACHE_TTL_SECS` guardrail has no test. This is the check that keeps the revocation window bounded, and neither the boundary nor the failure is verified.

Fix: add `init` tests using the existing `ConfigProvider` test pattern - 301 returns an error mentioning the maximum, 300 succeeds.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+
+impl InternalAuthGrpcLayer {
+    /// Build a layer around an optional `authenticator`.
+    ///
+    /// When `authenticator` is `Some`, enforcement defaults to
+    /// [`InternalAuthEnforcement::Required`] with the
+    /// [`DEFAULT_EXEMPT_PREFIXES`] allowlist. When `None`, the layer is a
+    /// no-op pass-through (Profile 1 / in-process).
+    #[must_use]
+    pub fn new(authenticator: Option<DynInternalAuthenticator>) -> Self {
```
MikeFalcon77: MEDIUM

`new(Option<DynInternalAuthenticator>)` conflates "no validator supplied" with "authentication deliberately off". That is what makes the fully-open configuration reachable by accident at any call site - and grpc-hub's `effective_auth_layer` fallback is exactly that accident.

Fix: take a non-optional `DynInternalAuthenticator` in `new` and add an explicit `InternalAuthGrpcLayer::disabled()` for the in-process pass-through, so disabling auth is always a deliberate call.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+        f.debug_struct("CachingInternalAuthenticator")
+            .field("ttl", &self.ttl)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<A> CachingInternalAuthenticator<A> {
+    /// Wrap `inner`, caching successful validations for `ttl`.
+    #[must_use]
+    pub fn new(inner: A, ttl: Duration) -> Self {
```
MikeFalcon77: MEDIUM

`new` accepts any `Duration` with no bound. The cap that makes this type safe to use lives in grpc-hub's `init`, so every other caller gets no protection, and `Duration::ZERO` builds a cache that can never hit.

Fix: export a `MAX_TOKEN_REVIEW_CACHE_TTL` from this crate, return `Result` from `new` rejecting zero and anything above it, and have grpc-hub rely on that instead of its own local constant.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -56,13 +68,49 @@ pub struct GrpcHubConfig {
...
+    pub internal_auth_exempt_methods: Option<Vec<String>>,
+
+    /// Seconds a **successful** platform-plane validation is cached, collapsing a
+    /// burst of calls carrying the same token into a single validation-backend
+    /// round-trip. Only positive results are cached, and only providers that make
+    /// a remote call benefit — in practice `kube` (the shared-secret provider is
+    /// a local comparison and is never wrapped). `0` disables caching (every call
+    /// re-validates). Bounded at boot by [`MAX_INTERNAL_AUTH_CACHE_TTL_SECS`] to
+    /// keep the token-revocation window tight.
+    pub internal_auth_cache_ttl_secs: u64,
```
MikeFalcon77, line 107: MEDIUM

All four new config keys are missing from the README. `gears/system/grpc-hub/README.md` still documents only `listen_addr`, so the entire inbound auth surface is undiscoverable for whoever writes the deployment YAML.

Fix: extend the README `Configuration` block with `internal_auth`, `internal_auth_enforcement`, `internal_auth_exempt_methods` and `internal_auth_cache_ttl_secs`, and note that `provider: kube` needs the `k8s-auth` feature.

gears/system/gear-orchestrator/src/server.rs
```diff
@@ -967,6 +914,7 @@ mod tests {
         // ephemeral port still bound.
         let server = tokio::spawn(async move {
             Server::builder()
+                .layer(auth_layer)
```
MikeFalcon77, line 937: MEDIUM

This is the only test that runs the layer behind a real tonic server, and it never checks that the handler saw `PlatformSecurityContext` or `PeerAuthenticated`. The claim that tonic propagates `http::Request` extensions into `tonic::Request` is only verified against the hand-written `Echo` service at the http layer. Since the per-handler `authenticate_peer` is gone, a propagation break would go unnoticed.

Fix: have one handler on this server read the extensions and assert the authenticated call sees peer `"peer"`.

gears/system/grpc-hub/src/gear.rs
```diff
@@ -600,6 +741,38 @@ impl Gear for GrpcHub {
...
+
+        // Build the inbound platform-plane middleware; an unset `internal_auth`
+        // yields a pass-through layer (Profile 1 / in-process).
+        let authenticator = build_internal_authenticator(
+            cfg.internal_auth.as_ref(),
+            cfg.internal_auth_cache_ttl_secs,
+        )
+        .await?;
+        if authenticator.is_some() {
+            tracing::info!(
```
MikeFalcon77: LOW

This logs only the enforcement mode - not the provider, the effective cache TTL, or whether the exempt list was overridden. And when `internal_auth` is absent the hub serves everything unauthenticated with nothing logged at all, which is the case an operator most needs to see.

Fix: include provider, cache TTL and effective exempt prefixes in this line, and add an explicit warning at init when `internal_auth` is unset.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    #[tokio::test]
+    async fn entry_expires_after_ttl() {
+        let cached =
+            CachingInternalAuthenticator::new(CountingAuth::new(), Duration::from_millis(20));
+
+        cached.authenticate("tok").await.unwrap();
+        assert_eq!(cached.inner.calls(), 1);
+
+        tokio::time::sleep(Duration::from_millis(40)).await;
```
MikeFalcon77: LOW

TTL expiry is proven with a real 40ms sleep against a 20ms TTL. The cache reads `std::time::Instant::now()` internally, so `tokio::time::pause` cannot virtualize it - the test is wall-clock dependent and slow by construction.

Fix: `lookup` and `store` already take `now: Instant`. Add an internal `authenticate_at(&self, token, now)` and keep `authenticate` as the `Instant::now()` wrapper, then assert expiry by passing `now + ttl + 1ms` instead of sleeping.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+    Present(SecretString),
+    /// No internal-token header was present.
+    Missing,
+    /// The header was present but malformed (non-ASCII) or empty.
+    Invalid,
+}
+
+/// Read the `x-toolkit-internal-token` header from an inbound request.
+fn read_token(headers: &http::HeaderMap) -> TokenOutcome {
+    let Some(value) = headers.get(INTERNAL_TOKEN_HEADER) else {
```
MikeFalcon77: LOW

`headers.get` returns only the first value, so if a caller sends `x-toolkit-internal-token` twice, the first is validated and both are forwarded to the handler untouched. Handlers should also never see the raw credential once the layer has consumed it.

Fix: reject the request when the header appears more than once, and remove the header from the request before calling the inner service.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    #[tokio::test]
+    async fn distinct_tokens_are_cached_independently() {
+        let cached = CachingInternalAuthenticator::new(CountingAuth::new(), Duration::from_mins(1));
+
+        cached.authenticate("a").await.unwrap();
+        cached.authenticate("b").await.unwrap();
+        cached.authenticate("a").await.unwrap();
+        assert_eq!(
```
MikeFalcon77: LOW

This asserts only the backend call count and never inspects the returned `PlatformIdentity`. A cache that handed token `"a"` the identity stored for `"b"` - identity confusion in security-critical code - would still pass. `CountingAuth` already sets the identity name to the token, so the check is free.

Fix: assert the returned identity name equals the token on each of the three calls.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,449 @@
...
+
+    fn call(&mut self, mut req: http::Request<ReqBody>) -> Self::Future {
+        let config = Arc::clone(&self.config);
+        // Tower readiness contract: `poll_ready` was called on `self.inner`, so
+        // the readiness reservation belongs to it. Move that instance into the
+        // future and leave a fresh clone behind for the next call.
+        let clone = self.inner.clone();
+        let mut inner = std::mem::replace(&mut self.inner, clone);
+
+        Box::pin(async move {
```
MikeFalcon77: LOW

Every inbound RPC pays a `Box::pin` heap allocation plus an inner-service clone before any branch is taken - including the no-authenticator pass-through, which is the default configuration, and the exempt health and reflection paths that do no async work at all.

Fix: replace the boxed `Self::Future` with an enum future carrying a `PassThrough(S::Future)` variant, and dispatch to it before allocating.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+            CacheEntry {
+                identity,
+                expires_at,
+            },
+        );
+    }
+}
+
+impl<A: InternalAuthenticator> InternalAuthenticator for CachingInternalAuthenticator<A> {
+    async fn authenticate(&self, token: &str) -> Result<PlatformIdentity, InternalAuthNError> {
```
MikeFalcon77: LOW

Per the Rust guidelines, every async fn should state its cancel safety. This one is polled inside tonic's abortable connection task and inside the backend's `tokio::time::timeout`; a drop between the backend returning and `store` running throws away a completed round-trip.

Fix: add a `// cancel-safe:` note on `authenticate` saying cancellation only loses the cache insert and leaves no partial state. Same for `InternalAuthGrpcService::call`.

libs/toolkit-k8s-auth/src/cache.rs
Artifizer: **[MAJOR] F-1-MAJOR-codex-default** in `libs/toolkit-k8s-auth/src/cache.rs`

Concurrent requests for the same uncached token all pass `lookup` before any result is stored and independently invoke `self.inner.authenticate(token)`. Consequently, the cache does not deliver its stated purpose of collapsing a burst into one Kubernetes `TokenReview` request. A newly observed or expired token can still generate one API-server call per concurrent RPC, creating a thundering herd precisely on the hot path this wrapper is intended to protect.

**Suggested fix:** Add per-token single-flight coordination so one request performs validation while concurrent callers await and reuse its result. Do not hold the synchronous cache mutex across `.await`; store shared in-flight futures or use a per-key async synchronization primitive, and remove the in-flight entry after either success or failure. Add a concurrency test that releases many tasks simultaneously for one token and asserts the backend was called once.

libs/toolkit-k8s-auth/src/cache.rs
```diff
@@ -0,0 +1,239 @@
...
+        let entry = cache.get(token)?;
+        (entry.expires_at > now).then(|| entry.identity.clone())
+    }
+
+    /// Store a freshly validated `identity` under `token` and drop expired entries.
+    fn store(&self, token: String, identity: PlatformIdentity, now: Instant) {
+        let expires_at = now + self.ttl;
+        let mut cache = self.cache.lock();
+        cache.retain(|_, entry| entry.expires_at > now);
+        cache.insert(
```
Artifizer: **[MAJOR] F-2-MAJOR-codex-default** in `libs/toolkit-k8s-auth/src/cache.rs`

The cache has no entry or memory bound and stores complete credential strings as keys. Expired entries are removed only when another successful validation reaches `store`; they are not removed during lookup, after failed validations, or by a periodic cleanup task. A high rate of distinct valid tokens can therefore retain an arbitrarily large amount of sensitive token material for at least the TTL, while a cache that becomes idle retains all expired credentials indefinitely until the authenticator is dropped or another successful validation occurs. This makes an authentication boundary vulnerable to memory growth based on caller-controlled credentials and contradicts the documented claim that entries themselves are TTL-bounded.

**Suggested fix:** Use a bounded cache with both TTL expiration and a maximum entry/weight limit, evict expired entries during lookup, and avoid retaining raw credentials where practical by keying on a collision-resistant digest. Add tests proving that expired entries disappear without requiring a later successful authentication and that exceeding the configured capacity evicts entries.

gears/system/grpc-hub/src/gear.rs
Artifizer: **[MAJOR] F-1-MAJOR-claude-sonnet-5** in `gears/system/grpc-hub/src/gear.rs`

This PR moves platform-plane authentication out of `DirectoryServiceImpl` entirely and centralizes it in a single server-wide Tower layer (`InternalAuthGrpcLayer`) built once in `GrpcHub::init` and stored in `self.auth_layer: OnceLock<InternalAuthGrpcLayer>`. All three serve paths (`serve_tcp`, the Unix-socket path, and the named-pipe path) fetch that layer via `effective_auth_layer()` before mounting routes. `effective_auth_layer()` (lines 292-299) treats an unset `auth_layer` as "no security context configured" and silently falls back to a pass-through layer, logging only a `tracing::warn!`. Since `DirectoryServiceImpl` and every other RPC handler no longer perform any of their own authentication, this is now the only enforcement point for `x-toolkit-internal-token`. If `serve_*` is ever reached without `init` having successfully populated `auth_layer` — e.g. a future refactor of gear lifecycle, a test harness, or any code path that calls `serve_tcp`/`serve_unix`/`serve_named_pipe` before `init` completes — every RPC on the hub silently becomes unauthenticated, even if `internal_auth` was configured with `Required` enforcement. This is inconsistent with the established convention in the same file: other required-but-possibly-unset `OnceLock`s (`instance_id` at lines 478-482 and 651-655) fail closed, refusing to proceed rather than silently defaulting. The cache-TTL validation added near `effective_auth_layer` explicitly states the philosophy: fail loud at init on an out-of-bound token-revocation window. `effective_auth_layer` does the opposite for the security control itself: fail open, log-only.

**Suggested fix:** Make `effective_auth_layer()` fail closed like the other required `OnceLock` accessors, e.g. have `serve_tcp`/`serve_uds`/`serve_named_pipe` propagate an error when `self.auth_layer.get()` is absent instead of silently constructing a permissive `InternalAuthGrpcLayer::new(None)`. If a genuine no-enforcement pass-through is needed, that state should be explicit by setting the layer during `init` even when `internal_auth` is `None`, distinguishing it from never initialized.

mattgarmon: Fixed, now returns `Result`

gears/system/gear-orchestrator/src/lib.rs
mattgarmon: Fixed, added migration message which fails loudly

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Fixed, validated in `init`

gears/system/grpc-hub/src/gear.rs
mattgarmon: Added test

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Fixed, `prefix_matches_boundary` added + full reflection prefixes

gears/system/grpc-hub/src/gear.rs
mattgarmon: Fixed, added `deny_unknown_fields`

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed, moved cache to `toolkit-security` with fixes

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Updated

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed, evict-on-lookup + amortized sweep

gears/system/grpc-hub/src/gear.rs
mattgarmon: Added shared helper

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed, moved to `toolkit-security`, feature-gated

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Added log

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Added test

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed

gears/system/gear-orchestrator/src/server.rs
mattgarmon: Already captured in follow-up: https://github.com/constructorfabric/gears-rust/issues/4554

gears/system/grpc-hub/src/gear.rs
mattgarmon: Added tests

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Fixed

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed

gears/system/grpc-hub/src/gear.rs
mattgarmon: Extended config in README

gears/system/gear-orchestrator/src/server.rs
mattgarmon: Fixed via `tower::util::MapRequestLayer` probe

gears/system/grpc-hub/src/gear.rs
mattgarmon: Updated

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Added `authenticate_at`

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Fixed

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Fixed. `Either<S::Future, Pin<Box<dyn Future>>>`; pass-through/exempt paths (`Either::Left`) no longer allocate, only the enforced path is boxed

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Duplicate of https://github.com/constructorfabric/gears-rust/pull/4642#discussion_r3831732249

Cache has been moved to `toolkit-security`, with single-flight coordination

libs/toolkit-k8s-auth/src/cache.rs
mattgarmon: Fixed. Added `MAX_CACHE_ENTRIES` with soonest-to-expire eviction

gears/system/grpc-hub/src/gear.rs
mattgarmon: Duplicate of https://github.com/constructorfabric/gears-rust/pull/4642#discussion_r3831732212

Fixed by returning `Result`

gears/system/gear-orchestrator/src/gear.rs
```diff
@@ -123,19 +60,25 @@ impl SystemCapability for GearOrchestrator {
...
+            .and_then(|raw| raw.get("config"))
+            .and_then(|config| config.get("internal_auth"))
+        {
+            anyhow::bail!(
+                "gear-orchestrator config still sets `internal_auth` ({internal_auth}), which no \
+                 longer has any effect: platform-plane enforcement moved to grpc-hub's own \
+                 `internal_auth` config (cpt-cf-adr-platform-plane-auth). Move this key under \
+                 `gears.grpc-hub.config` instead."
+            );
         }
```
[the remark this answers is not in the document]
mattgarmon: Fixed

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -0,0 +1,742 @@
...
+        match self.lookup(token, now) {
+            CacheLookup::Valid(identity) => return Ok(identity),
+            CacheLookup::Rejected => return Err(InternalAuthNError::InvalidToken),
+            CacheLookup::Miss => {}
+        }
+
+        // Single-flight: serialize concurrent misses for the same token so a
+        // burst of calls collapses into one backend round-trip.
+        let lock = self.token_lock(token);
+        let guard = lock.lock().await;
```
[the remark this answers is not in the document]
mattgarmon: Fixed by adding `ReleaseTokenLockOnDrop`. Added test

MikeFalcon77 (commented):
Re-review of the 27 comments from the previous round: 24 fixed, 2 partial, 1 still open. Details inline. CI is green on all 33 checks.

gears/system/gear-orchestrator/src/server.rs
MikeFalcon77, line 466: MEDIUM

Still open from the last round. Nothing in this file reads `PlatformSecurityContext` or `PeerAuthenticated` - the only mentions left are the doc comment at line 44 and the test at line 916. `register_instance`, `deregister_instance` and `heartbeat` validate the shape of `gear_name` but never check who is asking, so any holder of a valid platform-plane token can still register or deregister an arbitrary gear and hijack another gear's discovery entry.

The layer work in this PR makes authentication solid, so this is now purely the authorization half. It is fine to split it into a follow-up issue rather than grow this PR - but the identity is available in the extensions now and nothing consumes it, so please say which it is.

Fix: read `PlatformSecurityContext` in the three mutating RPCs and check the caller identity against `gear_name`, or open a tracked follow-up and note it in the module doc.

gears/system/gear-orchestrator/src/server.rs
```diff
@@ -1016,6 +982,12 @@ mod tests {
         assert_eq!(all.len(), 1);
         assert_eq!(all[0].gear, "billing");
 
+        assert!(
+            saw_expected_peer.load(std::sync::atomic::Ordering::SeqCst),
+            "InternalAuthGrpcLayer's PeerAuthenticated extension must propagate through tonic \
```
MikeFalcon77: LOW

This assertion does not check what its message says. `probe_layer` is a `MapRequestLayer` over `http::Request`, and `Server::builder().layer(auth_layer).layer(probe_layer)` puts both layers outside tonic's router - so the probe observes the request on its way out of the auth layer, which the `Echo`-based unit tests already cover. It never observes a `tonic::Request` inside a handler, so it does not verify the one hop that was previously untested: that tonic copies `http::Request` extensions into the extensions a handler sees.

Everything else about this test is right, and the rest of the round is fixed - this is just the claim being wider than the check.

Fix: either read the extension inside a real handler on this server and assert there, or reword the message to say what it actually proves (the layer populates the extension on the request it forwards).

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -0,0 +1,785 @@
...
+    fn insert(&self, token: String, entry: CacheEntry, now: Instant) {
+        let mut cache = self.cache.lock();
+        let count = self.sweep_counter.fetch_add(1, Ordering::Relaxed) + 1;
+        if count.is_multiple_of(SWEEP_INTERVAL) {
+            cache.retain(|_, e| e.expires_at() > now);
+        }
+        if cache.len() >= MAX_CACHE_ENTRIES && !cache.contains_key(&token) {
+            // No extra bookkeeping for real LRU order: the soonest-to-expire
+            // entry needs re-validation soonest regardless.
+            if let Some(victim) = cache
```
MikeFalcon77: LOW

New in this round. Once the cache reaches `MAX_CACHE_ENTRIES`, every subsequent insert for an unseen token runs `min_by_key` over the whole 10k-entry map while holding the mutex. That is the same O(n)-under-the-lock cost on the hot path that `SWEEP_INTERVAL` was introduced to amortize away, just moved to the capacity path - and it does not amortize, it happens on every miss for as long as the cache stays full.

It needs a large fleet of simultaneously-valid tokens to reach, so this is not urgent - but at that point it is a lock-contention cliff rather than a gradual slowdown.

Fix: force a `retain` sweep first when the map is full and only scan for a victim if that frees nothing, or pick the victim from a small bounded sample instead of the whole map.

gears/system/gear-orchestrator/src/server.rs
MikeFalcon77: not fixed. the register/deregister/heartbeat handlers still take gear_name from the request body and never check who is calling. any peer with a valid token can still hijack or overwrite another gear's registration. issue 4554 already covers this exact gap from an earlier PR and explains why it needs a bigger change (per-gear identity, which shared-secret mode does not have yet). fine to leave it there, just confirming this PR does not close it - it only made the token check itself solid.

gears/system/gear-orchestrator/src/server.rs
MikeFalcon77: partially fixed. the probe layer sits next to auth_layer on the tonic Server builder, both outside the router, so it only sees the http request as it leaves the auth layer - the same thing the existing unit tests already check. it does not prove tonic copies the extension into what a handler sees, which was the actual gap. to really close it, read PlatformSecurityContext or PeerAuthenticated inside a real handler on this server and assert there.

gears/system/grpc-hub/src/gear.rs
MikeFalcon77: partially fixed. the kube build and caching now go through one shared build_cached_k8s_authenticator in toolkit-k8s-auth, and oop.rs now errors instead of silently returning Ok(None), so the real bugs from the duplication are gone. what is still duplicated is the outer function itself - build_internal_authenticator in grpc-hub/gear.rs and in toolkit/oop.rs are still two separate ~30 line copies doing the same provider dispatch and feature gating. not a correctness problem anymore, just remaining duplication.

libs/toolkit-transport-grpc/src/internal_auth_server.rs
```diff
@@ -0,0 +1,686 @@
...
+
+/// gRPC method path prefixes exempt from platform-plane enforcement by default.
+///
+/// These are the infrastructure services a client or load balancer probes
+/// before (or independently of) authenticating: the standard health-checking
+/// service and both versions of the reflection service, spelled out in full
+/// (not a bare `grpc.reflection.` prefix) so an unrelated package that merely
+/// shares the string prefix (e.g. `grpc.reflection.evil.Svc`) is never
+/// accidentally exempted.
+pub const DEFAULT_EXEMPT_PREFIXES: &[&str] = &[
```
MikeFalcon77, line 64: Are these standard gRPC methods?

striped-zebra-dev (approved):

gears/system/gear-orchestrator/src/server.rs
mattgarmon: Confirmed. This PR is the authentication half only. The peer-vs-gear_name authorization for `register_instance`/`deregister_instance`/`heartbeat` is tracked in #4554

gears/system/gear-orchestrator/src/server.rs
mattgarmon: Reworded the doc and assertion to say what it proves: the layer populates `PeerAuthenticated` on the request it forwards downstream.

libs/toolkit-security/src/internal_auth_cache.rs
mattgarmon: Fixed. When full, `insert` now runs one `retain` sweep to reclaim expired entries first (amortized across the inserts it frees room for) and only falls back to the `min_by_key` scan if that frees nothing

libs/toolkit-transport-grpc/src/internal_auth_server.rs
mattgarmon: Yes, these are gRPC standard well-known service names, matched by full service prefix:

- `grpc.health.v1.Health`: The standard gRPC Health Checking Protocol
- `grpc.reflection.v1.ServerReflection` and `.v1alpha.ServerReflection`: The standard Server Reflection service

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -0,0 +1,846 @@
...
+            token,
+            lock: &lock,
+        };
+
+        // Another caller may have populated the cache while this one waited.
+        match self.lookup(token, now) {
+            CacheLookup::Valid(identity) => return Ok(identity),
+            CacheLookup::Rejected => return Err(InternalAuthNError::InvalidToken),
+            CacheLookup::Miss => {}
+        }
```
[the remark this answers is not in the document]
mattgarmon: Fixed, added test

PR #4648, opened 2026-08-21
Add Orders Lifecycle PRD

Corw1n-of-Amber:
## Summary

Adds the Orders Lifecycle PRD at `gears/bss/orders-lifecycle/docs/PRD.md`

**In scope:** the order document as System of Record, the order finite-state machine, catalog price pin and sellability gate at submit, versioning/amendment, cancel/hold/expiry, tenant axes, authorization, and the Lifecycle↔Workflow seam rules R1–R5 (single normative home in this PRD).

**Out of scope:** approval execution and fulfillment orchestration (Orders Workflow); price math (rating / Plan & Price); subscription state and OSS provisioning (Subscriptions); contract terms (Contracts).

## Validation

- [x] `cfs toc gears/bss/orders-lifecycle/docs/PRD.md`
- [x] `cfs validate --artifact gears/bss/orders-lifecycle/docs/PRD.md --skip-code --local-only --verbose` — **0 errors / 0 warnings**


dogbert2021:
From the CT Depot perspective, the Lifecycle↔Workflow ownership split and the two-wave/all-or-nothing saga are clear.

CT Depot channel facade →
  → CF Orders Lifecycle ↔ CF Orders Workflow
  → CF Subscriptions Lifecycle ↔ CT Depot OSS Subscription Manager specialization
  → CT OSS App Fulfillment
  → CT runtime App owners

The remaining **integration gap is the Workflow→Subscriptions contract**. 
Please require **draft-create, activation, draft-void and activated-cancel intents** and confirmations to **preserve orderId, orderVersion, orderLineId, wave, process correlationId, and an opaque caller-owned commercial-binding reference**. Please also include orderVersion in the idempotency identity; using only order ID + line + wave could collide after an amendment.

This lets Depot OSS Subscription Manager pin its exact commercial projection and OSS contract without leaking CT-specific internals into CF.  CT Depot OSS Subscription Manager and CT OSS App Fulfillment are CT consumers behind the generic CF Subscriptions lifecycle: draft create/void has no OSS effects, while activation/cancel uses pinned CT App Fulfillment operations. This prevents duplicate orchestration and a second order source of truth.

Finally, currently BSS roadmap still leaves Order Management [unscheduled](https://github.com/orgs/constructorfabric/projects/48/views/1?filterQuery=Order) while Depot has a first aggregators channel orders adapter target Q1'27 (f.ex. Ingram Micro purchase orders). 

**Please record this as a CT Q1'27 delivery dependency**, so CT Depot does not build a temporary replacement order engine.

Corw1n-of-Amber:
Thanks — this is useful, and the ownership picture is agreed. The Lifecycle↔Workflow split and the two-wave saga are intentional; they are unchanged in this PRD.

The remaining gap is the Workflow→Subscriptions intent contract, not order-document semantics, so it is addressed on the sibling PR (#4649) rather than restated here.

The Q1'27 Depot / channel-adapter timeline is noted as a delivery dependency. Dates and adopter commitments are not recorded in the PRD.

rostm:
> ## Summary
> Adds the Orders Lifecycle PRD at `gears/bss/orders-lifecycle/docs/PRD.md`
> 
> **In scope:** the order document as System of Record, the order finite-state machine, catalog price pin and sellability gate at submit, versioning/amendment, cancel/hold/expiry, tenant axes, authorization, and the Lifecycle↔Workflow seam rules R1–R5 (single normative home in this PRD).
> 
> **Out of scope:** approval execution and fulfillment orchestration (Orders Workflow); price math (rating / Plan & Price); subscription state and OSS provisioning (Subscriptions); contract terms (Contracts).
> 
> ## Validation
> * [x]  `cfs toc gears/bss/orders-lifecycle/docs/PRD.md`
> * [x]  `cfs validate --artifact gears/bss/orders-lifecycle/docs/PRD.md --skip-code --local-only --verbose` — **0 errors / 0 warnings**

## Subscription identity: one tenant, many apps

The model we need this boundary to support: a single platform with multiple apps,
each app carrying its own SKUs. A purchase provisions **one tenant** with the
corresponding apps enabled, and that tenant is managed through **one
subscription** — the enabled apps are its entitlement set, not separate
subscriptions.

The PRD's cardinality rules rule this out. Glossary *Order Line Item*: "one
fulfilled line item spawns **exactly one** subscription"; §6.1 fulfillment
mapping: "line → subscription is **1:1**". Selecting three apps produces three
subscriptions against one `resourceTenantId`, and no object represents "this
tenant's subscription".

### Why the two available answers don't fit

**Bundle plan.** The PRD's route to one subscription from several things,
captured as a single line, "never expanded into component lines at capture".
That makes every sellable app combination a pre-published catalog artifact with
its own price basis and rev-share. With N apps that is a combinatorial catalog
to author and maintain, and it still cannot express the ordinary case: enabling
one more app on a tenant that already exists. That would be a *different* bundle
plan — a plan change on a running subscription — for what is commercially just
"turn on app C".

**Grouping lines within an order.** Also insufficient, and worth being explicit
about why. A group is order-scoped; the tenant's subscription is not. A
follow-on order enabling app C has nothing to group into and would spawn a
second subscription. Whatever binds a line to a subscription has to be able to
name a subscription that already exists.

## Gap: add-on rules are assumed by the gate but not representable on a line

Gate predicate §6.1 (b) validates `qty` against "the declared bounds on add-on
rules and one-time plans". Add-on rules are plan-scoped in the catalog
(`cpt-cf-bss-plan-price-fr-addon-rules`: eligible add-on SKU, required/optional,
min/max/step quantity, optional price override reference). But the order line
carries only `skuId` + `planId` + `priceId` + `qty` + `catalogPricePin`, and
§9.1 defers shapes to Design. Nothing on the line expresses which add-ons were
selected against the base plan, so the gate checks bounds against data the line
model does not carry.

Either the line model gains add-on selection, or predicate (b) should drop the
add-on clause and the PRD should state that add-ons are out of scope for order
capture.

## Secondary: overlap key default

Gate predicate (f) defaults `overlapScopeKey` to
`(payerTenantId, catalogSubscriptionProductKey)` with cardinality 1. Keyed on
the *payer*, a partner buying the same app for several customer tenants collides
with themselves — the second tenant's order is rejected. For resource-tenant-
scoped provisioning the natural default key is `resourceTenantId`. Configurable
via `maxConcurrentActive`, so this is a question about the default, not the
mechanism.

## Asks

1. State whether a subscription may carry entitlements for multiple SKUs, and
   whether an order line may target a subscription (new or existing) rather than
   spawning one — coordinated with the Subscriptions PRD, not settled by the 1:1
   rule here.
2. Bring the "add an app to a running tenant" case into the `change` phasing
   question in §15 explicitly, distinguishing *augment* from *supersede*; the
   existing entry names quantity change and plan change only.
3. Resolve the add-on gap: line-level selection, or remove the add-on clause
   from predicate (b).
4. Confirm or change the `overlapScopeKey` default.

rostm:
## Flow: trial provisions first, contract and paid terms come later

The commercial flow we need supported end to end:

**deal → trial order → quote → contract → sales order (trial → paid)**

Deal registration and quotes are out of scope for this PRD and that is fine — we
will define them separately, with deal relating to quote and deal relating to
contract. What is *in* scope should not contradict this flow.

### What the flow does mechanically

1. A **trial order** is placed with no contract behind it, carrying what the
   buyer wants to evaluate.
2. Fulfilling it **provisions the tenant** and creates the **trial
   subscription**. The buyer now has a working environment to evaluate.
3. While the trial runs, the commercial track proceeds in parallel — the quote is
   negotiated and the contract is signed.
4. A **sales order converts that same subscription** from trial terms to paid
   terms. The tenant is not re-provisioned, the subscription is not replaced, and
   the buyer does not migrate anything. What changes is the commercial basis:
   trial plan/price becomes paid plan/price, under the contract signed in step 3.

The property the PRD currently cannot express is step 4: an order that targets an
**already-existing subscription and tenant** rather than spawning new ones.

### 1. The category enum has no conversion, and `new_sale` excludes it by definition

Glossary *Order Category*: "`new_sale` (this phase — **every line spawns a new
subscription**) or `change` (declared, phased — a line targets an existing
subscription for a commercially initiated **quantity/plan change**)."

The conversion order in step 4 fits neither:

- As `new_sale`, the definition forces a *new* subscription, which is exactly
  what must not happen — the tenant and subscription already exist and must
  survive. It also fails the gate: predicate §6.1 (f) evaluates the overlap rule
  "against existing subscriptions" at default cardinality 1 per
  `(payerTenantId, catalogSubscriptionProductKey)`, and the live trial
  subscription sits on that key, so the order is rejected.
- As `change`, it is declared-and-rejected until the phased path ships (§1.1),
  and the linkage named for it is `supersedesSubscriptionId` — *supersede*, where
  what is required is continuity of the same subscription and tenant with new
  commercial terms.

Ask: recognise trial-to-paid conversion as a distinct commercial transaction — an
order that binds new plan/price references and a new price pin to an existing
subscription without replacing it. Whether that is a third `category` or a
widening of `change` from supersede to rebind is a design call, but the current
enum admits neither.

### 2. Trial conversion is classified as system-driven — in this flow it is not

§1.1: "System-driven transitions (renewal, **trial conversion**, dunning-driven
suspension) are clockwork, stay with Subscriptions, and produce no order." The
same classification is repeated in §1.2, the §2 Insertion point note, §5.2 Out of
Scope, and §14 Assumptions.

That is true of auto-conversion on trial expiry with payment on file. It is not
true here: the buyer evaluates in order to decide, and the conversion is the
moment of commitment — quoted, contracted, and approved. It is commercially
initiated by definition, and it is the single highest-value transaction in the
flow.

The PRD's own argument in §1.2 applies to it directly: without an order artifact
there is "no price pin, no approval gate, and no booking record". A trial
becoming a paid commitment needs all three.

Ask: split the classification rather than remove it. Auto-conversion on expiry
stays clockwork with Subscriptions; **buyer-decided conversion is commercially
initiated and produces an order**. Five sections currently state the
undifferentiated version.

### 3. "Contract → Order → Subscription" is stated as canonical, but the trial order precedes the contract

§1.3 goals: "A new acquisition path integrates into the existing BSS monetization
sequence (**Contract → Order** → Subscription → Rating → Billing)". §2 Normative
alignment reinforces it: the manifest records "Order as a commercial artifact
**placed under a Contract**".

In this flow the trial order runs before any contract exists, and provisions real
resources while doing so. Contract → Order is not the canonical order of events.

The requirements themselves already allow this, which is why I read the sequence
statement as a framing error rather than a design one:

- Glossary *Order*: "An order **MAY** be placed under a Contract — where none is
  referenced, platform defaults govern commercial terms."
- §6.1 Order Creation: "An order **MAY** reference an active contract
  (`contractId`); where none is referenced, **platform defaults** govern
  commercial terms — uncontracted subscriptions on platform defaults are a
  first-class downstream state."

The plumbing supports trial-first; the narrative does not — and the narrative is
what downstream readers and the manifest alignment will enforce.

Ask: restate the sequence so the contract is optional and may attach before or
after the order — `Order → Subscription → Rating → Billing`, with Contract as an
optional governing artifact. Contract-first and trial-first are both valid entry
points.

### 4. Renewal has the same shape as conversion

Renewal is classified system-driven in the same five sections, but a renewal can
run **renewal deal → quote → order**, with re-negotiated terms — commercially
initiated, not a timer.

§6.1 Line Dates blocks the artifact outright: "**Renewal price** is likewise
governed by contract and the spawned subscription — never by the order; the
`catalogPricePin` is the acquisition pin, not a renewal-price election. The line
MUST NOT become a second authority for either."

If a re-quoted renewal is placed as an order, that order *is* the pricing
authority for the new term. As worded, a commercially initiated renewal is
unrepresentable.

Ask: apply the same auto-versus-negotiated split as trial conversion, and allow a
commercially initiated renewal order to carry its own price pin for the new term.

### 5. On the closed quote question

§15 closes it as "no quote artifact planned", reasoning that "CPQ, if it arrives,
is a front-end producing a draft order, **not a new SoR**".

As a scope statement for this PRD that is fine. As a platform-wide statement it
pre-empts work that will be defined separately. Two consequences worth
reflecting in the text:

1. **The trial order and the quote coexist in time**, not in sequence — the
   trial is live while the quote for the paid commitment is negotiated. The
   order should not be specified as though it were the only pre-commitment
   artifact.
2. **The order needs first-class references to the deal and quote** behind it.
   The existing per-line/per-order *external reference* is not the right home:
   §6.1 scopes it to buyer-side accounts-payable reconciliation and requires it
   to "propagate to billing documents". Deal and quote linkage is seller-side
   commercial provenance and belongs in its own field.

Ask: reword the §15 closure as scope-limited to this PRD rather than
platform-wide, and add optional `dealId` / `quoteId` references to the order
document.

## Summary of asks

1. Make trial-to-paid conversion expressible: an order that rebinds plan/price
   and pins a new price on an **existing** subscription, without spawning or
   superseding it. The current `new_sale` / `change` enum admits neither.
2. Split trial conversion into auto (clockwork, no order) and buyer-decided
   (commercially initiated, produces an order) across all five sections that
   state it undifferentiated.
3. Restate the canonical sequence with the contract optional and attachable after
   the order; trial-first and contract-first are both valid entry points.
4. Resolve the overlap-rule treatment of trial and paid on the same product key
   during conversion.
5. Apply the auto-versus-negotiated split to renewal, and let a commercially
   initiated renewal order carry its own price pin for the new term.
6. Reword the §15 quote closure as scope-limited, and add optional `dealId` /
   `quoteId` provenance references to the order.

rostm (commented):
## Quantity model: the order cannot express a soft or a hard limit

Quantity on a usage-bearing line needs two bounds, not one: a **soft limit**
(what the buyer is entitled to, and what gets displayed) and a **hard limit**
(where consumption is actually stopped). The commercial modes we need are the
three useful configurations of that pair:

| Mode | Soft | Hard | Behaviour |
|---|---|---|---|
| **PAYG** | none | none | Unlimited. Meter everything, bill it. |
| **Fixed** | N | N (= soft) | Cannot exceed what was purchased. Consumption stops at N. |
| **Mixed** | N | none, or M > N | N is shown as the limit; going above it is permitted and charged as overuse. |

Prepaid falls out of the same pair: soft = hard = purchased balance is Fixed
behaviour at zero; soft = balance with auto-recharge above it is Mixed.

The order line carries neither bound. It has `qty`, a floor, and an explicit
refusal to model any ceiling — so **mode is not a property the order can
express**. It ends up implicit in catalog configuration and quota configuration,
in two different subsystems, and invisible on the commercial document the buyer
signed.

### What is specified today

- The line carries `qty`, and "one fulfilled line item spawns **exactly one**
  subscription carrying the line's `qty`".
- Gate predicate §6.1 (b) enforces a **floor** only: `qty` "satisfies the
  purchase-quantity floor where the price row declares one (`minQtyThreshold` of
  type `purchase`)".
- Ceilings are refused and delegated: "there is **no plan-level maximum**: upper
  bounds are **resource quotas** owned by the quota/policy subsystem and enforced
  at fulfillment, not at the gate."

### 1. Mode by mode, what breaks

**PAYG** — works only by absence of bounds, and even then `qty` has no defined
meaning on a usage line (the PRD never says what to put there). There is no way
to *declare* unlimited, so PAYG is indistinguishable from an unbounded line that
should have had a cap and didn't.

**Fixed** — not expressible. `qty` records what was purchased but the PRD never
states that `qty` is also the enforcement ceiling; the relationship between
purchased quantity and enforced limit is simply unstated. The only ceiling
mechanism in the catalog is a closed top tier band, and exceeding it "**MUST**
fail closed at evaluation (no free, no clamp)" — a **rating failure**. So the
simplest and most common mode is achievable only as an error condition, or by an
undocumented handoff of `qty` into the quota subsystem.

**Mixed** — the catalog already has the pricing machinery and the order cannot
feed it. `reservedRate` with `reservationFlavor = consumption` is precisely this
mode: "matched usage at the reserved rate, remainder on-demand". But plan-price
states "the reserved/allocated quantity is supplied **at runtime**". So the soft
limit — the number that defines the deal — is a runtime input rather than an
order field, and the line carries no overage price reference either.
(`reservationFlavor = capacity`, "allocated quantity billed regardless of usage",
is the Fixed variant and has the same problem.)

### 2. A commercial cap is not a resource quota, but that is where it has been sent

The hard limit in Fixed and Mixed is a negotiated commercial term. By this PRD's
own §1.2 argument it needs a price pin, an approval gate, and a booking record —
which is what an order provides. Routing it to the quota/policy subsystem (a Core
Functionality gear per `docs/GEARS.md`) turns it into infrastructure
configuration: absent from the order, outside the approval path, owned by people
who do not negotiate deals.

A resource quota answers "how much may this tenant consume before we protect the
platform". A commercial hard limit answers "how much did this customer agree to be
billed for". Both are upper bounds; only the first belongs in the quota subsystem.

### 3. There is no soft-limit concept anywhere in the chain

Nothing in the order, the catalog, or Orders Workflow expresses a soft limit or a
crossing of one. The two nearest constructs are both something else:

- `minQtyThreshold` of type `usage` is a pricing **eligibility** floor that must
  "**fail closed** to the plan's defined fallback".
- The closed top tier band is a hard stop that surfaces as an evaluation error.

Mixed mode depends entirely on soft-limit crossing being a normal, observable
event — it is the moment the rate changes and the moment the buyer wants to know.
Today it is neither representable nor emittable.

### 4. Mixed mode is also foreclosed by a glossary statement

Resolved Order Total, on the usage charge kind: "**no committed amount**;
excluded from the total, flagged, priced at rating time".

In Mixed mode the soft limit *is* a committed amount — it is what earns the rate,
what the contract states, and what approval evaluates. The PRD declares it
absent.

Second-order consequence: "the single named figure exposed to approval policy is
the net pre-tax total-contract-value (TCV)", and usage is excluded from it. A
large Mixed-mode deal presents a TCV near zero to the approval policy owner.

### 5. "Includes N units" is not quotable

plan-price notes an included allowance is "representable today as a `$0` first
band — mathematically correct, but **display/quoting cannot label it 'includes N
units'** and reporting cannot separate included from billed consumption", with a
first-class `includedAllowance` (and `rolloverPolicy`) deferred to Future scope.

That is the soft limit again, from the catalog side: even where the entitlement
exists mathematically, nothing can display it as the limit — which is what Mixed
mode requires ("soft shown as the limit").

### 6. Changing either bound mid-term has no order

Raising a soft limit (a bigger commitment) or a hard limit is a quantity change,
and `category = change` "**MUST** be rejected until its path ships". For Fixed and
Mixed that is the routine motion.

### 7. Only add-ons get a maximum

Add-on rules carry "min/max/step quantity" and a required add-on "**MUST** have
`maxQty >= 1`". Base plans explicitly get no maximum — so the attachment can be
bounded at purchase and the thing it attaches to cannot.

## Asks

1. Add a **soft limit** and a **hard limit** to the order line as first-class
   fields, each independently nullable, so the three modes are expressible as
   (none, none), (N, N), and (N, none | M).
2. Add an **overuse price reference** on the line for the Mixed case, so the rate
   above the soft limit is pinned at purchase like any other price.
3. Define `qty` semantics per charge kind, and state its relationship to the soft
   limit — whether `qty` *is* the soft limit for a Fixed line, or a separate
   figure.
4. State that a commercial hard limit is distinct from a resource quota and keep
   it in the commercial chain, rather than delegating all upper bounds to the
   quota subsystem.
5. Make **soft-limit crossing an event** consumers can act on — Mixed mode
   depends on it — and keep it distinct from the closed-top-band rating failure.
6. Revise the "no committed amount" statement for usage, since the soft limit is
   a committed amount in Mixed mode.
7. Include the committed portion in TCV, or expose a second named figure, so
   Mixed-mode deals are visible to approval.
8. Say where the reserved/allocated quantity for `reservationFlavor` is captured
   at purchase, given the catalog prices against it but sources it at runtime.
9. Confirm whether a purchase-time maximum for base plans is intentionally absent,
   given add-on rules have one.

Corw1n-of-Amber:
Thanks — this is a careful read and the quotes are accurate. Four of the points were self-contradictions in this PRD and are fixed in the latest commit; the rest are recorded as Open Questions rather than settled unilaterally.

**Fixed**

- **Add-on gap** — agreed, and a real one: gate predicate (b) validated bounds the line model cannot express. Taken the second option offered — the add-on clause is removed from the gate, add-on selection at order capture becomes an explicit Out of Scope entry, and an Open Question carries whether the line should gain it.
- **Contract-first framing** — agreed, and a framing error rather than a design one. The sequence now reads `Order -> Subscription -> Rating -> Billing` with Contract as an optional governing artifact that MAY be referenced at submit or attached later. Contract-first and uncontracted entry points are both valid.
- **TCV and usage** — agreed and worth stating plainly. The glossary now says TCV is not total deal value where a line is predominantly usage-based, and that the approval policy owner MUST NOT read it as deal size.
- **Stale canon** — §17 pinned the sibling gears to a pre-merge checkout; repointed in-repo.

**Recorded as Open Questions**

- **Trial conversion and renewal classification.** Agreed on the substance: auto-conversion on expiry is clockwork, while a buyer-decided conversion after a quote is commercially initiated by this PRD's own §1.2 test. Worth adding that the downstream mechanism already exists — trials are a plan *phase* and `convertTrial` is a first-class `TransitionRequest`, "the phase twin of `changePlan`", so conversion advances a phase on the **same** subscription rather than spawning or superseding one. That also answers the overlap question: there is no collision at conversion, because there is one subscription throughout. Third `category` versus widening `change` from supersede to rebind is a product call and is now tracked with an owner.
- **Quantity model.** The distinction between a negotiated commercial cap and a resource quota is fair, and the consequence — a negotiated term sitting outside the order, the price pin and the approval path — is real. Tracked, along with `qty` semantics per charge kind and an overuse price reference. One scoping note: soft-limit crossing as an event is a rating/metering concern rather than an order one, so that part is unlikely to land here.
- **Subscription composition granularity.** The multi-product entitlement-set model is a subscription-aggregate question, as noted. Tracked here, but resolved in the Subscriptions gear first and then reflected — the 1:1 rule is inherited, not asserted.
- **Deal and quote provenance.** The §15 closure is scope-limited to this PRD and now says so; optional `dealId` / `quoteId` provenance is tracked separately from the external reference, which is buyer-side AP reconciliation and the wrong home for it.

**Not changed here**

- **`overlapScopeKey` default.** The partner-collision catch is correct and a good one. The default is adopted by reference from the subscriptions gear, which itself leaves overlap *dimension* binding open to its Design — so changing it locally would fork from the canonical gear and create exactly the drift this PRD exists to prevent. Raised upstream and tracked instead.
- **`includedAllowance`.** That quote is from a superseded revision. The pricing gear now carries `includedAllowance` as first-class, replacing the `\`-band-only stance, with "includes N units" display and included-versus-billed reporting separated.

PR #4649, opened 2026-08-21
Add Orders Workflow PRD

Corw1n-of-Amber:
## Summary

Adds the Orders Workflow PRD at `gears/bss/orders-workflow/docs/PRD.md`

**In scope:** process orchestration for commercially initiated orders — approval-requirement transport and gate execution via Generic Approval, two-wave fulfillment (draft-create then activation), saga/compensation, retries, manual tasks, process events, and a by-reference binding to Lifecycle R1–R5.

**Out of scope:** the order document and authoritative order state (Orders Lifecycle); price math (rating / Plan & Price); direct OSS provisioning (Subscriptions only); a module-local approval policy owner.

## Validation

- [x] `cfs toc gears/bss/orders-workflow/docs/PRD.md`
- [x] `cfs validate --artifact gears/bss/orders-workflow/docs/PRD.md --skip-code --local-only --verbose` — **0 errors / 0 warnings**


dogbert2021:
From the CT Depot perspective, the Lifecycle↔Workflow ownership split and the two-wave/all-or-nothing saga are clear.

CT Depot channel facade →
  → CF Orders Lifecycle ↔ CF Orders Workflow
  → CF Subscriptions Lifecycle ↔ CT Depot OSS Subscription Manager specialization
  → CT OSS App Fulfillment
  → CT runtime App owners

The remaining **integration gap is the Workflow→Subscriptions contract**. 
Please require **draft-create, activation, draft-void and activated-cancel intents** and confirmations to **preserve orderId, orderVersion, orderLineId, wave, process correlationId, and an opaque caller-owned commercial-binding reference**. Please also include orderVersion in the idempotency identity; using only order ID + line + wave could collide after an amendment.

This lets Depot OSS Subscription Manager pin its exact commercial projection and OSS contract without leaking CT-specific internals into CF.  CT Depot OSS Subscription Manager and CT OSS App Fulfillment are CT consumers behind the generic CF Subscriptions lifecycle: draft create/void has no OSS effects, while activation/cancel uses pinned CT App Fulfillment operations. This prevents duplicate orchestration and a second order source of truth.

Finally, currently BSS roadmap still leaves Order Management [unscheduled](https://github.com/orgs/constructorfabric/projects/48/views/1?filterQuery=Order) while Depot has a first aggregators channel orders adapter target Q1'27 (f.ex. Ingram Micro purchase orders). 

**Please record this as a CT Q1'27 delivery dependency**, so CT Depot does not build a temporary replacement order engine.

Corw1n-of-Amber:
Thanks — agreed, and this was a real hole in the Workflow→Subscriptions contract.

Approval requests were already keyed on `orderId` + `orderVersion`. Provisioning intents were not, so an amendment could collide with a superseded submit. The latest commit requires `orderId`, `orderVersion`, `orderLineId`, wave, process `correlationId`, and an opaque caller-owned binding reference on all four intents (draft-create, activation, draft-void, activated-cancel), includes `orderVersion` in the idempotency key, and requires confirmations/failures to echo that identity. The binding reference is opaque to this gear and to Subscriptions — it is not interpreted as order state.

The opaque binding reference is the generic pass-through. Pinning a commercial projection or App Fulfillment operations stays behind Subscriptions (R3) and is not specified in this PRD.

The Q1'27 Depot / channel-adapter timeline is noted as a delivery dependency. Dates and adopter commitments are not recorded in the PRD.

rostm (commented):
### The gap, concretely: a buyer purchases with a card via online shop

Trace a self-service card checkout through the two PRDs. The buyer picks a plan,
enters a card, and expects service if the charge succeeds and nothing if it fails.

**Where it breaks:**

1. **Payment is checked after the order is approved, and only as an
   authorization.** The order reaches `approved` before payment is looked at;
   the check then gates `approved → in_fulfillment`. Stripe's normal SaaS flow
   authorizes *and captures* at checkout. There is no capture step, and nowhere
   to record the `PaymentIntent`.

2. **Money moves in the opposite order.** At-sale money is "emitted by
   Subscriptions at activation and posted by Billing" — provision first, charge
   after. A card checkout charges first and provisions only on success. The
   supported ordering is the inverse of the scenario.

3. **3DS / SCA has no home.** A European card triggers a challenge: redirect,
   buyer authenticates, return. That is an async payment state lasting seconds
   to minutes, with a user-facing wait. But §5.1 states "no `payment_pending`
   order state", and process state "**MUST NOT** serve as a source of truth about
   order semantics" — so nothing the buyer's browser can read tells them where
   they are.

4. **A declined card leaves the order in `approved`, silently.** Workflow: "on
   failure without the seller tolerate-failure policy, begin-fulfillment **MUST
   NOT** be called (order stays `approved`)". Lifecycle's eleven events include
   none for payment, and the process events here are approval- and
   fulfillment-shaped. So a failed checkout is indistinguishable from an order
   about to fulfil.

5. **Retrying with another card has no operation.** Lifecycle §9.1 lists thirteen
   operations; none re-triggers payment. `Amend` creates a new version and re-runs
   the sellability gate — heavy, and payment is not order content anyway.

6. **The only exit is expiry.** The stuck order eventually hits the `approved`
   TTL and becomes `expired` via `OrderExpired` — the same terminal as an
   enterprise deal that went quiet, with nothing recording that a card was
   declined.

7. **If fulfillment fails after capture, the reversal is the wrong artifact.**
   Compensation reverses posted facts with "a Billing credit note /
   ChargeAdjustment". A captured Stripe charge needs a **refund** through Stripe —
   different artifact, different timing, different SoR.

8. **The buyer paid a different number than the order records.** Stripe Tax
   computes at checkout and the buyer approves a tax-inclusive total. Lifecycle:
   "Tax is not included ... explicitly pre-tax", and Preview's indicative tax
   "**MUST NOT** be stored on the order".

9. **Later Stripe webhooks have nowhere to land.** Disputes, refunds, and
   asynchronous payment confirmations arrive after the order is terminal. The
   trigger set is closed and Lifecycle-sourced, and Payments "**MUST NOT** drive
   order state transitions directly".

**What this adds up to:** the flow supports one payment model — invoice after
provisioning, with authorization as a risk check. A card purchase is not a
variant of that; it inverts it. Nothing here is a tuning problem.

**Minimum to close it:**

- A declared payment ordering (collect-then-provision vs provision-then-collect),
  chosen per instance or per order rather than implied by where billable facts
  are emitted.
- Payment outcome visible on the order — at least awaiting, failed, captured —
  since three different audiences need to see it and process state is explicitly
  not authoritative.
- A payment-failure event and a re-authorise/re-capture operation, so a declined
  card is recoverable without amending the order.
- Capture semantics, including refund as the reversal path when fulfillment fails
  after capture.
- An inbound path for post-terminal payment events (dispute, refund,
  async confirmation).
- A stored tax-inclusive total where the buyer commits against one.

Corw1n-of-Amber:
Thanks — the card-checkout trace is accurate, and the conclusion is agreed: this is not a tuning gap. The flow supports provision-then-collect with authorization as a risk check, and a checkout that collects first inverts it.

Rather than specify a Payments capability inside an Orders PRD, the latest commit does three things:

- **Payment collection is now an explicit Out of Scope entry** — capture, settlement, refunds, chargebacks and disputes, SCA challenges and their asynchronous return, retry with an alternative instrument, and PSP webhooks — with the consequence written into the text rather than left implied: only one payment ordering is expressible here.
- **Four Open Questions with named owners** carry the decisions: payment ordering as a declared per-instance or per-order property rather than an implicit consequence of where billable facts are emitted; the minimum payment-outcome surface (order-visible outcome, failure event, re-authorize/retry) and whether any of it belongs on the order document at all; refund versus credit note as the reversal path after capture; and whether post-terminal payment events need an inbound path here.
- **§16 records the limitation as a risk**, including the specific consequence raised — a declined instrument is currently indistinguishable from a healthy order and exits only by expiry.

Closing those needs a Payments capability spec, which does not exist in this repository today; this PRD will not invent one.

The tax-inclusive point is noted and belongs with the same decision: where a buyer commits against a tax-inclusive figure, something has to store what was agreed to, and that is a billing-chain and order-document question together rather than a workflow one.

PR #4655, opened 2026-08-24
feat(bss-pricing): pricing gear — money, catalog model and governed publish

diffora:
## What this gear is

`bss-pricing` authors, validates, governs and freezes price catalogs. **It computes no charge**: the arithmetic that turns a price row into an amount owed lives downstream (rating/evaluation and billing). This gear owns what a price *is*, that it is *valid*, and that once published it *never silently changes*.

## Money

- **Amounts are integer minor units** (`amountMinor`) — cents, not dollars. Never a float, never a decimal type: integers cannot accumulate rounding error.
- **The scale is per-currency, per ISO 4217** — there is no flat two-decimal rule. JPY takes 0 decimals, USD 2, BHD 3. Declaring a scale the currency cannot express is a first-class publish rejection (`PRECISION_EXCEEDED`), never a silent rounding.
- **Rates are a separate type at a separate scale**: a rate is a multiplier, not an amount, so it must be finer than the coin. Stored in 10⁻⁹ minor units (`unitRateNanoMinor`, `unitPriceNanoMinor`) — the `Nano` in the name is deliberate, so a reader can never mistake one scale for the other by a factor of 10⁹. Rounding happens once, on the multiplied result, downstream.
- **No add/subtract/multiply/round in the money module at all** — only the checks a publish needs. Charge arithmetic must have exactly one implementation, and it is not here.

## The catalog model

- **A price is addressed by a ten-axis canonical scope key** (plan, currency, region, overlay, phase, eligibility, charge kind, cohort, meter, dimension key). One priced line per `(meter, dimensionKey)` per slice — a duplicate key is an authoring fault, not a merge.
- **Published content is frozen.** The read model is INSERT-only and resolves through pinned catalog versions; a consumer that read a version once reads the same bytes forever (a 7+ year horizon). Corrections are new versions, never rewrites.
- **Pricing models**: flat, per-unit, graduated/volume tier bands, package (block pricing), plus phases, overlays (price lists as line containers), bundles with revenue-share, and included allowances that compile to an equivalent $0-band ladder.

## Validation & governance

- **Every rule refuses at the earliest stage that can judge it**: write-stage when the fault can never become legal on that key, publish-stage when a later edit could still resolve it. A 422 does not exist on this surface — every rejection is an RFC-9457 problem carrying a stable machine code.
- **Reports enumerate every violation, not the first** — an author remediates in one pass.
- **Publishing is governed**: two-person approval (submitter ≠ approver, enforced against a pinned content hash so what was approved is what ships), append-only hash-chained audit trail per aggregate, and DB triggers that make decided rows physically immutable.
- **Optimistic concurrency everywhere**: mutations require `If-Match`; the version bump happens in SQL under the row lock, never computed by a caller.
- **Idempotency as a first-class door**: client keys with at-most-once semantics; a replay returns the recorded status and body verbatim and executes nothing.
- **Sellability is fail-closed**: a six-predicate conjunction; an unanswerable predicate reports *who owes the answer* (`NotEvaluable { owed_to }`) rather than defaulting to sellable.
- **Price windows only open in the future** and may not strand a trailing coverage void; window history is immutable once elapsed.

## What is deliberately not here

- **No charge arithmetic.** No add, multiply, allocate or round on money anywhere in the gear — rating computes, billing invoices; this gear only authors and freezes what they read.
- **No product registry.** SKU existence, the registry `sellable` flag and GA gates live in the products-registry gear (not in this repo). Sellability doesn't fake those answers — it reports `NotEvaluable { owed_to: … }` naming exactly which gear owes the missing fact.
- **No tax computation.** A row carries `tax_category_ref` and a tax-inclusivity declaration; resolving them into an amount is downstream.

## Storage & testing

- **Two engines, one schema**: SQLite for the fast tier, Postgres in production. Schema equivalence is held by golden dumps of both engines; the migration chain is one migration per table, proven byte-identical against those goldens.
- **Every DB invariant is physical as well as typed** — CHECKs, exclusion constraints and triggers exist even where the only writer is already correct by construction.
- Gates on this branch: workspace clippy clean, fast tier 3349 tests across 120 binaries, Postgres tier 411/411, e2e suite against a live stand 125 passed / 2 skipped.


MikeFalcon77:
It's pretty imspossible to check ~400K lines. So we just merge it on green CI and pretty good coverage

PR #4667, opened 2026-08-27
feat(cluster): support cluster gear as a OOP deployable gear

asmith987:
## Summary

**New Features**
- The `cluster` gear can now be deployed **out of process** as its own gRPC service, not just embedded in-process.
- A single `ClusterClient` per process transparently routes coordination calls to a local backend or a remote cluster pod — **consumer code is identical in both topologies**.
- Added gRPC services for all cluster primitives (cache, lock, leader-election, profile) plus health/identity, with subscription streaming and server-side sweep.
- New `register_cluster_profile!` declaration lets a process name the profiles it expects, so a missing binding surfaces at startup instead of at the first coordination call.
- Cluster wiring now owns a lifecycle handle (`build_and_start` → `handle.stop()`), registering all backends atomically and running plugin shutdown hooks on teardown.

**Framework / toolkit**
- Added a gRPC projection and proto-bridge for contract types (`toolkit-contract` macros) so domain DTOs map to/from protobuf at the wire boundary.
- Postgres cluster plugin reworked for the remote seam (store-owned leases, lock/reaper changes) per the new ADRs.

**Bug Fixes / Safety**
- Resolution now defers descriptor validation on a cold start rather than blocking on cluster reachability (startup never hangs if the cluster pod isn't up yet).
- Capabilities are enforced at the descriptor boundary, and wire inputs (cache keys, election/profile names) are validated at the gRPC edge.
- Backend registration is atomic — a partial failure can't leave a half-registered hub.

**Documentation**
- New ADRs (remote-backend seam, store-owned leases, async-boundary rules), a rewritten DESIGN.md, and a consumer wiring README.

---

## What changes for a consumer

Almost nothing — this is mostly a **deployment/config** change. The same resolve code runs whether the cluster gear is embedded or a remote pod. Two source touch-points:

**1. `resolve()` is now `async`:**
```rust
// before
let cache = ClusterCacheV1::builder()
    .profile(AppProfile)
    .require(CacheCapability::Linearizable)
    .resolve()?;

// after — add .await
let cache = ClusterCacheV1::builder()
    .profile(AppProfile)
    .require(CacheCapability::Linearizable)
    .resolve()
    .await?;
```

**2. Declare the profiles the process expects (one line, optional but recommended):**
```rust
struct AppProfile;
impl ClusterProfile for AppProfile {
    const NAME: &'static str = "app";
}

// tells cluster's readiness contributor which profiles this process needs,
// so an unbound profile is reported at startup, not at first use.
cluster_sdk::register_cluster_profile!(AppProfile);
```

**That's it for source.** Going from embedded to a deployed cluster pod requires **no consumer code change** — a process that doesn't link the `cluster` gear automatically resolves over gRPC instead of a local hub. Which topology you get is an operator/wiring concern, not a code one.


striped-zebra-dev (approved):

libs/toolkit-contract-macros/src/proto_bridge.rs
```diff
@@ -136,13 +154,190 @@ fn parse_field_conversion(field: &Field) -> syn::Result<FieldConversion> {
...
+        // `message` adds nothing when the field is already `Option`: the stub is
+        // one `Option` deep either way.
+        (FieldConversion::Direct | FieldConversion::Message, true) => {
+            quote! { v.#field_ident.map(::std::convert::Into::into) }
+        }
+        (FieldConversion::ViaString, false) => {
+            let msg = format!("proto bridge: invalid string for field `{field_ident}`");
+            quote! {
+                <#inner_ty as ::std::str::FromStr>::from_str(&v.#field_ident)
+                    .expect(#msg)
```
MikeFalcon77: **MEDIUM**

The generated infallible `From<Proto> for Rust` impl for a `#[proto_bridge(via_string)]` field calls `.expect(...)`, so decoding a malformed string from an untrusted gRPC peer panics the process when a caller uses `.into()`/`From::from()` instead of the fallible path.

A single malformed wire field can crash the process, since the panicking conversion is reachable through the ordinary `.into()` API right alongside the fallible one.

Drop the panicking `From<Proto>` arm for `via_string` fields, and route all via_string decoding exclusively through the fallible `TryFromProto`/`try_from_proto` path.

libs/toolkit-contract-macros/src/proto_bridge.rs
```diff
@@ -136,13 +154,190 @@ fn parse_field_conversion(field: &Field) -> syn::Result<FieldConversion> {
...
+            }
+        }
+        (FieldConversion::ViaString, true) => {
+            let msg = format!("proto bridge: invalid string for field `{field_ident}`");
+            quote! {
+                v.#field_ident.map(|s| {
+                    <#inner_ty as ::std::str::FromStr>::from_str(&s)
+                        .expect(#msg)
```
MikeFalcon77: **MEDIUM**

Same panicking `.expect(...)` pattern as the non-optional case, generated here for `Option<T>` via_string fields inside the infallible `From<Proto> for Rust` impl.

A malformed string in an optional field still aborts the receiving process if code uses `.into()` instead of `try_from_proto`.

Apply the same fix as the non-optional branch: remove the panicking path from the infallible `From` impl for via_string fields.

gears/system/cluster/cluster/src/domain/registry_tests.rs
```diff
@@ -0,0 +1,195 @@
...
+    // thing that changed.
+    let (handle, bound) = wire(&["event-broker"]).await;
+    registry.publish(bound);
+    assert!(registry.resolve("event-broker").is_ok());
+
+    handle.stop().await;
+}
+
+#[tokio::test]
+async fn publish_increments_the_generation_on_every_swap() {
```
MikeFalcon77, line 62: **MEDIUM**

`ProfileRegistry::publish`'s doc comment states the generation bump happens under `ArcSwap::rcu` specifically so concurrent publishes get distinct generations, but every test here calls `publish()` sequentially from a single task.

The documented concurrent-CAS-retry guarantee is unverified, so a regression to a naive load-then-store race would not be caught by this suite.

Add a test that spawns multiple concurrent tasks calling `ProfileRegistry::publish()` on a shared registry and asserts the resulting generation count equals the number of publishes with no duplicates or gaps.

libs/toolkit-contract-macros/src/proto_bridge.rs
asmith987: Fixed. I dropped the panicking `From<Proto>` path for `via_string`. All decoding routes exclusively through `try_from_proto`/`TryFromProto`.

libs/toolkit-contract-macros/src/proto_bridge.rs
asmith987: Fixed alongside the non-optional case. The fallible path is now the only decode route

gears/system/cluster/cluster/src/domain/registry_tests.rs
asmith987: Added a new test that spawns multiple tasks and asserts the publish count.

PR #4674, opened 2026-08-27
feat(authz-resolver)!: migrate to toolkit contract with REST projection

mattgarmon:
Convert the `authz-resolver` from a hand-rolled client trait to a transport-agnostic `#[toolkit::contract]`, and update all PEP consumers. This lets out-of-process gears reach the PDP over REST via directory resolution instead of embedding the full authz stack (resolver + plugins) in every pod.
 
Changes:
- Rename `AuthZResolverClient` -> `AuthZResolverApi` (the `Api` suffix is the `#[toolkit::contract]` convention marking a remote-capable provided contract); `evaluate` now takes a `PlatformSecurityContext` (a compile-time platform-plane marker, not a tenant identity) and returns `CanonicalError`.
- Add the REST projection (rest module + internal, non-edge-exposed `/authz-resolver/v1/evaluate`) and wire local + rest transports via `#[toolkit::provides]`; DTOs gain JsonSchema/ToSchema.
- `PolicyEnforcer`/`EnforcerError` updated to the new signature/error type; add `PolicyEnforcer::from_hub`, a per-call PDP deadline (`PolicyEnforcer::with_deadline` / `DEFAULT_EVAL_DEADLINE` = 30s), and a REST round-trip test.
- Gate the generated REST client/server behind SDK cargo features (`rest-client` / `rest-server`) so consumers opt into transport.
- Update every PEP consumer to pass a `PlatformSecurityContext` and use the renamed contract; docs + dependency cleanups.
 
BREAKING CHANGE:
- `AuthZResolverClient` is renamed to `AuthZResolverApi`
- `AuthZResolverApi::evaluate` now requires a `PlatformSecurityContext` argument and returns `CanonicalError` instead of `AuthZResolverError`. `PolicyEnforcer::new` and `EnforcerError::EvaluationFailed` change accordingly.

## New Features
* Introduced the transport-agnostic `AuthZResolverApi` with optional REST support.
* Added lazy resolver discovery and configurable authorization evaluation deadlines.
* Added authenticated authorization evaluation over REST.
* Standardized evaluations around security context and canonical errors.
* Added OpenAPI and JSON Schema support for authorization models and endpoints.

## Documentation
* Updated authorization guides, architecture references, examples, and integration plans.

## Tests
* Added REST round-trip coverage and expanded timeout, wiring, and authorization scenarios.


gears/system/authz-resolver/authz-resolver/src/domain/local_client.rs
```diff
@@ -23,20 +23,86 @@ impl AuthZResolverLocalClient {
     }
 }
 
-fn log_and_convert(op: &str, e: DomainError) -> AuthZResolverError {
+/// Map an infrastructure `DomainError` onto the contract's `CanonicalError`.
+/// Access denial is never surfaced here — it rides in `EvaluationResponse`.
+fn log_and_convert(op: &str, e: &DomainError) -> CanonicalError {
     tracing::error!(operation = op, error = ?e, "authz_resolver call failed");
-    e.into()
+    CanonicalError::internal(e.to_string()).create()
```
MikeFalcon77: **HIGH**

`log_and_convert` maps every `DomainError` variant (`PluginNotFound`, `PluginUnavailable`, `InvalidPluginInstance`, `TypesRegistryUnavailable`, `Internal`) to `CanonicalError::internal`, discarding the semantic distinction the previous `AuthZResolverError` (`NoPluginAvailable`/`ServiceUnavailable`/`Internal`) preserved.

Now that `evaluate` is reachable over the new REST projection, a transient no-plugin-available condition surfaces as HTTP 500 instead of a retryable 503, and callers lose the ability to distinguish transient unavailability from a real internal bug.

Match on the `DomainError` variant and map `PluginNotFound`/`PluginUnavailable`/`TypesRegistryUnavailable` to `CanonicalError::service_unavailable()`, reserving `CanonicalError::internal()` for `InvalidPluginInstance`/`Internal`.

gears/system/authz-resolver/authz-resolver-sdk/src/rest.rs
```diff
@@ -0,0 +1,32 @@
+//! REST projection of [`AuthZResolverApi`].
+//!
+//! Carries the HTTP method/path annotations consumed by
+//! `#[toolkit::rest_contract]`. When the `rest-client` feature is enabled the
+//! macro also emits `AuthZResolverApiRestClient` (and its directory-resolving
+//! wrapper `AuthZResolverApiRestResolvingClient`) that implement
+//! [`AuthZResolverApi`] over HTTP; when `rest-server` is enabled it emits
+//! `register_auth_z_resolver_api_rest_routes` for the gear to host.
+//!
+//! The `evaluate` route is **internal** — authenticated on the tenant plane
```
MikeFalcon77: **HIGH**

The internal `/authz-resolver/v1/evaluate` route authenticates the caller's tenant-plane bearer via `SecurityContext` but never requires that `req.subject` match the authenticated caller's identity; `AuthZResolverLocalClient::evaluate` explicitly discards `ctx` (`_ctx: SecurityContext`) and decides purely from the attacker-suppliable `req.subject`.

Any in-cluster service holding a valid tenant-plane bearer can request a PDP decision — including row-level constraints — for an arbitrary subject, resource, or tenant, turning the PDP into an unrestricted authorization-decision oracle for identities other than the caller's own.

Cross-check `req.subject.id`/tenant against the authenticated `SecurityContext` before evaluating, or require platform-plane/service-identity auth for this endpoint, so correctness doesn't depend on `PolicyEnforcer` being the only well-behaved caller.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer.rs
```diff
@@ -391,7 +436,8 @@ impl PolicyEnforcer {
         let require = request.require_constraints.unwrap_or(true);
         let eval_request =
             self.build_request_with(ctx, resource, action, resource_id, require, request);
-        let response = self.authz.evaluate(eval_request).await?;
+        let authz = self.resolve_authz()?;
+        let response = authz.evaluate(ctx.clone(), eval_request).await?;
```
MikeFalcon77: **HIGH**

`access_scope_with` awaits `authz.evaluate(ctx.clone(), eval_request)` with no PEP-specific deadline, even though this call can now be a REST round-trip to `authz-resolver` instead of the previously guaranteed-fast in-process trait dispatch.

Every authorization check across all PEP consumers now inherits only the generic REST client's default timeout, so a slow or unresponsive `authz-resolver` stalls every dependent gear's request handling with no faster fail-safe on this hot, request-gating path.

Wrap the `evaluate` call in an explicit `tokio::time::timeout`, or configure a PEP-specific client timeout tighter than the generic default, so a degraded PDP cannot cascade into blocked consumers.

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer_tests.rs
```diff
@@ -23,11 +23,12 @@ const RESOURCE: &str = "33333333-3333-3333-3333-333333333333";
 struct AllowAllMock;
 
 #[async_trait]
-impl AuthZResolverClient for AllowAllMock {
+impl AuthZResolverApi for AllowAllMock {
     async fn evaluate(
         &self,
+        _ctx: SecurityContext,
```
MikeFalcon77: **MEDIUM**

Every mock added for the new `evaluate(ctx: SecurityContext, ...)` signature (`AllowAllMock`, `DenyMock`, `FailMock`) declares the context parameter as `_ctx` and never inspects it, so no test asserts that `PolicyEnforcer` actually threads the caller's `SecurityContext` through to the resolved client.

The core new behavior this PR adds — passing `ctx` into `evaluate` — has no coverage verifying the correct context reaches the resolver.

Add a mock that captures the `SecurityContext` argument and assert it matches the `ctx` passed into `access_scope`/`access_scope_with`.

gears/system/authz-resolver/authz-resolver-sdk/tests/rest_roundtrip.rs
```diff
@@ -0,0 +1,130 @@
...
+    });
+    format!("http://{addr}")
+}
+
+fn http_client(base_url: &str) -> AuthZResolverApiRestClient {
+    AuthZResolverApiRestClient::new(ClientConfig::new(base_url.to_owned()))
+        .expect("default toolkit-http client build is infallible in tests")
+}
+
+#[tokio::test]
```
MikeFalcon77, line 125: **MEDIUM**

The REST round-trip test only exercises the `Ok(EvaluationResponse)` path; it never drives a resolver that returns `Err(CanonicalError)` through the generated server/client.

The main new failure-handling behavior introduced by migrating from `AuthZResolverError` to `CanonicalError` — how errors map across the REST boundary — is untested.

Add a test where the hosted `AuthZResolverApi` impl returns `Err(CanonicalError::...)` and assert the REST client surfaces the expected variant/HTTP status after the round trip.

gears/system/authz-resolver/authz-resolver/src/domain/local_client.rs
mattgarmon: Added mapping

gears/system/authz-resolver/authz-resolver-sdk/src/rest.rs
mattgarmon: Fixed by switching to platform-plane auth. The route now takes `PlatformSecurityContext` instead of `SecurityContext`. Tenant-plane PEPs pass `PlatformSecurityContext::outbound_marker()`

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer.rs
mattgarmon: Added deadline

gears/system/authz-resolver/authz-resolver-sdk/src/pep/enforcer_tests.rs
mattgarmon: Updated because context was switched to `PlatformSecurityContext`
Added assertions for the passed context.

gears/system/authz-resolver/authz-resolver-sdk/tests/rest_roundtrip.rs
mattgarmon: Added test `rest_evaluate_surfaces_error_variant`

striped-zebra-dev (approved):

gears/system/authz-resolver/authz-resolver-sdk/src/rest.rs
aviator5, line 1: Suggest having gRPC instead of REST for `AuthZResolver`. It's supposed to be used only internally by another gears. I would also suggest removing REST implementation for now to avoid any questions.

p.s.: @mattgarmon I don't want to block this PR merge since it touches a lot of places, so propose to merge this as is and add gRPC client in the next one.

aviator5 (approved):

gears/system/authz-resolver/authz-resolver-sdk/src/rest.rs
mattgarmon: Agreed, it can be added later

MikeFalcon77 (approved):

PR #4675, opened 2026-08-27
feat(toolkit-oop): lazy connect DirectoryService client

mattgarmon:
OoP bootstrap eagerly connected to the DirectoryService via `DirectoryGrpcClient::connect[_with_interceptor]`, which blocked (and on failure crashed) startup when the directory was not yet reachable - making it a hard startup dependency and offloading retries onto k8s `CrashLoopBackOff`. This violates `cpt-cf-adr-eventual-readiness`.
 
Add lazy-connecting constructors that build the channel without an eager TCP/HTTP2 connect, deferring the actual connection to the first RPC where the background presence loop's exponential backoff absorbs the startup window; tonic reconnects transparently after transient failures.
 
 Changes:
- `toolkit-transport-grpc`: add `connect_lazy()` (no eager connect, fails fast only on a malformed URI)
- directory SDK: add `DirectoryGrpcClient::connect_lazy[_with_interceptor]`
- `toolkit/bootstrap/oop.rs`: use the lazy client so gears start immediately even against a cold DirectoryService

## Summary by CodeRabbit

* **New Features**
  * Added lazy gRPC connection support for directory services.
  * Connections begin on the first request and reconnect automatically when needed.
  * Invalid service addresses still report errors immediately.

* **Bug Fixes**
  * Startup no longer fails or blocks when the directory service is temporarily unreachable.
  * Background retries handle unavailable directory services with exponential backoff.

* **Documentation**
  * Updated startup documentation and status messages to clarify lazy connection behavior.

### DIFF
--- a/libs/system-sdks/sdks/directory/src/grpc/client.rs
+++ b/libs/system-sdks/sdks/directory/src/grpc/client.rs
@@ -14,7 +14,7 @@ use crate::api::{
 };
 use std::collections::BTreeMap;
 use toolkit_transport_grpc::InternalAuthInterceptor;
-use toolkit_transport_grpc::client::{GrpcClientConfig, connect_with_retry};
+use toolkit_transport_grpc::client::{GrpcClientConfig, connect_lazy, connect_with_retry};
 
 use crate::{
     DeregisterInstanceRequest, DirectoryServiceClient, GetOpenApiSpecRequest, GrpcServiceEndpoint,
@@ -109,6 +109,51 @@ impl DirectoryGrpcClient {
     ) -> Result<Self> {
         let cfg = GrpcClientConfig::new("directory");
         let channel: Channel = connect_with_retry(uri, &cfg).await?;
+        Ok(Self::from_channel_with_interceptor(channel, interceptor))
+    }
+
+    /// Create a directory client with a **lazily-connecting** channel.
+    ///
+    /// Performs **no** eager connection: the channel connects on the first RPC
+    /// and transparently reconnects on failure. This is the eventual-readiness
+    /// entry point (`cpt-cf-adr-eventual-readiness`) for `OoP` bootstrap — the
+    /// process starts even when the `DirectoryService` is not yet reachable, and
+    /// the presence loop's backoff retry absorbs the startup window instead of
+    /// the process crashing (which would offload retries onto a k8s
+    /// `CrashLoopBackOff`).
+    ///
+    /// # Runtime context
+    /// Must be called from within a Tokio runtime context: building the lazy
+    /// channel initialises the hyper reactor. Calling it outside a runtime
+    /// returns an error rather than panicking (it still does not connect).
+    ///
+    /// # Errors
+    /// Returns an error if called outside a Tokio runtime context, or if `uri`
+    /// is malformed — never for an unreachable peer.
+    pub fn connect_lazy(uri: impl Into<String>) -> Result<Self> {
+        let cfg = GrpcClientConfig::new("directory");
+        let channel: Channel = connect_lazy(uri, &cfg)?;
+        Ok(Self::from_channel(channel))
+    }
+
+    /// Create a directory client with a **lazily-connecting** channel, attaching
+    /// `interceptor`'s platform-plane credential to every outbound call.
+    ///
+    /// The lazy counterpart of [`connect_with_interceptor`](Self::connect_with_interceptor);
+    /// see [`connect_lazy`](Self::connect_lazy) for the connection semantics.
+    /// The URI is validated before `interceptor` is consumed, so the credential
+    /// is only moved into the client on success.
+    ///
+    /// # Errors
+    /// Returns an error only if `uri` is malformed — never for an unreachable
+    /// peer.
+    pub fn connect_lazy_with_interceptor(
+        uri: impl Into<String>,
+        interceptor: InternalAuthInterceptor,
+    ) -> Result<Self> {
+        let cfg = GrpcClientConfig::new("directory");
+        // Validate the URI (build the channel) before consuming `interceptor`.
+        let channel: Channel = connect_lazy(uri, &cfg)?;
         Ok(Self::from_channel_with_interceptor(channel, interceptor))
     }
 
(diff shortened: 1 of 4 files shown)


libs/toolkit-transport-grpc/src/client.rs
```diff
@@ -241,6 +241,48 @@ where
...
+///
+/// # Runtime context
+/// Although this does not connect, it must be called from within a Tokio
+/// runtime context: building the lazy `Channel` initialises the hyper reactor.
+/// Calling it outside a runtime will panic.
+///
+/// # Errors
+/// Returns an error only if `uri` is not a valid endpoint (parse failure). It
+/// never errors for an unreachable peer — that is deferred to the first RPC.
+pub fn connect_lazy<TClient>(
```
MikeFalcon77, line 265: **MEDIUM**

`connect_lazy` panics if called outside a Tokio runtime context, but this precondition is documented only in a doc comment, not enforced by the type system or a runtime check.

A caller who misses the doc comment gets a runtime panic instead of a compile-time error or a clean `Result::Err`.

Guard the call with `tokio::runtime::Handle::try_current()` and return an error when no runtime is active, or accept an explicit `&Handle` parameter instead of relying on caller discipline.

libs/toolkit-transport-grpc/src/client.rs
```diff
@@ -241,6 +241,48 @@ where
...
+) -> anyhow::Result<TClient>
+where
+    TClient: From<Channel>,
+{
+    let uri_string = uri.into();
+    let endpoint = build_endpoint(uri_string, cfg)?;
+    let channel = endpoint.connect_lazy();
+
+    if cfg.enable_tracing {
+        tracing::debug!(
```
MikeFalcon77: **MEDIUM**

The lazy-connect path logs client creation at `debug!`, while the equivalent success event in `connect_with_stack` (eager path) logs at `info!`.

This hides, unless debug logging is enabled, the operationally significant fact that connectivity to the dependency was never verified at startup.

Log the lazy-channel creation at `info!` to match `connect_with_stack`, including the same `connect_timeout_ms`/`rpc_timeout_ms` fields.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -534,11 +537,21 @@ pub async fn run_oop_with_options(opts: OopRunOptions) -> Result<()> {
...
-    // Connect to DirectoryService. When the gear configures a platform-plane
-    // credential, attach it (as `x-toolkit-internal-token`) to every outbound
-    // system call via an InternalAuthInterceptor (`cpt-cf-adr-two-plane-auth`).
+    // Create the DirectoryService client with a LAZY channel. Per
+    // `cpt-cf-adr-eventual-readiness`, an OoP gear must start immediately and
+    // unconditionally even when the DirectoryService is not yet reachable —
+    // blocking (or crashing) on it at bootstrap would make it a hard startup
+    // dependency (the SPOF the ADR rejects). A lazy channel performs no eager
+    // connection: it fails fast only on a malformed endpoint, and defers the
+    // actual connect to the first RPC, where the background presence loop's
```
MikeFalcon77: **MEDIUM**

This comment states unconditionally that the background presence loop's exponential backoff absorbs the startup window for the lazy DirectoryService client, but the legacy branch further down (no `oop_http` config) only runs a fixed-interval heartbeat loop with no registration retry or backoff.

A reader relying on this comment would wrongly assume every OoP gear gets backoff-based resilience against a cold DirectoryService, when the legacy path does not.

Scope the comment to the `oop_http` (HTTP-serving) lifecycle, or add a note that the legacy branch has no presence/backoff loop and depends on external registration.

libs/toolkit-transport-grpc/src/client.rs
```diff
@@ -412,4 +454,44 @@ mod tests {
         let result = build_endpoint(String::new(), &cfg);
         assert!(result.is_err(), "build_endpoint should fail with empty URI");
     }
+
+    #[tokio::test]
+    async fn test_connect_lazy_does_not_connect_eagerly() {
```
MikeFalcon77, line 474: **MEDIUM**

The `connect_lazy` tests only cover successful lazy construction and a build-time malformed-URI error; none exercises the deferred-connection failure path (an actual RPC issued through the lazy channel against an unreachable peer).

That deferred-failure behavior is the core guarantee the eventual-readiness design depends on the caller's retry loop to absorb, and it is currently unverified.

Add a test that builds a channel with `connect_lazy` against an unreachable address, drives it through a real RPC call, and asserts the call returns an error instead of hanging.

libs/system-sdks/sdks/directory/src/grpc/client.rs
```diff
@@ -470,6 +512,46 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn connect_lazy_succeeds_against_unreachable_peer() {
```
MikeFalcon77, line 519: **MEDIUM**

`connect_lazy`/`connect_lazy_with_interceptor` tests only assert that construction succeeds against an unreachable peer or fails on a malformed URI; no test drives an actual `DirectoryClient` RPC through the lazily-built client.

Without this, there is no verification that a deferred connection failure surfaces as a proper lookup/call error rather than hanging indefinitely.

Add a test that calls a `DirectoryClient` trait method (e.g. `resolve_grpc_service`) on a `connect_lazy` client pointed at an unreachable address and asserts the result is an `Err`.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -552,19 +565,20 @@ pub async fn run_oop_with_options(opts: OopRunOptions) -> Result<()> {
...
-                .await?;
+        let client = DirectoryGrpcClient::connect_lazy_with_interceptor(
+            &opts.directory_endpoint,
+            interceptor,
+        )?;
         (client, provider)
     } else {
         (
-            DirectoryGrpcClient::connect(&opts.directory_endpoint).await?,
+            DirectoryGrpcClient::connect_lazy(&opts.directory_endpoint)?,
```
MikeFalcon77: **MEDIUM**

`run_oop_with_options` now builds the DirectoryService client with `connect_lazy` instead of awaiting `connect`/`connect_with_interceptor`, removing the bootstrap failure on an unreachable directory — but no test in this diff verifies that bootstrap now proceeds when the directory endpoint is unreachable.

This is the core behavioral change of the PR and it is currently unverified at the level where a regression would actually be caught.

Add or extend a bootstrap-level test that runs `run_oop_with_options` (or the relevant slice of it) against an unreachable directory endpoint and asserts it no longer fails at the connect step.

libs/toolkit-transport-grpc/src/client.rs
```diff
@@ -412,4 +454,44 @@ mod tests {
...
+        let cfg = GrpcClientConfig::default();
+        let result: anyhow::Result<Channel> = connect_lazy("http://127.0.0.1:1", &cfg);
+        assert!(
+            result.is_ok(),
+            "connect_lazy must return Ok for an unreachable peer (no eager connect)"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_connect_lazy_without_tracing_skips_debug_log() {
```
MikeFalcon77: **LOW**

`test_connect_lazy_without_tracing_skips_debug_log` never captures or inspects tracing output, so it cannot actually detect whether the debug log is skipped when `enable_tracing` is false — its only assertion (`result.is_ok()`) is identical to `test_connect_lazy_does_not_connect_eagerly`.

The test name promises coverage of the tracing branch that the test body does not provide, which can mask a regression in that branch.

Install a tracing test subscriber before the call and assert no debug event was emitted, or remove the test since it currently adds no coverage beyond the default-config case.

libs/system-sdks/sdks/directory/src/grpc/client.rs
```diff
@@ -112,6 +112,48 @@ impl DirectoryGrpcClient {
...
+    /// Create a directory client with a **lazily-connecting** channel, attaching
+    /// `interceptor`'s platform-plane credential to every outbound call.
+    ///
+    /// The lazy counterpart of [`connect_with_interceptor`](Self::connect_with_interceptor);
+    /// see [`connect_lazy`](Self::connect_lazy) for the connection semantics.
+    ///
+    /// # Errors
+    /// Returns an error only if `uri` is malformed — never for an unreachable
+    /// peer.
+    pub fn connect_lazy_with_interceptor(
```
MikeFalcon77, line 150: **LOW**

`connect_lazy_with_interceptor` takes ownership of `interceptor` but silently drops it if `connect_lazy(uri, &cfg)?` fails on a malformed URI.

This loses the caller's credential material on the one error path this constructor has — a minor but avoidable API footgun (flagged against the external `RUST_GUIDELINES.md` rule to return consumed arguments to the caller on failure).

Validate/parse the URI before consuming `interceptor`, or surface it back to the caller via a typed error.

libs/toolkit-transport-grpc/src/client.rs
mattgarmon: Fixed. Now guarded with `Handle::try_current()`, returns Err instead of panicking. Added a test.

libs/toolkit-transport-grpc/src/client.rs
mattgarmon: Done. Now logs at `info!` with `connect_timeout_ms`/`rpc_timeout_ms`, matching `connect_with_stack`.

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Done. Comment now notes backoff applies to the `oop_http` presence loop; legacy path is heartbeat-only with no retry.

libs/toolkit-transport-grpc/src/client.rs
mattgarmon: Test added

libs/system-sdks/sdks/directory/src/grpc/client.rs
mattgarmon: Added a test

libs/toolkit/src/bootstrap/oop.rs
mattgarmon: Done, extracted `build_directory_client` and added bootstrap-level tests

libs/toolkit-transport-grpc/src/client.rs
mattgarmon: Removed the test

libs/system-sdks/sdks/directory/src/grpc/client.rs
mattgarmon: Done, URI is validated before the interceptor is consumed

striped-zebra-dev (approved):

PR #4676, opened 2026-08-27
feat(api-gateway): advertise bound REST endpoint for directory registration

mattgarmon:
The runtime's directory-register phase advertises in-process REST providers (via `#[toolkit::provides]`) so out-of-process consumers can resolve them, but `ApiGatewayCapability::bound_endpoint()` returned the default `None` - so the phase found no endpoint, timed out, and never registered the providers.
 
Publish the gateway's reachable base URL once the main listener binds and override `bound_endpoint()` to expose it. Prefer the explicitly configured `advertise_uri` (required in Kubernetes, where the pod binds 0.0.0.0); otherwise fall back to the bound address.
 
 Changes:
- config: add `ApiGatewayConfig.advertise_uri` (optional)
- gear: store `bound_endpoint` (`parking_lot::Mutex`), set it in `serve()` via a `publish_bound_endpoint` helper, and implement the capability accessor

## Summary by CodeRabbit

* **New Features**
  * API Gateway deployments can advertise a configurable base URL for their REST endpoint.
  * When no URL is configured, the gateway publishes its active bound address, including any configured path prefix.
  * Other services can retrieve the gateway’s available endpoint after startup.

* **Bug Fixes**
  * Endpoint discovery now reflects the configured or active gateway address, including dynamically assigned ports.
  * Gateways bound to wildcard addresses or using invalid path prefixes avoid publishing unreachable endpoints unless an advertised URL is provided.

### DIFF
--- a/gears/system/api-gateway/src/config.rs
+++ b/gears/system/api-gateway/src/config.rs
@@ -22,6 +22,13 @@ fn default_gateway_sync_interval_secs() -> u64 {
 #[allow(clippy::struct_excessive_bools)]
 pub struct ApiGatewayConfig {
     pub bind_addr: String,
+
+    /// Base URL other pods use to reach this gateway's REST endpoint, published
+    /// for directory registration (required in Kubernetes, where the pod binds
+    /// `0.0.0.0`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub advertise_uri: Option<String>,
+
     #[serde(default)]
     pub enable_docs: bool,
     #[serde(default)]
@@ -189,6 +196,7 @@ impl Default for ApiGatewayConfig {
     fn default() -> Self {
         Self {
             bind_addr: String::default(),
+            advertise_uri: None,
             enable_docs: false,
             cors_enabled: false,
             cors: None,
--- a/gears/system/api-gateway/src/gear.rs
+++ b/gears/system/api-gateway/src/gear.rs
@@ -94,6 +94,13 @@ pub struct ApiGateway {
     // task and read by the Forwarder fallback; empty/unused when
     // `gateway_proxy` is disabled.
     pub(crate) proxy_registry: Arc<toolkit_gateway::ProxyRegistry>,
+
+    // Base URL other pods use to reach this gateway, published once the main
+    // listener binds (from `serve`). Read by the runtime's directory-register
+    // phase (via `ApiGatewayCapability::bound_endpoint`) to advertise in-process
+    // REST providers. Write-once/read-many, so `OnceLock` (lock-free reads),
+    // matching `healthcheck_registry`/`health_router`; unset until the server binds.
+    pub(crate) bound_endpoint: OnceLock<String>,
 }
 
 impl Default for ApiGateway {
@@ -110,6 +117,7 @@ impl Default for ApiGateway {
             registered_routes: DashMap::new(),
             registered_handlers: DashMap::new(),
             proxy_registry: Arc::new(toolkit_gateway::ProxyRegistry::new()),
+            bound_endpoint: OnceLock::new(),
         }
     }
 }
@@ -201,6 +209,7 @@ impl ApiGateway {
             registered_routes: DashMap::new(),
             registered_handlers: DashMap::new(),
             proxy_registry: Arc::new(toolkit_gateway::ProxyRegistry::new()),
+            bound_endpoint: OnceLock::new(),
         }
     }
 
@@ -693,7 +702,10 @@ impl ApiGateway {
 
         // Bind the main socket.
         let listener = tokio::net::TcpListener::bind(addr).await?;
-        tracing::info!("HTTP server bound on {}", addr);
+        let bound_addr = listener.local_addr().unwrap_or(addr);
+        tracing::info!("HTTP server bound on {}", bound_addr);
+
+        self.publish_bound_endpoint(&cfg, bound_addr);
 
         // Bind the separate health listener (if `serve` = separate|both) BEFORE signalling
         // ready, so readiness reflects every listener the pod must accept traffic on.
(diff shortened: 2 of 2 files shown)


gears/system/api-gateway/src/config.rs
```diff
@@ -22,6 +22,11 @@ fn default_gateway_sync_interval_secs() -> u64 {
 #[allow(clippy::struct_excessive_bools)]
 pub struct ApiGatewayConfig {
     pub bind_addr: String,
+
+    /// Base URL to reach this gateway's REST endpoint
```
MikeFalcon77: **HIGH**

The `advertise_uri` doc comment doesn't say the value must already include `prefix_path`.

Unlike the bound-address fallback (which appends the normalized `prefix_path` automatically), `publish_bound_endpoint` uses `advertise_uri` verbatim, so an operator following this comment can publish a directory endpoint missing the path prefix routes are actually served under, breaking out-of-process requests silently.

Update the doc comment to state that `advertise_uri` must already include the gateway's `prefix_path` (e.g. `http://host:8087/cf`, not just `http://host:8087`).

gears/system/api-gateway/src/gear.rs
```diff
@@ -94,6 +94,12 @@ pub struct ApiGateway {
     // task and read by the Forwarder fallback; empty/unused when
     // `gateway_proxy` is disabled.
     pub(crate) proxy_registry: Arc<toolkit_gateway::ProxyRegistry>,
+
+    // Base URL other pods use to reach this gateway, published once the main
+    // listener binds (from `serve`). Read by the runtime's directory-register
+    // phase (via `ApiGatewayCapability::bound_endpoint`) to advertise in-process
+    // REST providers. `None` until the server binds.
+    pub(crate) bound_endpoint: Mutex<Option<String>>,
```
MikeFalcon77: **MEDIUM**

`bound_endpoint` is a `parking_lot::Mutex<Option<String>>` even though it is written exactly once (in `publish_bound_endpoint`) and only ever read afterward.

The struct already uses `OnceLock` for the same write-once/read-many shape (`healthcheck_registry`, `health_router`), so this introduces an unnecessary lock-based access pattern where a lock-free one is the established convention.

Replace with `bound_endpoint: OnceLock<String>`, using `.set()` in `publish_bound_endpoint` and `.get().cloned()` in the `bound_endpoint()` getter.

gears/system/api-gateway/src/gear.rs
```diff
@@ -729,6 +740,43 @@ impl ApiGateway {
...
+    /// otherwise falls back to the bound address plus the normalized
+    /// `prefix_path`, so the advertised base URL matches the path under which
+    /// `rest_finalize` actually serves the routes.
+    ///
+    /// A wildcard bind address (`0.0.0.0` or `::`) is not reachable by other
+    /// pods, so when `advertise_uri` is unset we refuse to publish it rather
+    /// than advertise an unusable endpoint; operators must set `advertise_uri`.
+    fn publish_bound_endpoint(&self, cfg: &ApiGatewayConfig, bound_addr: SocketAddr) {
+        let advertised = match cfg.advertise_uri.clone() {
+            Some(uri) => uri,
```
MikeFalcon77: **MEDIUM**

The `Some(uri) => uri` branch accepts the operator-supplied `advertise_uri` verbatim with no check against the configured `prefix_path`.

A misconfigured `advertise_uri` that omits the prefix `rest_finalize` actually serves routes under is published and logged as a successful registration, giving no signal that the endpoint is wrong.

Compare `uri` against the normalized `prefix_path` and emit a `tracing::warn!` (or reject at config-validation time, mirroring `GatewayProxyConfig::validate`) when they are inconsistent.

gears/system/api-gateway/src/gear.rs
```diff
@@ -729,6 +740,43 @@ impl ApiGateway {
...
+                    "REST host bound on a wildcard address with no `advertise_uri`; \
+                     skipping directory registration endpoint (set `advertise_uri`)"
+                );
+                return;
+            }
+            None => {
+                // The router is served under `prefix_path` (see `rest_finalize`),
+                // so the fallback base URL must include it. The prefix is already
+                // validated by `get_or_build_router` before we get here.
+                let prefix = Self::normalize_prefix_path(&cfg.prefix_path).unwrap_or_default();
```
MikeFalcon77: **MEDIUM**

The `Err` from `Self::normalize_prefix_path(&cfg.prefix_path)` is discarded via `unwrap_or_default()`, silently falling back to an empty prefix if validation ever fails for this config snapshot.

This would mask a real configuration error as a successful (but wrong) endpoint publication instead of surfacing it.

Propagate or log the error at `warn`/`error` and skip publishing, mirroring the wildcard-address branch above it, instead of silently defaulting to an empty prefix.

gears/system/api-gateway/src/gear.rs
```diff
@@ -1165,6 +1217,85 @@ mod tests {
...
+        api.publish_bound_endpoint(&ApiGatewayConfig::default(), addr);
+
+        assert_eq!(
+            api.bound_endpoint(),
+            Some("http://127.0.0.1:8087".to_owned())
+        );
+    }
+
+    #[test]
+    fn publish_bound_endpoint_prefers_advertise_uri() {
```
MikeFalcon77: **MEDIUM**

No test covers `publish_bound_endpoint` when both `advertise_uri` and a non-empty `prefix_path` are configured together.

It is unverified whether the advertised URL is expected to include the prefix or intentionally ignore it in that case, which is exactly the ambiguity that makes the `advertise_uri` precedence risky.

Add a test that sets both `advertise_uri` and a non-empty `prefix_path` and asserts the exact value `bound_endpoint()` returns.

gears/system/api-gateway/src/gear.rs
```diff
@@ -693,7 +701,10 @@ impl ApiGateway {
 
         // Bind the main socket.
         let listener = tokio::net::TcpListener::bind(addr).await?;
-        tracing::info!("HTTP server bound on {}", addr);
+        let bound_addr = listener.local_addr().unwrap_or(addr);
```
MikeFalcon77: **LOW**

The new `bound_addr` derivation in `serve()` (`listener.local_addr().unwrap_or(addr)`) is only exercised indirectly through `publish_bound_endpoint` with manually constructed addresses.

No test verifies the `serve()` path itself when the OS-assigned address differs from the configured `bind_addr`, e.g. binding to an ephemeral `:0` port.

Add a test that binds to an ephemeral port and asserts `bound_endpoint()` reflects the actual OS-assigned address, not the configured one.

gears/system/api-gateway/src/config.rs
mattgarmon: Updated

gears/system/api-gateway/src/gear.rs
mattgarmon: Fixed, changed to `OnceLock<String>`

gears/system/api-gateway/src/gear.rs
mattgarmon: Done - now compares against the normalized `prefix_path` and emits `tracing::warn!` when the URI omits it

gears/system/api-gateway/src/gear.rs
mattgarmon: Fixed

gears/system/api-gateway/src/gear.rs
mattgarmon: Added test: `publish_bound_endpoint_advertise_uri_used_verbatim_with_prefix`

gears/system/api-gateway/src/gear.rs
mattgarmon: Added test: `serve_publishes_os_assigned_ephemeral_port`

striped-zebra-dev (approved):

PR #4693, opened 2026-09-01
feat(event-broker)!: model topics and event types as derived GTS types

striped-zebra-dev:
A topic and an event type are both derived GTS type schemas. Their governing configuration - partition count, retention, storage selection, owning topic, admissible subject types - is `x-gts-traits` on the schema, because it steers broker behaviour rather than describing data. Trait keywords are schema-only, so none of it can reach an event payload.

An event therefore carries no `topic`: its `type` names a derived type whose `topic` trait names the stream. `cf.core.events.event_type.v1~` is gone - a type whose instances were types.

Both base schemas are generated from their Rust declarations and held to the committed files by tests, so what ships is what gets registered. Reads return flat DTOs projected from the resolved schemas, keeping trait plumbing off the wire. Every committed GTS document is named after the identifier it declares.

Adds `Iso8601Duration` to toolkit-utils so a duration is validated at the boundary and carried as one, not as a string.


gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
```diff
@@ -1,134 +1,146 @@
...
+      "maxLength": 512,
       "pattern": "^[\\x20-\\x7E]+$",
-      "maxLength": 512
-    },
-    "partition_key": {
-      "description": "Optional producer-supplied routing key for partition selection. When present: partition = murmur3_32(ascii_bytes(partition_key)) % topic.partitions. When absent: same hash applied to `tenant_id` (per-tenant ordering by default). There is no explicit producer-set partition override. See ADR/0002-partition-selection.md. ASCII only, max 1024 bytes.",
-      "type": "string",
-      "pattern": "^[\\x20-\\x7E]*$",
-      "maxLength": 1024
+      "type": "string"
```
Artifizer: I think there must be something like:

    "x-gts-ref": "gts."

since it's not just string, it must be subject type in GTS format

gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
```diff
@@ -1,134 +1,146 @@
...
-      "type": "integer",
-      "format": "int64",
-      "minimum": 0,
-      "readOnly": true
-    },
-    "sequence_time": {
-      "description": "Server-stamped timestamp recording when `sequence` was assigned. Surfaced on read. Producers MUST NOT supply this field on publish; doing so is rejected with 400 BadRequest.",
+    "type": {
+      "description": "GTS type identifier",
+      "format": "gts-type-id",
```
Artifizer: Is `gts-type-id` a known type that Event Broker can recognize?
It's not a valid JSON schema type

gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
```diff
@@ -1,134 +1,146 @@
...
-          "description": "Producer registered via POST /v1/producers; bound to the calling principal. Required for chained / monotonic mode publishes. Omit for stateless. ASCII (UUID format).",
-          "type": "string",
-          "format": "uuid"
-        },
-        "previous": {
-          "description": "Predecessor's sequence for chain dedup. Required and only valid in chained mode (validated server-side against the registered mode of meta.producer_id). On the first chained-mode publish for a (producer_id, topic, partition), set to 0.",
-          "type": "integer",
-          "format": "int64",
-          "minimum": 0
+      "x-gts-ref": "gts.*"
```
Artifizer: I believe it must be:

`gts.cf.core.events.type.v1~`

because you can't use any GTS type as event type, ...

gears/system/event-broker/docs/schemas/gts.cf.core.events.topic.v1~.schema.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "required": [
+    "id"
+  ],
+  "type": "object",
+  "x-gts-abstract": true,
+  "x-gts-traits-schema": {
+    "additionalProperties": false,
+    "properties": {
+      "partitions": {
```
Artifizer: mmm, should we have `x-gts-traits-schema` at all? I think we can put partititions, retention, streaming, etc into the properties, not into the traits, because in any cases topics will be defined by vendors as instances

gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
striped-zebra-dev: Agreed on the substance — it's a GTS type, not a bare string. It now reads:

```json
"subject_type": { "type": "string", "maxLength": 512, "pattern": "^[\\x20-\\x7E]+$", "x-gts-type": true }
```

Not `"gts."` because `gts-validator` rejects a bare prefix under *any* keyword, `x-gts-ref` included — `Invalid GTS identifier: gts.: no segments found`. Wildcards fare no better: `gts.*` passes only under `x-gts-ref`, which is the one keyword exempt from the wildcards-in-pattern-contexts rule. So `true` states what we can enforce and what's actually true here: it must be a type identifier, and the broker can't name the family — an event's subject belongs to some other domain.

gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
striped-zebra-dev: Correct, and removed. It isn't a registered JSON Schema format and nothing acted on it. The constraint moved to `x-gts-type`.

Worth flagging that it wasn't hand-written: it comes from the `gts` crate's own `JsonSchema` impls, which `gts-macros` substitutes wholesale for the property. So the same defect was on the topic's `id` as `format: "gts-instance-id"` — you hadn't seen that one, and it's gone too.

gears/system/event-broker/docs/schemas/gts.cf.core.events.event.v1~.schema.json
striped-zebra-dev: Agreed on the narrowing — the previous `x-gts-ref: "gts.*"` matched every identifier, so it constrained nothing. Now:

```json
"type": { "type": "string", "...": "...", "x-gts-type": "gts.cf.core.events.event.v1~" }
```

Narrowed to the event base rather than `…events.type.v1~`, because that identifier isn't registered anywhere — it appears only in `guidelines/GTS.md`, and an error or constraint naming it is a dead end for anyone who tries to resolve it. This change also removes its last two uses: the event-type canonical errors now report `gts.cf.core.events.event.v1~` and are told apart by `reason`.

gears/system/event-broker/docs/schemas/gts.cf.core.events.topic.v1~.schema.json
striped-zebra-dev: Agreed, and done — the topic now declares no traits at all:

```
keys:       $id, $schema, description, type, additionalProperties, properties, required
properties: id, description, retention        required: id, description
```

A concrete topic is an instance of the base, as you say. `retention` became a property. `partitions` and `streaming` left the type entirely rather than moving to properties — they describe how a *deployment* serves a stream, not what the stream is, so they're broker configuration (`partitions` defaults to 8, retention to 30 days). A topic author registering a stream from another gear's init has no view of either.

Artifizer (approved):

Issue #4628, opened 2026-08-20
[Epic] Types Registry P1 — tenancy, the platform plane, authorization and projections


aviator5:
## Goal

P0 delivered a durable registry for **global** entities: async admission, materialized effective
artifacts, a new SDK contract, authenticated but *unauthorized* access, and one listener. P1
closes that gap — tenant-owned entities with real visibility rules, a genuine two-plane split
with `PlatformIdentity`, PDP-gated reads and writes, tenant availability verdicts, and
caller-chosen projections.

After P1 the registry is the full DESIGN contract minus federation and the P2 extension
mechanisms: a tenant can own its own type contracts, a subject sees exactly what the tenant
hierarchy says it may see, and every access is a policy decision rather than a convention.

**A note on numbering.** `PRD.md` and `DESIGN.md` call the whole first release *P1* and the
extension release *P2*. Our delivery split is different: **P0** was the durable global core
(shipped as #4619), **this P1** is the remainder of DESIGN's first release except federation,
and **P2** is federation plus the platform data-retention operations. Where a document says
"P1 requires…" it usually means "the first full release requires…", which is this epic.

## What we get at the end of P1

1. **Tenant ownership** — `ownership_scope = 2` and `owner_tenant_id` are populated for the
   first time. The columns and CHECK constraints already exist from P0, so there is no schema
   migration for ownership itself (ADR-0009).
2. **Visibility as a directed descendant relation** — a tenant-owned entity is visible to the
   owning tenant and all of its descendants, computed from the **subject's** ancestor chain,
   never the Context Tenant's. Identifier uniqueness stays global, and the disclosure boundary
   is honoured: an out-of-scope reverse resolution is indistinguishable from an unissued
   reference, a name conflict reveals only that the name is unavailable, and a blocked deletion
   reports a count without identities.
3. **A real platform plane** — a separate listener, `X-ToolKit-Internal-Token` /
   `PlatformIdentity`, `PlatformSecurityContext` as the first argument of the platform trait,
   and two separate OpenAPI documents over the same paths. Misrouting fails authentication
   because workload and tenant identities cannot cross planes. This closes P0 ceiling **C8**.
4. **PDP authorization on all four actions** — `register` and `delete` per distinct candidate
   identifier at acceptance, `read` and `list` once per request. One resource type
   (`gts.cf.toolkit.types_registry.entity.v1~`) with the identifier in the `gts_id` resource
[...]

PR #4694, opened 2026-09-01
Types Registry Implementation P0-2

### LINKED ISSUE #4622: [Types Registry P0] Phase 2 — revisions and concurrency
Parent: #4619 · Depends on: Phase 1 (#4621)

## Goal

Make a logical entity evolve safely: second and later revisions, optimistic concurrency, and
the version-family rules that must hold under concurrent admissions.

## Tasks

- [ ] **T11** — content revisions and compare-and-swap: `expected_resource_version`
      preconditions, immutable revision insert, current-state pointer move, and the `unchanged`
      outcome when authored content equals current.
- [ ] **T12** — version-family shape and contiguity rules enforced under the family lock:
      minor shape uniform within a major, minors contiguous from `M.0`. Keyed lookups, not scans.

## Exit criteria — Checkpoint 2

- [ ] Equal content reports `unchanged` without creating a revision
- [ ] A stale `expected_resource_version` fails `precondition_failed`
- [ ] Family shape and contiguity refusals hold under concurrency
- [ ] Gear tests on three backends
- [ ] `make dylint` — full workspace, once for the phase
- [ ] Human review

## References

- `gears/system/types-registry/docs/p0/SPEC.md` — D4 (multi-pod commit re-read)
- `gears/system/types-registry/docs/p0/todo.md` — T11–T12

Already on `main`:

- `gears/system/types-registry/docs/ADR/0004-cpt-cf-types-registry-adr-gts-minor-version-identity-evolution.md`
- `gears/system/types-registry/docs/ADR/0005-cpt-cf-types-registry-adr-type-schema-revisions.md`
- `gears/system/types-registry/docs/ADR/0006-cpt-cf-types-registry-adr-registered-instance-revisions.md`

aviator5:
closes #4622

## New Features
- Added support for revising existing schemas and instances with exact expected-version checks.
- Unchanged submissions now complete without creating duplicate revisions.
- Added version-family validation for kind, version shape, and predecessor continuity.
- Added retry guidance when family-lock contention temporarily prevents processing.

## Bug Fixes
- Prevented stale or concurrent revisions from overwriting newer content.
- Prevented revisions from resurrecting deleted entities.
- Minor-bearing type schemas cannot be content-revised.
- Improved handling of internal processing failures with opaque server errors.

## Documentation
- Clarified revision preconditions, security boundaries, and version-family rules.
- Documented family-lock timeout configuration and compatibility-force limitations.


gears/system/types-registry/docs/p0/SPEC.md
```diff
@@ -711,7 +711,7 @@ because other documents cite the numbers.
...
+| C6 | **No PDP.** Reads and writes are authenticated but not authorized, deviating from `06`'s *"every sensitive DB access MUST be covered by a PDP decision"*. Entities are `#[secure(unrestricted)]`, so a tenant-scoped query fails closed rather than leaking. **The sharpest edge is the revision path**: §8.1 step 3 asks the registration policy of creations only — correctly, since the policy governs which regions gain members — and nothing takes its place for an edit, so a caller that reaches the submit route can replace the authored content of any entity the registry holds, a platform-seeded `cf.core.*` schema included, in a region the deployment has closed. Bounded in P0 by transport rather than by policy: the mutation routes are internal-only (C8) | The deferred identity-to-permission binding; then an owner/principal check before `unit::commit_revision`, and `tenant_col` + `PolicyEnforcer` (§12) |
```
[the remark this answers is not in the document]
aviator5: This comment is anchored on ceiling **C6** itself — the paragraph it flags is the one that documents this exact gap, with the same remediation ("an owner/principal check before `unit::commit_revision`"). Acknowledged and tracked, not new.

For the record:

- P0 does have the revision path (T11/T12 shipped content revisions + CAS), so this is a real edit surface — but the missing authorization is identical for creations. The registration policy governs which Regions gain members, not which principal writes; it was never the check being bypassed.
- The P0 bound is transport, and it is stated in C8: mutation routes ship `exposed = false`, in-process only. "Untrusted internal caller" is not in P0's threat model — an in-process gear is compiled into the same binary. The real surface opens with out-of-process gears and the platform listener.
- No cheap partial guard exists here. The obvious candidate — "only the entity's `owning_gear` may revise it" — is self-declared on the inventory record (D11, T22–T25), not an authenticated attribute, so it is not an authorization primitive.

Closing C6 requires the identity-to-permission binding, which is itself still deferred in `docs/arch/authorization/PERMISSION_GTS_TYPE.md`. Both are scoped in the P1 epic #4628 (blocking prerequisite 1, and "PDP authorization on all four actions"); C8 is prerequisite 2 in the same epic. C6's upgrade path now names #4628 explicitly.

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
```diff
@@ -116,6 +136,155 @@ pub async fn run_operation(
...
+async fn lock_families(
+    db: &DBProvider<WorkerError>,
+    family_keys: &[FamilyKey],
+) -> Result<Vec<DbLockGuard>, WorkerError> {
+    let handle = db.db();
+    let mut guards = Vec::with_capacity(family_keys.len());
+    for key in lock_order(family_keys) {
+        let lock_key = family_lock_key(&key);
+        match handle
+            .try_lock(LOCK_GEAR, &lock_key, LockConfig::default())
```
Artifizer: **Severity: High** — the 30s family-lock wait sits on the synchronous HTTP request path, and its timeout strands the operation.

`LockConfig::default()` is `max_wait: 30s`, `max_retries: None`, backoff capped at 5s (`libs/toolkit-db/src/advisory_locks.rs:213`). Meanwhile `registry_service.rs:252` runs `run_operation` **inline** on `POST /entities` when `admission_mode == Inline`. So a contended family key can block a client request for a full 30 seconds — well past any reasonable api-gateway timeout — and the budget is hardcoded here rather than drawn from `WorkerSettings`, which already carries `operation_timeout` (currently unread).

The second half is worse than the latency. When the budget does expire, `FamilyLockUnavailable` propagates out of `lock_families` → `commit_evaluated` → `process_item` → `run_operation`, which means `mark_completed` never runs. The caller gets an opaque `500` (per the new mapping in `api/rest/error.rs`) **and** the operation row is left `running` with its items `pending`. The only recovery is a replay under the same `Idempotency-Key`, and nothing in the `500` tells the caller to do that.

This is the same stranding failure mode that `commit_evaluated` explicitly guards against a few lines below:

> Without the retry, a deadlock on the entity compare-and-swap propagates out of `process_item` before `mark_completed`, stranding the operation row in `running` with its items `pending` and nothing to re-drive it.

The lock path has that property and no equivalent handling.

Suggested, as one change:
1. derive the lock budget from config, defaulted well under the gateway timeout (~2–5s rather than 30s);
2. decide deliberately what a lock timeout returns — either terminalize the item so the operation completes, or map it to a retryable status (`503` + `Retry-After`) that names the replay contract, rather than an opaque `500` over a stranded row.

---

gears/system/types-registry/docs/p0/todo.md
```diff
@@ -722,56 +722,117 @@ which is a technical property, not a concept. A directory whose name promises a
...
+that it **is not** one — the secure query builder exposes no lock clause. So two concurrent
+admissions of two *different new* members of one family can both pass their shape check and both
+commit, leaving a family that violates minor-shape uniformity. This is **not new to T12**: T10's
+kind rule has the identical hole, and the unique key on `gts_id` closes only the same-identifier
+race. The mechanism exists — `Db::try_lock`, an advisory lock working on all three backends — and
+`VersionFamilyRepo::lock_order` is already its ordering half, so the fix is service-layer wiring
+rather than new infrastructure. It is not free: the `SQLite` lock scope is keyed on the DSN, and
+every test using `sqlite::memory:` shares one scope through markers in a **cross-process** cache
+directory, so contending test binaries would serialize against each other. Decide at Checkpoint 2
+whether that wiring belongs to T15 (which already owns bounded retry) or to its own task.
```
Artifizer: **Severity: Medium** — this paragraph contradicts the code the same PR ships.

It says *"there is no family lock"* and points at `VersionFamilyRepo::lock_order` as "already its ordering half", deferring the wiring to a Checkpoint 2 decision between T15 and a new task. But this PR **implements the lock**:

- `worker.rs::lock_families` takes the advisory lock per family, in `lock_order`'s order;
- `worker.rs::family_lock_key` / `LOCK_GEAR` define the key namespace;
- `WorkerError::FamilyLockUnavailable` is the new timeout arm, with a REST mapping;
- `family_test.rs::a_creation_holds_the_family_lock_across_its_commit` pins it, mutation-checked;
- and `VersionFamilyRepo::lock_order` — the method this paragraph cites — is **deleted** here, moved to `domain::family::lock_order`.

So the gap is closed, the cited API no longer exists, and no task section records the work. Given how much this repo leans on `todo.md` as the audit trail — and how carefully every other marker in this diff is justified — a materially false "not done" on shipped code is worse than a missing one: the next reader plans work that already exists.

Either fold this into T12 as a closed criterion, or open the standalone task T12 asked Checkpoint 2 to decide about and mark it done here. The `SQLite` DSN-scoped-lock caveat in the last two sentences is still worth keeping either way — it is a real property of the test topology, and it survived the implementation.

---

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
```diff
@@ -465,9 +486,327 @@ pub async fn commit_creation(
...
+                    tx,
+                    scope,
+                    NewRevision {
+                        entity_id: entity.id,
+                        revision_no,
+                        raw_schema: unit.canonical_body.clone(),
+                        content_hash: unit.content_hash.clone(),
+                        gts_spec_version: GTS_SPECIFICATION_VERSION.to_owned(),
+                        gts_impl_version: GTS_IMPLEMENTATION_VERSION.to_owned(),
+                        compat_forced: false,
```
Artifizer, line 706: **Severity: Medium** — `force` is accepted on the revision path and then silently discarded.

Acceptance still validates `force` (`acceptance.rs` step 6), and `has_cross_minor_check` returns `true` for any `vM.n~` with `M ≥ 1, n ≥ 1` — it reads the identifier only, so it cannot tell a creation from a revision. With T11 opening the revision path, a caller can now submit `force: true` on a revision of such an identifier, pass acceptance, and land here — where `compat_forced` is hardcoded `false` and no compatibility comparison runs at all.

The result is that the caller's declared intent is dropped with no signal: they get a `202`, and the persisted `type_schema_revision` row asserts `compat_forced = false`. That row is the permanent provenance record, and T17's criterion is explicit about it:

> - [ ] Every admitted revision records `gts_spec_version`, `gts_impl_version` and `compat_forced`

Two clean options:
1. carry `candidate.force` through `EvaluatedUnit` onto the revision row now, so the record is honest before T17 gives it teeth; or
2. refuse `force` on a revision until T17 lands — which is precisely the `AcceptanceError::RevisionNotAccepted` pattern this PR just retired, and it worked well: a loud refusal beats a field that silently means nothing.

Either is fine; silently writing `false` over a caller's `true` is the one option I would not keep. Note the creation path at line 399 has the same literal, but there it is honest — a first revision has no baseline to waive.

---

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
```diff
@@ -465,9 +486,327 @@ pub async fn commit_creation(
...
+                        operation_item_id: unit.operation_item_id,
+                        now,
+                    },
+                )
+                .await?;
+            if !stores
+                .update_current_schema(
```
Artifizer, line 713: **Severity: Medium** — sequencing: this makes Type Schemas editable in place before anything checks what the edit breaks.

This statement moves the current pointer and re-materializes D3's artifacts for the new revision. What it does not do — because the tasks do not exist yet — is:

- compare the new revision against its baseline (**T17**, `compare_documents` / `CompatibilityVerdict`), and
- refresh or re-validate anything downstream (**T14**, reverse-impact worklist).

So between T11 and T17 a revision of `…v1.2~` can narrow a schema out from under Registered Instances that were admitted against the previous revision, and nothing notices. The instances keep their recorded `(type_schema_entity_id, type_schema_revision_no)` — which is the right design, and is exactly why the breakage is silent: those rows still point at a revision that validated them, while the *current* schema no longer would.

I suspect this is the intended P0 ordering rather than an oversight — T17 and T14 are sequenced after Checkpoint 2 deliberately. My concern is only that it is not written down anywhere. T11's section documents the concurrency shape in detail and says nothing about the compatibility hole it opens, so today it reads as complete rather than as complete-but-unguarded.

Suggest one explicit line in T11 naming the window, and a matching row in SPEC §9's ceilings table beside C6 — the same treatment C6 itself just got in this PR. That turns it into a decision with an owner instead of a gap someone rediscovers at T17.

---

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
```diff
@@ -465,9 +486,327 @@ pub async fn commit_creation(
...
+        return Ok(Err(ItemFailure::new(
+            "precondition_failed",
+            format!(
+                "'{}' moved past resource_version {expected_resource_version}, or was deleted, \
+                 while this revision was being admitted",
+                unit.gts_id
+            ),
+        )));
+    }
+    let resource_version = expected_resource_version.saturating_add(1);
```
Artifizer: **Severity: Low** — this is a second spelling of a fact the database already owns.

`EntityRepo::compare_and_swap_version` writes `ResourceVersion + 1` and returns a `bool`. This line re-derives the same arithmetic in Rust, and every value reported to the caller — the `CommittedUnit`, the item's `result_resource_version`, the REST receipt — comes from the re-derivation rather than from the write. The two agree today, and nothing in the type system keeps them agreeing: a future change to the CAS statement (say, a version that skips on some condition) would leave this returning a number no row holds.

The PR itself argues against exactly this shape, twice:

> Re-parsed rather than carried on `EvaluatedUnit`, which would hold two spellings of one fact

> `NewEntity` now takes `family.ownership_scope` / `family.owner_tenant_id` instead of re-reading the request — a copy taken from the row it is verified against **cannot** disagree with it, which is stronger than verifying two independent readings

Same reasoning applies here. Having `compare_and_swap_version` return `Option<i64>` — `None` for the lost race, `Some(new_version)` otherwise — collapses the duplication and lets the caller report what was actually committed. It also removes the `saturating_add`, which papers over an overflow the SQL would reject anyway (`i64::MAX + 1` saturates in Rust and errors in the database — the two disagree precisely at the edge the saturation exists to handle).

`revision_no` on the next line has the same shape but is lower-stakes: it derives from the current pointer read inside the transaction, which the CAS has already pinned.

---

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
```diff
@@ -116,6 +136,155 @@ pub async fn run_operation(
...
+            .await?
+        {
+            Some(guard) => guards.push(guard),
+            None => {
+                return Err(WorkerError::FamilyLockUnavailable {
+                    family_key: key.as_str().to_owned(),
+                });
+            }
+        }
```
Artifizer: **Severity: Low** — the partial-acquisition paths leak guards to `Drop`, which this module elsewhere treats as insufficient.

Both early exits here abandon already-acquired guards: the `None` arm returns `FamilyLockUnavailable`, and the `?` on `try_lock` at line 179 returns the database error. In each case the `guards` vector is dropped rather than released, and the doc comment above accepts that:

> Any guard already taken is dropped on the way out, which releases it best-effort.

But `commit_evaluated` goes out of its way to do the opposite, for a stated reason:

> Released deterministically rather than on `Drop`, which the toolkit documents as best-effort: a sibling admission is waiting on this key.

Both halves of that reasoning apply on the failure path too — arguably more so, since a contended lock is exactly when a sibling is waiting, and a best-effort release that does not land holds the key until the session expires. The asymmetry looks unintentional rather than argued.

This is unreachable in P0: `commit_evaluated` passes `std::slice::from_ref(&evaluated.family_key)`, so `lock_order` yields one key and there is never a second iteration to fail on. It becomes reachable at **T19** (dependency-aware partial admission), when a unit spans several families — which is also when the ordering that `lock_order` exists to provide starts mattering. Cheap to fix now while the invariant is fresh: release the accumulated guards explicitly before returning, reusing the same warn-and-continue loop `commit_evaluated` already has.

Worth extracting that loop into a small helper so both call sites share one release policy rather than two.

---

Artifizer (commented):
Reviewed the full diff against the gear's SPEC/todo and the toolkit lock primitives. Overall this is strong work — the concurrency design is mutation-checked against real PostgreSQL and MySQL interleavings rather than asserted, and the doc comments carry the *reasoning* rather than restating the mechanics. CI is green across all 31 checks, clippy included. Most of what follows is about what the PR opens up rather than what it gets wrong.

Six inline comments, by severity:

| # | Severity | Where | What |
|---|---|---|---|
| 1 | **High** | `worker.rs:178` | `LockConfig::default()`'s 30s wait sits on the synchronous `POST /entities` path, and its timeout strands the operation in `running` behind an opaque `500` |
| 2 | Medium | `todo.md:821` | T12 still says "there is no family lock" and cites a method this PR deletes — the lock is implemented here and no task records it |
| 3 | Medium | `unit.rs:696` | `force` is accepted on the revision path and silently written as `compat_forced = false` |
| 4 | Medium | `unit.rs:703` | Type Schemas become editable in place before T17 (compatibility) and T14 (reverse-impact) exist; the window is undocumented |
| 5 | Low | `unit.rs:680` | `expected + 1` re-derives what the CAS statement already computed — two spellings of one fact |
| 6 | Low | `worker.rs:181-187` | Partial lock acquisition leaks guards to `Drop`, which `commit_evaluated` deliberately avoids; reachable at T19 |

Only #1 looks merge-blocking to me; #2 is cheap and worth doing in the same push since it misleads the next reader about work that already exists.

Two things I checked and am satisfied with, recorded so they don't get re-litigated:

- **The C6 authorization gap** (a caller reaching the submit route can rewrite any entity's authored content, seeded `cf.core.*` schemas included, in a region the deployment has since closed) is deliberate, reasoned, and documented in three places — `acceptance.rs` step 3, `unit::commit_revision`, and the SPEC ceilings table, which this PR expands specifically to say so. It is bounded by C8 rather than by policy, and tracked to the P1 epic. I would confirm C8 genuinely holds in every deployment shape before merge, but I would not block on C6 itself.
- **The `unchanged` re-read under `READ COMMITTED`** — I re-derived the argument in T11's "concurrency shape" note. It holds: a re-read that still sees `expected` linearizes this pass ahead of the competitor, so the remaining window after the re-read is not a defect.

---

gears/system/types-registry/types-registry/src/api/rest/routes.rs
```diff
@@ -228,7 +228,9 @@ pub fn register_routes(
         .error_415(openapi)
         .error_422(openapi)
         // `require_registry` answers 503 where no database is bound
-        // (`no-db.yaml`, `--mock`): declared, not merely enforced.
+        // (`no-db.yaml`, `--mock`); inline admission also answers 503 with
+        // `Retry-After` when the family-lock budget expires. Both are declared,
+        // not merely enforced.
```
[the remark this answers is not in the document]
aviator5: Fixed in `6b836e2`. `error_503(openapi)` now chains `.response_header(ResponseHeaderSpec::new("Retry-After", …, Integer))`, and `submission_response_headers_are_declared` gained the `(503, "Retry-After", Integer)` row. Declared optional: only the family-lock arm carries it, `require_registry`'s 503 does not.

gears/system/types-registry/docs/p0/SPEC.md
```diff
@@ -711,9 +713,10 @@ because other documents cite the numbers.
...
+| C9 | **Implementation sequencing only.** T11 makes revisions executable before T14 refreshes reverse impact and T17 compares compatibility. Minor-bearing Type Schema revisions remain permanently refused by ADR-0004; during this window every effective `force` also fails closed, and C8 keeps the database mutation path internal | T14 and T17 close the two gaps at Checkpoints 3 and 4, before T24 exposes any consumer. Strike this row when both checkpoints are complete; striking it removes only the temporary `force` refusal, not the ADR-0004 invariant |
```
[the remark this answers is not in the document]
aviator5: Fixed in `6b836e2` — wording only; the code never disagreed. C9 now reads "Content revisions **of** minor-bearing Type Schemas remain permanently refused by ADR-0004 — creating one is admissible (§8.1 step 4), editing it is not". That matches `acceptance.rs:246-255`, which gates on `kind == Registration && expected == Precondition::Version(_) && is_minor_bearing_type_schema`. One decision, taken once in acceptance, so the worker cannot diverge.

gears/system/types-registry/types-registry/src/infra/storage/repo/entity_repo.rs
```diff
@@ -281,31 +288,38 @@ impl EntityRepo {
         entity_id: i64,
         expected_resource_version: i64,
         now: OffsetDateTime,
-    ) -> Result<bool, ScopeError> {
+    ) -> Result<Option<i64>, ScopeError> {
+        let next_resource_version = expected_resource_version.checked_add(1).ok_or_else(|| {
+            ScopeError::Db(DbErr::Custom(
+                "resource_version cannot advance past i64::MAX".to_owned(),
+            ))
+        })?;
```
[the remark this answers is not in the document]
aviator5: Not taking this one — the scenario is unreachable. Two domain guards sit above the repo:

- `unit.rs:610` compares `entity.resource_version != expected_resource_version` and returns `stale_precondition` **before** any CAS, so "client submits `i64::MAX` against a version-1 row" already yields terminal `precondition_failed`.
- `unit.rs:663` has an explicit `expected_resource_version == i64::MAX` arm returning `ResourceVersionExhausted`, reached only when the row genuinely is at the ceiling.

`checked_add` in the repo is belt-and-braces for direct callers, and `repo_test.rs:257` tests that repo contract in isolation. Folding the ceiling into `Ok(None)` would make the domain unable to tell version exhaustion from a lost race — one is an operator alert, the other a `412`.

gears/system/types-registry/types-registry/tests/repo_backends_test.rs
```diff
@@ -385,6 +387,120 @@ async fn current_documents_reads_the_current_revision_only(
...
+    let item1 = seed_operation_item(&conn, gts_id, 1, NOW).await;
+    seed_type_schema_revision(&conn, entity.id, 1, item1, first, NOW).await;
+    seed_current_type_schema(&conn, entity.id, 1, first, NOW).await;
+    let item2 = seed_operation_item(&conn, gts_id, 2, NOW).await;
+    seed_type_schema_revision(&conn, entity.id, 2, item2, second, NOW).await;
+
+    // A fingerprint that is not the seed's, so a pointer move that failed to
+    // rebind the binary column would read back as the old value rather than as
+    // an absent one.
+    let fingerprint = vec![0xDE, 0xAD, 0xBE, 0xEF, 0x00, 0xFF];
```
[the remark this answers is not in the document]
aviator5: Not taking this one — there is no `BINARY(16)` column. `m20260817_000001_initial.rs` declares `resolution_fingerprint` as `bytea` (PG, L328), `BLOB` (SQLite, L641) and `VARBINARY(64)` (MySQL, L941). None is fixed-width and `VARBINARY` does not pad, so the six-byte value round-trips on all three and the assertion at L447-450 cannot fail for the stated reason.

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
aviator5: Fixed in `6b836e2`: the lock budget is configurable (5s default), timeout returns `503` + `Retry-After` with same-key replay guidance, and recovery is tested.

gears/system/types-registry/docs/p0/todo.md
aviator5: Fixed in `6b836e2`: T12 now records the implemented family lock, timeout/recovery behavior, deterministic cleanup, and SQLite caveat.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
aviator5: Fixed in `6b836e2`: effective `force` now fails closed until T17, so caller intent is never silently persisted as `false`.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
aviator5: Fixed in `6b836e2`: T11 now names the interim window and SPEC §9 tracks it as ceiling C9 through T14/T17.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
aviator5: Fixed in `6b836e2`: CAS now returns `Option<i64>` with the exact stored version, uses repository-side `checked_add`, and has overflow coverage.

gears/system/types-registry/types-registry/src/domain/admission/worker.rs
aviator5: Fixed in `6b836e2`: all acquired guards now use the shared explicit release path on success, timeout, and acquisition error.

MikeFalcon77 (approved):

PR #4701, opened 2026-09-02
Add zone-based throttling middleware (rate + in-flight limits per route) — supersedes #4171

parviz-93:
## Description
Zone-based throttling middleware for the API gateway: named `rate_limit_zones` /
`in_flight_limit_zones` in gateway config, bound per-operation via `ThrottlingSpec`
(GCRA rate limiting, per-key in-flight gates, dry-run, background pruning).

This is a continuation of the abandoned #4171: the original commit is cherry-picked
with authorship preserved, all 13 review threads from that PR are addressed
(see the reply on each thread), and three follow-up defects are fixed on top:
the toolkit contract no longer depends on `axum`, `max_keys` is actually enforced
for rate zones, and the `x-rate-limit-*` / `x-in-flight-limit` OpenAPI extensions
are restored. Design docs: `docs/arch/throttling/` (ADR-0001, ADR-0002, DESIGN.md).

## Type of Change
- [x] Documentation update

Breaking: the old inline `defaults.rate_limit { rps, burst, in_flight }` gateway
config block is replaced by named zones; a gateway hitting the fallback router
path with auth enabled and no AuthN Resolver client now fails startup (fail-fast)
instead of silently serving a stale default router. `max_keys` on
`in_flight_limit_zones` is now enforced on admission: a request with a
never-seen key beyond the cap is rejected with the zone's
`response_status_code` and `Retry-After: 10` until the periodic prune frees
idle gates. Dry-run serves such requests and counts them in the new
`throttling.dry_run_rejections` counter (attributes `zone`/`kind`, next to
`throttling.rejections`).

## Testing
- [x] Unit tests pass
- [x] Integration tests pass

`cargo test --workspace`: 15627 passed, 0 failed. New tests cover `max_keys`
admission (enforce + dry-run), OpenAPI zone-name emission and gateway enrichment,
the fallback router/pruner pairing, in-flight `max_keys` admission (enforce,
dry-run, concurrent), and the `throttling.dry_run_rejections` metric.

## Documentation

## Checklist
- [x] Self-review completed

## Related Issues
Supersedes #4171 (continuation of that abandoned PR; maintainers — please close it
in favor of this one).

## Summary by CodeRabbit

* **New Features**
  * Added configurable, zone-based API throttling for request rates and concurrent requests.
  * Added per-IP throttling, retry guidance, rate-limit response headers, and configurable rejection status codes.
  * Added throttling metadata to the API specification for the mini-chat chat creation endpoint.
  * Added observe-only throttling to monitor potential rejections without blocking requests.
  * Added bounded key tracking and automatic cleanup to support predictable resource usage.

* **Documentation**
  * Added architecture and design documentation covering distributed throttling, configuration, limits, and observability.


config/quickstart.yaml
```diff
@@ -99,15 +99,58 @@ gears:
...
+      # by the `ThrottlingSpec` attached to the operation IN CODE, which references
+      # these zone names. That binding is already wired in the mini-chat gear at
+      # gears/mini-chat/.../api/rest/routes/chats.rs (create_chat):
+      #   ThrottlingSpec {
+      #     rate_limit_zone:        "rl_mini_chat_chat",
+      #     in_flight_limit_zone:   "ifl_mini_chat_chat",
+      #     key_type:               "ip",
+      #     identity_key_func:      None,
+      #     require_security_context: false,   // IP-keyed => enforced before auth
+      #   }
```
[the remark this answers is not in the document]
parviz-93: Fixed in b72d26d03: the comment now shows the real ThrottlingSpec fields (`key_type` / `identity_key_func` no longer exist) and explicitly calls out that the mini-chat binding sets `dry_run: true`, i.e. these zones only observe and log until the gear flips it.

gears/system/api-gateway/src/gear.rs
```diff
@@ -388,6 +397,24 @@ impl ApiGateway {
...
         ));
 
+        // Build both throttling partitions together so a zone referenced from
+        // both pre-auth and post-auth operations shares a single limiter/gate
+        // instance. The pre-auth map is layered later (step 9b).
+        let (throttling_map, throttling_map_noauth, throttle_key_pruner) =
+            middleware::throttling::build_maps(&specs, &config)?;
+        // Hand the keyed-store pruner to `serve`, which owns the lifecycle
+        // CancellationToken needed to bound the background prune task.
+        *self.throttle_key_pruner.lock() = Some(throttle_key_pruner);
```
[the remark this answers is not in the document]
parviz-93: Fixed in b72d26d03. `apply_middleware_stack` now returns the pruner instead of stashing it, so the pairing is the caller's responsibility: `rest_finalize` stores the pruner together with `final_router`, the `serve` fallback path stores the one from its own build, and `rebuild_and_cache_router` deliberately discards its rebuild's pruner instead of overwriting the serving pairing (its router goes only to the cache and is never served). Worth noting the desync was latent rather than live — `rebuild_and_cache_router` currently has no callers — but the trap is now structurally closed.

docs/arch/throttling/ADR/0001-distributed-throttling-cluster-cache.md
```diff
@@ -0,0 +1,207 @@
...
+  new cluster-cache-backed path. A zone selects its backend via configuration;
+  absence of configuration keeps today's in-process behavior.
+* The gateway must resolve `ClusterCacheV1` from the ClientHub during
+  initialization (the same mechanism already used to resolve the AuthN resolver
+  client) and declare a dependency on the cluster capability. Deployments that
+  do not provide a cluster cache can only use in-process zones.
+* Per-key state must be serialized to and from cache bytes as the GCRA arrival
+  time, and each write must carry a TTL equal to the bucket's full replenish
+  window so idle keys self-expire cluster-wide. This replaces the local pruner's
+  role for distributed zones and provides the `max_keys` bound implicitly.
```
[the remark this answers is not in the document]
parviz-93: Addressed in b72d26d03 on both sites. Code: dry-run no longer feeds unadmitted keys to the limiter — past the cap the would-be rejection is logged and check_key is not called, so the keyed store stays capped in observe mode too (regression test dry_run_does_not_grow_limiter_past_max_keys asserts limiter len stays within the cap). Docs: ADR-0001 now describes TTL as idle-key cleanup only and requires an explicit distributed cardinality policy; DESIGN.md documents the remaining admission/limiter desynchronization honestly — combined state is transiently bounded by ~2x max_keys right after a sweep, converging within one interval — and records the custom StateStore (admission and eviction in one structure) as the upgrade path if that window matters in production.

docs/arch/throttling/ADR/0001-distributed-throttling-cluster-cache.md
```diff
@@ -0,0 +1,207 @@
...
+* Cache unavailability handling must be an explicit per-zone policy. Default is
+  **fail-open** (serve the request, emit a warning and a metric) so the cache is
+  not a single point of failure for all traffic; operators may opt into
+  fail-closed.
+* Observability must distinguish enforced rejections from cache-error /
+  fail-open events (extend the existing `throttling.rejections` counter with a
+  reason/label), so operators can see when distributed enforcement degrades.
+* Each throttled request on a distributed zone incurs (amortized) one cache
+  round-trip plus bounded CAS retries under contention; this is acceptable for
+  identity-keyed budgets and must be load-tested for hot pre-auth IP zones.
```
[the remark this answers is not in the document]
parviz-93: Fixed in b72d26d03: the estimate now reads at least two cache operations on the success path (a get of the current arrival time followed by a compare_and_swap), plus additional get+CAS rounds under contention.

docs/arch/throttling/DESIGN.md
```diff
@@ -0,0 +1,414 @@
...
+
+## 1. Architecture Overview
+
+### 1.1 Architectural Vision
+
+ADR-0002 chose Option A — a declarative, config-owned throttling policy with a
+runtime-agnostic contract. That ADR deliberately records intent only and defers
+concrete types, config schema, and migration to a follow-up design. This is that
+follow-up, narrowed to the two items that are defects in the shipped code rather
+than shape improvements:
```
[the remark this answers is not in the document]
parviz-93: Fixed in b72d26d03: the overview now says three defects and lists the OpenAPI regression (W3) as the third.

gears/system/api-gateway/src/gear.rs
```diff
@@ -1281,64 +1322,130 @@ impl OpenApiRegistry for ApiGateway {
...
+                .as_ref()
+                .and_then(|ext| ext.get("x-throttling-rate-limit-zone"))
+                .and_then(serde_json::Value::as_str)
+                .map(str::to_owned)
+            else {
+                continue;
+            };
+            let Some(zone) = config.rate_limit_zones.get(zone_name.as_str()) else {
+                continue;
+            };
```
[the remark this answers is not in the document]
parviz-93: Good catch — old main did emit x-in-flight-limit, my "main never emitted in-flight numbers" note was wrong. Fixed in 42515fab8: the gateway enrichment now also resolves x-throttling-in-flight-limit-zone against config.in_flight_limit_zones and emits x-in-flight-limit; the test gained an in-flight-only operation asserting the numeric value (and that rate-limit extensions are absent), plus the inverse assertion on the rate-only operation.

docs/arch/throttling/DESIGN.md
```diff
@@ -0,0 +1,428 @@
...
+* Accounting is per zone (`Arc<RateZone>`), shared across both auth partitions,
+  consistent with `shared_zone_arc_across_partitions`.
+* Tests: admission refuses a new key beyond `max_keys` while admitted keys keep
+  flowing, and admission reopens after the sweep resets the set (unit +
+  end-to-end through the middleware).
+* The cap is documented as approximate: the check-then-insert is not atomic
+  (concurrent races can overshoot by a few entries), dry-run operations still
+  feed unadmitted keys to the limiter, and the sweep resets admission while the
+  limiter may retain recently-active keys. Bounding memory growth, not a
+  precise count, is the goal; making it exact would require locking the hot
```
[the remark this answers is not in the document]
parviz-93: Fixed in 42515fab8 by explicitly changing the requirement, as you suggested: section 3.2 now states the invariant as an approximate memory cap, not an exact count, with the two tolerated deviations named and bounded (admission-race overshoot; ~2x combined state transiently after a sweep, converging within one prune interval). The stale dry-run sentence is corrected to match the code and the earlier bullet: unadmitted keys never reach the limiter in any mode — enforce rejects, dry-run logs and serves without consulting it (regression-tested). Acceptance is admission behavior plus the dry-run bound, not an exact count; the custom StateStore recorded in the rejected alternatives remains the upgrade path to an exact bound.

docs/arch/throttling/ADR/0002-throttling-model-config-boundary-and-portability.md
```diff
@@ -0,0 +1,220 @@
...
+  and `dry_run` in code).
+* **Option C — Status quo** — keep per-operation `with_throttling`, the closure
+  key, and code-level binding/`dry_run`.
+
+## Decision Outcome
+
+Chosen direction: **Option A**. This ADR records the *intent* and the specific
+questions the follow-up toolkit DESIGN must resolve; concrete types, config
+schema, and migration steps are deferred to that DESIGN. The follow-up must
+decide:
```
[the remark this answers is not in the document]
parviz-93: Fixed in 42515fab8: ADR-0002's Decision Outcome now allows staged follow-ups — it names docs/arch/throttling/DESIGN.md as the stage covering the shipped-code defects (contract runtime dependency, max_keys, OpenAPI extensions) and explicitly defers the three larger decisions (config/code boundary for dry_run and binding, the declarative key model, group-level zones) to a later design, so the two documents no longer state conflicting acceptance criteria.

gears/system/api-gateway/src/gear.rs
```diff
@@ -601,20 +635,36 @@ impl ApiGateway {
...
+    fn build_router_with_pruner(
+        &self,
+    ) -> Result<(Router, Option<middleware::throttling::ThrottleKeyPruner>)> {
         // If the cached router is currently held elsewhere (e.g., by the running server),
         // return it without rebuilding to avoid unnecessary allocations.
         let cached_router = self.router_cache.load();
         if Arc::strong_count(&cached_router) > 1 {
             tracing::debug!("Using cached router");
-            return Ok((*cached_router).clone());
+            return Ok(((*cached_router).clone(), None));
```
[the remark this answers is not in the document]
parviz-93: Confirmed and fixed in 2361d10ba — and it ran deeper than the reuse branch being reachable: RouterCache::load() is ArcSwap::load_full(), so the swap always retains its own reference and Arc::strong_count > 1 was unconditionally true. The shortcut never rebuilt anything (this predates the PR — it behaved the same on main), and after our refactor it also swallowed the pruner. The shortcut is removed: the fallback path now always builds fresh and stores the pruner paired with the router it serves; build_router_with_pruner returns the pruner unconditionally (no Option). One deliberate behavior change ships with this, called out in the commit: a gateway hitting the fallback path with auth enabled and no AuthN Resolver client previously served the stale default router silently — it now fails startup fail-fast.

docs/arch/throttling/DESIGN.md
```diff
@@ -0,0 +1,434 @@
...
+  operators use for tuning.
+* The background sweep clears the admission set in the same tick it calls
+  `retain_recent()` / `shrink_to_fit()`, keeping the two views from drifting.
+  Clearing rather than selectively pruning is deliberate: without a containment
+  test on `governor`'s side there is no way to prune the two structures in
+  agreement, and a full clear is self-correcting within one interval. The known
+  residual desynchronization: keys still active enough to survive
+  `retain_recent()` stay in the limiter while admission reopens, so combined
+  state is transiently bounded by ~2× `max_keys` immediately after a sweep,
+  converging within one interval. Eliminating that window requires the
```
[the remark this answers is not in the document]
parviz-93: Partially right, and the docs are corrected in 2361d10ba. Where the scenario breaks down: an unadmitted key never reaches check_key (the admission gate precedes the limiter in both enforce and dry-run modes), so a retained-but-not-readmitted key gets no state refreshes — its GCRA state replenishes untouched and retain_recent drops it on the first sweep after the zone's replenish window W = (burst_limit + 1)/rps. Each cohort therefore dies after W; for max_keys=1 your A, B, C accumulate only while within their replenish window, then expire. Steady-state limiter size is bounded by max_keys x (1 + ceil(W/T)), T = 10s — bounded, not unbounded. You are right that our previous "~2x, converging within one interval" claim was wrong in general: it holds only for W <= T, and a zone with burst=100, rps=1 pays 11x. DESIGN.md and the admit() doc now state the parameterized bound with the derivation, and the risk section no longer claims established clients lose limiter state at the sweep. The custom bounded StateStore (admission and eviction in one structure) remains the recorded upgrade path to an exact bound.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,865 @@
...
+                KeyValue::new("kind", kind.to_owned()),
+            ],
+        );
+    }
+    tracing::warn!(
+        method = %key.0,
+        path = %key.1,
+        kind,
+        zone,
+        key = %id,
```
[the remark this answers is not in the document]
parviz-93: Thanks. We reviewed this from both the security and the operations side and are keeping the `key` field, with a doc note added in b6bc26450. The facts:

- `subject_id` is a `Uuid` by construction (`libs/toolkit-security/src/context.rs`), never an email or name. The key is never attacker-authored text: `compute_key` yields a parsed `IpAddr`, a UUID, `"unknown"` or `"anonymous"`.
- For IP zones with `trusted_proxy_hops = 0` this adds no new exposure: the access log already emits `remote_addr_ip` at `info` for every request, including 429s.
- The #4171 review asked to raise rejections to `info` and keep the key out of the message body (https://github.com/constructorfabric/gears-rust/pull/4171#discussion_r3598662637). That is what the code does: the key is a structured field, never message text and never a metric attribute. The toolkit sinks are JSON, so a pipeline can redact or shorten retention for that one field. The request id is attached through the `http_request` span.
- Removing or hashing the key would defeat dry-run. The counters carry only `zone`/`kind`; operators need to see how many distinct clients would be throttled before they switch a zone to enforce.

The honest part of the finding: for identity zones and for `trusted_proxy_hops >= 1` this is the first `info`/`warn` path that exposes the subject UUID or the XFF client IP. We now document that on both functions and in DESIGN.md. We rate this Minor rather than Major.

gears/mini-chat/mini-chat/src/api/rest/routes/chats.rs
```diff
@@ -19,6 +19,20 @@ pub(super) fn register_chat_routes(
...
+        // runs before auth (require_security_context = false).
+        //
+        // dry_run: the gateway observes limits and logs would-be rejections but
+        // never returns 429, so functional tests and clients are not throttled.
+        // Operators can enforce for real by flipping this to false.
+        .with_throttling(ThrottlingSpec {
+            rate_limit_zone: Some("rl_mini_chat_chat".to_owned()),
+            in_flight_limit_zone: Some("ifl_mini_chat_chat".to_owned()),
+            require_security_context: false,
+            dry_run: true,
```
aviator5, line 34: Show we have such throttling configuration parameters in the gear's config according to `docs/arch/throttling/ADR/0002-throttling-model-config-boundary-and-portability.md`?

aviator5:
I'd like us to clarify the operational model and next steps for throttling. This PR replaces the existing per-route limiter with named, per-key zones. Since the project is pre-release, removing the old default policy is not a compatibility concern.

I suggest addressing these changes introduced by this PR:

1. **The new dry-run mode waits for an in-flight slot.** The middleware calls `gate.acquire(backlog_timeout).await` before checking `dry_run`. This can delay a request or exhaust the gateway request timeout even when enforcement is disabled. It should record a would-reject event and continue immediately when no slot is available.
2. **The new identity-keyed zones need validation against the effective authentication policy.** `require_security_context` only selects the middleware stage; it does not require authentication. A public route gets an anonymous context, so all its clients share one identity bucket. I suggest rejecting identity-zone bindings for routes that allow anonymous access, and for `auth_disabled`, at startup. An unexpectedly missing context at runtime should produce an internal error instead of silently falling back to an anonymous key.
3. **Proxy-header trust needs an explicit boundary.** When `trusted_proxy_hops > 0`, the gateway selects an address from `X-Forwarded-For` by position, with an `X-Real-IP` fallback, but does not verify that the immediate peer is a trusted proxy. If the gateway is reachable directly, a client can supply these headers and choose or rotate its IP throttling key. Should we validate trusted proxy addresses/CIDRs, or require and document an enforced deployment topology where only trusted proxies can reach the gateway and forwarding headers are sanitized? The fallback behavior should follow the same trust policy.

There are also two pre-existing issues that this rewrite carries forward: in-flight permits are released when the response future returns, before a streaming body finishes, and automatic `Retry-After` uses `.as_secs()`, rounding positive subsecond delays down to zero. These are not regressions introduced by this PR. They still deserve fixes or explicit follow-up issues: retain the permit through body completion/cancellation and round `Retry-After` up.

I also propose changing the execution order to **in-flight admission → rate check → handler**. The current rate-before-in-flight order is inherited from the previous middleware, but it spends rate quota on requests subsequently rejected because concurrency is exhausted. As a result, a client can still be rate-limited after an in-flight slot becomes available, even though its intervening requests never reached the handler. Acquire the in-flight permit first (including any configured backlog wait), then check the rate limit immediately before invoking the handler, and release the permit if that check rejects the request. A regression test should verify that an in-flight rejection does not consume rate quota.

For the next iteration, can we agree on the following policy model and capture it in a concrete design?

- **Configuration-owned route bindings and dry run.** Operators should be able to start throttling an existing endpoint, change its zone bindings, and enable enforcement without rebuilding. Do we extend `route_policies.rules` or introduce dedicated throttling rules? We need to define matching, precedence, exclusions, and how configuration updates are applied, including whether a restart is required. Dry run should be configurable per zone/policy so rate and in-flight limits can be rolled out independently.
- **Composable zones and key strategies.** We currently have only one rate zone and one in-flight zone per operation, and keys are limited to IP and `subject_id`. Useful cases include a global service budget combined with per-client limits and a key derived from a configured header. The design should also define how zones can be applied before and after authentication on the same endpoint.
- **Selective key filtering.** Currently, `excluded_keys` supports only exact string matches and only for in-flight zones. Rate zones have no key exclusions, and neither zone kind supports `included_keys` or glob matching. We should support both use cases explicitly: bypass throttling for selected keys, and apply throttling only to selected keys. These policies should work consistently for rate and in-flight zones, with defined matching semantics, including whether glob patterns are supported.
- **Dry-run traffic can affect enforced routes.** Routes referencing the same zone share state even when their `dry_run` settings differ. For the same computed key, a dry-run request consumes available rate tokens and can hold in-flight permits/backlog slots. It can therefore cause another operation with enforcement enabled to return a throttling error. Fixing the dry-run wait alone does not remove this interaction. We should either reject mixed dry-run/enforcement bindings to shared state at startup or isolate observational state from enforced budgets; the intended behavior needs an explicit decision and a regression test.
- **Explicit accounting semantics.** We should define whether budgets can be shared or independent per route. Also, reaching `max_keys` rejects new clients even if they have not exceeded any per-client limit; this saturation behavior needs to be documented.
- **Actionable observability.** The rejection and dry-run counters already expose `zone` and `kind`. It would help to distinguish a full backlog from a backlog timeout in metrics and structured logs, and identify the affected rule/operation in metrics. Queue length and wait duration would make tuning easier. The response currently uses the same error details for rate and in-flight rejection; a structured reason would help clients distinguish them too.

ADR-0002 points toward configuration-owned policy and group-level zone bindings, but its status is still `proposed`, and the concrete schema and migration are deferred. The current DESIGN also leaves this work out of scope.

I propose creating an **umbrella issue for the throttling roadmap**, linking this PR and both ADRs, with sub-issues for:

- Correctness: dry-run behavior and isolation from enforcement, identity/auth validation, proxy-header trust, limiter ordering, streaming permit lifetime, and `Retry-After`.
- Configuration-owned policy: route rules, zone bindings, per-zone/policy dry run, configuration update behavior, and migration from `with_throttling`.
- Zone composition and keys: global and per-client budgets, multiple zones per operation, pre-/post-auth application, header keys, and include/exclude filters.
- Accounting and observability: sharing semantics, `max_keys` saturation, rejection reasons, and backlog metrics/logging.
- Distributed rate limiting: the cluster-cache backend proposed in ADR-0001, including consistency and failure behavior.

Each sub-issue should have a defined scope, acceptance criteria, priority, and dependencies. The umbrella should distinguish work completed by this PR, fixes required before merging, and later milestones, and identify which design decisions still need agreement. This would make the development direction and delivery sequence visible beyond the ADRs.

There is an existing [Go throttling implementation](https://github.com/acronis/go-appkit/tree/master/httpserver/middleware/throttle) that can serve as a reference for supported use cases and behavior. We can use it to inform the design while keeping the Rust API and implementation idiomatic.

@parviz-93, @MikeFalcon77 what do you think?

gears/mini-chat/mini-chat/src/api/rest/routes/chats.rs
parviz-93: Not in this PR, on purpose. The goal here was to finish the abandoned #4171, which was almost complete, and fix its three defects. The zone limits are in the gateway config. The operation-to-zone binding and `dry_run` stay in code. ADR-0002 is `proposed` and defers config-owned bindings to a follow-up design: route matching, precedence, and config update semantics are not decided yet. DESIGN.md §4 lists this move as out of scope for the same reason. If you think the binding must move now, I am open to revisit the ADR and the design before this PR merges.

parviz-93:
@aviator5 thanks. I checked each point against the code. All nine technical findings are correct. All of them come from the original #4171 commit, not from the follow-up commits in this PR. The PR is mine now, so I fix the small ones here and put the rest in the roadmap.

**Confirmed:** 1 (dry-run waits), 2 (identity zones not checked against auth), 3 (no peer check for proxy headers), 6 (rate before in-flight), 7 (shared zone mixes dry-run and enforce), 8 (`excluded_keys`), 9 (same error body).
**Confirmed, pre-existing:** 4 (permit released before the body ends), 5 (`Retry-After` rounds down). Both exist in the old `rate_limit.rs` too.

**Fixed in `ab2ebb863`**

1. Dry-run uses `try_acquire` and never waits in the backlog. If no permit is free, the request is counted and served at once.
2. An identity zone on an anonymous operation, or with `auth_disabled: true`, fails the router build. A missing `SecurityContext` at runtime returns an internal error. The `"anonymous"` fallback is removed.
3. `Retry-After` rounds up. A sub-second wait gives `1`.
4. Mixed dry-run and enforce bindings on one zone fail the router build. This is my pick of your two options. State isolation goes to the roadmap. Objections welcome.
5. The 429 body gets the kind (`rate_limit`, `max_keys`, `in_flight`) and `retry_after_seconds`. `subject` stays `"throttling"`.
6. Docs: with `trusted_proxy_hops > 0` the gateway must sit behind a trusted proxy that sanitizes forwarding headers, `X-Real-IP` included. `max_keys` saturation and the shared-zone rule are documented.

**Your decision**

- Order change (in-flight → rate → handler). Feasible here, with a test that an in-flight rejection does not use rate quota. I do it in this PR if you and @MikeFalcon77 agree.
- Trusted proxy list. A std `IpAddr` prefix match needs no new crate. I suggest a sub-issue.
- Permit lifetime through the body. A body wrapper that owns the permit is moderate work. I suggest a sub-issue.

**Roadmap**

I agree with the umbrella issue and the five sub-issues: correctness, config-owned policy (ADR-0002), zone composition and keys, accounting and observability, distributed limiting (ADR-0001). `go-appkit` throttle is a good reference. After you confirm the split, I can draft the issues.

gears/system/api-gateway/src/config.rs
```diff
@@ -392,6 +410,231 @@ pub struct RoutePolicyRule {
...
+            Self::Seconds(n) => serializer.serialize_u64(*n),
+        }
+    }
+}
+
+/// The keying strategy used by a throttling zone.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum KeyType {
+    /// Key by a per-operation identity (supplied in code via `ThrottlingSpec`).
```
MikeFalcon77: **HIGH**

This says the identity is "supplied in code via `ThrottlingSpec`", but `ThrottlingSpec` doesn't carry or supply an identity, it only binds a zone name and a couple of flags. `compute_key` in `throttling.rs` always derives the key from `SecurityContext::subject_id()`, independent of anything in `ThrottlingSpec`.

gears/system/api-gateway/src/gear.rs
```diff
@@ -238,6 +243,11 @@ impl ApiGateway {
 
     /// Force rebuild and cache of the router.
     ///
+    /// Replaces only the cached router; the serving router (and the throttling
+    /// key pruner paired with it, handed to `serve`) is unaffected, so the
+    /// rebuild's own pruner is deliberately discarded rather than allowed to
```
MikeFalcon77, line 248: **LOW**

Withdrawing the original framing: the claim that a hot-swapped cached router would serve traffic with unpruned zones does not hold, because the cached router never serves traffic at all.

The actual finding is that `RouterCache` is a dead mechanism. `rebuild_and_cache_router` and `get_cached_router` have no callers anywhere in the tree; the serving path is `get_or_build_router`, which keeps the router/pruner pair together (`gear.rs:758`). `build_router_with_pruner` also unconditionally stores into `router_cache` on every call, so the cache is written on every build and read by nobody.

The doc comment here already states the discard is deliberate, so the code is fine as written. Suggest either removing `RouterCache`, `rebuild_and_cache_router` and `get_cached_router`, or — if a hot-swap path is genuinely planned — leaving a note that wiring the cached router into serving must also carry its pruner, since today that pairing is only maintained on the `get_or_build_router` path.

gears/system/api-gateway/src/gear.rs
```diff
@@ -601,22 +635,39 @@ impl ApiGateway {
             crate::middleware::request_id::MakeReqId,
         ));
 
-        Ok(router)
+        Ok((router, throttle_key_pruner))
```
MikeFalcon77, line 638: **LOW — pre-existing, out of scope for this PR**

Correcting the original claim: this PR does not rework the layer stack. The only change to `apply_middleware_stack` is replacing the single `// 10) Per-route rate limiting` layer with two throttling layers (`9b` pre-auth, `11b` post-auth). CORS, MIME validation, auth, error mapping and the rest are untouched, and `apply_middleware_stack` already exists on `main` (there at `gear.rs:361`).

The underlying fact still stands: a grep over `gears/system/api-gateway/src` and `libs/toolkit/src` finds no `Strict-Transport-Security`, `X-Content-Type-Options`, `X-Frame-Options`, CSP or `Referrer-Policy`, and nothing strips `Server` / `X-Powered-By`. But that gap predates this PR and is unrelated to throttling.

Not a blocker here. Worth a separate issue for a `SetResponseHeaderLayer`-based security-headers layer, unless these are already terminated at the ingress — in which case a one-line note in the middleware-order comment saying so would close the question.

gears/system/api-gateway/src/gear.rs
```diff
@@ -601,22 +635,39 @@ impl ApiGateway {
...
 
+    /// Build the router together with the throttling key pruner created from
+    /// the same zone instances.
+    ///
+    /// Always builds fresh. A previous revision short-circuited to the cached
+    /// router when `Arc::strong_count(...) > 1`, but `RouterCache::load()` is
+    /// `ArcSwap::load_full()` — the swap always retains its own reference, so
+    /// the count was at least 2 and the "reuse" branch was taken
+    /// unconditionally, returning the stale cached router with no pruner.
+    fn build_router_with_pruner(
```
MikeFalcon77, line 668: **HIGH**

The doc comment here says a previous revision unconditionally reused the stale cached router because `RouterCache::load()` keeps its own strong reference, so `strong_count > 1` was always true, and that this is now fixed by always rebuilding. No test calls `build_router()` twice with a spec change in between to confirm the second call reflects the new spec instead of the old cached one.

gears/system/api-gateway/src/gear.rs
```diff
@@ -715,6 +774,17 @@ impl ApiGateway {
...
 
+        // Bound the throttling keyed stores: spawn the periodic pruner produced
+        // during middleware build, tied to the lifecycle token so it stops on
+        // shutdown. Without this the keyed stores grow one entry per distinct
+        // (attacker-influenced) key with no eviction, and pruning off the hot
+        // path avoids per-request all-shard write-locking scans under a flood.
+        if let Some(pruner) = self.throttle_key_pruner.lock().take() {
+            // The task self-terminates when `cancel` fires, so the handle is
+            // intentionally discarded rather than joined.
+            drop(pruner.spawn(cancel.clone()));
```
MikeFalcon77: **LOW**

Correcting the impact stated in the original version of this comment — it was inverted.

The premise stands: the sweep task's `JoinHandle` is dropped, so if the task panics it stops silently and nothing reports it. But the consequence is not unbounded keyed-store growth. Both zone kinds cap growth on the admission path (`RateZone::admit`, `InFlightZone::gate`), and in production the pruner task is the *only* caller of `reset_admitted` / `prune_idle_keys`. With the task dead, `admitted_len` saturates at `max_keys` and never resets, so every never-seen key is refused from then on with `reason = "max_keys"` (`throttling.rs:720-726` and `:746-752`) for the remaining life of the process.

So the failure mode is the opposite of a leak: a permanent self-inflicted denial of service for new clients, silent except for the rejection metric. Memory stays bounded.

Panic likelihood is low — the loop body has no `unwrap`, no indexing, only governor/DashMap calls — so this is a robustness nit, not a blocker. If you want to close it cheaply, either log on task exit (`if let Err(e) = handle.await` in a small supervisor, or a `defer`-style `tracing::error!` at the end of the task body) or wrap the sweep body so a panic in one zone does not take the ticker down.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1000 @@
...
+    /// builds an isolated partition (its zone runtimes are not shared with any
+    /// pre-auth map) and is intended for standalone use such as tests.
+    ///
+    /// # Errors
+    /// Returns an error if an entry references an undefined zone, an invalid
+    /// (e.g. zero-limit) zone, an identity-keyed zone bound to an operation
+    /// that allows anonymous access, or an identity-keyed zone while
+    /// `auth_disabled` is true. The dry-run consistency check is skipped here;
+    /// only [`build_maps`] performs it.
+    pub fn from_specs(specs: &[OperationSpec], cfg: &ApiGatewayConfig) -> Result<Self> {
```
MikeFalcon77: **HIGH**

`from_specs` is `pub` on both `ThrottlingMap` and `ThrottlingMapNoAuth`, but the doc comment says it's "intended for standalone use such as tests" and explicitly skips the dry-run consistency check and zone sharing that `build_maps` provides. The only caller is `throttling_tests.rs`, which is a child module of this file and already has access to private items, so this doesn't need to be `pub` at all.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1000 @@
...
+    headers.insert("X-RateLimit-Remaining", h.remaining.clone());
+}
+
+/// Compute the throttling key for a request according to the zone key type.
+///
+/// Returns `None` only for an identity-keyed zone without a `SecurityContext`.
+/// Startup validation guarantees such zones run after authentication on
+/// authenticated operations, so `None` is an invariant violation, not a limit;
+/// dry-run does not soften it.
+fn compute_key(kind: KeyType, req: &Request, trusted_proxy_hops: usize) -> Option<String> {
```
MikeFalcon77, line 828: **HIGH**

No test ever calls `compute_key(KeyType::Identity, ...)` with a `SecurityContext` present in the request. Every identity-keying test (`compute_key_identity_requires_security_context`, `identity_zone_without_context_is_internal_error`) only exercises the missing-context branch. The actual success path, resolving to `subject_id()`, has no assertion anywhere.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -376,8 +376,10 @@ pub struct OperationSpec {
     /// Independent of [`authenticated`](Self::authenticated) — an exposed route
     /// may still require a JWT.
     pub exposed: bool,
-    /// Optional rate & concurrency limits for this operation
-    pub rate_limit: Option<RateLimitSpec>,
+    /// Optional zone-based throttling configuration for this operation.
+    /// Binds the operation to gateway throttling zones and supplies the
+    /// identity extractor for identity-keyed zones.
```
MikeFalcon77: **HIGH**

This says `ThrottlingSpec` "supplies the identity extractor for identity-keyed zones", but the struct only has `rate_limit_zone`, `in_flight_limit_zone`, `require_security_context` and `dry_run`, nothing that supplies an extractor. Identity keying is hardcoded to `SecurityContext::subject_id()` in `throttling.rs`'s `compute_key`, regardless of what an operation configures.

gears/system/api-gateway/src/middleware/throttling.rs
```diff
@@ -0,0 +1,1000 @@
...
+    inflight: Arc<Semaphore>,
+    backlog: Arc<Semaphore>,
+}
+
+impl KeyGate {
+    /// Acquire an in-flight permit, optionally waiting in the backlog.
+    ///
+    /// Returns `None` when the request should be rejected (no in-flight slot and
+    /// either no backlog capacity or the backlog wait timed out).
+    async fn acquire(&self, backlog_timeout: Duration) -> Option<OwnedSemaphorePermit> {
```
MikeFalcon77, line 142: `acquire` takes a backlog slot then races a timed wait for the in-flight permit, and it's driven from `enforce`, which runs under the gateway's outer 30s tower `timeout` layer, so this future can be dropped mid-poll on a slow request. There's no `// cancel-safe:` note here saying that a drop here cleanly releases `_backlog_slot` and never leaves an orphaned in-flight permit.

MikeFalcon77 (commented):
9 findings from nonstop-reviewer (run 8).

gears/system/api-gateway/src/config.rs
parviz-93: Fixed in 05a5606. The `KeyType::Identity` doc now says that the key is `SecurityContext::subject_id()`. The old wording came from the removed `IdentityKeyFn`.

libs/toolkit/src/api/operation_builder.rs
parviz-93: Fixed in 05a5606. The doc now says that the spec binds the operation to zones and carries `require_security_context` and `dry_run`. The "identity extractor" sentence is removed. It came from the same removed `IdentityKeyFn`.

gears/system/api-gateway/src/middleware/throttling.rs
parviz-93: Fixed in 05a5606. `from_specs` on both maps is now private and test-only. `throttling_tests.rs` is a child module, so it has direct access.

gears/system/api-gateway/src/middleware/throttling.rs
parviz-93: Fixed in 05a5606. A new test inserts a `SecurityContext` into the request extensions. It makes sure that `compute_key(KeyType::Identity, ..)` returns the `subject_id()` string.

gears/system/api-gateway/src/gear.rs
parviz-93: Fixed in 05a5606. A new regression test builds the router once. Then it stores a poisoned router with a `/stale` route into `router_cache`. Then it builds again and makes sure that the second router returns 404 on `/stale`. This proves that the second build is fresh and not from the cache. The fallback router has no main routes, so a spec change is not visible there. The poisoned cache targets the exact bug from the doc comment.

gears/system/api-gateway/src/middleware/throttling.rs
parviz-93: Added a `// cancel-safe:` note in 05a5606. Both permits are `OwnedSemaphorePermit` RAII guards, and `acquire_owned` in tokio is cancel-safe. If the future drops during the wait, `_backlog_slot` is released and no in-flight permit is lost.

gears/system/api-gateway/src/gear.rs
parviz-93: Agreed on the failure mode: silent saturation at `max_keys`, not a leak. In 05a5606 a supervisor task now joins the pruner. If the pruner panics, the supervisor logs a `tracing::error!` and restarts the pruner after 1 s. As a result, eviction continues and the process does not stay at `max_keys` saturation.

gears/system/api-gateway/src/gear.rs
parviz-93: Confirmed. `RouterCache`, `get_cached_router` and `rebuild_and_cache_router` have no callers outside `gear.rs`, and all three already exist on `main`. Their removal is not related to throttling, so this PR keeps its scope. If you do not mind, I can open a separate issue to remove the dead cache.

gears/system/api-gateway/src/gear.rs
parviz-93: Agreed. This gap exists before this PR and is not related to throttling. If you do not mind, I can open a separate issue for a security-headers layer based on `SetResponseHeaderLayer`.

aviator5:
@parviz-93 thanks for the details in https://github.com/constructorfabric/gears-rust/pull/4701#issuecomment-5659097866

> + Order change (in-flight → rate → handler). Feasible here, with a test that an in-flight rejection does not use rate quota. I do it in this PR if you and @MikeFalcon77 agree.

Actually, I would provide configuration for this (dedicated sub-issue) and keep current behavior as is for now.

> + Trusted proxy list. A std IpAddr prefix match needs no new crate. I suggest a sub-issue.

Agree

> + Permit lifetime through the body. A body wrapper that owns the permit is moderate work. I suggest a sub-issue.

Agree

MikeFalcon77:
> @parviz-93, @MikeFalcon77 what do you think?
Overall I agree with the proposed direction and the roadmap split.

One point I'd like to clarify before changing the middleware order though: I'm not sure `in-flight -> rate -> handler` is always preferable.

With rate limiting first, excessive traffic is rejected before it can consume in-flight/backlog capacity. If we acquire/wait for an in-flight slot first, traffic that will eventually be rate-limited can still occupy the backlog and potentially affect legitimate requests.

So I think this depends on the accounting semantics we want:

* if rate limit means requests admitted by the gateway, the current `rate -> in-flight -> handler` order looks reasonable;
* if it means requests actually reaching the handler, then `in-flight -> rate -> handler` is correct.

For a general-purpose gateway I would lean toward protecting the in-flight queue with the rate limiter, but in any case I think we should define this explicitly in DESIGN.md before changing the order.

I agree with keeping trusted proxy validation as a follow-up issue.

I also agree that holding the in-flight permit through streaming body completion can be a follow-up, but we should explicitly document the current limitation, since for SSE/streaming the current in-flight limit does not represent the number of actually active requests.

The proposed umbrella issue and sub-issue split look good to me.

aviator5 (approved):

docs/arch/throttling/DESIGN.md
```diff
@@ -0,0 +1,483 @@
...
+  custom-`StateStore` variant below, where admission and eviction live in one
+  structure by construction.
+
+Cost: one extra hash lookup on the throttling path, and a duplicated copy of the
+key string per tracked key. The duplication is the price of a bound that
+`governor` cannot provide; peak memory becomes `O(max_keys)` in the admission
+set and `O(max_keys × (1 + ⌈W/T⌉))` in the limiter — zone-config-dependent, but
+no longer unbounded in either.
+
+**Rejected alternatives**
```
aviator5, line 305: Does it worth to make ADR doc for this decision?

docs/arch/throttling/DESIGN.md
```diff
@@ -0,0 +1,483 @@
...
+  partition, or the bound doubles.
+* The distributed backend (ADR-0001) will replace per-key state entirely for
+  rate zones. Whatever accounting is added here must be cheap to delete.
+
+## 3. Key Decisions
+
+### D1: Remove the closure rather than replace it with `enum Key`
+
+ADR-0002 proposes replacing `IdentityKeyFn` with a serializable
+`enum Key { Ip, Header(String), JwtClaim(String), … }`. This design does **not**
```
aviator5, line 119: I think we will need to have some mechanism for implementing custom throttling, probably via plugins.
It's not a blocker for now, just an idea for the future api-gateway scope.

MikeFalcon77:
Approved with comment

MikeFalcon77 (approved):

gears/system/api-gateway/src/gear.rs
MikeFalcon77: `gears/system/api-gateway/src/gear.rs:253` — `rebuild_and_cache_router` still calls `build_router()`, and `build_router` still drops the paired pruner at line 656 (`let (router, _pruner) = self.build_router_with_pruner()?;`). I confirmed there is no caller outside `gear.rs`, but both methods are `pub` on `ApiGateway`, so the unpruned-zone path stays reachable from outside the crate. Minimal fix: make `rebuild_and_cache_router` also store the fresh pruner into `throttle_key_pruner` and respawn the sweep, or make the cache API private/removed in this PR.

gears/system/api-gateway/src/gear.rs
MikeFalcon77: `gears/system/api-gateway/src/gear.rs:638` — `apply_middleware_stack` still ends at the request-id layers with no response-header layer, and a search for `strict-transport`, `x-content-type-options`, `x-frame-options`, `content-security-policy`, `referrer-policy` or `SetResponseHeaderLayer` across `gears/system/api-gateway/src` returns no matches. Minimal fix: add a `SetResponseHeaderLayer` group (HSTS, nosniff, frame-options, referrer-policy, CSP) alongside the CORS layer at line 528.

parviz-93:
@aviator5 @MikeFalcon77 the follow-up issues from this review are filed. Please make a short check that the split and the open questions match what you asked for.

Umbrella: #4879 — throttling roadmap (what shipped, five open questions, order of work)

Sub-issues:
- #4880 — the in-flight limit does not cover streaming response bodies (documentation, first)
- #4881 — throttling accounting semantics and backlog observability are undefined (remaining scope of #4754)
- #4882 — throttling correctness follow-ups: proxy trust, dry-run state, permit lifetime, limiter order
- #4883 — throttling route bindings cannot be set from configuration (remaining scope of #4755)
- #4884 — throttling supports only one zone per operation and two key types (remaining scope of #4756)
- #4885 — record the bounded key-admission decision as an ADR (answers the `DESIGN.md:305` question)
- #4886 — throttling has no distributed backend (ADR-0001, last)

Cleanup, independent of the roadmap:
- #4887 — `RouterCache` and its accessors are dead code
- #4888 — responses carry no security headers

#4754, #4755 and #4756 stay open. Each one is linked from the sub-issue that carries its remaining scope.

docs/arch/throttling/DESIGN.md
parviz-93: Yes. Filed as #4885: promote the bounded key-admission decision (admission set with a full clear, not a custom `governor` `StateStore`) to ADR-0003, with the consequences and the upgrade path.

docs/arch/throttling/DESIGN.md
parviz-93: Agreed as a future item. It is listed in the umbrella #4879 as a backlog idea without its own issue. When a consumer for custom throttling appears, I can open a dedicated issue.

PR #4710, opened 2026-09-04
docs(settings-service): refine service design

shaneemberton:
## Description

  This PR aligns the Settings Service PRD and design around three related decisions.

  ### Per-tenant setting access

  - Replace declaration-wide `tenant_visible` / `tenant_overridable` flags with sparse restrictions on each `(setting, tenant)` pair.
  - Store only `read_only` and `hidden`; absence of a row means `overridable`.
  - Compute effective access as the strictest restriction on the root-to-tenant path.
  - Allow only strict ancestors to set or clear a tenant's restriction.
  - Check the writer's effective access when managing descendant values.
  - Preserve stored values and permission rows across restriction changes and declaration retire/revive.
  - Require `If-Match` for permission writes so stale decisions cannot overwrite newer ones.

  ### Declaration evolution and revival

  - Define `POST /declarations` in terms of the version-stripped setting identity.
  - Evolve an active declaration by creating the next major version, copying and re-validating its values, and retiring its predecessor.
  - Keep exactly one active major for a setting.
  - Treat exact retries and metadata-only differences as conflicts instead of accidentally creating another major.
  - Revive an exact retired declaration when no active version exists, preserving its UUID and retained state.
  - Re-validate retained values during revival and reject changes that cross the secret-storage boundary or alter Scope Class.
  - Reject module-contributed value-type changes without an explicit major-version change.

  ### GTS ownership

  - Use the gear-owned abstract base type `gts.cf.core.settings.setting_type.v1~` for setting keys.
  - Move Settings gear control-plane entities, events, errors, and authorization resources under `gts.cf.core.settings.*`.
  - Keep toolkit-owned value-schema types under `gts.cf.toolkit.settings.type_*~`.
  - Synchronize the Settings Activation authorization model with the new namespace.

  This PR changes documentation and proposed contracts only; it contains no Rust implementation or crate-version changes.

  ## Type of Change

  - [ ] Bug fix
  - [ ] New feature
  - [ ] Breaking implementation change
  - [x] Documentation update

  ## Testing

  - [x] `git diff --check`
  - [x] PRD and design consistency reviewed manually
  - [x] GTS namespaces checked across the Settings Service documentation
  - [ ] Unit tests — not run; Markdown-only changes
  - [ ] Integration tests — not run; Markdown-only changes

  ## Documentation

  - [x] Settings Service PRD updated
  - [x] Settings Service design updated
  - [x] Settings Activation design synchronized
  - [x] No README, rustdoc, or generated API documentation changes required

  ## Checklist

  - [x] Documentation follows the project conventions
  - [x] Self-review completed
  - [x] No dependencies or executable code changed
  - [x] Version bump is not applicable because no crate or implemented public surface changed

  ## Related Issues

  N/A

## Summary by CodeRabbit

* **Documentation**
  * Clarified the per-setting, per-tenant access model, including read-only, overridable, and hidden states.
  * Documented scope enforcement, inheritance rules, tenant overrides, and preconditions for changing restrictions.
  * Defined activation visibility and status counts using retained await-records, with one record per subscriber.
  * Clarified authorization responses: `403` for denied read permission and `404` for out-of-scope or entirely invisible change sets.
  * Documented declaration ownership, lifecycle, and compatibility checks through the Types Registry.
  * Updated value encoding guidance to reference JSON Schema draft-07.


gears/settings-service/docs/DESIGN-activation.md
```diff
@@ -763,11 +763,11 @@ The load-bearing per-`(change_set_id, subscriber, key)` state. Created at publis
 
 #### Authorization model
 
-Mirrors the Settings Service model (design DESIGN.md §4.8); enforced server-side via `PolicyEnforcer` over the AuthZ Resolver (fail-closed). Activation is part of the Settings Service, so its control-plane resources sit under `gts.cf.toolkit.settings.*`.
+Mirrors the Settings Service model (design DESIGN.md §4.8); enforced server-side via `PolicyEnforcer` over the AuthZ Resolver (fail-closed). Activation is part of the Settings gear, so its gear-owned control-plane resources sit under `gts.cf.core.settings.*`.
 
 | Operation | Required permission | Scope | Unauthorized |
 |-----------|---------------------|-------|--------------|
-| Read activation facet / responses (`GET /settings-service/v1/change-sets/{change_set_id}/activation`, `…/responses`) | `read` on `gts.cf.toolkit.settings.change_set.v1~` | The change set's tenant subtree | `404` if not visible |
+| Read activation facet / responses (`GET /settings-service/v1/change-sets/{change_set_id}/activation`, `…/responses`) | `read` on `gts.cf.core.settings.change_set.v1~` | The change set's tenant subtree | `404` if not visible |
```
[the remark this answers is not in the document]
shaneemberton: Taken, with the shape stated rather than implied.

The facet is now **defined over the members the caller may see** — not a wider answer with a filter applied to it. Hidden members contribute no row to `/responses`, the counts cover the same members, and `overall_status` is the status of that set: `success` means every member the caller can see was activated, and claims nothing about the rest. §4.3 *Reading a change set the caller sees only part of*, with the authorization row pointing at it.

Two notes on the reasoning, since neither is obvious from the diff.

**The counts are filtered for consistency, not because a count is dangerous.** A count names nothing on its own. But a truthful count beside a filtered list is worse than either: `expected_records = 10` against seven rows reports exactly what was withheld and how much of it, and watched over time it reports how often an ancestor changes settings the reader cannot see. One consistent view discloses neither, and it matches what the settings listing already does — absent, not marked, not counted in the total (DESIGN.md §4.8).

**A hidden member that fails is not silently swallowed.** It was written by an ancestor, who sees the whole change set — keys and failures alike — so the failure is visible to exactly the party able to diagnose it. Reporting `failed` to a caller who cannot see the member would give them an alarm they can neither explain nor clear, on top of disclosing that the member exists.

Not taken: the alternative of denying the read outright when the caller cannot see every member. It would cost that caller the activation view of their **own** settings whenever an ancestor happened to mix visible and hidden ones into one request, which is a normal way for a provider to work.

Two tests cover it — a mixed change set read by the restricted caller and by the platform administrator, and a hidden failure alongside a visible success.

gears/settings-service/docs/DESIGN.md
```diff
@@ -391,9 +391,9 @@ Settings Service is **supplied as a Constructor Fabric Gear** — an SDK crate p
...
+The setting `key` is a **GTS type identifier** `gts.cf.core.settings.setting_type.v1~<vendor>.<package>.<category>.<name>.v1~` for **both** authors. The base type is owned and registered by the Settings gear; toolkit does not own it. The derived half is the setting and has four segments before the version — `<vendor>.<package>.<namespace>.<type>` per the GTS grammar (§4.7), with no `gts.` prefix of its own. The category is always the third: a **module** provides its own half and the category is extracted from `<namespace>`; an **admin** setting uses `<vendor>.settings.<category>.<name>.v1`. The trailing `~` makes the key a type rather than an instance, which lets a policy name one setting (§4.8). The value's schema is separate: `value_type_id` names a toolkit-owned catalog type.
```
Artifizer:
```suggestion
The setting `key` is a **GTS type identifier** `gts.cf.core.settings.setting_type.v1~<vendor>.<package>.<category>.<name>.v1~` for **both** authors. The base type is owned and registered by the Settings gear. The derived types can be registered by other gears specifying `<vendor>.<package>.<namespace>.<type>` per the GTS grammar (§4.7).
```

Artifizer: I think it's an overkill to repeat such GTS specifics in every PRD, so I'd make it shorter

gears/settings-service/docs/DESIGN.md
```diff
@@ -628,15 +655,13 @@ erDiagram
  }
  SettingDeclaration {
  uuid id PK
- string key "GTS type id: settings.value.v1~<derived>~ (both authors)"
+ string key "GTS type id: core.settings.setting_type.v1~<derived>~ (both authors)"
```
Artifizer, line 658: Services must not store raw GTS strings in database and use GTS type UUID obtained from Type Registry SDK. The reasons are that GTS string is a "presentation" layer:
1. it may change or may have an alias in the future and we do not want to have migration
2. UUIDs are shorter

@aviator5, we probably need to have a static validation rule (python script for md files, dylint for Rust code) to prohibit use of GTS as strings

gears/settings-service/docs/DESIGN.md
```diff
@@ -816,13 +847,15 @@ Defaults (`default_value`) are re-validated against the new value type the same
 
 **Worked example.** A `bool` setting upgraded to a `string`:
 
-```
-before: gts.cf.toolkit.settings.type_bool_flag.v1~cf.toolkit.settings.cat.sett1.v1
-after: gts.cf.toolkit.settings.type_string.v1~cf.toolkit.settings.cat.sett1.v2
- └──── new value type ────────────────────┘└─ same stripped path, new major ─┘
+```text
+before key:           gts.cf.core.settings.setting_type.v1~cf.toolkit.cat.sett1.v1~
+before value_type_id: gts.cf.toolkit.settings.type_bool_flag.v1~
```
Artifizer, line 852: I just realized probably some design flaw that we missed earlier. 

I remember we discussed to have settings as GTS, but it automatically means it must be a struct and never a 'value', so then it seems like we do not need to keep `value_type_id` at all

1. If somebody needs to define just a `bool` setting, it would probably look like this:

https://github.com/GlobalTypeSystem/gts-spec/tree/main/examples/settings

Base setting type definition:
https://github.com/GlobalTypeSystem/gts-spec/blob/main/examples/settings/types/gts.x.core.settings.type.v1~.schema.json

Derived setting type definition (specific setting type):
https://github.com/GlobalTypeSystem/gts-spec/blob/main/examples/settings/types/gts.x.core.settings.type.v1~x.platform._.feature_new_dashboard.v1~.schema.json

Setting value example:
https://github.com/GlobalTypeSystem/gts-spec/blob/main/examples/settings/instances/gts.x.core.settings.type.v1~x.platform._.feature_new_dashboard.v1~.examples.jsonc

2. Then, since GTS schema is a JSON schema we can control backward compatible/incompatible schema changes. Backward incompatible changes would allow a major version bump. Also, in general, such changes may require a complicated migration and I think it must be executed by some external workflows with tenant hierarchy traversal and proper content "transformation". These workflows from my perspective must be out of settings service responsibility and developed and executed by the vendors that define actual setting values and then change it. So I would keep settings service simple - as JSON object storage with some semantic rules

Artifizer (changes requested):

gears/settings-service/docs/DESIGN.md
shaneemberton: Two things.

**On the rename.** Storing the UUID gives us nothing here. The GTS key maps to the UUID deterministically — `Uuid::new_v5(&GTS_NS, id.as_bytes())`, and the SDK calls it a *"Deterministic UUID v5 derived from the type id"*. So if the key is renamed, its UUID changes with it. We migrate either way.

**On the size.** Size does matter in audit — that table takes ≥ 50,000,000 records a year. But there we will have to write the long key anyway, so that a record is still readable a year later, when the gear that wrote it may be gone. So the only place left to save is `setting_declarations`, and there are at most 5,000 of those. Values and permissions do not store the key at all — they point at the declaration row by `declaration_id`. So the saving is little.

That said — if you are sure you want the UUID, I can do it.

gears/settings-service/docs/DESIGN.md
shaneemberton: Eight questions on the first half, and one objection to the second.

## On removing `value_type_id`

**1. Who writes the derived schema for a setting an administrator creates in the UI?**

A gear does not write one by hand: `#[gts_type_schema]` derives it from a Rust struct and the link-time inventory registers it. An administrator has no struct. Today they fill a form — vendor, name, category, a shape picked from the catalogue, a default — and no schema exists anywhere. Under this model the same action has to produce a ~35-line JSON Schema and register it. Is the intent (a) a schema editor in the admin UI, (b) a schema builder we generate the document from, or (c) that admin-authored settings go away and every setting comes from a gear? The answer decides whether this is a data-model change or a new subsystem.

**2. Must `value` always be an object?**

The base declares `value` as `type: object`, and the example stores `{"enabled": false}` for a feature flag. Ours stores `false` — a value is "scalar or structured". If the object is mandatory, who names the field inside it? Two gears will call the same boolean `enabled` and `is_enabled`. And the in-process reader returns an object where a consumer wanted a `bool`.

**3. Where does the Schema Default live?**

The derived example puts `"default": false` inside the schema. Our decision is the opposite, with a reason: `default_value` is a column, authoritative, read locally with the values so that **effective-value resolution never depends on the registry**. If the default lives in the schema, a read that falls through to the default needs the registry to be up. Is that intended?

**4. What happens to metadata edits?**

`description`, `mode` and `domain_affinity` are editable in place today — immediate, no step-up, no version change, because none of them alters an effective value. As traits on a registered type, each edit becomes a re-registration. Is a trait change a compatible schema change, or does it need a major bump?

**5. `tenant_visible` / `tenant_overridable` are traits in your base type.**

That makes them declaration-level and platform-managed. This PR moves them onto the `(setting, tenant)` pair, so an ancestor can withhold a setting from one reseller without declaring it non-overridable for every tenant on the platform. Which of the two stands? This one blocks the rest — until it is settled, the shape of the declaration is unknown.

**6. Does the category leave the identifier?**

Your example is `x.platform._.feature_new_dashboard.v1` — the namespace segment is `_`. Ours carries the category there, and both category-shaped grants and `UNIQUE(category_id, leaf_slug)` rely on it.

**7. `barrier_traversable` is a trait we deliberately do not have.**

We ruled out a per-declaration opt-in for barrier traversal: a standalone tenant inherits as any other tenant, "with no marking, exemption or per-declaration opt-in required". Is the trait mandatory in the base?

**8. Volume and timing.**

5,000 settings means 5,000 registered types, each with its own schema, registered at runtime as declarations are created. Everywhere else in the platform a GTS type is declared at compile time from a struct and registered by the inventory at init. Is runtime type registration an intended usage pattern for types-registry?

## On moving migration out to vendor workflows

**It postpones the problem rather than solving it.**

There is a working proposal on the table, and it covers most of the simple cases. On a type change every existing value is copied and **re-validated against the new type**. What still validates goes live unchanged. What does not is flagged `needs_review`, kept, excluded from resolution — the read falls through to the ancestor or the Schema Default — and shown to the administrator who can fix it. That handles the common shapes: a compatible minor revision, a widened enum, an added optional field, a `bool` that coerces to a `string`.

And the gap if we leave it out: between the type change and the workflow run, values neither validate nor are marked. The service then either serves values that do not conform to the current type, or fails reads that used to work. `needs_review` is what bounds that window — it defines the state of the system before any vendor workflow runs.

shaneemberton: 1. Agreed.

With this small exceptions:

1.1. **Not taking the declaration traits.** `scope_class`, `mode`, `domain_affinity`, `data_classification`, `tenant_visible`, `tenant_overridable` describe our access rules, not the value, and each is read where it is enforced — the resolver reads Scope Class on every resolve, search filters on classification, and clearing `requires_step_up` is verified and audited in the transaction that writes it. In the type each would still need its copy in the database, which is the second home we refused `default`. `barrier_traversable` is a property of the tenant besides. `x-gts-traits` keeps what describes the value: `secret`, `multiline`, cron dialect.

1.2. **`value` should stay unconstrained.** Forcing an object gives every scalar setting an invented wrapper field, and the in-process reader returns an object where a consumer asked for a `bool`. Not a GTS constraint — draft-07 describes scalars.

1.3. **`subject_type` demands `subject_id`**, and the schema should enforce that rather than state it in a description: `dependencies: { "subject_type": ["subject_id"] }`.

1.4. **Platform scope is the root tenant's id, not an all-zeros sentinel.**

1.5. **Registration stays with the Settings gear.** A gear that registers its own type composes its own schema, and the curated value-type catalogue stops constraining the value shape. A bad derived half cannot be refused after the fact either — `types-registry` has no unregister.

One gap: an administrator creating a setting in the UI has no struct to derive a schema from. The catalogue can become a template library we generate from — small, but a decision.

2. **On migration.** Little reason to push onto vendor workflows what the service already does. The design covers it: on an upgrade every stored override is re-validated against the new schema, and what fails is flagged `needs_review` — excluded from resolution, still visible to the admin, never silently coerced. It is already in place; no reason to design it out in advance.

gears/settings-service/docs/PRD.md
```diff
@@ -222,9 +222,9 @@ The platform has no configuration service yet: each gear owns its own configurat
...
-- **Availability Expectation**: Required at declaration-authoring and value-write time for type resolution/validation. If briefly unavailable, already-validated effective values continue to be servable from cache per [§6.1 Performance: Read-Path Caching](#61-gear-specific-nfrs).
+- **Role**: The platform `types-registry` gear. Holds each Setting Declaration's own registered GTS type, which is what makes a setting referenceable and policy-addressable platform-wide.
+- **Integration Direction**: Outbound — Settings Service registers a setting's type there when its declaration is created, revived or evolved, and takes the registry's compatibility verdict on a re-registration; `types-registry` does not call into Settings Service.
+- **Availability Expectation**: Required at declaration-authoring time — creating, reviving or evolving a setting fails closed without it. Value writes and reads do not depend on it: a value is validated against the schema held with its own declaration.
```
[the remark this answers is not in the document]
shaneemberton: We reach the registry through its SDK trait, not over HTTP. `TypesRegistryClient::register_type_schemas` returns `Result<Vec<RegisterResult>, CanonicalError>` — a per-item verdict, synchronously, with `FailedPrecondition` for an unregistered chain parent exactly as this design describes. `register_entities`, `RegistrationOperation`, `get_operation` and `IdempotencyKey` exist in ADR-0012 and in the registry's own DESIGN, but not in `types-registry-sdk`. There is no pending admission to model against the interface we actually call. When the registry moves its write path over, this section moves with it.

PR #4711, opened 2026-09-04
feat(toolkit): JSON log lines carry the service identity

cyberantonz:
**Why.** A log collector cannot tell which service or release wrote a JSON line — the fmt subscriber emits no resource identity, and tracing-subscriber's layer has no way to add static fields to every line.

**What changed.** `GearsJsonFormat` — a gears-owned `FormatEvent` — produces the JSON line in the same shape tracing-subscriber's `Json` emits, plus top-level `service` and `version` entries serialized like every other key (fields via `tracing-serde`, spans via `FormattedFields`, one serialization pass). Identity comes from `OpenTelemetryResource::effective_service_version()` — the new `service_version` field, falling back to `attributes."service.version"` — the same value `build_resource` exports, so logs and the OTel resource cannot diverge; the OoP path folds the rendered master telemetry section into `final_config` for the same reason. Text output is local-dev and carries no identity keys; an empty `service_name` omits both keys — documented in configure.md.

**One entry point: `init_logging_unified(&AppConfig, otel_layer)`.** It derives logging config, home dir and identity itself — a breaking signature change for direct callers. `build_oop_config_and_db` now folds merged logging and rendered telemetry into `final_config`, so the OoP host uses the same function.

**Verified.** 7 new unit tests (identity keys, JSON-escape round-trip, span chain, attribute fallback, non-JSON span fields, empty identity, `build_resource` version once from either source); full toolkit lib suite (508) passes; `cargo check --workspace --all-features`, clippy pedantic, rustfmt clean.

### DIFF
--- a/docs/web-docs/build-with-gears/configure.md
+++ b/docs/web-docs/build-with-gears/configure.md
@@ -32,6 +32,7 @@ database:
 logging:
   default:
     console_level: info
+    console_format: text # or: json (one object per line, for container log collectors)
     file: "logs/cf-gears.log"
     file_level: info
 
@@ -52,6 +53,13 @@ gears:
 - **API gateway** — `bind_addr`, `enable_docs`, `prefix_path`, CORS, rate limits, timeouts.
 - **Deployment shape** — `gears.<name>.runtime.type: local | oop` selects in-process vs
   out-of-process. See [Run a gear out-of-process](../out-of-process/).
+- **Logging** — `logging.default.console_level` and `console_format: text | json`.
+  Every JSON line (console and file) carries top-level `service` and `version`
+  keys from `opentelemetry.resource.{service_name,service_version}`
+  (`version` falls back to an `attributes."service.version"` entry; an empty
+  `service_name` omits both keys). `text` output is for local development —
+  one known binary in a terminal — and carries no identity keys; collectors
+  parse `json`.
 - **Tracing** — a `tracing:` block points at an OTLP backend. See
   [Add observability](../add-observability/).
 - **Database** — `database.servers.<name>` connection templates (`engine`, `params`, `pool`) that gears inherit; SQLite, PostgreSQL, and MariaDB engines are supported.
--- a/libs/toolkit/src/bootstrap/host/logging.rs
+++ b/libs/toolkit/src/bootstrap/host/logging.rs
@@ -1,4 +1,5 @@
-use super::super::config::{ConsoleFormat, LoggingConfig, Section};
+use super::super::config::{AppConfig, ConsoleFormat, LoggingConfig, Section};
+use crate::telemetry::config::OpenTelemetryResource;
 use anyhow::Context;
 use std::io::Write;
 use std::path::Path;
@@ -15,6 +16,121 @@ pub type OtelLayer = tracing_opentelemetry::OpenTelemetryLayer<
 >;
 #[cfg(not(feature = "otel"))]
 pub type OtelLayer = ();
+
+// ================= per-line service identity =================
+
+/// The gears-owned JSON line format: the same shape tracing-subscriber's
+/// `Json` emits, plus top-level `service` / `version` identity entries.
+struct GearsJsonFormat {
+    service: String,
+    version: Option<String>,
+}
+
+impl GearsJsonFormat {
+    fn new(resource: &OpenTelemetryResource) -> Self {
+        Self {
+            service: resource.service_name.clone(),
+            version: resource.effective_service_version().map(str::to_owned),
+        }
+    }
+}
+
+/// A span's `FormattedFields` (stored by the layer as a JSON object string)
+/// flattened next to its `name`, as the upstream Json format renders spans.
+struct SerializableSpan<'a, S, N>(
+    tracing_subscriber::registry::SpanRef<'a, S>,
+    std::marker::PhantomData<N>,
+)
+where
+    S: for<'lookup> tracing_subscriber::registry::LookupSpan<'lookup>;
+
+impl<S, N> serde::Serialize for SerializableSpan<'_, S, N>
+where
+    S: for<'lookup> tracing_subscriber::registry::LookupSpan<'lookup>,
+    N: for<'writer> fmt::FormatFields<'writer> + 'static,
+{
+    fn serialize<Ser>(&self, serializer: Ser) -> Result<Ser::Ok, Ser::Error>
+    where
+        Ser: serde::Serializer,
+    {
+        use serde::ser::SerializeMap;
+
+        let mut map = serializer.serialize_map(None)?;
+        let extensions = self.0.extensions();
+        let formatted = extensions
+            .get::<fmt::FormattedFields<N>>()
+            .map(|fields| fields.fields.as_str())
+            .unwrap_or_default();
+        match serde_json::from_str::<serde_json::Value>(formatted) {
+            Ok(serde_json::Value::Object(fields)) => {
+                for (key, value) in fields {
+                    map.serialize_entry(&key, &value)?;
+                }
+            }
+            _ => map.serialize_entry("fields", formatted)?,
+        }
+        map.serialize_entry("name", self.0.metadata().name())?;
+        map.end()
+    }
+}
+
+impl<S, N> fmt::FormatEvent<S, N> for GearsJsonFormat
+where
+    S: tracing::Subscriber + for<'a> tracing_subscriber::registry::LookupSpan<'a>,
+    N: for<'a> fmt::FormatFields<'a> + 'static,
+{
+    fn format_event(
+        &self,
+        ctx: &fmt::FmtContext<'_, S, N>,
+        mut writer: fmt::format::Writer<'_>,
+        event: &tracing::Event<'_>,
+    ) -> std::fmt::Result {
+        use serde::ser::{SerializeMap, Serializer as _};
+        use tracing_log::NormalizeEvent;
+        use tracing_serde::fields::AsMap;
+
+        let normalized = event.normalized_metadata();
+        let meta = normalized.as_ref().unwrap_or_else(|| event.metadata());
+        let span_marker: std::marker::PhantomData<N> = std::marker::PhantomData;
+
+        let serialize = || {
+            let mut buf = Vec::with_capacity(256);
+            let mut serializer = serde_json::Serializer::new(&mut buf);
+            let mut map = serializer.serialize_map(None)?;
+
+            let timestamp = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Micros, true);
+            map.serialize_entry("timestamp", &timestamp)?;
+            map.serialize_entry("level", meta.level().as_str())?;
+            if !self.service.is_empty() {
+                map.serialize_entry("service", &self.service)?;
+                if let Some(version) = &self.version {
+                    map.serialize_entry("version", version)?;
+                }
+            }
+
+            map.serialize_entry("fields", &event.field_map())?;
+            map.serialize_entry("target", meta.target())?;
+
+            if let Some(span) = ctx.event_scope().and_then(|mut scope| scope.next()) {
+                map.serialize_entry("span", &SerializableSpan(span, span_marker))?;
+            }
+            if let Some(scope) = ctx.event_scope() {
+                let spans: Vec<_> = scope
+                    .from_root()
+                    .map(|span| SerializableSpan::<S, N>(span, span_marker))
+                    .collect();
+                map.serialize_entry("spans", &spans)?;
+            }
+
+            map.end()?;
+            Ok::<_, serde_json::Error>(buf)
+        };
+
+        let buf = serialize().map_err(|_| std::fmt::Error)?;
+        writer.write_str(std::str::from_utf8(&buf).map_err(|_| std::fmt::Error)?)?;
+        writer.write_char('\n')
+    }
+}
 
 // ================= level helpers =================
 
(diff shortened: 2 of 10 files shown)


docs/web-docs/build-with-gears/configure.md
```diff
@@ -52,6 +53,11 @@ gears:
 - **API gateway** — `bind_addr`, `enable_docs`, `prefix_path`, CORS, rate limits, timeouts.
 - **Deployment shape** — `gears.<name>.runtime.type: local | oop` selects in-process vs
   out-of-process. See [Run a gear out-of-process](../out-of-process/).
+- **Logging** — `logging.default.console_level` and `console_format: text | json`.
+  Every JSON line (console and file) carries the service identity from
+  `opentelemetry.resource` — `service_name` as a top-level `service` key, and
+  `service_version` (when set) as `version` — so a collector can label lines
+  without unwrapping span context.
```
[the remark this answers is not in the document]
cyberantonz: Fixed in cfbc9b7: the doc now states the exact contract — version falls back to attributes."service.version", an empty service_name omits both keys, and text output carries no identity keys (local-dev format; collectors parse json).

libs/toolkit/src/bootstrap/run.rs
```diff
@@ -311,8 +311,18 @@ pub fn init_procedure(config: &AppConfig) -> Result<()> {
...
     let otel_layer = None;
 
-    // Initialize logging + otel in one Registry
-    init_logging_unified(&config.logging, &config.server.home_dir, otel_layer);
+    // Initialize logging + otel in one Registry. Every JSON line carries the
+    // same identity the OTel resource does.
+    let identity = LineIdentity {
+        service: config.opentelemetry.resource.service_name.clone(),
+        version: config.opentelemetry.resource.service_version.clone(),
+    };
```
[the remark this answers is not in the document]
cyberantonz: Fixed in cfbc9b7: OpenTelemetryResource::effective_service_version() (field, falling back to attributes."service.version") is now the single source for both build_resource and the log identity; the OoP path folds the rendered master opentelemetry section into final_config — the same section init_tracing reads — so logs and the OTel resource cannot diverge. Covered by new tests on both paths.

libs/toolkit/src/telemetry/config.rs
```diff
@@ -42,6 +42,9 @@ pub struct OpenTelemetryResource {
     /// Logical service name.
     #[serde(default = "default_service_name")]
     pub service_name: String,
+    /// Release version (`service.version` semconv); also on every JSON log line.
+    #[serde(default)]
```
MikeFalcon77, line 46: With a derived `Serialize` and no skip, this puts `"service_version": null` into every rendered gear config, and `OpenTelemetryResource` is `deny_unknown_fields`. An OoP gear built against an older toolkit then fails `RenderedGearConfig::from_json`, and `oop.rs` swallows that with `.ok()`, so the gear silently loses master's database, logging and telemetry config. `#[serde(default, skip_serializing_if = "Option::is_none")]` keeps the wire readable by older binaries, same as `cardinality_limit` below.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -193,9 +309,14 @@ fn create_rotating_writer_at_path(
 // the background flush thread and silently loses buffered log output.
 static CONSOLE_GUARD: OnceLock<WorkerGuard> = OnceLock::new();
 
-/// Unified initializer used by both functions above.
+/// Unified initializer: logging config, home dir and the per-line service
+/// identity all come from the app config.
 #[allow(unknown_lints, de1301_no_print_macros)] // runs before tracing subscriber is installed
-pub fn init_logging_unified(cfg: &LoggingConfig, base_dir: &Path, otel_layer: Option<OtelLayer>) {
+pub fn init_logging_unified(config: &AppConfig, otel_layer: Option<OtelLayer>) {
```
MikeFalcon77, line 315: `init_logging_unified` is `pub` and re-exported by `bootstrap::host` from a published crate, so this signature change breaks out-of-crate callers, and the PR is a plain `feat` with no breaking marker. Either mark the change breaking so the release bump reflects it, or keep the old three-argument form as a `#[deprecated]` wrapper.

libs/toolkit/src/bootstrap/oop.rs
```diff
@@ -185,11 +185,17 @@ fn build_oop_config_and_db(
...
     // Merge logging: master logging (base) + local logging (override by key)
-    let final_logging = merge_logging_configs(
+    final_config.logging = merge_logging_configs(
         rendered_config.as_ref().and_then(|r| r.logging.as_ref()),
         &local_config.logging,
     );
 
+    // Telemetry comes from master when rendered (the same section init_tracing
+    // reads), so the log identity cannot diverge from the OTel resource.
+    if let Some(otel) = rendered_config.and_then(|r| r.opentelemetry.as_ref()) {
```
MikeFalcon77, line 195: `render_gear_config` only fills `opentelemetry` when the master has tracing or metrics enabled, so with OTel off this branch never runs and the log identity comes from the gear's own local config, or from the `cf-gears` default. So the identity can diverge from the master's resource, and in the default setup every OoP gear logs `service: "cf-gears"` instead of its own name.

libs/toolkit/src/bootstrap/oop_tests.rs
```diff
@@ -636,7 +636,8 @@ mod full_oop_config {
         let result = build_oop_config_and_db(&local_config, "test_gear", Some(&rendered));
 
         assert!(result.is_ok());
-        let (final_config, merged_logging, _) = result.unwrap();
+        let (final_config, _) = result.unwrap();
```
MikeFalcon77, line 639: All five `RenderedGearConfig` fixtures in this module still pass `opentelemetry: None`, so the new branch that copies the master OTel section into `final_config.opentelemetry` never runs in tests. Can we add a case with `opentelemetry: Some(...)` asserting the resource comes from master, and one where rendered has no section so the local resource survives?

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -762,4 +881,151 @@ mod tests {
...
+    #[test]
+    fn an_empty_identity_injects_nothing() {
+        let resource = OpenTelemetryResource {
+            service_name: String::new(),
+            ..OpenTelemetryResource::default()
+        };
+
+        let line = capture_json_line(&resource, || tracing::info!("plain probe"));
+
+        assert!(line.get("service").is_none(), "no service key: {line}");
```
MikeFalcon77, line 1028: This case leaves `service_version` unset, so it does not cover the part that can break: `version` is serialized inside the `!self.service.is_empty()` branch, and the docs promise an empty `service_name` omits both keys. Set `service_version: Some("1.2.3")` here and assert `version` is absent too.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -16,6 +17,121 @@ pub type OtelLayer = tracing_opentelemetry::OpenTelemetryLayer<
...
+        use serde::ser::{SerializeMap, Serializer as _};
+        use tracing_log::NormalizeEvent;
+        use tracing_serde::fields::AsMap;
+
+        let normalized = event.normalized_metadata();
+        let meta = normalized.as_ref().unwrap_or_else(|| event.metadata());
+        let span_marker: std::marker::PhantomData<N> = std::marker::PhantomData;
+
+        let serialize = || {
+            let mut buf = Vec::with_capacity(256);
```
MikeFalcon77, line 97: Every log line allocates a fresh `Vec`, then gets UTF-8 validated and copied into `writer`. Upstream `Json` serializes straight into the `Writer` through a small `io::Write` adaptor, which avoids both the extra buffer and the copy per event.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -16,6 +17,121 @@ pub type OtelLayer = tracing_opentelemetry::OpenTelemetryLayer<
...
+            }
+
+            map.serialize_entry("fields", &event.field_map())?;
+            map.serialize_entry("target", meta.target())?;
+
+            if let Some(span) = ctx.event_scope().and_then(|mut scope| scope.next()) {
+                map.serialize_entry("span", &SerializableSpan(span, span_marker))?;
+            }
+            if let Some(scope) = ctx.event_scope() {
+                let spans: Vec<_> = scope
```
MikeFalcon77, line 118: `from_root()` already buffers the scope internally, so this `collect()` is a second allocation on every log line. Serializing the sequence directly with `serialize_seq` + `serialize_element` over the iterator produces the same JSON without it.

libs/toolkit/src/telemetry/config.rs
```diff
@@ -42,6 +42,9 @@ pub struct OpenTelemetryResource {
...
+    pub service_version: Option<String>,
```
MikeFalcon77, line 47: This struct is `pub` with `pub` fields in a published crate, so adding a field breaks every out-of-crate struct literal, exactly as the literals in `telemetry/init.rs` had to be fixed in this PR. If the resource block keeps growing, `#[non_exhaustive]` belongs here now, while the release is already breaking.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -425,9 +548,7 @@ fn install_subscriber(
                     .json()
                     .with_writer(nb_stderr)
                     .with_ansi(false)
-                    .with_target(true)
-                    .with_level(true)
-                    .with_timer(fmt::time::UtcTime::rfc_3339())
+                    .event_format(GearsJsonFormat::new(resource))
```
MikeFalcon77, line 551: `.json()` is load-bearing here only for the field formatter: `event_format` throws away its `Json` event format, but `JsonFields` stays, and `SerializableSpan` depends on span fields being stored as a JSON object. Writing it as `.fmt_fields(JsonFields::default())` makes that dependency visible, and `.with_ansi(false)` above is now inert since the custom format never reads the ansi flag.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -16,6 +17,121 @@ pub type OtelLayer = tracing_opentelemetry::OpenTelemetryLayer<
...
+
+        let serialize = || {
+            let mut buf = Vec::with_capacity(256);
+            let mut serializer = serde_json::Serializer::new(&mut buf);
+            let mut map = serializer.serialize_map(None)?;
+
+            let timestamp = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Micros, true);
+            map.serialize_entry("timestamp", &timestamp)?;
+            map.serialize_entry("level", meta.level().as_str())?;
+            if !self.service.is_empty() {
```
MikeFalcon77, line 104: An empty `service_name` doubles as the off switch, and it also drops a configured `version` even though the two are independent settings. Resolving the identity once in `new()` into an `Option` of a small identity struct would make both the intent and that coupling explicit instead of re-deciding it per event.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -762,4 +881,151 @@ mod tests {
...
+        };
+
+        let line = capture_json_line(&resource, || {
+            tracing::info_span!("outer", request_id = "rid-1")
+                .in_scope(|| tracing::info!("span probe"));
+        });
+
+        assert_eq!(line["span"]["name"], "outer");
+        assert_eq!(line["span"]["request_id"], "rid-1");
+        assert_eq!(line["spans"][0]["name"], "outer");
```
MikeFalcon77, line 971: With one span the `spans` array has a single element, so this passes whether the chain is written root first or innermost first, and it says nothing about how nested spans flatten. Nest a second span and assert `spans[0]` is the outer one while `span` is the inner one.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -762,4 +881,151 @@ mod tests {
...
+            ..OpenTelemetryResource::default()
+        };
+
+        let line = capture_json_line(&resource, || tracing::info!("identity probe"));
+
+        assert_eq!(line["service"], "probe-service");
+        assert_eq!(line["version"], "1.2.3");
+        assert_eq!(line["fields"]["message"], "identity probe");
+        assert_eq!(line["level"], "INFO");
+        assert!(line["timestamp"].is_string(), "timestamp survives: {line}");
```
MikeFalcon77, line 940: `is_string()` holds for any string, so the timestamp format can change without a test noticing, and no test asserts `target` at all even though this PR dropped `.with_target(true)` and `.with_timer(...)` from both layers. Parse the value with `chrono::DateTime::parse_from_rfc3339` and add an assertion on `line["target"]`.

libs/toolkit/src/bootstrap/host/logging.rs
```diff
@@ -762,4 +881,151 @@ mod tests {
...
+    fn identity_values_are_json_escaped() {
+        let resource = OpenTelemetryResource {
+            service_name: "probe \"quoted\" service".to_owned(),
+            service_version: None,
+            ..OpenTelemetryResource::default()
+        };
+
+        let line = capture_json_line(&resource, || tracing::info!("escape probe"));
+
+        assert_eq!(line["service"], "probe \"quoted\" service");
```
MikeFalcon77, line 953: This is a point example for a serializer whose one-object-per-line contract is what log collectors depend on, and `SerializableSpan` also parses span field strings back with `serde_json::from_str`. A proptest over arbitrary service names, versions and message text, asserting exactly one output line that parses back with the same `service` value, would cover newlines and control characters that a single quoted-string example misses.

libs/toolkit/src/bootstrap/oop.rs
MikeFalcon77: **CRITICAL**

`service_name` is not new here. It already exists in `OpenTelemetryResource` with a `cf-gears` default, so this adds a third name for the same thing, next to `server.name` and `gear_name`. Both ways through this branch get the identity wrong:

- OTel off, which is the default: `render_gear_config_for_oop` returns no `opentelemetry` section, so this branch never runs and the gear logs `service: "cf-gears"`.
- OTel on: the section is `app.opentelemetry.clone()` from the host with its `resource` included, so this line overwrites the gear identity with the master's name and every OoP gear logs under it.

The second case is the opposite of what this PR is for. The name is available on both sides already: `gear_name` is a parameter of this function and is used three lines below for the DB paths, and `render_gear_config_for_oop` takes it too.

Take master's exporter, tracing and metrics settings, but not its `resource`. Set `resource.service_name` from `gear_name` unless the gear's own config sets it, so precedence is gear config first, then `gear_name`, and `cf-gears` stops being reachable for a named gear. `server.name` is the matching default on the in-process side, which also closes the `app` vs `service` split at `run.rs:335`.

MikeFalcon77 (changes requested):

PR #4717, opened 2026-09-07
Add Change Orders PRD

Corw1n-of-Amber:
## Summary

Adds the Change Orders PRD at `gears/bss/orders-changes/docs/PRD.md` — the second phase of the Orders capability: the `change` order category, scoped this phase to **increases**. A commercially initiated purchase that adds capacity or a component to a subscription that already exists, without replacing it.

Raised by adopter feedback on #4648 / #4649: buying more against a running subscription had no commercial artifact, so growth was transacted without a price pin, an approval gate, or a booking record. Three questions those PRDs deferred converge here and are answered together — add-on selection on the order line, change-order phasing, and how a line targets an existing subscription.

**In scope:** `targetSubscriptionId` and change kind (`increase_quantity` / `add_component`) on the line, augment never supersede, one target per order; add-on selection on the line, valid at acquisition capture as well; delta sellability gate including add-on rule bounds; delta catalog price pin; delta TCV for approval; transactional application; overlap-rule exemption for the target; change events carrying the applied delta.

**Out of scope:** decreases, removals, and plan change (deferred with Open Questions); proration, credits, and invoice impact (Subscriptions and the billing chain); subscription composition semantics (Subscriptions); approval decision logic (Generic Approval).

## Notes for review

- **Consequential amendment (§2).** This PRD activates the `change` category that `orders-lifecycle` currently requires to be **rejected**, and lifts its add-on exclusion. The two must land together; a half-adopted pair is recorded as a risk in §16.
- **Why one target per order (§6.1).** It is what makes the increase applicable as a single transactional intent. The acquisition path compensates a partial failure by voiding drafts, because nothing is live until activation; here the target is live from the start, so the only compensation would be a decrease — which this phase does not specify. The constraint buys atomicity instead of building a saga over an unwritten reversal path.
- **Upstream asks use a `CHG-S*` namespace (§9.2)**, not `SUB-O*`. The `SUB-O*` register is authored in `gears/bss/subscriptions/docs/SEAMS.md`; numbering into it from this side has already produced one collision, which is recorded in §16.
- **Grandfathering (§6.2).** Whether a protected cohort's price extends to quantity added later is named as a **pricing** decision that Orders consumes and must not infer. The determination does not exist yet and carries the earliest target date of the Open Questions.

## Validation

- [x] `cfs toc gears/bss/orders-changes/docs/PRD.md`
- [x] `cfs validate --artifact gears/bss/orders-changes/docs/PRD.md --skip-code --local-only --verbose` — **0 errors / 0 warnings**
- [x] Full-tree `cfs validate` — 0 errors; warning count unchanged from `main`

## Summary by CodeRabbit

- **Documentation**
  - Documented “change” orders for increasing capacity or adding components to existing subscriptions without replacing them.
  - Defined eligibility and classification rules for recurring and usage-based changes, while excluding one-time purchases.
  - Documented pricing, sellability, approval, acceptance, effective dating, transactional application, events, and audit requirements.
  - Clarified the boundary between change orders and acquisition flows, including out-of-scope decreases, removals, and plan changes.
  - Updated lifecycle requirements for overlap validation, fulfillment, cancellation, subscription linkage, retries, and line-level add-on selection.
  - Defined public operations, use cases, acceptance criteria, dependencies, assumptions, risks, and open questions.


rostm:
## One-time consumable top-up against a live subscription

This is a case the two change kinds don't reach.

**Use case.** A customer with a live subscription buys extra consumable units —
1000 more exam attempts, another block of credits. One-time charge, no change to
the recurring price, units land on the existing subscription's balance, drawn down
by consumption, and the purchase **repeats** whenever the balance runs low.

**Why it isn't covered.**

1. **The two kinds partition on one test — does the target already carry it — and
   a repeated purchase breaks it.** First 1000 exams is `add_component` ("the
   target does **not yet** carry"). The second cannot be: the target now carries
   it, leaving only `increase_quantity` ("more of what the target already
   carries"), and "more of" a completed one-time charge has no defined meaning.
   The model admits the first top-up and misclassifies every one after. That
   breaks for any one-time line bought twice against one target, consumables
   aside.

2. **A balance isn't a composition element.** §5.2 delegates composition to
   `PlanLink` / `AddOn` — both structural. A consumable grant is a drawdown
   balance with an amount and an expiry. The acquisition PRD sells prepaid packs
   as lines on a one-time plan that spawn their **own** subscription, so routing
   this through `add_component` implies a second subscription — the outcome the
   augment rule exists to prevent.

3. **Nothing records the result.** The delta pin freezes price rows. Nothing says
   "the target's entitled quantity is now N + 1000, expiring D".

**Suggestions.**

1. State whether one-time consumable top-ups are in scope. If not, put them in
   §5.2 with an owner and target, as decreases and plan change are — for
   certification and exam products this is the primary growth motion.
2. Fix the partition either way: add a third kind whose validity doesn't depend on
   what the target already carries, or restate `increase_quantity` to cover a
   repeated one-time purchase and define "more of" for a `one_time` charge kind.
3. Say where the resulting entitled quantity and expiry are recorded.
4. Confirm the downstream shape — balance on the target, or a separate one-time
   subscription? The augment rule and the acquisition PRD currently disagree.

Corw1n-of-Amber:
The partition is broken regardless of the consumable question. Defining the two kinds by "does the target already carry it" makes the first purchase valid and every repeat unclassifiable, and it breaks for any `one_time` line bought twice against one target, consumables aside. The test should be on charge kind, not on prior carriage. That gets fixed either way.

The contradiction is real too: the acquisition PRD sells prepaid packs as lines on a one-time plan, and its 1:1 rule makes each spawn its own subscription — the opposite of the augment rule here.

On where the units land: pricing defines the prepaid grant but never tracks balance; subscriptions supersedes credit pools to "balance/drawdown = Billing/Rating", GA-gated; there is no Billing gear. A balance on the target has no owner today.

Question before I write it up: does a separate one-time subscription per top-up work — purchased through an order, linked to the main subscription, drawn down by the same consumption — or do you need the units on the existing subscription's balance?

Separate subscription is buildable now. Balance on the target stays blocked on the grant/balance owner, and I'd record it with that dependency.

Corw1n-of-Amber:
Following up on our DM, where you confirmed that a separate one-time subscription bound to the tenant works for you.

One-time top-ups are out of the change path: §5.2 now routes them as an acquisition line on a one-time plan carrying the same `resourceTenantId`, spawning their own subscription.

Change kinds are scoped by charge kind rather than by what the target already carries, so both apply only to `recurring` and `usage`. A `one_time` line is rejected at submit with a reason distinct from "not sellable", so the caller can re-route it.

One thing to watch on your side: the overlap default is one active subscription per `(payerTenantId, catalogSubscriptionProductKey)`. Two top-ups of the same pack share that key, so the second is rejected as a duplicate unless the catalog template for that product sets `maxConcurrentActive` above one. That's in §5.2 now.

Provenance — an optional reference from the top-up to the subscription it supplements — is an open question on the acquisition path, not here.

rostm (approved):

PR #4736, opened 2026-09-08
feat(rbac,authz-resolver-plugin): add the RBAC gear and its PDP

bit4flip:
## Description

Adds the **RBAC gear** (`cf-gears-rbac`, `cf-gears-rbac-sdk`) and the **AuthZ Resolver Plugin** (`cf-gears-authz-resolver-plugin`) that acts as the platform's PDP.

- **RBAC gear** — role definitions and role assignments over Postgres/SQLite: REST surface under `/rbac/v1/`, OData listing, strong `ETag` optimistic concurrency, a built-in role catalogue seeded idempotently, a permission evaluator with scope/`not_permissions` handling, and an in-process `dyn RbacServiceClientV1` published in `ClientHub`.
- **AuthZ Resolver Plugin** — a 10-step evaluation pipeline (request validation → GTS type validation → token-scope enforcement → RBAC policy evaluation → assignment-scope provenance → scope materialization → non-empty enforcement → `require_constraints` branch → constraint generation → audit), with a bounded LRU + singleflight in front of hierarchy reads, and constraint generation that fails closed rather than degrading into an unconstrained allow.
- **Ledger seller predicate** — `bss-ledger`'s `seller_tenant_types` becomes a required, deployment-supplied list instead of carrying a built-in default (see *Behaviour change* below).

Specs ship with the code per the SDD rules in CONTRIBUTING §2.2: `gears/system/rbac/docs/{PRD,DESIGN}.md`, `gears/system/authz-resolver/plugins/authz-resolver-plugin/docs/{PRD,DESIGN}.md`, plus JSON schemas and an ADR for the ledger predicate.

## Type of Change

- [x] Documentation update

## Version bump justification

Required by CONTRIBUTING §Enforcement ("if you touched a public surface, you must justify the version bump category").

| Crate | Version | Category | Why |
|---|---|---|---|
| `cf-gears-rbac` | `0.1.0` (new) | n/a | New crate; nothing downstream to break. |
| `cf-gears-rbac-sdk` | `0.1.0` (new) | n/a | New crate. |
| `cf-gears-authz-resolver-plugin` | `0.1.0` (new) | n/a | New crate. |
| `cf-gears-bss-ledger` | no `[package] version` | none owed | Declares neither `version` nor `publish`, so it is not a release target and `check_packaging_metadata.py` skips it. The SemVer rules in CONTRIBUTING §3 govern published crates; no bump is owed here. |

No existing published crate's Rust API, REST path, proto, or CLI surface changes in this PR. The workspace `Cargo.toml` and `Makefile` changes are additive (new members, new workspace deps, a new `test-rbac-pg` target).

### Behaviour change to note (not SemVer-breaking)

`bss-ledger`'s `seller_tenant_types` no longer has a built-in default (it was partner + platform). It is now required, and an empty list is rejected by `BssLedgerConfig::validate` at `init()` instead of degrading into "no tenant owns a ledger" at the first provisioning request. Rationale: tenant-type ids are deployment-specific, so any built-in list names one product and silently rejects every seller on every other.

This is the "more strict validation returning errors in previously accepted edge cases" case from CONTRIBUTING §"What Is Not Breaking (But Must Be Noted)". **Operational impact:** a `server.yaml` that omitted `seller_tenant_types` used to boot with the default and will now fail at `init()` with a config error. It needs a release-note entry and a migration line for anyone running the ledger gear.

## Testing

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing completed

Verified on the current head (rebased onto `main`):

| Check | Result |
|---|---|
| `cargo nextest run -p cf-gears-rbac -p cf-gears-rbac-sdk -p cf-gears-authz-resolver-plugin` | 1031 / 1031 passed |
| `cargo clippy` (three crates, `-D warnings -D clippy::perf`) | clean |
| `cargo fmt --all --check` | clean |
| `make lint` (workspace, `-D warnings`) | clean |

Verified on the pre-rebase head (same tree contents, older base):

| Check | Result |
|---|---|
| `make test-no-macros` (whole workspace) | 16754 / 16754 passed |
| `make test-rbac-pg` (Postgres tier, Docker) | 142 / 142 passed |
| `make test-macros`, `make test-fixtures-narrow` | pass |
| `cargo gears lint --dylint` | clean |
| `make deny` | advisories / bans / licenses / sources ok |
| `cargo xtask check-test-container-pins` | 3914 files, no violations |
| `validate_gear_names.py` | 18 / 18 |

The whole-workspace suite has **not** been re-run since the rebase onto `main`, which brought a breaking SecureORM change (`fix(secure-orm)!: own the syntax error, opt into the anchor, check membership by table`) and several SDK major bumps. The three affected crates and a workspace-wide `cargo check` are clean; CI on this head is the remaining signal.

Test tiers added: unit tests colocated per module, SQLite integration tests, and a Postgres tier (`postgres_*`, 142 tests) gated `#[ignore]` behind Docker. This PR also adds the CI job that runs that tier — `make test-rbac-pg` in the `integration` workflow — because the tier previously would have compiled on every run and executed on none.

Not run locally: `make check-packaging-metadata` and `make check-release-config` (the venv bootstrap needs Python ≥ 3.10 for its pinned `pytest`; this machine has 3.9.6), and the `cargo hack` half of `make clippy` (`cargo-hack` not installed — it scopes toolkit crates outside this diff). Both run in CI.

## Documentation

- [x] README updated (if applicable)

PRD + DESIGN for both new gears, an ADR for the ledger seller predicate, per-crate `README.md`, JSON schemas for the RBAC resources and events, `docs/GEARS.md` catalogue entry, and the two new docs directories added to the `lychee` link-check paths.

## Checklist

- [x] Self-review completed
- [x] Commits signed off (DCO, CONTRIBUTING §2.7)
- [x] Up-to-date with `main` (rebased)
- [ ] **`cargo-shear` `ignored` entry for `rbac`** — see *Before merge* below

## Before merge

1. **Add `"rbac"` to `[workspace.metadata.cargo-shear] ignored` in the workspace `Cargo.toml`.**
   `authz-resolver-plugin` declares `deps = [types_registry, authz_resolver, rbac, tenant_resolver, resource_group]`, and the `#[toolkit::gear]` macro consumes that dep only through generated `pub use` re-exports that `cargo-shear` cannot trace. Its four siblings are already in the list; `rbac` is not. Per CONTRIBUTING §2.3, `cargo-shear` does not run on PRs — it runs nightly — so a missing entry breaks `main` *after* this merges rather than failing here.
2. **Release note for the ledger config change** (see *Behaviour change* above) with a migration line: deployments running `bss-ledger` must add an explicit `seller_tenant_types` list.
3. Merge strategy: **squash and merge** per CONTRIBUTING §2.10. Note the PR *title* still carries the `!` breaking marker while the commit message no longer does — squash-merge takes the PR title, so align the two before merging.

## Related Issues

_None linked._


gears/system/authz-resolver/plugins/authz-resolver-plugin/docs/PRD.md
```diff
@@ -0,0 +1,950 @@
+Created:  2026-08-27 by Constructor Fabric
```
entropyshift: replace with proper value

gears/system/authz-resolver/plugins/authz-resolver-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,620 @@
+Created:  2026-08-27 by Constructor Fabric
```
entropyshift: replace with proper value

gears/system/rbac/docs/PRD.md
```diff
@@ -0,0 +1,1003 @@
+Created:  2026-08-20 by Constructor Fabric
```
entropyshift: replace with proper value

gears/system/rbac/docs/DESIGN.md
```diff
@@ -0,0 +1,2356 @@
+Created:  2026-08-20 by Constructor Fabric
```
entropyshift: replace with proper value

.github/workflows/ci.yml
```diff
@@ -393,6 +393,16 @@ jobs:
       - name: Test coord with Postgres (integration)
         run: make test-coord-pg
 
+      # The RBAC gear shipped with its entire Postgres tier `#[ignore]`d and no
```
entropyshift: please fix merge conflicts

entropyshift (changes requested):

gears/system/rbac/rbac/src/domain/role_definition/service.rs
```diff
@@ -0,0 +1,1199 @@
...
+        // Reject names containing `,` or `)` — the PG
+        // `Key (col)=(val) already exists` detail string is parsed by
+        // `extract_quoted_value` in `infra::canonical_mapping` to
+        // recover the conflicting name from a uniqueness violation;
+        // the parser splits on the first `,` or `)`, so a name
+        // containing either truncates the recovered string and the
+        // typed `RoleDefinitionNameTaken { name }` variant carries a
+        // lie. Reject at the domain boundary so the parser's input
+        // stays bounded.
+        validate_name_charset(&request.name)?;
```
ainetx: ### **MINOR**: Role-definition name/description accepted without runtime length bounds

**Severity**: **MINOR**

**Problem**
validate_name_charset in role_definition/service.rs only checks for the presence of ',' or ')' characters in the name; there is no length check for name and no validation at all for description, despite the published schema documenting maxLength 256/4096 and sibling permission-rule fields being explicitly length-bounded in permission_matcher.rs.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Call RoleDefinitionService::create with an authorized WRITE caller.
2. Supply `name` under 256 chars containing no ',' or ')' but very long, or an arbitrarily long `description` (megabytes).
3. On SQLite (text columns, no length constraint) the row persists unbounded; on Postgres the insert fails with a raw varchar-overflow DB error surfaced as an internal/500 rather than a clean validation error.

**Expected behavior**
Domain-layer validation should reject oversized name/description at the service boundary (mirroring the documented schema maxLength and the pattern already used for permission rule fields) before any DB round-trip.

**Actual behavior**
No length bound is enforced in Rust; only Postgres's DB column type incidentally caps size (with a poor error path), and SQLite has no bound at all.

```text
caller -> RoleDefinitionService::create -> validate_name_charset(name) [char-set only] -> NewRoleDefinition{name, description} -> repo.create() -> DB (varchar cap or unbounded text)
```

**Impact**
Unbounded persisted/echoed text increases storage, response-payload, and log/resource-amplification risk, and produces inconsistent runtime behavior versus the published schema contract, especially on SQLite deployments.

**Suggested correction**
Add explicit length checks (e.g. name <= 256, description <= 4096) in validate_name_charset or a sibling validator, executed before persistence, matching the schema's stated bounds.

**How to verify**
Add a unit test creating a role definition with an oversized name/description and assert a typed Validation error is returned instead of a DB-level failure or silent success.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/gts_type_validator.rs
```diff
@@ -0,0 +1,285 @@
...
+                None
+            }
+        };
+
+        if let Some(result) = cached {
+            return self.apply_result(result, gts_type, kind);
+        }
+
+        // Cache miss — consult the registry.
+        let result = map_registry_result(self.registry.get_type_schema(gts_type).await);
```
ainetx: ### **MINOR**: Unvalidated caller-supplied GTS type string reaches the registry API and log/deny messages

**Severity**: **MINOR**

**Problem**
validate() in domain/validation.rs rejects only an empty resource.type; validate_request/validate_type in gts_type_validator.rs pass that string straight to `self.registry.get_type_schema(gts_type)` and into `format!("unknown {kind} gts type: '{gts_type}'")` and `warn!(gts_type = %gts_type,...)` without any character-set/grammar validation.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Send an EvaluationRequest with resource.resource_type set to an arbitrary non-empty string containing control characters or bytes outside the GTS grammar.
2. validate() passes it through (only checks non-empty).
3. GtsTypeValidator::validate_type calls the registry with the raw string and later logs it verbatim in warn!/deny messages.

**Expected behavior**
resource.type (and subject.subject_type) should be checked against the GTS identifier grammar/character set before being sent to the external registry client or interpolated into log/deny text.

**Actual behavior**
The value is forwarded unvalidated to the registry client and to log output.

```text
caller request -> validate() [non-empty only] -> GtsTypeValidator::validate_type -> registry.get_type_schema(raw_string) + warn!/deny message(raw_string)
```

**Impact**
Malformed/control-character identifiers can produce forged or split log lines, unnecessary upstream registry load, and inconsistent registry error behavior; the documented GTS identifier contract is not enforced at the point of acceptance.

**Suggested correction**
Validate resource.type (and subject.subject_type) against the GTS identifier grammar/character allow-list in validation.rs before it reaches the registry client or any log/deny text.

**How to verify**
Add a test sending a resource.type with control characters and assert it is rejected as InvalidRequest before any registry call is attempted.

</details>

gears/system/rbac/rbac/src/domain/ports/permission_catalog.rs
```diff
@@ -0,0 +1,346 @@
...
+            // The `DependencyUnavailable` variant has no `cause` slot,
+            // so the upstream `Registry(msg)` body would otherwise be
+            // dropped on conversion. Log it before discarding so
+            // operators still have the diagnostic for audit
+            // correlation (same philosophy as the bonus-1 fix in
+            // `error_mapping.rs::AuthorizationDenied`).
+            PermissionCatalogError::Registry(msg) => {
+                tracing::warn!(
+                    target: "rbac.permission_catalog",
+                    cause = %msg,
```
ainetx: ### **MINOR**: Upstream registry error text logged unsanitized as a tracing field

**Severity**: **MINOR**

**Problem**
The `Registry(msg)` -> DomainError conversion in permission_catalog.rs logs `cause = %msg` where msg is `e.to_string()` from the TypesRegistryClient error in types_registry_permission_catalog.rs; msg is never checked for CR/LF or other control characters before being embedded in a structured log line.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Trigger a TypesRegistryClient failure whose error Display text contains a newline or carriage return (e.g. from a malformed/compromised upstream response echoed into an error message).
2. PermissionCatalogError::Registry(msg) is constructed and converted to DomainError.
3. tracing::warn!(cause = %msg,...) emits the raw text, splitting or forging the structured log record.

**Expected behavior**
Upstream error text embedded in log fields should be sanitized (control characters stripped/replaced, length capped) before being logged, consistent with the redaction pattern used elsewhere in the diff.

**Actual behavior**
The raw upstream message is logged verbatim via the `%msg` field.

```text
TypesRegistryClient error -> e.to_string() -> PermissionCatalogError::Registry(msg) -> tracing::warn!(cause = %msg) [unsanitized]
```

**Impact**
A malformed or compromised upstream registry response can forge or split log lines, disrupting log ingestion and operator correlation.

**Suggested correction**
Sanitize/redact the message (strip control chars, cap length) before logging, matching the pattern already used for DB diagnostics elsewhere in this PR.

**How to verify**
Unit test constructing a Registry error with an embedded newline and asserting the logged/derived text has no raw control characters.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/audit_emitter.rs
```diff
@@ -0,0 +1,239 @@
...
+/// any control character (CR, LF, tab, ANSI ESC, other C0/C1) with the Unicode
+/// replacement character and cap the length. Benign values (no control chars,
+/// under the cap) pass through unchanged, so normal audit output is unaffected.
+fn sanitize_for_audit(value: &str) -> String {
+    /// Generous bound: real action/resource-type ids are short; this only
+    /// guards against an attacker padding an audit line to absurd length.
+    const MAX_LEN: usize = 256;
+    value
+        .chars()
+        .map(|c| if c.is_control() { '\u{FFFD}' } else { c })
```
ainetx: ### **MINOR**: Audit field sanitizer does not neutralize bidi-override/format characters

**Severity**: **MINOR**

**Problem**
sanitize_for_audit only replaces characters for which `c.is_control()` is true; U+061C and other Unicode bidi format characters are not control characters and are left untouched, so caller-supplied action/resource_type/deny_details values can carry direction-override characters into the emitted `cf-authz.audit` record used for security review.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Submit an evaluation request whose action name embeds U+061C or another bidi override character.
2. AuditRecord::from_response captures the action field and AuditEmitter::emit calls sanitize_for_audit(&record.action).
3. is_control() returns false for the bidi character, so it is preserved in the audit event exactly as supplied.

**Expected behavior**
Audit sanitization intended to protect the integrity of security-audit text should also neutralize known bidi-reordering/format characters, not just control characters, given the project's existing precedent (name_confusables.rs) of defending against Unicode-based spoofing in security-sensitive text.

**Actual behavior**
Only is_control()-classified characters are replaced; bidi override characters pass through unmodified.

```text
caller-controlled action/resource_type -> sanitize_for_audit() [is_control() only] -> audit log field (bidi char preserved)
```

**Impact**
A crafted request can produce a visually reordered or misleading audit record, weakening log-review integrity for a field explicitly designed to support security correlation.

**Suggested correction**
Extend sanitize_for_audit to also replace known bidi format characters (U+061C, U+200E/F, U+202A-E, U+2066-2069) in addition to control characters.

**How to verify**
Unit test passing a string containing U+061C through sanitize_for_audit and asserting it is replaced.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/error.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use crate::domain::deny::{service_errors, validation_messages as vm};
+use crate::domain::metrics_port::{ErrorType, FailClosedReason};
+
+/// A failure raised inside the evaluation pipeline.
+///
+/// Grouped by classification rather than by producing module, because the
+/// classification is what the variant exists to carry.
+///
+/// `Clone` — unlike `AuthZResolverError` — so [`crate::domain::hierarchy_cache`]
```
ainetx: ### **MINOR**: AuthZ PluginError bypasses the project's thiserror error-derive convention

**Severity**: **MINOR**

**Problem**
PluginError in domain/error.rs derives `Debug, Clone, PartialEq, Eq` only and manually implements `std::fmt::Display` and `std::error::Error`, instead of using `#[derive(Error, Debug)]` with per-variant `#[error("...")]` messages as the sibling RBAC DomainError in the same diff does.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Open gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/error.rs.
2. Observe `#[derive(Debug, Clone, PartialEq, Eq)]` on PluginError with no `Error` derive.
3. Observe the hand-written `impl std::fmt::Display for PluginError` (lines 170-200) and `impl std::error::Error for PluginError {}` (line 202), duplicating what thiserror would generate declaratively.

**Expected behavior**
PluginError should use `#[derive(Error, Debug)]` with `#[error("...")]` per variant, matching the convention already used for RBAC's DomainError in the same PR.

**Actual behavior**
Display/Error are implemented manually, separating the error-message contract from the enum declaration and diverging from the established pattern.

```text
PluginError enum -> manual impl Display (message text) + manual impl Error (marker) [instead of #[derive(Error)] + #[error(...)] per variant]
```

**Impact**
Error-message and variant evolution can drift from the enum declaration over time, and the domain error surface is inconsistent with the workspace's established structured-error convention, hurting maintainability.

**Suggested correction**
Add `#[derive(Error, Debug)]` (thiserror) to PluginError and move each Display arm's text into a per-variant `#[error("...")]` attribute, removing the manual Display/Error impls.

**How to verify**
Recompile after the refactor and confirm `Display`/`Error` behavior (message text, `source()`) is unchanged for every variant via existing error_tests.rs coverage.

</details>

gears/system/rbac/rbac/src/api/rest/role_definitions.rs
```diff
@@ -0,0 +1,544 @@
...
+            "integer",
+        )
+        .query_param("cursor", false, "Opaque base64url pagination cursor")
+        .handler(list_role_definitions)
+        .json_response_with_schema::<Page<RoleDefinitionDto>>(
+            openapi,
+            StatusCode::OK,
+            "Page of role definitions",
+        )
+        .with_odata_filter::<RoleDefinitionFilterField>()
```
ainetx: ### **MINOR**: RBAC OpenAPI metadata omits supported $orderby and undeclared ETag/Location response headers

**Severity**: **MINOR**

**Problem**
role_definitions.rs and role_assignments.rs list operations register only with_odata_filter::<...FilterField>(), never with_odata_orderby, yet infra/odata_err.rs's list_query_with_default_order explicitly preserves a caller-supplied $orderby (verified by its own unit test). Additionally, no call to OperationBuilder::response_header exists anywhere in the crate, even though several response descriptions state the response 'carries ETag and Location headers'.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Inspect role_definitions.rs / role_assignments.rs list route builders: only.with_odata_filter is called. 2. Inspect infra/odata_err.rs: an explicit $orderby is preserved, not overwritten. 3. Search the crate for response_header: zero matches despite ETag/Location being documented in response text.

**Expected behavior**
The generated OpenAPI document should declare $orderby support (allowed fields/order) via with_odata_orderby, and declare the ETag/Location response headers via response_header so generated clients can discover them.

**Actual behavior**
Generated OpenAPI spec omits sortable-field metadata and response header declarations that the implementation actually supports/returns.

```text
handler honors $orderby & sets ETag/Location headers --> OperationBuilder registers only $filter, no orderby/header metadata --> published OpenAPI contract under-documents real behavior
```

**Impact**
Generated SDK/API clients cannot discover supported sort fields or the presence of ETag/Location response headers from the published contract, though the runtime behavior actually supports/returns them.

**Suggested correction**
Add.with_odata_orderby::<RoleDefinitionFilterField>() /::<RoleAssignmentFilterField>() to the list operations, and add.response_header(...) declarations for ETag/Location on the relevant responses.

**How to verify**
Regenerate OpenAPI spec and confirm $orderby allowed fields and ETag/Location headers appear for the affected operations.

</details>

gears/system/rbac/rbac/src/odata.rs
```diff
@@ -0,0 +1,170 @@
...
+/// on string fields. `role_definition_id` is `Uuid` because the column is
+/// `uuid` and a text parameter bound against it is a database-level type
+/// error, not a filter that matches nothing.
+///
+/// A caller should not have to know that. Both spellings — `'<uuid>'`
+/// quoted and `<uuid>` bare — are accepted for both fields, because
+/// `infra::odata_normalize::normalize_filter_literals` coerces each literal
+/// into the kind its field declares before the filter is validated. The
+/// declared kinds here stay the ones the *column* needs.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
```
ainetx: ### **MINOR**: RBAC OData filter fields use hand-written FilterField impls instead of the ODataFilterable derive

**Severity**: **MINOR**

**Problem**
RoleAssignmentFilterField and RoleDefinitionFilterField in gears/system/rbac/rbac/src/odata.rs are manually declared enums with hand-written `impl FilterField` blocks (name()/kind()/FIELDS), rather than being generated via #[derive(ODataFilterable)] from a query-DTO struct as documented in docs/toolkit_unified_system/07_odata_pagination_select_filter.md and the TOOLKIT-ODATA-001 compliance rule.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Open gears/system/rbac/rbac/src/odata.rs. 2. Observe RoleAssignmentFilterField/RoleDefinitionFilterField are plain enums with a manual FilterField impl, no #[derive(ODataFilterable)] anywhere in the file.

**Expected behavior**
New filterable fields should be declared through a typed #[derive(ODataFilterable)] struct so field enum, allowed-operator metadata, and OpenAPI $filter documentation are generated from one source.

**Actual behavior**
Field names/kinds are duplicated by hand across the enum body and its FilterField impl, risking future drift between the two.

```text
new filterable field added --> must be added in both the enum arm and the manual impl match arms --> a missed arm silently diverges filter metadata from filter behavior
```

**Impact**
Future field additions can update one manual match but omit another, though this is consistent with a repo-wide existing pattern rather than a regression unique to this diff.

**Suggested correction**
Migrate to #[derive(ODataFilterable)] on a query struct, consistent with the documented toolkit convention, when convenient; not blocking given existing precedent elsewhere.

**How to verify**
Confirm generated FilterField enum via the derive macro produces the same field/kind set as today's hand-written enum.

</details>

gears/system/rbac/rbac/src/api/rest/permissions.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Listing the catalog still requires an authenticated caller —
+    // without this guard a mis-wired deployment lacking the upstream
+    // `AuthN` middleware would expose the full catalog unauthenticated.
+    let _ = require_authenticated(extension_ctx)?;
+
+    let limit = match query.limit {
+        Some(0) | None => DEFAULT_LIMIT,
+        Some(v) => v,
+    };
+    if limit > MAX_LIMIT {
```
ainetx: ### **MINOR**: Permission catalog page-size cap is enforced only at the REST layer, not on the shared trait

**Severity**: **MINOR**

**Problem**
list_permissions in api/rest/permissions.rs enforces MAX_LIMIT (200) before calling the catalog, but PermissionCatalog::list_permissions and paginate_catalog_page accept/consume a raw u32 limit with no cap of their own.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Inspect api/rest/permissions.rs: `if limit > MAX_LIMIT { return Err(...) }` runs before `state.catalog.list_permissions(...)`. 2. Inspect domain/ports/permission_catalog.rs: `paginate_catalog_page(..., limit: u32,...)` converts to usize with no upper bound check.

**Expected behavior**
The page-size bound should live on the shared PermissionCatalog port/paginate_catalog_page itself so any caller (REST or future in-process/SDK caller) is bounded uniformly.

**Actual behavior**
Only the REST handler enforces MAX_LIMIT; a hypothetical future in-process caller bypassing the REST handler could request an oversized page.

```text
REST handler checks limit <= MAX_LIMIT --> calls catalog.list_permissions(limit) --> trait/impl applies no bound of its own --> future non-REST caller could pass an arbitrarily large limit
```

**Impact**
Low risk today since the only caller is the REST handler which already enforces the bound, but the shared contract itself provides no defense-in-depth against future callers or misuse.

**Suggested correction**
Move/duplicate the MAX_LIMIT check into paginate_catalog_page or the PermissionCatalog trait's default behavior so it applies regardless of caller.

**How to verify**
Add a test calling PermissionCatalog::list_permissions directly with limit > 200 and assert it is rejected or clamped.

</details>

gears/system/rbac/rbac/src/config.rs
```diff
@@ -0,0 +1,545 @@
...
+/// rather than to RBAC, so both lists are configured rather than compiled in.
+///
+/// The list a grant appears in decides its `principal_type`:
+/// [`RbacServiceConfig::service_principal_grants`] writes `ServicePrincipal`,
+/// [`RbacServiceConfig::user_grants`] writes `User`. That type must match what
+/// the caller's token classifies as — a grant filed under the wrong one is
+/// never found, and the request is denied with no diagnostic anywhere.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct RoleGrant {
```
ainetx: ### **MINOR**: RoleGrant can be constructed with blank role/principal_id, bypassing its own invariant

**Severity**: **MINOR**

**Problem**
RoleGrant exposes public `role` and `principal_id` String fields with a documented nonblank/must-resolve invariant, but the struct itself has no `new()`/checked constructor -- the invariant is enforced only later by RbacServiceConfig::validate, so any caller (config deserialization aside) can build an invalid RoleGrant via a bare struct literal.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Open gears/system/rbac/rbac/src/config.rs. 2. Observe `pub struct RoleGrant { pub role: String, pub principal_id: String }` derives Deserialize only. 3. Observe no `impl RoleGrant { fn new(...) -> Result<Self,...> }` exists; blank-value rejection lives only in RbacServiceConfig::validate.

**Expected behavior**
RoleGrant should expose its own checked constructor validating nonblank role/principal_id, consistent with self-validating-constructor practice for invariant-bearing types.

**Actual behavior**
A RoleGrant{role: String::new(), principal_id: String::new()} literal compiles and is only rejected if and when RbacServiceConfig::validate happens to run over it.

```text
RoleGrant{..} literal built directly --> no local invariant check --> only caught later if RbacServiceConfig::validate is invoked on the containing config
```

**Impact**
Any future code path (tests, other config assembly, a different construction site) that builds a RoleGrant directly without routing through RbacServiceConfig::validate can carry an invalid grant silently until it fails deep in startup grant-seeding logic.

**Suggested correction**
Add an inherent `RoleGrant::new(role, principal_id) -> Result<Self, RbacConfigError>` that checks non-blank fields, and route deserialization + validate() through it or duplicate the check inline.

**How to verify**
Add a unit test constructing RoleGrant::new("", "x") and asserting it errors instead of compiling silently.

</details>

gears/system/rbac/rbac/src/gts/permissions.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::domain::actions;
+use crate::domain::resource_types::{
+    ROLE_ASSIGNMENT as ROLE_ASSIGNMENT_RESOURCE_TYPE,
+    ROLE_DEFINITION as ROLE_DEFINITION_RESOURCE_TYPE,
+};
+use toolkit_gts::{AuthzPermissionV1, gts_instance};
+
+// role-definition resource permissions
+
+gts_instance! {
```
ainetx: ### **MINOR**: RBAC permission catalog tests are inline and don't verify per-permission payload shape

**Severity**: **MINOR**

**Problem**
gears/system/rbac/rbac/src/gts/permissions.rs declares six AuthzPermissionV1 instances via gts_instance! but, unlike the sibling ledger/pricing gears which split catalog tests into a `permissions_tests.rs` file included via `#[path =...]`, defines its tests inline. The inline tests (all_rbac_permissions_are_registered_in_inventory, rbac_permission_inventory_covers_every_expected_id) only check the count of registered instances, membership of instance_id strings, and type_id — they never read each instance's `action`, `resource_type`, or `display_name` payload fields.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Change one gts_instance! declaration's `action: actions::WRITE.to_owned()` to `actions::READ.to_owned()` while leaving its `id` string unchanged. 2. Run the existing test suite. 3. Both tests still pass because they never inspect the action/resource_type/display_name payload.

**Expected behavior**
A sibling permissions_tests.rs (matching the ledger/pricing convention) asserting each permission's action and resource_type payload fields, not just its id string and count.

**Actual behavior**
Only id-string and count/type_id assertions exist, and they live inline in permissions.rs rather than in a paired _tests.rs file.

```text
gts_instance! declares (id, action, resource_type) -> inventory::submit! -> test only checks id string set & count -> action/resource_type drift undetected
```

**Impact**
A REST-enforced action/resource_type pair can silently diverge from its registered AuthzPermission metadata without any test catching it, producing incorrect permission-catalog data for RBAC consumers.

**Suggested correction**
Extract the test module into gears/system/rbac/rbac/src/gts/permissions_tests.rs (matching the established convention) and add assertions on each permission's action and resource_type payload fields, not just its instance_id.

**How to verify**
Mutate one permission's action/resource_type field, confirm the augmented test fails; confirm the new file follows the same `#[path=...]` inclusion pattern as ledger's permissions.rs.

</details>

gears/system/rbac/rbac/src/module.rs
```diff
@@ -0,0 +1,874 @@
...
+        let cancel = self.ctx.cancellation_token().clone();
+        tokio::spawn(async move {
+            let metrics = RbacMetricsMeter::from_global();
+            let mut ticker = tokio::time::interval(INVENTORY_REFRESH_INTERVAL);
+            loop {
+                tokio::select! {
+                    biased;
+                    () = cancel.cancelled() => break,
+                    _ = ticker.tick() => {
+                        match role_definition_repo::count_role_inventory(&inventory_db).await {
```
ainetx: ### **MINOR**: Detached RBAC inventory-refresh task has no bound on its per-tick DB await and no retained handle

**Severity**: **MINOR**

**Problem**
gears/system/rbac/rbac/src/module.rs spawns a background gauge-refresh task with `tokio::spawn` and never stores the returned JoinHandle. Cancellation is only observed by `tokio::select!` between ticks; once the `ticker.tick()` branch is selected and `count_role_inventory(&inventory_db).await` starts, the cancellation token is not polled again until that database call resolves.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Start the RBAC gear so the inventory task is spawned. 2. Trigger a database stall/hang during a tick (e.g. connection pool exhaustion or a long-running lock) exactly while `count_role_inventory` is executing. 3. Signal gear/process shutdown via the cancellation token. 4. Observe the task keeps holding its captured `inventory_db` handle and does not terminate until the stalled query eventually returns or errors.

**Expected behavior**
The per-tick database call should be bounded by a timeout (or otherwise made cancellation-aware), and/or the spawned task's JoinHandle should be retained so shutdown code can await/abort it with a bound.

**Actual behavior**
The task is fire-and-forget; a stalled COUNT query can keep the task (and its DB handle) alive indefinitely past a cancellation signal, since nothing polls the cancellation token again until the await resolves.

```text
spawn(task) [handle dropped] -> loop { select(cancel.cancelled(), ticker.tick()) } -> tick branch -> await count_role_inventory (unbounded, not select'd) -> cancel signal fires but is not observed until await completes
```

**Impact**
A stalled database query during shutdown/init failure leaves a background task and its DB resources alive beyond the gear's own lifecycle, a lifecycle/resource-leak gap on an otherwise low-risk metrics-only path.

**Suggested correction**
Wrap `count_role_inventory(&inventory_db)` in `tokio::time::timeout(...)` inside the tick branch, or race it against `cancel.cancelled()` via a nested select so cancellation can interrupt an in-flight query; optionally retain the JoinHandle so any future serve()/shutdown hook can await/abort it with a bound.

**How to verify**
Inject a DB call that never returns during a tick, assert cancelling the token still allows the task to observe cancellation within a bounded time instead of blocking on the stalled future.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/module.rs
```diff
@@ -0,0 +1,182 @@
...
+                     and `deny_unknown_fields` rejects typos): {err}"
+                ));
+            }
+        };
+
+        // `vendor` is required: the AuthZ Resolver Gateway selects this plugin
+        // by its `vendor`, so a missing/empty value would silently register
+        // the plugin under no vendor and leave it unselectable. Fail fast
+        // rather than guessing a default.
+        if config.vendor.is_empty() {
```
ainetx: ### **MINOR**: Whitespace-only AuthZ plugin vendor passes startup validation but breaks gateway plugin selection

**Severity**: **MINOR**

**Problem**
AuthZResolverPluginGear::init in gears/system/authz-resolver/plugins/authz-resolver-plugin/src/module.rs only checks `config.vendor.is_empty()` before registering the plugin under that vendor string. A value consisting solely of whitespace is non-empty, so it passes validation and gets registered, but the gateway's exact-match vendor selection (as documented in the surrounding comments) would never match it, effectively disabling the intended PDP for that vendor without any startup error.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Deploy with `modules.authz-resolver-plugin.config.vendor: " "` (e.g. from a templating bug that leaves whitespace instead of a real value). 2. Gear init succeeds because `is_empty()` is false. 3. The plugin registers under the whitespace vendor string. 4. The gateway's configured vendor selector (a real, non-whitespace value) never matches this instance, so the intended PDP plugin is effectively unreachable/unselected.

**Expected behavior**
Vendor validation should reject blank/whitespace-only strings the same way the empty-string case is rejected, e.g. `config.vendor.trim().is_empty()`.

**Actual behavior**
Only the literal-empty-string case is rejected; whitespace-only strings silently pass and get registered as a functionally-unmatchable vendor.

```text
YAML vendor: " " -> deserialize OK -> vendor.is_empty() == false -> registered under whitespace vendor -> gateway's exact-match selector never matches -> plugin effectively inert
```

**Impact**
A malformed deployment config can silently disable the intended authorization PDP plugin for a vendor with no startup failure, risking either an authorization gap or unexpected fallback behavior discovered only in production.

**Suggested correction**
Change the guard to `if config.vendor.trim().is_empty() {... }` (and consider the same treatment for other operator-supplied identifier strings in this config).

**How to verify**
Add a unit/integration test deploying with a whitespace-only vendor and assert `init()` now fails fast with the same error as the empty-string case.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/docs/DESIGN.md
```diff
@@ -0,0 +1,620 @@
...
+`subject_type`) pair, so borrowing the tag onto another id buys nothing: the **resource**
+leg is still validated, because a resource type is an ordinary registered type no matter
+who is asking.
+
+### 3.9 Audit
+
+- [x] `p2` - **ID**: `cpt-cf-authz-plugin-component-audit-emitter`
+
+One structured event per completed evaluation, allow or business deny, gated by
+`audit.enabled` (**default on**). A PDP that decides without leaving an audit trail is a
```
ainetx: ### **MAJOR**: AuthZ/RBAC PRD and DESIGN documents contradict each other and the shipped implementation on audit default and HTTP status contracts

**Severity**: **MAJOR**

**Problem**
The authz-resolver-plugin PRD normatively requires audit emission disabled by default (cpt-cf-authz-plugin-fr-audit-disabled-by-default: 'MUST be disabled by default'), but DESIGN.md §3.9 says the opposite ('default on') and the shipped config.rs implements 'default on' (AuditConfig::default() -> enabled: true). Independently, RBAC's DESIGN.md unit-test table documents CannotModifyBuiltInRole and RoleHasAssignments as mapping to HTTP 409, but api/rest/error.rs deliberately maps both to failed_precondition/400 (with an inline comment explaining the status was moved off 409), so the DESIGN.md test rows are stale relative to the shipped API contract.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Read authz-resolver-plugin/docs/PRD.md 'Disabled by Default' requirement (MUST be disabled). 2. Read authz-resolver-plugin/docs/DESIGN.md §3.9 ('default on'). 3. Read src/config.rs AuditConfig::default() -> enabled:true. 4. Read rbac/docs/DESIGN.md test rows U-61/U-62 (409). 5. Read rbac/src/api/rest/error.rs RoleDefinitionAssignmentsExist/BuiltInRoleNotModifiable handling -> failed_precondition (400).

**Expected behavior**
PRD, DESIGN, and the shipped code should agree on the audit default and on the HTTP status codes for built-in-role and has-assignments conflicts, or the divergence should be explicitly called out and reconciled in the docs.

**Actual behavior**
A deployment operator reading the PRD believes audit is off by default; the shipped plugin actually emits an audit record for every decision by default. An SDK/API consumer reading RBAC DESIGN.md's test catalog believes these two conflicts return 409, while the real API returns 400 failed_precondition.

```text
PRD.md (MUST default off) --x-- config.rs (default on) --> operators get unexpected audit volume
DESIGN.md test table (409) --x-- error.rs (400 failed_precondition) --> client-side status handling written against docs breaks
```

**Impact**
Operators and API consumers who implement against the documented contract (audit-off-by-default, 409 statuses) will build incorrect assumptions -- unexpected audit log volume in production, and client error-handling code that never matches on the actual 400 category returned for these two conflicts.

**Suggested correction**
Update authz-resolver-plugin DESIGN.md §3.9 (and/or the PRD's disabled-by-default requirement) so both documents and config.rs agree on one audit default; update RBAC DESIGN.md's U-61/U-62 test rows to reflect the actual 400 failed_precondition status now returned by api/rest/error.rs.

**How to verify**
Re-read the three documents plus config.rs/error.rs after the fix and confirm all sources state the same audit default and the same HTTP status for CannotModifyBuiltInRole/RoleHasAssignments.

</details>

gears/system/rbac/rbac/src/infra/storage/repo/role_definition_repo.rs
```diff
@@ -0,0 +1,860 @@
...
+/// * `Ok(None)` — the row vanished (concurrent delete). Returns
+///   `Ok(String::new())` so the caller surfaces `412` with an empty
+///   etag and the client follows the normal not-found retry path.
+/// * `Err(_)` — secondary DB failure (pool exhaustion, network blip,
+///   etc.). Surfaces as `Err(DomainError::*)` via `map_scope_err` so
+///   operators see a real `ServiceUnavailable`/`Internal` instead of
+///   a misleading `StaleEtag { current_etag: "" }` that pretends the
+///   DB is healthy.
+async fn fetch_current_etag(db: &impl DBRunner, id: Uuid) -> Result<String, DomainError> {
+    match role_definition::Entity::find_by_id(id)
```
ainetx: ### **MINOR**: CAS-recovery ETag lookup fetches unused JSONB payload columns

**Severity**: **MINOR**

**Problem**
fetch_current_etag() in role_definition_repo.rs uses role_definition::Entity::find_by_id(id).one(db), transferring and deserializing the full row (name, description, permissions, not_permissions, assignable_scopes JSONB) even though it only reads updated_at and id to construct the ETag.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. A client update()/delete() loses a CAS race (result.rows_affected == 0). 2. fetch_current_etag(db, id) is called to recover the current ETag. 3. The query fetches and deserializes the entire row including three JSONB columns that are never used.

**Expected behavior**
The recovery query should use select_only()/column() (or a narrow FromQueryResult struct) to fetch only id and updated_at.

**Actual behavior**
The full entity, including wide JSONB payloads, is transferred and deserialized on every lost-CAS recovery read.

```text
update()/delete() CAS lost -> fetch_current_etag() -> Entity::find_by_id().one() [fetches all columns] -> only.id/.updated_at used -> JSONB payload discarded
```

**Impact**
Extra DB transfer and JSON deserialization cost on a concurrency-sensitive recovery path; limited because it only triggers on lost CAS, not the common successful path.

**Suggested correction**
Use select_only().columns([Column::Id, Column::UpdatedAt]) with a narrow projection struct instead of the full entity in fetch_current_etag().

**How to verify**
Confirm the recovery query's generated SQL only selects id and updated_at, and that existing update/delete CAS-loss tests still pass.

</details>

gears/system/rbac/rbac/src/infra/storage/repo/role_definition_repo.rs
```diff
@@ -0,0 +1,860 @@
...
+        // row is rebuilt from `existing + patch + advanced` to avoid a
+        // follow-up SELECT that could observe a third writer.
+        let existing = role_definition::Entity::find_by_id(id)
+            .secure()
+            .scope_with(&AccessScope::allow_all())
+            .one(db)
+            .await
+            .map_err(|err| map_scope_err("update", err))?
+            .ok_or_else(|| {
+                DomainError::internal(format!(
```
ainetx: ### **MAJOR**: Concurrent deletion during update/delete surfaces as an internal 500 instead of a normal concurrency outcome

**Severity**: **MAJOR**

**Problem**
role_definition_repo.rs's update()/delete() re-read the row for CAS validation after the service layer's own pre-fetch. If a concurrent delete removes the row in that window, `.ok_or_else(|| DomainError::internal(...))` fires, producing a redacted Internal error (HTTP 500) instead of the StaleEtag/not-found outcome the same module already treats as a normal race elsewhere (fetch_current_etag's Ok(None) path).

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Client A calls GET role-definition (service pre-fetch via repo.find_by_id in role_definition/service.rs). 2. Client B deletes the same role-definition. 3. Client A calls PATCH/DELETE with the ETag it obtained. 4. update()/delete() in role_definition_repo.rs re-reads the row, finds None, and returns DomainError::internal(...) instead of StaleEtag/NotFound.

**Expected behavior**
A concurrent delete discovered during the repository's CAS pre-read should map to a client-facing StaleEtag (empty etag, as done in fetch_current_etag) or a not-found outcome, consistent with the CAS-lost recovery path in the same file.

**Actual behavior**
The row-vanished branch in update()/delete()'s initial find_by_id call returns DomainError::Internal, which the REST layer renders as HTTP 500.

```text
Service pre-fetch (row exists) -> [race: concurrent delete] -> repo.update/delete's own find_by_id -> None -> DomainError::internal -> HTTP 500 (should be StaleEtag/404)
```

**Impact**
Clients cannot reliably distinguish or recover from a normal concurrent deletion during optimistic-concurrency mutations and instead receive a misleading server-failure response, undermining the documented CAS contract.

**Suggested correction**
In update()/delete(), map the initial find_by_id's None case to the same graceful outcome used by fetch_current_etag (e.g. Err(DomainError::StaleEtag { current_etag: String::new() }) or a not-found domain error) instead of DomainError::internal.

**How to verify**
Add a test that deletes the row between the service's read and the repo's update/delete call and assert the caller receives StaleEtag/NotFound rather than Internal/500.

</details>

gears/system/rbac/rbac/src/module.rs
```diff
@@ -0,0 +1,874 @@
...
+            rg_read,
+            Arc::new(RbacMetricsMeter::from_global()),
+        ));
+
+        let local_client: Arc<dyn RbacServiceClientV1> =
+            Arc::new(RbacServiceLocalClient::new(Arc::clone(&evaluator)))
+                as Arc<dyn RbacServiceClientV1>;
+        self.ctx
+            .client_hub()
+            .register::<dyn RbacServiceClientV1>(local_client);
```
ainetx: ### **MAJOR**: RbacServiceClientV1 is published to ClientHub before schema registration, seeding, and bootstrap succeed

**Severity**: **MAJOR**

**Problem**
RbacRuntimeBuilder::with_evaluator_and_local_client() registers the dyn RbacServiceClientV1 in ClientHub synchronously, but init() only registers GTS schemas, seeds built-in roles, and runs bootstrap/configured grants afterward, and only then publishes the ArcSwapOption runtime slot. If any of those later steps fails, init() returns Err, but the ClientHub registration that already occurred is not rolled back.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. RBAC gear's init() runs with_evaluator_and_local_client(), publishing dyn RbacServiceClientV1 into ClientHub. 2. register_schemas()/seeding/bootstrap subsequently fails (e.g. registry outage, bad grant config). 3. init() returns Err, but any consumer that already resolved dyn RbacServiceClientV1 from ClientHub during this window (or that continues operating post-failure, depending on gear-failure handling) holds a client backed by an evaluator whose built-in roles/bootstrap grants are not guaranteed to exist.

**Expected behavior**
The dyn RbacServiceClientV1 should be published to ClientHub only after schema registration, built-in role seeding, and bootstrap/configured grants have all succeeded -- mirroring the same completeness guarantee already implemented for the ArcSwapOption runtime slot.

**Actual behavior**
ClientHub registration happens unconditionally and early in the builder chain, well before the side-effecting steps that this same file's own comments say must complete before anything is considered 'ready'.

```text
with_evaluator_and_local_client() -> ClientHub.register(client) [visible now] -> register_schemas()/seed()/bootstrap() [can still fail] -> runtime.store(Some(...)) [only this is gated]
```

**Impact**
Dependent consumers resolving the RBAC client via ClientHub may observe a partially initialized RBAC surface (missing seeded roles/grants) if RBAC's own init later fails, inconsistent with the documented all-or-nothing completeness contract.

**Suggested correction**
Defer client_hub().register::<dyn RbacServiceClientV1>() until after register_schemas(), the seeding transaction, and bootstrap/configured grants have all succeeded -- e.g. build the local client but only register it in ClientHub at the same point the ArcSwapOption runtime is published.

**How to verify**
Add a test/integration scenario where seeding or bootstrap fails after with_evaluator_and_local_client() runs, and assert that ClientHub does not expose a functioning RbacServiceClientV1 in that failure case.

</details>

gears/system/rbac/rbac/src/infra/bootstrap.rs
```diff
@@ -0,0 +1,265 @@
...
+pub enum BootstrapDecision {
+    /// Proceed and run the bootstrap with the given subject identifier.
+    Run(String),
+    /// Skip the bootstrap because no subject identifier was configured.
+    Skip,
+}
+
+/// Decide whether the bootstrap should run, given the optional subject id.
+#[must_use]
+pub fn evaluate_bootstrap_decision(platform_admin_subject_id: Option<&str>) -> BootstrapDecision {
```
ainetx: ### **MAJOR**: Blank platform_admin_subject_id passes config validation and creates an unusable Owner assignment

**Severity**: **MAJOR**

**Problem**
RbacServiceConfig::validate() checks grant lists for blank role/principal_id fields but has no equivalent check for platform_admin_subject_id. evaluate_bootstrap_decision(Some("")) returns Run(""), and BootstrapPlatformAdmin::run/build_role_assignment_active_model insert an Owner-at-/ role_assignment row with principal_id="" without any non-empty validation, reporting BootstrapOutcome::Created as success.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Deploy RBAC gear with platform_admin_subject_id set to an empty string (e.g. an unresolved env var placeholder). 2. cfg.validate() passes because it never checks this field. 3. init() calls evaluate_bootstrap_decision(Some("")) -> Run("") -> BootstrapPlatformAdmin::run(&conn, "") inserts a role_assignments row with principal_id="". 4. Logs report 'rbac: platform-admin bootstrap created Owner-at-/ assignment' as if successful, but no real subject can ever match this row.

**Expected behavior**
Config validation should reject a blank/whitespace-only platform_admin_subject_id the same way it rejects blank grant fields, failing init() loudly rather than silently seeding an unusable assignment.

**Actual behavior**
The blank subject id is accepted, treated as a valid bootstrap subject, and persisted as an empty-string principal_id, leaving the deployment without a functioning platform administrator while reporting success.

```text
config.validate() [no check on platform_admin_subject_id] -> evaluate_bootstrap_decision(Some("")) -> Run("") -> INSERT role_assignments(principal_id='') -> Created (false success)
```

**Impact**
A deployment can start successfully but have no functional platform administrator, requiring manual database inspection/repair before RBAC can be administered at all -- a self-inflicted lockout with a misleading success log.

**Suggested correction**
In RbacServiceConfig::validate(), reject a present-but-blank/whitespace-only platform_admin_subject_id (mirroring validate_grants' blank checks), and/or add a non-empty guard in build_role_assignment_active_model/BootstrapPlatformAdmin::run.

**How to verify**
Add a config validation test asserting Some("") or Some(" ") for platform_admin_subject_id returns an RbacConfigError, and a bootstrap unit test asserting run() with an empty subject_id errors rather than inserting a row.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/evaluate.rs
```diff
@@ -0,0 +1,610 @@
...
+            .capabilities
+            .iter()
+            .map(|c| match c {
+                Capability::TenantHierarchy => "tenant_hierarchy",
+                Capability::GroupMembership => "group_membership",
+                Capability::GroupHierarchy => "group_hierarchy",
+            })
+            .collect();
+        caps.sort_unstable();
+        self.metrics.inc_capability_negotiation(&caps.join(","));
```
ainetx: ### **MINOR**: Capability-negotiation metric label is not deduplicated, allowing unbounded cardinality from a caller-supplied Vec

**Severity**: **MINOR**

**Problem**
`emit_request_metrics` maps `request.context.capabilities` (a `Vec<Capability>`) to strings, sorts, and joins them into one OTel attribute value via `inc_capability_negotiation`. There is no deduplication and no length cap; `validation.rs` does not constrain the size or uniqueness of this Vec.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. A PEP/caller constructs an `EvaluationRequest` with `context.capabilities` containing the same `Capability` variant repeated N times. 2. `evaluate()` reaches `emit_request_metrics`. 3. `caps.join(",")` produces a string of length proportional to N, and each distinct N yields a distinct label value stored as an OTel `KeyValue` attribute.

**Expected behavior**
The capability set should be deduplicated (and/or capped) before being used as a metric label, since only 3 known capability values are meaningful.

**Actual behavior**
The raw (possibly repeated) capability Vec is joined verbatim into the metric label, producing unbounded label cardinality from caller input.

```text
caller -> Vec<Capability> (dup'd) -> caps.sort_unstable(); caps.join(",") -> OTel attribute (unbounded strings)
```

**Impact**
A faulty or misbehaving internal caller can generate high-cardinality metric series / large attribute strings, degrading the metrics backend, without affecting the authorization decision itself.

**Suggested correction**
Deduplicate (e.g. via a `BTreeSet`) before sorting/joining, and/or cap the number of capabilities considered.

**How to verify**
Add a test asserting that a request with repeated capabilities produces the same metric label as one with the deduplicated set.

</details>

gears/bss/ledger/ledger/src/module.rs
```diff
@@ -908,24 +908,17 @@ impl Gear for BssLedgerGear {
         // `MissingConfigSection` arms both fall through here, so
         // `unwrap_or_default()` yields the parsed config or the all-defaults
         // config respectively (the `GearNotFound` / parse-error arms already
-        // returned). Abort init loudly on a present-but-invalid jobs cadence
-        // (a zero tick would panic `tokio::time::interval` in `serve`).
+        // returned). Abort init loudly on an invalid or incomplete config: a
+        // zero job cadence would panic `tokio::time::interval` in `serve`, and
+        // an absent `seller_tenant_types` would turn every provisioning request
+        // into a "not a ledger owner" rejection. Note this makes a gear entry
+        // with no `config:` block a startup error, not a defaults-only run.
```
ainetx: ### **MINOR**: No explicit upgrade/rollout note for the newly mandatory seller_tenant_types field

**Severity**: **MINOR**

**Problem**
`BssLedgerGear::init` now calls `cfg.validate()` which rejects an empty `seller_tenant_types`, turning a previously-valid defaults-only `bss-ledger:` gear entry into a hard startup failure (module.rs comment: 'this makes a gear entry with no `config:` block a startup error, not a defaults-only run'). The README documents the new required field and its validation behavior but does not call out this as a breaking change for existing deployments or provide upgrade guidance beyond the new config example.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. An existing deployment has `bss-ledger:` configured with no `config:` block (previously valid, ran on defaults). 2. Deploy this change. 3. `ctx.config()` returns `MissingConfigSection`, `cfg` becomes `BssLedgerConfig::default()` with empty `seller_tenant_types`. 4. `cfg.validate()` returns `ConfigError::Required`, and `init()` returns `Err`, failing gear startup.

**Expected behavior**
Documentation should explicitly flag this as a breaking change for deployments upgrading from a pre-this-PR version and give guidance on how to discover valid seller tenant type IDs before upgrading (beyond the general config example).

**Actual behavior**
The README's Configuration section reads as a description of the current desired state; there is no 'Upgrading' or 'Breaking changes' note warning existing deployments they must add this field before/during upgrade.

```text
old deployment (no config: block) -> upgrade -> init() -> cfg.validate() -> ConfigError::Required -> gear fails to start (no advance warning in docs)
```

**Impact**
Every existing deployment relying on defaults-only ledger startup fails at the next deploy until an operator reads the error and adds the field, causing avoidable deployment downtime/toil.

**Suggested correction**
Add a short 'Upgrading' note in the README or release notes calling out the new mandatory field and pointing at how to discover valid tenant type IDs (e.g. via the types-registry endpoint already mentioned).

**How to verify**
Check the README for an explicit upgrade/compatibility section referencing this change.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/audit_emitter.rs
```diff
@@ -0,0 +1,239 @@
...
+        // empty/zero default rather than the string "None".
+        let subject_tenant_id = record
+            .subject_tenant_id
+            .map(|id| id.to_string())
+            .unwrap_or_default();
+        let resource_id = record
+            .resource_id
+            .map(|id| id.to_string())
+            .unwrap_or_default();
+        let constraints_count = record.constraints_count.unwrap_or(0) as u64;
```
ainetx: ### **MINOR**: Audit records don't flag degraded fields, and the emitter's own doc comment misstates the audit-enabled default

**Severity**: **MINOR**

**Problem**
`compute_constraints_hash` returns `None` for both `constraints.is_empty()` and JSON-serialization failure (only warn-logging the latter outside the audit record). `sanitize_for_audit` silently truncates at 256 chars and replaces control characters with no indication in the emitted record that alteration occurred. Separately, audit_emitter.rs's module doc says audit is 'Gated by `config.audit.enabled` (default `false`)' while `AuditConfig::default()` in config.rs sets `enabled: true`.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Enable audit. 2. Trigger an allow decision whose constraint set fails to serialize (or is legitimately empty) -- both cases show `constraints_hash=""` in the audit event, though `constraints_count` differs and only the failure path emits a separate `warn!`. 3. Trigger a request with an oversized/control-char action name -- the emitted `action` field is silently altered/truncated with no flag distinguishing it from a genuinely short value.

**Expected behavior**
The audit record (or accompanying doc) should let a consumer distinguish a legitimately-empty hash from a degraded/failed one, mark truncated/sanitized fields, and the doc comment should match the actual default.

**Actual behavior**
Both cases render identically; the doc comment claims a false default that contradicts config.rs's Default impl.

```text
allow w/ 0 constraints -> hash=None -> "" | allow w/ N constraints, serde_json::to_writer fails -> warn! + hash=None -> "" (same audit output)
```

**Impact**
Weakens incident investigation and audit-trail reliability, and could mislead an operator reading the source about whether audit is on by default.

**Suggested correction**
Fix the doc comment to say default true; consider adding a degraded/serialization_failed marker to AuditRecord distinguishable from a legitimate zero-count allow.

**How to verify**
Grep audit_emitter.rs doc comment vs AuditConfig::default in config.rs for consistency; add a test asserting the audit record distinguishes both cases.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/evaluate.rs
```diff
@@ -0,0 +1,610 @@
...
+        tenant_resolver: Arc<dyn TenantResolverClient>,
+        resource_group: Arc<dyn ResourceGroupReadHierarchy>,
+        types_registry: Arc<dyn TypesRegistryClient>,
+        metrics: Arc<AuthZMetrics>,
+    ) -> Self {
+        // Clone the config / rbac Arc handles once before the moves so the
+        // ScopeEnforcer and PolicyEvaluator can keep their own references.
+        let scope_enforcer = ScopeEnforcer::new(Arc::clone(&config));
+        // Compiled once from config: the PDP's trusted-actor bypass surface.
+        let trusted_system_actors = crate::domain::subject_type::TrustedSystemActors::from_config(
```
ainetx: ### **MINOR**: No documented restart/reload requirement for revoking a trusted-system-actor bypass

**Severity**: **MINOR**

**Problem**
`TrustedSystemActors::from_config` is called once during `AuthZResolverPlugin` construction in `module.rs::init()` and the resulting set lives for the process/gear lifetime. Neither the README's 'Trusted system actors' section nor the PRD states that removing/changing an entry requires a gear/process restart before the change takes effect.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Deploy with a trusted_system_actors entry. 2. Edit config to remove the entry (intending to revoke the bypass). 3. Without restarting the running process, the in-memory `TrustedSystemActors` set still contains the old entry and continues bypassing policy for that subject.

**Expected behavior**
Documentation should explicitly state that config changes to trusted_system_actors require a restart (this is the standard toolkit gear-config model) to take effect/be revoked.

**Actual behavior**
No restart/reload/revocation statement exists in README.md, PRD.md, or code comments for this security-sensitive bypass list.

```text
config edit (remove actor) -> running process's in-memory TrustedSystemActors (unchanged) -> bypass still active until restart (undocumented)
```

**Impact**
An operator could believe a privileged bypass was revoked by editing config alone, while a running instance continues to honor it until the next restart.

**Suggested correction**
Add one sentence to the README/PRD trusted-actor sections stating that changes take effect only after a full gear/process restart.

**How to verify**
Check README/PRD for an explicit restart-to-apply statement in the trusted system actors section.

</details>

gears/system/rbac/rbac/src/api/service/error_mapping.rs
```diff
@@ -0,0 +1,203 @@
...
+            // layer logs that and returns a generic envelope.
+            DomainError::ServiceUnavailable {
+                detail,
+                retry_after,
+                cause,
+            } => {
+                log_dropped_cause(Level::Warn, cause.as_ref(), &detail);
+                Self::ServiceUnavailable {
+                    detail,
+                    retry_after_seconds: retry_after.map(|d| d.as_secs()),
```
ainetx: ### **MINOR**: Sub-second RBAC retry_after hints are truncated to zero at the SDK boundary

**Severity**: **MINOR**

**Problem**
`From<DomainError> for RbacServiceError` converts `retry_after: Option<Duration>` via `Duration::as_secs()`, which floors any value under one second to 0. The `ServiceUnavailable.retry_after` field is a general-purpose `Option<Duration>` with no documented whole-second constraint, so any future producer supplying e.g. 250ms would have that intent silently discarded.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. A future/foreign dependency failure constructs `DomainError::ServiceUnavailable { retry_after: Some(Duration::from_millis(250)),.. }`.
2. The boundary mapping in error_mapping.rs converts it with `.as_secs()`.
3. The wire response carries `retry_after_seconds: Some(0)` instead of an actionable delay.

**Expected behavior**
A sub-second retry hint should round up (e.g. `as_secs_f64().ceil()` or a minimum of 1) rather than floor to 0, or the field should be typed/documented to reject sub-second values at construction.

**Actual behavior**
`as_secs()` truncates toward zero, turning any positive delay under 1s into `Some(0)`.

```text
DomainError::ServiceUnavailable{retry_after: Some(250ms)} -> as_secs() -> retry_after_seconds: Some(0) -> client sees near-immediate retry hint
```

**Impact**
If any future producer supplies a sub-second retry hint, clients following the hint could retry immediately rather than backing off, though no current call site in this diff exercises this path.

**Suggested correction**
Use `d.as_secs_f64().ceil() as u64` (or `.max(1)`) instead of `as_secs()` when converting a positive Duration to whole seconds.

**How to verify**
Add a test asserting `Duration::from_millis(250)` maps to a non-zero `retry_after_seconds`.

</details>

gears/system/authz-resolver/plugins/authz-resolver-plugin/src/domain/gts_type_validator.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        // Cache write: only Known and Unknown are stable; RegistryUnavailable
+        // is transient and must not poison the cache.
+        if matches!(result, ValidationResult::Known | ValidationResult::Unknown) {
+            let entry = CacheEntry {
+                result: result.clone(),
+                valid_until: now + self.ttl,
```
ainetx: ### **MINOR**: GTS type-validator cache has no concurrent-miss coalescing or write-time freshness check

**Severity**: **MINOR**

**Problem**
Two concurrent cache misses for the same GTS type id independently call `self.registry.get_type_schema()` and each unconditionally `put()`s its own result into the shared LRU with no version/ordering check, so a slower response can overwrite a fresher one written by a faster concurrent request.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Type X is unregistered; two concurrent requests validate type X.
2. Both miss the cache and call the registry concurrently.
3. Between the two calls, type X gets registered.
4. The request that started before registration (sees Unknown) finishes after the one that started later (sees Known) and overwrites the cache with Unknown.
5. Subsequent requests read the stale Unknown entry for up to the full TTL.

**Expected behavior**
Concurrent misses for the same key should coalesce (singleflight, as done in hierarchy_cache.rs) or at least avoid overwriting a newer entry with an older result.

**Actual behavior**
Cache writes are unconditional last-writer-wins with no ordering guarantee.

```text
ReqA(slow, pre-registration)->Unknown; ReqB(fast, post-registration)->Known->cache.put(Known); ReqA finishes later->cache.put(Unknown) overwrites Known
```

**Impact**
During a type-registration rollout under concurrent load, Strict mode can deny an already-registered type (or continue allowing a just-removed one) for up to one TTL window.

**Suggested correction**
Add singleflight coordination for the GTS validator cache (mirroring hierarchy_cache.rs) or record a monotonic fetch-start token and only accept a write if it is not older than the current cached entry.

**How to verify**
Add a test with two overlapping registry calls returning different results and assert the cache converges on the response of the call that started later, not the one that finished last.

</details>

gears/system/rbac/rbac/src/domain/service/permission_evaluator.rs
```diff
@@ -0,0 +1,1249 @@
...
+}
+
+/// In-process permission evaluation engine behind
+/// `RbacServiceClientV1::get_subject_roles` and `evaluate_permission`.
+///
+/// Generic over the two repositories rather than holding them as trait
+/// objects: their methods take the executor as `&C where C: DBRunner`,
+/// which is not dyn-compatible. `db` is the evaluator's own connection
+/// source — every read here is a single statement, so the evaluator takes
+/// a connection per call and never opens a transaction.
```
ainetx: ### **MINOR**: PermissionEvaluator combines assignment and role-definition reads from independent, non-atomic connections

**Severity**: **MINOR**

**Problem**
`get_subject_roles`/`evaluate_permission` fetch role assignments and then role definitions as two separate statements, each taking its own connection from `DBProvider`, with no shared transaction or snapshot. The evaluator's own doc comment acknowledges this explicitly as a design choice for simplicity, but it means the two reads are not mutually consistent.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Subject S has an assignment to role R.
2. Concurrently, an admin revokes S's assignment to R (deletes the row) right after step 1's assignment query has run but before the role-definition lookup for R executes.
3. The evaluator still uses the already-fetched assignment for R and evaluates the permission as if it were still granted.

**Expected behavior**
Either an explicit documented consistency contract stating the acceptable staleness window and its bound, or a single transaction/snapshot for one evaluation call.

**Actual behavior**
Reads are two independent statements over separate connections with no coordination.

```text
T1: SELECT assignments (sees R) | T2 (concurrent): DELETE assignment R | T1: SELECT role_definitions(R) still succeeds -> permission evaluated using a revoked assignment
```

**Impact**
A revocation racing an evaluation can be missed for the duration of that single evaluation call; bounded and narrow, but present in the core authorization path.

**Suggested correction**
Wrap the assignment + role-definition reads for one evaluation in a single transaction/snapshot (e.g. REPEATABLE READ) or explicitly document the accepted staleness bound and why it's safe.

**How to verify**
Add a concurrency test that revokes an assignment between the two evaluator reads and asserts the expected (documented) behavior.

</details>

gears/system/rbac/rbac/src/api/rest/dto.rs
```diff
@@ -0,0 +1,399 @@
...
+// ---------------------------------------------------------------------------
+// Role-definition DTOs
+// ---------------------------------------------------------------------------
+
+/// REST mirror of [`PermissionRule`]. The Allow / Deny class is encoded
+/// by which array (`permissions` vs `not_permissions`) it lives in on
+/// the surrounding entity — the DTO itself only carries the
+/// `(operation, target_type)` pair.
+#[derive(Debug, Clone)]
+#[toolkit_macros::api_dto(request, response)]
```
ainetx: ### **MINOR**: Nested PermissionRuleDto silently accepts unknown fields despite strict top-level and schema contracts

**Severity**: **MINOR**

**Problem**
PermissionRuleDto lacks `#[serde(deny_unknown_fields)]`, so unknown members inside `permissions`/`not_permissions` array entries in CreateRoleDefinitionRequest/PatchRoleDefinitionRequest are silently ignored by serde, even though the top-level request DTOs reject unknown fields and the vendored JSON Schema's permission_rule definition sets additionalProperties: false.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. POST /rbac/v1/role-definitions with a body containing `"permissions": [{"operation": "read", "target_type": "gts.foo.*", "typo_field": true}]`.
2. The top-level deny_unknown_fields does not apply to the nested array element.
3. `typo_field` is silently dropped; the request succeeds as if the field were never sent.

**Expected behavior**
Unknown fields inside a permission rule object should be rejected with a 400, matching the top-level request strictness and the vendored schema's additionalProperties: false.

**Actual behavior**
serde silently ignores the unrecognized nested field and the request succeeds.

```text
Client sends permissions:[{...,unknown_field}] -> top-level deny_unknown_fields checks only top-level keys -> PermissionRuleDto (no deny_unknown_fields) drops unknown_field -> 200 OK, field silently ignored
```

**Impact**
Clients may believe an attribute was accepted/applied when it was silently dropped, and REST behavior diverges from the documented/vendored schema contract; low security impact since no privilege-relevant data is affected.

**Suggested correction**
Add `#[serde(deny_unknown_fields)]` to PermissionRuleDto (or to a request-specific variant if the response side must stay lenient).

**How to verify**
Add a test posting a permission rule with an extra unknown key and assert the request is rejected with 400 rather than succeeding.

</details>

ainetx (changes requested):
Requesting changes — this PR introduces two new gears (RBAC and the authz-resolver-plugin) with solid architecture, but four issues need to be resolved before this can merge, spanning both documentation/contract integrity and correctness of concurrency and startup behavior.

- **AuthZ/RBAC PRD, DESIGN, and implementation disagree on audit defaults and HTTP status contracts** — The authz-resolver-plugin PRD mandates audit *disabled* by default, DESIGN.md says *enabled* by default, and the shipped `config.rs` implements enabled-by-default — pick one and align all three. Separately, RBAC's DESIGN.md still documents `CannotModifyBuiltInRole`/`RoleHasAssignments` as HTTP 409 even though `api/rest/error.rs` deliberately moved both to 400/`failed_precondition`. Docs need to catch up to the shipped contract (or vice versa, if the shipped behavior is the mistake). ([discussion](https://github.com/constructorfabric/gears-rust/pull/4736#discussion_r4084364999))
- **Concurrent deletion during update/delete leaks as a 500 instead of a normal race outcome** — `role_definition_repo.rs`'s CAS re-read in `update()`/`delete()` treats a row disappearing between the service's pre-fetch and the repo's re-read as `DomainError::internal(...)`, producing an opaque 500. The same file already handles this exact race gracefully elsewhere (`fetch_current_etag`'s `Ok(None)` path) — this path should resolve to StaleEtag/not-found instead. ([discussion](https://github.com/constructorfabric/gears-rust/pull/4736#discussion_r4084365642))
- **`RbacServiceClientV1` is published to ClientHub before it's actually ready to serve** — `RbacRuntimeBuilder::with_evaluator_and_local_client()` registers the client synchronously, but schema registration, built-in role seeding, and bootstrap/configured grants all happen afterward in `init()`. If any of those later steps fail, `init()` returns `Err` but the already-registered ClientHub entry is never rolled back, leaving a client other gears can call against incomplete state. ([discussion](https://github.com/constructorfabric/gears-rust/pull/4736#discussion_r4084366040))
- **Blank `platform_admin_subject_id` sails through config validation and produces a broken Owner grant** — `RbacServiceConfig::validate()` checks grant lists for blank principal/role fields but has no such check for `platform_admin_subject_id`. An empty string flows through `evaluate_bootstrap_decision` and `BootstrapPlatformAdmin::run`, inserting an Owner-at-`/` role assignment with `principal_id=""` and reporting success. Add the missing non-empty validation. ([discussion](https://github.com/constructorfabric/gears-rust/pull/4736#discussion_r4084366419))

PR #4740, opened 2026-09-09
feat(toolkit): projectable streaming with fallible open (#4734)

### LINKED ISSUE #4734: toolkit-contract: projectable streaming for multipart/mixed, and for a stream whose open can fail
## Summary

`#[toolkit::rest_contract]` can express exactly one streaming shape: an infallible `text/event-stream` client, with no server side at all. This requests two additions - a transport selector on `#[streaming]` covering `multipart/mixed`, and a streaming method whose open is awaited and can fail - so a long-lived streaming endpoint becomes projectable on both ends.

## Motivation

event-broker publishes two consumption transports, `GET /v1/events:stream` (`multipart/mixed; boundary=...`) and `GET /v1/events:sse` (`text/event-stream`). Both stay hand-registered, and neither gets a generated client.

- **No server side for either.** A `#[streaming]` method that is not `#[server_manual]` is a `compile_error!` (`libs/toolkit-contract-macros/src/rest_contract.rs:1291-1303`). `generate_unary_handler` (`:1527-1663`) is the only handler generator, so both routes are hand-written `OperationBuilder` chains (`gears/system/event-broker/event-broker/src/api/rest/routes/mod.rs:264-306`) over hand-written handlers (`.../handlers/delivery/streaming.rs:73-109`).
- **No client for multipart.** The generated streaming client hardcodes `.header("accept", "text/event-stream")` (`rest_contract.rs:946-951`) and the runtime has one parser, `runtime/sse.rs`. There is no multipart reader or writer anywhere in the workspace: `grep -rn 'multipart\|boundary' libs/toolkit-http/src/` is empty; `operation_builder.rs:928-943` is request-side form-data OpenAPI metadata only; `gears/system/oagw/oagw-sdk/src/multipart.rs` is a request-side form-data writer over a finite part list (`:239-241`, `:325-330`).
- **The SSE client does not work against the SSE endpoint either.** The gear puts the frame kind on the `event:` line (`.../handlers/delivery/encode.rs:22-29`, `:57-63`) and emits no `id:`, no `retry:`, no `event: done` and no `event: error`. The parser ignores every named event (`libs/toolkit-contract/src/runtime/sse.rs:464-468`, the `_ => false` arm) and then errors on the close for want of a `done` (`runtime/client.rs:359-363`). So today a generated client would silently drop every frame and end with a transport error.
- **A streaming method cannot fail at open.** `rewrite_streaming_signature` strips `async` and the outer `Result` (`libs/toolkit-contract-macros/src/projection.rs:108-116`), the base macro does the same (`codegen.rs:49-60`), and `#[streaming] async fn` is a hard error (`parse.rs:124-132`). event-broker's real method is `async fn stream(..) -> Result<FrameStream, EventBrokerError>` (`gears/system/event-broker/event-broker-sdk/src/api.rs:543-547`) and the consumer branches on three distinct open-time outcomes (`.../event-broker-sdk/src/consumer/dispatcher.rs:474-507`, `:1332-1372`): `PositionsNotSet` re-seeks and retries, anything else re-JOINs. Opening the stream is where `404`, `409 PositionsNotSet`, `409 StreamingInProgress` and `410` are raised, all before the first frame.

## Why the obvious implementation is inadequate

**Reporting the open failure as the stream's first item** is what the current signature forces, and it is wrong on this contract. The frame vocabulary is closed and the first frame is specified to be the open-time `topology` baseline (`gears/system/event-broker/docs/features/0004-consumption-transport.md`), so an error frame has nowhere to go. It also demotes four HTTP statuses with `Problem` bodies to a `200`, losing what gateways and `Retry-After` read, and it cannot express "no stream was created" - a `200` is a created stream. The exclusion lease makes this concrete: it is acquired by the open (`.../domain/delivery.rs:607-614`) and owned by the returned stream's lifetime, so a lazily-connecting stream records no lease until first poll and a dropped-unpolled stream releases nothing.

**A generic `Frame`/`FrameKind` abstraction in the toolkit** buys nothing. The frame set is already one serde-taggable enum (`.../handlers/delivery/dto.rs:149-171`, `#[serde(tag = "kind")]`), both transports serialize the same DTO (`encode.rs:4-7`), and a toolkit envelope would force `topology`/`control` - whose fields are inline, not nested under `payload` - into a shape they do not have. The toolkit should carry framing and media type; frame typing belongs to the contract's `Item` type.

**A strict RFC 2046 parser** does not fit the wire. No closing `--<boundary>--` is ever written, and the boundary opening part N+1 is written only when frame N+1 exists, so a parser keyed on the next delimiter runs a full frame behind a live stream - the e2e reader documents exactly this and scans for a complete JSON object instead (`testing/e2e/suites/event_broker/conftest.py:323-334`).

## Requested capability (for discussion)

- **Transport on the marker**, parsed like `#[idempotency(NonIdempotentWrite)]` (`parse.rs:180-207`): `#[streaming]` (unchanged, SSE), `#[streaming(sse)]`, `#[streaming(multipart_mixed)]`. Additive, so every existing projection keeps its meaning. It selects three things and nothing else: the client's `Accept`, the server's emitted `Content-Type`, the OpenAPI media-type key.
- **Fallible open needs no new syntax.** Let `async fn` on a `#[streaming]` method mean an awaited, fallible open - replacing the hard error at `parse.rs:124-132`, which already inspects `sig.asyncness` at that exact point. `fn` keeps today's infallible shape. Emitted signature: `async fn .. -> Result<Pin<Box<dyn Stream<Item = Result<Item, E>> + Send>>, E>`. On the client this removes code: the four `stream::once(async { Err(..) })` bail-outs at `rest_contract.rs:886-935` collapse to `?`.
- **One framing enum, matched, not a `dyn` decoder trait** (a decoder trait needs `fn decode<T>(..)` and cannot be object-safe):

```rust
// libs/toolkit-contract/src/runtime/framing.rs
#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[non_exhaustive]
pub enum StreamFraming { ServerSentEvents, MultipartMixed }

impl StreamFraming {
    /// The media type the client advertises in `Accept` and the server emits.
    pub const fn media_type(self) -> &'static str;
}
```

- **Split the runtime helper**, so the connect and status check happen eagerly:

```rust
pub async fn open_streaming<F, T>(request: StreamRequest<F>)
    -> Result<Pin<Box<dyn Stream<Item = Result<T, TransportError>> + Send>>, TransportError>;
```

  `StreamRequest` carries factory, framing, reconnect policy and idle deadline, built through a builder - the current `send_streaming(factory, reconnect, timeout)` (`runtime/client.rs:205-213`) is a positional list with an `Option` argument. `send_streaming` stays for the `fn` form, implemented over `open_streaming`.
- **A multipart reader** (`runtime/multipart.rs`, new): boundary read from the response `Content-Type`, per-part header block skipped, a part emitted when its body parses rather than when the next delimiter arrives, a bare close treated as a clean end, and the same 16 MiB accumulation guard the SSE parser carries (`runtime/sse.rs:81`, `:292-299`).
- **A server-side framer** returning `impl IntoResponse`, plus `OperationBuilder::multipart_json::<T>()` beside `sse_json` (`libs/toolkit/src/api/operation_builder.rs:1532`, `:1744`). The generated handler is the hand-written one with the service call generated: `let items = svc.stream(ctx, q).await?;` then framed. Note this makes the gear's stream emit a closing `--<boundary>--`, which it does not today.
- **Relax `dispatch_event`** (`runtime/sse.rs:449-468`) so a named `event:` decodes as `T`, keeping only `done` and `error` reserved. ~10 lines, and it is what makes an SSE endpoint that names its frames consumable at all - the kind is redundant with the body's own discriminant. Undecodable frames then surface as a `Serialization` error rather than vanishing, which is the trade `sse.rs:324-328` already argues for.
- **`HttpMethodBindingIr`**: add `#[serde(default)] stream_framing` rather than replacing `streaming: bool` - the struct is deliberately not `#[non_exhaustive]` because the macro emits a struct literal of it (`libs/toolkit-contract/src/ir/binding.rs:5-8`, `:39-42`).

Rough size: ~1,100 lines of production code across `rest_contract_parse.rs`, `parse.rs`, `codegen.rs`, `projection.rs`, `rest_contract.rs`, two new `runtime/` modules, `runtime/client.rs`, `runtime/sse.rs`, `ir/binding.rs`, `operation_builder.rs`; ~600 of tests (server + client codegen, a multipart parser suite, one trybuild `fail` fixture for an unknown transport ident).

## What this is not

- Not general multipart support. `multipart/form-data` requests, mixed per-part content types and `Content-Disposition` stay out of scope. This is streaming response framing only, one JSON item per part.
- Not content negotiation. Two transports at two paths are two operations, not variants of one - `Accept: text/event-stream` against `:stream` is a `406` by design (`gears/system/event-broker/scenarios/consumer/stream/1.08-guardrail-sse-from-stream-endpoint.md`). ADR-0003's 1:1 delegation rule therefore means two base methods. Accept enforcement stays with the author: the `406` body carries a domain problem with a `supported` list (`.../handlers/delivery/streaming.rs:22-60`), which is not framework behaviour.
- Not author-declared response headers. The multipart `boundary=` is a parameter on the one platform-standard header, generated by the framer, and is not a step toward `#[response_header]`.
- Not the `#[rest(status, tag, license)]` group, still unimplemented.
- Not a toolkit frame model. See above.

## ADR amendment

Required, in `docs/arch/toolkit-contract-binding/ADR/0002-cpt-cf-binding-adr-openapi-spec-limits.md`, which currently scopes out "Multipart, URL-encoded-form, and octet-stream request or response bodies" (`:60-62`) and whose Option B rejection names `#[multipart]` interacting with `#[streaming]` as a combinatorial hazard (`:175`). The ADR pre-authorises the shape of the fix at `:125` - "Extending the macro's scope later ... is a non-breaking change provided the attribute is optional and the default behavior remains the Phase 1 semantics" - and already carries the precedent for the edit: the "Amendment: repeated query parameters are now in scope" section (`:68-103`), which narrowed a wholesale exclusion to the part that genuinely cannot work. A parallel amendment should put `multipart/mixed` response streaming in scope and restate what stays out.

`0003-cpt-cf-binding-adr-projection-server-gen.md` needs edits rather than an amendment: `:302-303`, `:350-353` and `:365-366` are status statements that go stale, and the attribute-shape rule at `:305-324` needs one sentence on why the transport parameterises `#[streaming]` instead of joining the `#[rest(..)]` group - it is intrinsic to what the marker means, where the group is for orthogonal projection config. Same for `DESIGN.md:158`, `:687`, `:693`.

asmith987:
Closes #4734

## Summary by CodeRabbit

- **New Features**
  - Added typed `multipart/mixed` response streaming alongside SSE, including OpenAPI schema support.
  - Added payment feed streaming over REST and gRPC.
  - Streaming operations can now report opening errors before delivering items.

- **Improvements**
  - Stream configuration now uses framing-neutral names while accepting legacy SSE names.
  - Added reconnect, timeout, and framing support for multipart streams.
  - Improved handling of typed stream errors and tenant-scoped payment feeds.
  - Added clearer validation for streaming configuration and framing options.

- **Documentation**
  - Expanded streaming and multipart documentation, including server-registration limitations.


libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
+                // visible at the one place it decides anything: resuming a
+                // multipart stream is the caller's own reopen loop, not the
+                // transport's.
+                StreamFraming::MultipartMixed => None,
+            };
+            // URL-build / serialization failure on the request side — not
+            // eligible for reconnect (config-shape error).
+            let built = self.factory.build(snapshot.as_deref());
+            let attempted = match built {
+                Ok(builder) => attempt_open(builder, self.idle_timeout).await,
```
MikeFalcon77: **HIGH**

`StreamOpener::open` passes `self.idle_timeout` as the deadline for `attempt_open`, and `StreamRequest::new` defaults that field to `None`.

With the default, the connect, the response headers and the error-body read have no deadline, so `open_streaming` hangs forever against a peer that accepts the socket and never answers.

Add a separate `open_timeout` field to `StreamRequest`, default it from `ClientConfig::timeout`, and pass that to `attempt_open`. Leave `idle_timeout` to bound only the item loop.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -374,22 +709,103 @@ where
             // also restarts the backoff at its base delay, which is what you
             // want after a healthy period rather than resuming a 30s ceiling.
             if delivered_an_event {
-                attempt = 0;
+                opener.reset_budget();
```
MikeFalcon77, line 819: **MEDIUM**

The reconnect budget is reset after any attempt that delivered at least one item.

A peer that sends one item and then drops the connection never exhausts `max_attempts`: the client reopens at `base_delay` (500 ms) forever, re-sending the auth token on every attempt.

Reset the budget only after a connection stayed up for a minimum duration, and add an absolute cap on reopens so a one-item-then-fail peer cannot loop indefinitely.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
+                    }
+                }
+            }
+        }
+    }
+
+    /// Consume one reconnect attempt for `err`, backing off before returning
+    /// `true`. Returns `false` when the budget is spent or the error class is
+    /// not retryable, meaning the caller must surface `err`.
+    async fn consume_attempt(&mut self, err: &TransportError) -> bool {
```
MikeFalcon77, line 517: **MEDIUM**

`client.rs`, `multipart.rs` and the SSE parser emit no tracing.

A reconnect attempt, its backoff delay, budget exhaustion and an open failure are all invisible; an operator sees only the final error, and nothing at all when a reconnect succeeds.

Emit `tracing::warn!` in `consume_attempt` with attempt number, `max_attempts`, delay, framing and error, and `tracing::debug!` on a successful reopen.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
+            // Multipart is the opposite (#4734 D3/Q4): a graceful EOF is a
+            // clean end, with or without the closing `--<boundary>--`. The
+            // contract's own item type carries the completion signal here, so
+            // manufacturing a transport-level one would give two sources of
+            // truth for "the stream is finished". A consumer that must tell
+            // "fully delivered" from "the connection went away" checks for its
+            // own terminal item — see `runtime::multipart`'s module docs. An
+            // *aborted* body is a different thing and still arrives as a
+            // `Network` error item, not as an end of stream.
+            Self::Multipart(_) => None,
```
MikeFalcon77: **MEDIUM**

`end_of_stream_error` returns `None` for multipart on any EOF, so `MultipartStream::saw_close_delimiter` is never checked on the public path.

`send_streaming` and `open_streaming` return an opaque boxed stream, so a truncated multipart response looks identical to a complete one and the caller silently processes partial data.

Return a `TransportError::Framing` item when the close delimiter is missing, matching what the SSE arm already does.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
     factory: F,
+    framing: StreamFraming,
     reconnect: ReconnectConfig,
+    idle_timeout: Option<Duration>,
+}
+
+impl<F: StreamRequestFactory> StreamRequest<F> {
+    /// Start from a request factory, with SSE framing, reconnect disabled and
+    /// no idle deadline.
+    pub fn new(factory: F) -> Self {
```
MikeFalcon77, line 230: **MEDIUM**

`StreamRequest::new` has no `#[must_use]` while every builder method on the type has one.

A constructed request that is never passed to `send_streaming` compiles without a warning.

Add `#[must_use]` to `StreamRequest::new`.

libs/toolkit-contract/src/runtime/multipart.rs
```diff
@@ -0,0 +1,689 @@
...
+// `Pin<&mut MultipartStream<...>>` without pinning the type itself.
+impl<T, S: Unpin> Unpin for MultipartStream<T, S> {}
+
+impl<T, S> MultipartStream<T, S>
+where
+    T: DeserializeOwned + 'static,
+{
+    /// Run the state machine over the currently buffered bytes.
+    fn drain_buffer(&mut self) {
+        while matches!(self.step(), Step::Progress) {}
```
MikeFalcon77: **MEDIUM**

`drain_buffer` keeps stepping until the whole buffer is consumed.

One wire chunk carrying many parts runs every `serde_json::from_slice` inside a single `poll_next` and pushes all decoded items into the unbounded `pending` queue, so a large chunk blocks the executor thread and the queue has no upper bound.

Stop stepping once `pending` is non-empty, and call `drain_buffer` at the top of `poll_next` so buffered bytes are drained before the wire is polled again.

libs/toolkit-contract/src/runtime/multipart.rs
```diff
@@ -0,0 +1,689 @@
...
+}
+
+/// Parse a part's `Content-Length`, rejecting a value the accumulation guard
+/// would not let us buffer anyway. `Err` carries the framing-error message.
+fn parse_content_length(raw: &[u8]) -> Result<usize, String> {
+    let text = std::str::from_utf8(raw)
+        .map_err(|_| "part `Content-Length` is not valid UTF-8".to_owned())?;
+    let len: usize = text
+        .parse()
+        .map_err(|_| format!("part `Content-Length` is not a byte count: `{text}`"))?;
```
MikeFalcon77: **MEDIUM**

The peer-supplied `Content-Length` value is interpolated raw into the framing error message.

`step_headers` splits on CRLF only and `trim_ascii` leaves interior control bytes, so a hostile peer puts up to 16 MiB of arbitrary bytes, including bare CR/LF, into a string that is propagated and logged.

Truncate the value to a fixed short length and escape non-printable bytes before formatting it, or drop the value and report only the reason.

libs/toolkit-contract/src/runtime/multipart.rs
```diff
@@ -0,0 +1,689 @@
...
+        return None;
+    }
+    Some(rest[1..].trim_ascii())
+}
+
+/// Parse a part's `Content-Length`, rejecting a value the accumulation guard
+/// would not let us buffer anyway. `Err` carries the framing-error message.
+fn parse_content_length(raw: &[u8]) -> Result<usize, String> {
+    let text = std::str::from_utf8(raw)
+        .map_err(|_| "part `Content-Length` is not valid UTF-8".to_owned())?;
```
MikeFalcon77: **LOW**

`map_err(|_| ...)` discards the `Utf8Error` and `ParseIntError`.

The `source()` chain ends at a `DisplayError` string, so the caller cannot see what actually failed to parse.

Pass the underlying error as the boxed source of `TransportError::Framing` instead of flattening it into a `String`.

libs/toolkit/src/http/multipart.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+
+    fn with_validated_boundary(stream: S, boundary: String) -> Self {
+        let value = format!("multipart/mixed; boundary={boundary}");
+        let content_type = HeaderValue::from_str(&value)
+            // Unreachable: RFC 2046 boundary characters are all legal header
+            // value characters. Falling back to the bare media type keeps this
+            // total rather than panicking.
+            .unwrap_or_else(|_| HeaderValue::from_static("multipart/mixed"));
```
MikeFalcon77: **MEDIUM**

When `HeaderValue::from_str` fails, the fallback emits `multipart/mixed` with no `boundary` parameter.

The framer still writes `--<boundary>` delimiters, so the response is unparseable by any client, and nothing is logged to say why.

Log the failure at `error` and fall back to a generated, known-valid boundary, or make the constructor return `Result` so this header cannot be built.

libs/toolkit/src/http/multipart.rs
```diff
@@ -0,0 +1,316 @@
...
+                    reason,
+                    "multipart/mixed boundary is not RFC 2046-legal; using a generated boundary instead"
+                );
+                Self::with_validated_boundary(stream, generated_boundary())
+            }
+        }
+    }
+
+    fn with_validated_boundary(stream: S, boundary: String) -> Self {
+        let value = format!("multipart/mixed; boundary={boundary}");
```
MikeFalcon77: **MEDIUM**

The boundary is written unquoted into `Content-Type`, but `validate_boundary` accepts RFC 2046 bchars that are tspecials: space, `,`, `:`, `=`, `?`, `(`, `)`, `/`.

A boundary such as `a b` or `x,y` produces a media type that an RFC 2045 conformant parser reads differently from this crate's own reader, which splits parameters on `;` only.

Quote the boundary when it contains a character outside the HTTP token set, or restrict `validate_boundary` to token characters.

libs/toolkit/src/http/multipart.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// boundary is logged at `error` and replaced with a generated one, since
+    /// substituting a working boundary keeps the stream readable (the value
+    /// travels in the `Content-Type` header) where emitting the illegal one
+    /// would not.
+    ///
+    /// The boundary must also not occur inside any item's JSON. A generated
+    /// boundary makes that impossible in practice; a caller-chosen one makes it
+    /// the caller's responsibility.
+    #[must_use]
+    pub fn with_boundary(stream: S, boundary: impl Into<String>) -> Self {
```
MikeFalcon77: **MEDIUM**

`with_boundary` validates its argument, and on failure logs and substitutes a generated boundary.

The caller gets a stream with a boundary it did not ask for and can only detect the swap by comparing `boundary()` against its own input.

Return `Result<Self, BoundaryError>` and let the caller decide whether to fall back to `MultipartJsonStream::new`.

libs/toolkit/src/http/multipart.rs
```diff
@@ -0,0 +1,316 @@
...
+        let boundary = boundary.into();
+        match validate_boundary(&boundary) {
+            Ok(()) => Self::with_validated_boundary(stream, boundary),
+            Err(reason) => {
+                tracing::error!(
+                    boundary = %boundary,
```
MikeFalcon77: **LOW**

The rejected boundary is logged with `%boundary` (Display).

This branch runs exactly when the value failed validation, so a value with CR/LF writes raw line breaks into the log record; the crate's own test uses `"not\r\nlegal"`.

Log it with `?boundary` so control characters are escaped.

libs/toolkit/src/api/openapi_registry.rs
```diff
@@ -270,9 +270,13 @@ impl OpenApiRegistryImpl {
                 // empty `content_type`. Emit just `description` — attaching a
                 // `content` block would make code-generators expect a body.
                 if !r.content_type.is_empty() {
+                    // Streaming media types are json-like here too: their
+                    // declared schema is the *item* type, so it must render as
+                    // a `$ref` rather than as an opaque string blob.
                     let is_json_like = r.content_type == "application/json"
                         || r.content_type == problem::APPLICATION_PROBLEM_JSON
-                        || r.content_type == "text/event-stream";
+                        || r.content_type == "text/event-stream"
```
MikeFalcon77: **MEDIUM**

`is_json_like` is a hand-written list of media-type string literals.

A new `StreamFraming` variant renders its item schema as an opaque string with a custom format instead of a `$ref`, and nothing fails to compile.

Derive the streaming media types from `StreamFraming::media_type()` across all variants, or match on an enum so an unhandled framing is a compile error.

libs/toolkit-contract/src/wiring.rs
```diff
@@ -25,9 +25,14 @@ pub struct ClientTuning {
...
-    /// Override for the SSE-stream reconnect policy.
-    #[serde(default)]
-    pub sse_reconnect: Option<ReconnectSettings>,
+    /// Override for the stream reconnect policy, of any framing.
+    ///
+    /// `alias = "sse_reconnect"` keeps already-deployed config files working:
+    /// this struct has no `rename_all`, so the Rust field name *is* the JSON
+    /// key, and the field was called `sse_reconnect` before the policy covered
+    /// framings other than SSE.
+    #[serde(default, alias = "sse_reconnect")]
```
MikeFalcon77, line 40: **MEDIUM**

The `serde(alias = "sse_reconnect")` compatibility path does not define what happens when a document sets both `stream_reconnect` and `sse_reconnect`.

`#[serde(alias)]` treats the pair as a duplicate field and fails the parse, and `ClientTuning` is `#[serde(flatten)]`ed into `ClientWiring` where duplicate handling differs; a deployment mid-rename produces exactly that document.

State the both-keys outcome in the field doc comment and add a test in `tests/wiring_config.rs` that parses such a document and asserts the resulting error.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
+    /// Wire-activity counter, so the idle deadline stays *idle* rather than
+    /// merely *quiet* under either framing.
+    fn activity_handle(&self) -> StreamActivity {
+        match self {
+            Self::Sse(s) => s.activity_handle(),
+            Self::Multipart(s) => s.activity_handle(),
+        }
+    }
 
+    async fn next(&mut self) -> Option<Result<T, TransportError>> {
```
MikeFalcon77, line 631: **LOW**

`FramedItems::next` is dropped by `tokio::time::timeout` on every idle elapse, and carries no cancel-safety annotation; `send_open_request`, `open_status_failure`, `StreamOpener::open` and `consume_attempt` are also unannotated.

A future that is dropped mid-frame must document whether buffered state survives, otherwise the next maintainer cannot tell if the timeout path loses a partial frame.

Add a `// cancel-safe: <reason>` or `// NOT cancel-safe: <reason>` line to each of these async fns.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,180 +200,506 @@ where
...
+            match attempted {
+                // A framing-setup failure (a `200` whose `Content-Type` names
+                // no usable boundary) is not retried: it is a response-shape
+                // error that repeats identically, the same argument as the
+                // request-shape branch above. Returning it from here is also
+                // what makes it an `Err` from `open_streaming` rather than the
+                // stream's first item.
+                Ok(response) => return open_response(self.framing, response),
+                Err(OpenFailure::Fatal(e)) => return Err(e),
+                Err(failure) => {
```
MikeFalcon77: **LOW**

The match handles `Err(OpenFailure::Fatal(e))` explicitly and catches the rest under a wildcard `Err(failure)` binding.

A third `OpenFailure` variant would be treated as retryable without any compile error.

Replace the wildcard with `Err(OpenFailure::Retryable(err))`.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -300,6 +321,29 @@ impl ReconnectConfig {
...
+    ///
+    /// Generated clients pass this for a method whose open is fallible
+    /// (`#[streaming] async fn`). Such an open carries domain semantics the
+    /// client must not blindly repeat: any exclusion lease it acquired is owned
+    /// by the returned stream's lifetime, resume may be specified through the
+    /// contract's own cursor rather than `Last-Event-ID`, and a reconnect-time
+    /// failure would arrive as a stream *item* — past the caller's open-time
+    /// error handling, which is the whole reason the fallible shape exists.
+    #[must_use]
+    pub fn disabled() -> Self {
```
MikeFalcon77: **LOW**

`ReconnectConfig::disabled()` builds itself with `..Self::default()`.

A field added later with an enabling default is silently absorbed into a constructor documented as never reconnecting.

Name every field explicitly and make it `const fn`, matching `RetryConfig::off()`.

libs/toolkit-contract-macros/src/projection.rs
```diff
@@ -177,9 +239,14 @@ where
 }
 
 /// Render the unary or streaming return type for a generated client method.
-pub fn render_method_return_ty(ok: &Type, err: &Type, streaming: bool) -> TokenStream {
+pub fn render_method_return_ty(
```
MikeFalcon77: **MEDIUM**

`render_method_return_ty` takes `streaming: bool` next to `open: StreamOpen`.

The combination `(streaming = false, open = Awaited)` is representable and meaningless, and the same bool+enum pair is repeated in `Delegation::for_method`, `RestMethodModel` and `GrpcMethodModel`.

Replace the pair with a single enum such as `MethodShape { Unary, Stream(StreamOpen) }` and thread it through both call sites.

libs/toolkit-contract-macros/src/rest_contract_parse.rs
```diff
@@ -356,12 +371,24 @@ fn parse_method(method: &TraitItemFn) -> syn::Result<RestMethodModel> {
     )?);
     let optional = method.default.is_some();
 
+    // Mirrors the base macro's rule (`parse.rs`): `async` on a streaming
+    // method selects the fallible-open shape. Until now this parser did not
+    // inspect `asyncness` at all, so an author's `async` on a REST streaming
+    // projection was silently normalised away by `rewrite_streaming_signature`.
+    let open = if streaming && method.sig.asyncness.is_some() {
```
MikeFalcon77: **MEDIUM**

The REST projection parser now reads `method.sig.asyncness` to pick the open shape.

An existing `#[streaming] async fn` that was normalised to `fn -> Stream` now generates `async fn -> Result<Stream, E>` instead, changing the trait signature with no diagnostic.

Document the behaviour change in the macro module docs and in DESIGN.md, and call it out in the PR description so existing projections with a stray `async` get audited.

examples/toolkit/api-contracts/api-contracts/src/api/grpc.rs
```diff
@@ -199,3 +245,204 @@ fn canonical_to_status(err: CanonicalError) -> Status {
...
+/// Hand-written because the derive does not expose the category it already
+/// knows at expansion time. That is the one piece of this file that wants to
+/// move into `#[derive(ContractError)]`: the mapping is a restatement of the
+/// attributes, so it is the only part that can drift from them.
+///
+/// The wildcard arm is required — `PaymentError` is `#[non_exhaustive]` and
+/// lives in another crate — and it **logs**. A silently mis-categorised variant
+/// is the exact failure mode #4734 is about, and a new variant defaulting to
+/// `Internal` would turn a client fault into a server fault with no signal.
+fn payment_error_category(err: &PaymentError) -> ProblemCategory {
```
MikeFalcon77: **MEDIUM**

`payment_error_category` restates the `#[canonical(..)]` category of every `PaymentError` variant inside the gear crate.

`PaymentError` is `#[non_exhaustive]` and owned by the SDK, so a variant added there falls into the wildcard arm and goes on the wire as `Code::Internal`.

Generate a `category()` accessor from `#[derive(ContractError)]` in the SDK and call it here instead of the local match.

examples/toolkit/api-contracts/api-contracts/src/api/rest/routes.rs
```diff
@@ -102,5 +104,36 @@ pub fn register_manual_routes(
...
+    // over `multipart/mixed`, with a fallible open (#4734).
+    //
+    // `multipart_json` is `sse_json`'s counterpart: it declares a `200` whose
+    // content type is `multipart/mixed` and whose declared schema is the *item*
+    // type, since the framing is what turns one item into one part. The runtime
+    // `boundary=` parameter is generated per response by
+    // `MultipartJsonStream` and is deliberately not in the spec — the OpenAPI
+    // media-type key is the bare `multipart/mixed`.
+    let router = OperationBuilder::get("/api-contracts/v1/payments:feed")
+        .operation_id("api_contracts.payment_feed")
```
MikeFalcon77: **LOW**

The operation_id is `api_contracts.payment_feed` while the contract method is `stream_payments`.

The sibling manual SSE route names its method (`api_contracts.list_payments`), so this one breaks the mapping from spec id to trait method.

Rename the operation_id to `api_contracts.stream_payments`.

examples/toolkit/api-contracts/api-contracts/src/domain/service.rs
```diff
@@ -166,30 +166,100 @@ impl PaymentDomainService {
...
+    ///   position cannot be opened, and the caller needs to know *which* ones
+    ///   (`unseeded`) to seed them.
+    /// - [`PaymentError::InvalidFilter`] when `filter.currency` is not a
+    ///   three-letter ISO 4217 code — the pre-existing "unacceptable filter"
+    ///   open failure, now typed. Its category is `InvalidArgument`, the other
+    ///   one whose 400 degrades to `Internal` / 500 through `CanonicalError`,
+    ///   so the two open failures cover both of the badly-degrading categories.
+    pub fn open_feed(
+        self: &Arc<Self>,
+        _ctx: &SecurityContext,
```
MikeFalcon77: **LOW**

`open_feed` takes the `SecurityContext` as `_ctx` and never reads it.

`snapshot` filters only on status and currency, so every caller sees every record; `charge_v2` in the same file keys on `ctx.subject_tenant_id()`, so this example teaches the weaker pattern.

Filter `snapshot` by the caller's tenant from `ctx`, and store the owning tenant on `PaymentRecord`.

examples/toolkit/api-contracts/api-contracts/src/domain/service.rs
```diff
@@ -166,30 +166,100 @@ impl PaymentDomainService {
...
+        filter: &ListPaymentsFilter,
+    ) -> Result<
+        api_contracts_sdk::contract::PaymentStream<PaymentSummary, PaymentError>,
+        PaymentError,
```
MikeFalcon77: **LOW**

`open_feed` returns the SDK wire error `PaymentError` straight out of the domain layer.

The documented chain is domain error -> SDK error -> Problem; returning the SDK type here makes the domain depend on the transport envelope.

Define a local domain error enum for the open failures and convert it to `PaymentError` at the SDK boundary.

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -867,116 +1042,20 @@ fn generate_streaming_method_body(
...
-            let __timeout = ::std::option::Option::Some(self.config.sse_idle_timeout);
+            #prelude
+
+            // A stream uses the per-item idle deadline (NOT the unary
+            // `timeout`), so a healthy slow stream is not killed between items.
+            //
+            // Per-client reconnect policy (D6: only the immediate form gets
+            // one — a fallible open carries domain semantics the client must
+            // not blindly repeat).
+            let __request = #support::runtime::client::StreamRequest::new(__factory)
```
MikeFalcon77, line 1059: **MEDIUM**

No test drives a `#[streaming(multipart_mixed)] fn` (immediate-open) method against a server.

Every multipart runtime test uses the awaited `async fn` shape, so the lazy multipart open, the `Accept: multipart/mixed` header on that path, a boundary failure surfacing as the stream's first item, and multipart reopen under `stream_reconnect` never execute.

Add a `#[streaming(multipart_mixed)] fn` method to the `multipart_roundtrip.rs` contract and cover three cases: successful items, a mislabelled `200` arriving as the first item, and a reopen with `stream_reconnect` enabled.

libs/toolkit-contract-macros/src/grpc_contract.rs
```diff
@@ -988,14 +1035,59 @@ mod tests {
...
+        );
+        assert!(!platform.contains("attach_bearer"), "got:\n{platform}");
+        // The attach must precede the call that opens the stream.
+        let attach = platform
+            .find("attach_internal_token")
+            .expect("attach is emitted");
+        let open = platform
+            .find("watch_thing (__request)")
+            .or_else(|| platform.find("watch_thing(__request)"));
+        if let Some(open) = open {
```
MikeFalcon77: **MEDIUM**

The ordering assertion is wrapped in `if let Some(open) = open`.

If neither `watch_thing (__request)` nor `watch_thing(__request)` is found in the expansion, the test skips its only assertion and still passes, so a change in the emitted formatting disables the check instead of failing it.

Replace the `if let` with `.expect("the open call is emitted")` on the `find`.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -326,6 +370,22 @@ mod tests {
         assert_eq!(r.max_attempts, 1);
     }
 
+    /// `disabled()` is what generated code passes for a fallible open (D6),
+    /// so the guarantee it has to carry is "never a second attempt" — pinned
+    /// here rather than left to `Default`, which is where the value happens to
+    /// come from but not where the requirement lives.
+    #[test]
+    fn disabled_reconnect_makes_no_attempts_and_matches_the_default() {
```
MikeFalcon77: **MEDIUM**

`disabled_reconnect_makes_no_attempts_and_matches_the_default` reads back the fields of `ReconnectConfig::disabled()`, which is `Self { max_attempts: 0, ..Self::default() }`.

The test restates the constructor and cannot fail unless struct-update syntax breaks; the guarantee in its own doc comment, that there is never a second attempt, is not asserted.

Drop this test. `reconnect_is_derived_from_the_open_shape_not_from_client_config` already pins the behaviour by counting server connections.

examples/toolkit/api-contracts/api-contracts/src/api/grpc.rs
```diff
@@ -199,3 +245,204 @@ fn canonical_to_status(err: CanonicalError) -> Status {
...
+    ///
+    /// The regression guard for a real defect: this mapping was first written
+    /// as a lookup on the problem-type URI that named only seven categories and
+    /// sent the rest to `Code::Internal`. A `#[canonical(Unauthenticated)]` or
+    /// `#[canonical(DeadlineExceeded)]` contract error would therefore have
+    /// reached a non-toolkit peer as a server fault, inverting its retry
+    /// classification. Inert at the time only because `PaymentError` happened to
+    /// use four categories the table covered.
+    #[test]
+    fn every_canonical_category_maps_to_its_own_grpc_code() {
```
MikeFalcon77: **LOW**

`every_canonical_category_maps_to_its_own_grpc_code` restates the sixteen match arms of `code_for` as a literal table.

It re-implements the code under test, and nothing forces exhaustiveness, so a new `ProblemCategory` falls to the wildcard arm and the test still passes: exactly the defect it claims to guard.

Drive the loop from an exhaustive `match` on `ProblemCategory` inside the test, or from a `ProblemCategory::ALL` constant, so a new category breaks the build.

libs/toolkit-contract/tests/multipart_roundtrip.rs
```diff
@@ -0,0 +1,830 @@
...
+    // frame N the reader must have *nothing* — the server has not written
+    // frame N+1 yet. A framing that waited for the next delimiter would
+    // deliver frames in pairs and this would find frame 1 already waiting.
+    let base = start_server().await;
+    let mut stream = Box::pin(open(&format!("{base}/slow?count=3")).await);
+
+    let first = stream.next().await;
+    assert_eq!(first.map(Result::unwrap), Some(Frame { seq: 0 }));
+
+    let too_soon = tokio::time::timeout(Duration::from_millis(50), stream.next()).await;
```
MikeFalcon77: **LOW**

`a_part_arrives_before_the_next_parts_bytes_exist` asserts a negative against wall-clock time: the server sleeps 300 ms per frame and the test asserts nothing arrives inside a 50 ms window.

A 250 ms scheduling stall on a loaded runner fails a correct implementation, and the test spends about 900 ms sleeping.

Drive the server from a `tokio::sync::mpsc` channel: send frame 0, assert the client yields it, assert `now_or_never()` is `None` on the next poll, then send frame 1 and assert it arrives.

libs/toolkit/src/api/operation_builder.rs
```diff
@@ -1760,6 +1795,28 @@ where
         self
     }
 
+    /// Additional `multipart/mixed` response (if the operation already has
+    /// one). See [`OperationBuilder::multipart_json`] on the first-response
+    /// builder for what `T` and the media-type key mean.
+    pub fn multipart_json<T>(
```
MikeFalcon77, line 1801: **LOW**

The `multipart_json` additional-response method on `WithResponse` has no caller and no test.

The only multipart OpenAPI test builds a `ResponseSpec` by hand and mutates `content_type`, and the example gear uses the first-response overload, so the `upsert_response` path never runs.

Add a test that chains `.json::<T>()` then `.multipart_json::<Item>()` and asserts the `200` carries both media-type keys with the item `$ref` under `multipart/mixed`.

libs/toolkit-contract/src/runtime/multipart.rs
```diff
@@ -0,0 +1,689 @@
...
+
+        let mut scan = self.scan_from;
+        let hit = find_from(&self.buf, CRLF, &mut scan);
+        self.scan_from = scan;
+        match hit {
+            Some(at) => {
+                // RFC 2046 permits transport padding (linear whitespace)
+                // between the boundary and its CRLF; anything else means the
+                // boundary token was a prefix of some longer token.
+                if self.buf[..at].iter().any(|b| !matches!(b, b' ' | b'\t')) {
```
MikeFalcon77, line 471: **LOW**

`step_boundary_tail` accepts RFC 2046 transport padding (spaces and tabs) between the boundary token and its CRLF, but only the reject path is tested.

A change that turned padding into a framing error would pass CI, since no test feeds a padded delimiter.

Add a test in `multipart_tests.rs` that feeds `--<boundary>  \t\r\n` before a part and asserts the part is still read.

libs/toolkit-contract/src/runtime/client.rs
asmith987: Fixed. I added a dedicated `open_timeout` field to satisfy this issue

libs/toolkit-contract/src/runtime/client.rs
asmith987: Done. Added logic to reset the budget after a connection has stayed up fo ra minimum duration and added an absolute cap.

libs/toolkit-contract/src/runtime/client.rs
asmith987: Added tracing::warn and debug

libs/toolkit-contract/src/runtime/client.rs
asmith987: Done

libs/toolkit-contract/src/runtime/client.rs
asmith987: Added `#[must_use]` to `StgreamRequest::new`

libs/toolkit-contract/src/runtime/multipart.rs
asmith987: Done. Drain at the top of `poll_next` and stops stepping once `pending` is non-empty.

libs/toolkit-contract/src/runtime/multipart.rs
asmith987: Done.

libs/toolkit-contract/src/runtime/multipart.rs
asmith987: Done.

libs/toolkit/src/http/multipart.rs
asmith987: This branch is currently unreachable, but hardened it by logging and falling back to a generated valid boundaary rather than emitting an unparseable header.

libs/toolkit/src/http/multipart.rs
asmith987: Done

libs/toolkit/src/http/multipart.rs
asmith987: Fixed.

libs/toolkit/src/api/openapi_registry.rs
asmith987: Added `StreamFraming::media_type()`

libs/toolkit-contract/src/wiring.rs
asmith987: Stated the outcome and added a test.

libs/toolkit-contract/src/runtime/client.rs
asmith987: Added

libs/toolkit-contract/src/runtime/client.rs
asmith987: Fixed. Replaced wild card with `Err(OpenFailure::Retryable(err))`

libs/toolkit-contract/src/runtime/config.rs
asmith987: Named every field explicitly

libs/toolkit-contract-macros/src/projection.rs
asmith987: Replaced the par with `MethodShape` enum.

libs/toolkit-contract-macros/src/rest_contract_parse.rs
asmith987: Documented the behavior in DESIGN.md

examples/toolkit/api-contracts/api-contracts/src/api/grpc.rs
asmith987: Added this.

examples/toolkit/api-contracts/api-contracts/src/api/rest/routes.rs
asmith987: Renamed the operation id.

examples/toolkit/api-contracts/api-contracts/src/domain/service.rs
asmith987: Updated all of the examples.

libs/toolkit-contract-macros/src/rest_contract.rs
asmith987: Added the tests

libs/toolkit-contract/src/runtime/config.rs
asmith987: Removed the test

examples/toolkit/api-contracts/api-contracts/src/api/grpc.rs
asmith987: Added the constant

libs/toolkit/src/api/operation_builder.rs
asmith987: Added the test

libs/toolkit-contract/tests/multipart_roundtrip.rs
asmith987: Fixed.

libs/toolkit-contract/src/runtime/multipart.rs
asmith987: Fixed

libs/toolkit-contract-macros/src/grpc_contract.rs
asmith987: Fixed

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -191,213 +200,719 @@ where
...
+                StreamFraming::ServerSentEvents => self.last_id.current(),
+                // `Last-Event-ID` is an SSE mechanism and `multipart/mixed`
+                // has no resume token of its own, so the factory is handed
+                // `None` on every attempt. Stated as a branch rather than left
+                // to the cell simply never being advanced, so the reason is
+                // visible at the one place it decides anything: resuming a
+                // multipart stream is the caller's own reopen loop, not the
+                // transport's.
+                StreamFraming::MultipartMixed => None,
+            };
```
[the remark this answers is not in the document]
asmith987: Fair point — I've documented this rather than restricting it.

The behavior is intentional: reconnect is opt-in (`ReconnectConfig::disabled()` by default), and for a framing with no resume token a reopen is a blind re-issue, so at-least-once is the semantics by construction. Refusing reconnect under `MultipartMixed` outright would drop a useful recovery path for idempotent consumers.

But you're right that the docs stated only "no resume token" and not the duplicate-delivery consequence. Fixed: `ClientConfig::stream_reconnect` now spells out that a multipart reopen is a blind restart (not a resume) — the server replays from the first part and redelivers already-yielded items (at-least-once, with no marker for the restart) — and that an exactly-once consumer should leave reconnect disabled and run its own reopen loop with application-level dedup. `send_streaming`'s doc cross-references it.

libs/toolkit-contract/src/runtime/multipart.rs
```diff
@@ -0,0 +1,784 @@
...
+                    // A graceful end of the byte stream is a clean end of this
+                    // stream, with or without the closing `--<boundary>--`.
+                    // Whatever partial construct is buffered is discarded
+                    // rather than surfaced as a truncation error — see the
+                    // module docs on why completion detection is the
+                    // consumer's, and why an *aborted* body (the `Err` arm
+                    // below) is loud instead.
+                    this.done = true;
+                    this.state = State::Closed;
+                    this.buf.clear();
```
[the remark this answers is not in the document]
asmith987: This is already handled — one layer up, by design.

`MultipartStream::poll_next` intentionally yields a clean `None` on graceful EOF and exposes the fact via `saw_close_delimiter()`. The driver is what turns truncation into an error: `FramedItems::end_of_stream_error` in `runtime/client.rs` returns `TransportError::Framing("multipart body ended without a closing --<boundary>-- delimiter")` whenever `!saw_close_delimiter()`, and `drive` yields that before ending the stream. So on the public path (`send_streaming` / `open_streaming`) a proxy half-close is *not* indistinguishable from a complete stream — it surfaces as a framing error, symmetric with the SSE `done`-event check.

Applying the suggested diff would duplicate that detection inside the parser and change the documented contract of the concrete stream (the `saw_close_delimiter()` accessor exists precisely so the layer that owns reconnect decides what a graceful end means). The deferred-serialization-error precedence you mention is already preserved in this branch, and because the truncated stream still errors via the driver, the discarded unsized part is not a silent clean end. Leaving as-is.

libs/toolkit/src/http/multipart.rs
```diff
@@ -0,0 +1,433 @@
...
+            // Clean end of the source stream: write the close delimiter, then
+            // let the body end gracefully on the next poll.
+            state.finished = true;
+            let close = format!("--{}--\r\n", state.boundary);
+            return Some((Ok(Bytes::from(close)), state));
+        };
+        match serde_json::to_vec(&item) {
+            Ok(json) => {
+                let bytes = encode_part(&state.boundary, &json);
+                Some((Ok(bytes), state))
```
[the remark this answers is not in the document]
asmith987: Good catch — fixed.

You're right that a server item above the reader's ceiling yields a stream the toolkit client cannot consume, and since `toolkit` depends on `toolkit-contract` the limit is reachable. `frame()` now checks `json.len()` against `MAX_PART_BYTES` — bound to the reader's `toolkit_contract::runtime::multipart::MAX_ACCUMULATED_BYTES` as the single source of truth, so there's no duplicate value — before `encode_part`, and aborts the body with a clear size error (logged at `error`) exactly the way the serialize-failure path does. Added a test asserting an oversized item aborts the body rather than emitting an unreadable part.

examples/toolkit/api-contracts/api-contracts/src/api/rest/handlers.rs
```diff
@@ -73,3 +74,92 @@ pub async fn list_payments_handler(
...
+    Delivered(PaymentSummary),
+    /// The service failed mid-stream. Serializing this is an error on purpose.
+    Failed,
+}
+
+impl serde::Serialize for FeedItem {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        match self {
+            FeedItem::Delivered(summary) => summary.serialize(serializer),
+            FeedItem::Failed => Err(serde::ser::Error::custom(
```
MikeFalcon77: **MEDIUM**

A mid-stream failure is signalled by making `Serialize` return an error, so the framer aborts the body instead of ending it.

This makes an application enum's `Serialize` impl load-bearing for transport framing, and nothing in the type says so. Anyone who logs a `FeedItem`, builds one in a test, or hands it to a different serializer gets a failure with no obvious cause. This file is the example other gears copy, so the pattern spreads.

Aborting a response body is a real transport operation, so give it a name instead of deriving it from a broken `Serialize`. Let `MultipartJsonStream` accept `Stream<Item = Result<T, Abort>>` and abort the body on `Err`. Then `FeedItem` goes away and the service's own `Result` passes straight through.

libs/toolkit-contract-macros/src/rest_contract_parse.rs
```diff
@@ -356,12 +377,28 @@ fn parse_method(method: &TraitItemFn) -> syn::Result<RestMethodModel> {
     )?);
     let optional = method.default.is_some();
 
+    // Mirrors the base macro's rule (`parse.rs`): `async` on a streaming
+    // method selects the fallible-open shape. Until now this parser did not
+    // inspect `asyncness` at all, so an author's `async` on a REST streaming
+    // projection was silently normalised away by `rewrite_streaming_signature`.
+    let shape = if streaming {
+        let open = if method.sig.asyncness.is_some() {
```
MikeFalcon77: **MEDIUM**

`async` on the method is what selects the open shape: `fn` generates `-> Stream`, `async fn` generates `-> Result<Stream, E>`.

Nothing at the declaration site tells the author that the keyword carries this meaning. Someone who deletes an `async` that looks unused changes the generated client API for every consumer, and this parser still compiles it without a word. The comment four lines above asks people to audit existing projections for exactly this case, which is the running cost of the rule, not a one-time migration.

Put the shape in the attribute instead: `#[streaming(multipart_mixed, open = fallible)]`. The selector becomes greppable, and an `async` that disagrees with it can be rejected with `compile_error!` rather than silently changing the output.

examples/toolkit/api-contracts/api-contracts-sdk/src/contract.rs
```diff
@@ -15,7 +15,11 @@ use toolkit_security::SecurityContext;
 use crate::models::{ChargeRequest, ChargeResponse, Invoice, ListPaymentsFilter, PaymentSummary};
 
 /// Boxed stream type returned by streaming `PaymentApi` methods.
-pub type PaymentStream<T> = Pin<Box<dyn Stream<Item = Result<T, CanonicalError>> + Send + 'static>>;
+///
+/// The error parameter defaults to `CanonicalError`, which every streaming
+/// method declares.
+pub type PaymentStream<T, E = CanonicalError> =
```
MikeFalcon77, line 31: **MEDIUM**

`E` promises that stream items can carry a `CanonicalError`. Over `multipart/mixed` they never do.

`MultipartStream::Item` is `Result<T, TransportError>`, and the decode path only runs `serde_json::from_slice::<T>`. There is no error part on the wire. Once the open has succeeded, a domain failure can only reach the client as a synthetic framing or truncation error. SSE could carry an error frame and gRPC has a trailing status, so "one base method, two framings" holds for the open and stops holding after it. A caller reading this type writes a match arm for `Err(CanonicalError)` that never runs on the multipart path.

Either add a typed error part to the multipart framing, or state in this doc comment that under `multipart_mixed` every post-open failure arrives as a transport error.

examples/toolkit/api-contracts/api-contracts/src/api/rest/handlers.rs
asmith987: Done. `MultipartJsonStream` takes `Stream<Item = Result<T, E>>` now. I removed `FeedItem`, so `Result<Paymentsummary, CanonicalError>` flows straight throw. This also gets rid of `Serialize`.

libs/toolkit-contract-macros/src/rest_contract_parse.rs
asmith987: Done. Update to the suggestion of using `#[streaming(multipart_mixed, open = fallible)]` and rejects via `compile_error!` whenever async disaagrees.

examples/toolkit/api-contracts/api-contracts-sdk/src/contract.rs
asmith987: Added a typed `application/problem+json` error part, so a post-open failure is represented as `Err(CanonicalError)` with the category preserved.

PR #4737, opened 2026-09-08
docs(credstore): credential record / secret value split, PUT+PATCH, type-only scope, immutable value versions (ADR-0004/0005/0006)

ffedoroff:
## The plan

This PR is the design half of a pair. It records **what CredStore will become and why** — the ADRs, the PRD (requirement level) and DESIGN (the contract) that follow them. The other half, **#4741, is the implementation**; it is stacked on this branch and carries the technical detail (code layout, migration, protocol steps, tests, CI). Read this one for the decisions, that one for how they were built. Nothing here is repeated there.

**Documentation only.** No code. Because `main` still runs the previous design, DESIGN and PRD here keep two layers — the shipped behaviour marked *superseded*, and the decided one — so a reviewer can see exactly what changes; #4741 collapses them to the current state once the code lands.

## Why

CredStore had one read surface returning value and metadata together under a single `read` action, and no collection read. Three needs did not fit: an **integration administrator** who configures and rotates credentials without ever reading them; a **catalogue** that shows what a tenant has and whether it is its own or inherited; an **application** that reads its credentials one by one or as a set.

## Decisions

Ten ADRs, all `accepted`; ADR-0002 is `superseded`. Each is one decision, cut to the template's sections (≤ 140 lines); the endpoint tables, request/response examples, roles, reason codes and DDL live in DESIGN only, and the ADRs link there.

| ADR | Decides |
|---|---|
| 0004 — Credential: metadata with a selectable secret | One entity `credential` (`gts.cf.core.credstore.credential.v1~`); the default representation omits the secret; `secret` is a `$select` field on the point read and the collection alike; read actions follow the projection (`read`/`list` for record fields, `read_secret` for the secret). No separate secret address. |
| 0005 — Upward-rooted collection read | The listing walks the caller's ancestor chain only; the tenant predicate gates, `reference`/`type` clamp in SQL, reduction in memory; cursor pagination on reference boundaries; **secret mode** (`$select=…,secret`) is the bounded bulk read. |
| 0006 — Immutable value versions | Every secret write is a new immutable backend entry under a fresh id; the row switches its pointer in one transaction; the reaper is withdrawn for a periodic job (the pattern of Vault KV v2 / cloud secret managers, shadow paging, Git objects + GC). Supersedes 0002, amends 0001/0003. |
[...]

PR #4741, opened 2026-09-09
feat(credstore): credential record / secret value split, PUT+PATCH, six PDP actions, upward-rooted listing, immutable value versions

ffedoroff:
## The implementation

This PR is the code half of a pair. **#4737 is the plan** — the ADRs, the PRD and the reasoning; this PR implements it and carries the technical detail only. It is **stacked on #4737** (a fork-based PR cannot name a fork branch as its base), so those documentation commits appear here too; merge #4737 first, then this PR's own diff is the implementation plus the docs brought to the current state. For *what* and *why*, read #4737 (ten ADRs, one decision each); nothing from it is repeated here.

### REST (`gears/credstore/credstore/src/api/rest/`)

| Method and path | PDP action | Behaviour |
|---|---|---|
| `GET /credstore/v1/credentials` | `list` per distinct type | SQL-first query in three steps: the distinct types visible to the caller across the ancestor chain, one PDP evaluation per type → the permitted set (empty → empty page, no further query); a keyset page of `DISTINCT reference` clamped by `secret_type_uuid IN (permitted ∩ $filter)` and `reference` (eq/in); then the page's rows whole, unclamped by type, reduced in memory (shared `reduce_reference`; a winner outside the permitted set is an override-type-consistency violation, dropped and counted), `sharing`/`fallback`/`expires_at` filtered after reduction, `$orderby` on `reference` only, `limit` (default 50, cap `list.max_limit` 200) + `cursor` (toolkit `CursorV1`, validated against filter and order), `$select` projection. |
| `GET /credentials?$select=…,secret` | `read_secret` per distinct type (+ `list` when a record field is selected too) | Secret mode: `reference`/`type` selector only, no `limit`/`cursor`/`$orderby`, cap+1 fetch of permitted references with `TOO_MANY_MATCHES` above `list.secret_mode_cap` (25), winners' values read concurrently (at most 8 in flight, order preserved, one backend error fails the request) through the fence-checked read path, refused items omitted, `next_cursor` absent. |
| `GET /credentials/{ref}` | actions follow `$select`: `read` for record fields (or no `$select`), `read_secret` when `secret` is named, both when both | One item of the collection's shape; `secret` appears only when selected; a value-only projection over a value-less winner is the canonical 404; strong `ETag "<id>.<version>"` or weak `W/"…"` whatever the projection; retry-once when the pointer moved under the read. `GET …/{ref}/secret` no longer exists — `$select=reference,type,expires_at,secret` is its equivalent (the SDK's `get_secret`). |
| `PUT /credentials/{ref}` | `write`; `write_secret` when `secret` is a string or a `null` removes an existing secret | Exactly one of `If-None-Match: *` (201, `Location`, `ETag`) or `If-Match` (204); `secret` tri-state — absent 400 `SECRET_REQUIRED`, string written, explicit `null` creates a `declared` row (no backend call, `write` only) or removes the secret on an `active` row in one transaction; type required on create, immutable after; create over an inherited record of another type → 409 `TYPE_MISMATCH_WITH_INHERITED`. |
| `PATCH /credentials/{ref}` | `write` for metadata keys, `write_secret` for `secret` | `application/merge-patch+json` (else 415), `If-Match` mandatory, `If-None-Match` → 400; tri-state fields (absent / null / string); `secret: null` → one transaction to `declared` + gc row; metadata-only no-op keeps the version; `{}` → 400 `EMPTY_PATCH`. |
| `DELETE /credentials/{ref}` | `delete` | `If-Match` mandatory; one transaction (row + gc row); the reference is free at once. |

Authorization runs before any response that would reveal a row detail; on reads the action set follows the projection (`Service::get_item`), on writes the body. Permissions are six `gts_instance!` declarations in `src/gts/permissions.rs`, registered through the toolkit-gts inventory. Reason codes and config keys are listed in DESIGN §10.

### Write protocol (`domain/secret/service.rs`, `infra/storage/repo_impl/writes.rs`)

1. validate, authorize (both actions where the body requires), resolve the type, read the row for the precondition;
2. `INSERT credstore_value_gc(new_id, pending)` — the intent;
3. `plugin.put(tenant_id, new_id, value)`;
4. one transaction: row `INSERT … active` or `UPDATE … SET value_id, version+1, value_fp, fp_key_id, <metadata> WHERE id AND version` under `FOR UPDATE`, `DELETE gc(new_id)` — which must affect exactly one row, otherwise the transaction rolls back as retryable (the intent was reclaimed) — and `INSERT gc(old_id, superseded)`;
5. best-effort `plugin.delete(old_id)` + `DELETE gc(old_id)`.

A lost CAS marks the uploaded version `aborted` and deletes it; a failed `plugin.put` leaves the intent marked `aborted` (never deleted, so an ack lost after a persisted write cannot leave an untracked orphan); a second `NotFound` on a live pointer is a 503, not a 404. `Service::run_gc` (expiry pass, gc drain, pending reclaim past `gc.pending_max_age_secs` with claim-before-delete) is exposed as `CredStoreMaintenanceV1` in `ClientHub` — no REST, no timer; the host schedules it.

### Storage (`infra/storage/`)

- `m0002_value_versions`: `value_id UUID NULL` + partial unique index, `fallback SMALLINT CHECK (1,2)`, `idx_credstore_type (tenant_id, secret_type_uuid)`, `CHECK (status IN (2, 4))`, `ck_credstore_fp_with_value` (`value_id IS NULL ⇔ value_fp IS NULL ⇔ fp_key_id IS NULL`), `credstore_value_gc (value_id PK, tenant_id, reason CHECK (1..4), enqueued_at)`. PostgreSQL: additive, the anonymous m0001 CHECKs located via `pg_constraint`; SQLite: table rebuild. No `idx_credstore_pending`.
- Resolution predicate `status = 2 AND not expired OR (status = 4 AND fallback = 2)`; ancestors contribute `shared` rows only; every multi-row write is one `DBProvider::transaction` with the row locked and re-read inside it.

### SDK and consumers

- `credstore-sdk`: `CredStoreClientV1 { get, get_secret, put, patch, list, delete }`, models `Credential`, `Secret`, `CredentialWrite`, `CredentialPatch`/`PatchField`, `PutPrecondition`, `CredentialListItem`, `ValueId`, `FENCE_KEY_VALUE_ID`; `CredStoreMaintenanceV1 { run_gc } → GcReport`; plugin SPI `get/put/delete(ctx, tenant_id, value_id)` — `put` on an existing id MUST be `Conflict`; GTS types renamed to `…credential.v1~…`.
- `static-credstore-plugin`: keyed `(tenant, value_id)`, immutable entries, YAML value seeding removed.
- `keycloak-idp-plugin`, `oagw`: adapted to the new client trait. Static-plugin `secrets:` blocks removed from every config; the oagw and mini-chat e2e fixtures provision through the API.

### Configuration

`gc.pending_max_age_secs` (3600), `gc.batch_size` (256), `list.max_limit` (200), `list.secret_mode_cap` (25). The `reaper.*` block is gone and is rejected by `deny_unknown_fields`.

## Verification

- `cargo clippy --all-targets --all-features -D warnings` and `cargo test`: `cf-gears-credstore` 307, `cf-gears-credstore-sdk` 36, `cf-gears-static-credstore-plugin` 16, `cf-gears-keycloak-idp-plugin` 357, `cf-gears-oagw` 658 — all green.
- Service tests cover the T1→T2→T3 walkthrough (override, rotate, suppress, delete → inherit), body-derived actions, torn write / lost CAS / racing readers / `run_gc` idempotence; repository tests run the migration and the CHECKs on SQLite.
- `make e2e-local SUITE=credstore`: 4/4 (`testing/e2e/suites/credstore/test_integration_seams.py` — route smoke, lifecycle incl. `$select=secret` reads, tenant-isolation seam incl. one-request suppression with `"secret": null`, RFC 9457 shape) against a live server.
- `cfs validate`: 0 errors. `docs/api/api.json` regenerated from the example server, `gears/credstore/docs/api/openapi.yaml` re-extracted; QUICKSTART and the READMEs describe the current API; DESIGN and PRD describe the current system only; the API field is `secret` on read and write, storage keeps `value_id`/`value_fp`.

## CI

`API Contract Checks` fails on purpose: the four `/credstore/v1/secrets*` routes (and, since the review amendments, `/credentials/{ref}/secret`) are removed without a deprecation period (the plan in #4737 chose no compatibility layer — greenfield). It needs the `breaking-api-acknowledged` label from a maintainer.

## Not in this PR

- A scheduler that calls `CredStoreMaintenanceV1::run_gc` (a `gc` subcommand of the application binary under a CronJob, or a scheduler gear).
- The PostgreSQL arm of `m0002` is reviewed, not executed: every test here runs on SQLite.


gears/credstore/docs/migration/value-migration.md
```diff
@@ -0,0 +1,291 @@
+Created:  2026-09-15 by Constructor Tech
+Updated:  2026-09-15 by Constructor Tech
+
+# Moving existing values to the ADR-0006 address
```
ffedoroff, line 4: @diffora please review migration

Issue #4052, opened 2026-06-09
resource-group: external (non-RG-prefix) membership types are validated but cannot be registered/resolved (#1929 necessary but not sufficient)


diffora:
## Summary

[PR #1929](https://github.com/cyberfabric/cyberware-rust/pull/1929) relaxed **validation** of `allowed_membership_types` entries (`validate_membership_type_code` no longer requires the `gts.cf.core.rg.type.v1~` prefix), so the *syntax* of an external membership type such as `gts.cf.core.am.user.v1~` is now accepted.

However, an external membership type still **cannot be registered or resolved** end-to-end. cyberfabric/cyberware-rust#1929 is necessary but **not sufficient**: the registration and resolution paths still require every membership type to be a row in `gts_type`, and the only way to create such a row (`create_type`) still rejects non-RG-prefixed codes. This blocks the AM-side follow-up [#1930](https://github.com/cyberfabric/cyberware-rust/issues/1930).

## Background

The intent (review thread https://github.com/cyberfabric/cyberware-rust/pull/1906#discussion_r3257560655) is:

> RG module should allow using any kind of resource types (e.g. `gts.cf.core.am.user.v1~`, `gts.cf.core.am.tenant.v1~`, `gts.cf.core.oagw.upstead.v1~`, etc.) in the membership records.

cyberfabric/cyberware-rust#1929 delivered the **validation** half of that for `allowed_membership_types`. The **registration + resolution** half is missing.

## The gap

Three points in `resource-group` still hard-require the membership type to be a concrete `gts_type` row, and that row can only carry an RG-prefixed code:

1. **`create_type` / `update_type` validate the type's own `code` with `validate_type_code`** (`domain/type_service.rs`), which enforces the `gts.cf.core.rg.type.v1~` prefix. So an external type (e.g. `gts.cf.core.am.user.v1~`) **cannot be registered** as a `gts_type` row in the first place — `create_type` rejects it. `validate_membership_type_code` is applied only to the *entries* of `allowed_membership_types`, never to the type's own code.

2. **`resolve_ids(allowed_membership_types)`** (`infra/storage/type_repo.rs`) does `SELECT id FROM gts_type WHERE schema_id IN (codes)` and fails with `Referenced types not found` if any membership code has no `gts_type` row. So registering a container type with `allowed_membership_types: ["gts.cf.core.am.user.v1~"]` fails unless that bare type was already registered — which (1) forbids.

[...]

PR #4744, opened 2026-09-09
fix(authz): type native group predicates

### LINKED ISSUE #4644: authz: make native group predicates type-safe and PostgreSQL-compatible

fdlockgraf:
## Description
Fix native `InGroup` and `InGroupSubtree` scope compilation so resource-group membership is both type-safe and PostgreSQL-compatible.

- carry an explicit RG member-handle GTS type from `ResourceType` into compiled scope filters
- suppress group capabilities for resource descriptors without that mapping and reject native predicates that were not negotiated
- require UUID hierarchy keys, `id` as the group-scoped property, and `owner_tenant_id` in the same AND constraint
- enforce the same tenant/type invariants for direct SecureORM scope construction
- deny empty or unsupported group selectors instead of widening them to tenant-only scope
- document canonical opaque-ID encoding and the consuming-gear expression-index pattern
- qualify `resource_group_membership` through `gts_type.schema_id` / `gts_type_id`
- cast the querying entity ID to `TEXT` instead of casting RG's opaque `resource_id` to UUID
- retain the existing low-level compiler and group-filter constructors for source compatibility; legacy untyped paths now deny safely
- make the bundled TR PDP honor capability negotiation; RG advertises no group capabilities (its own group resource has no RG member-handle type - group nesting is hierarchy, not membership rows)
- add PostgreSQL execution tests for UUID/text matching, duplicate IDs across member types, opaque non-UUID IDs, type and tenant isolation, unknown types, and descendant groups
- document the mapping and projection requirements

## Type of Change
- [x] Breaking change (adds a public ConstraintCompileError variant and #[non_exhaustive]; authz-resolver-sdk goes to 0.5.0)
- [x] Documentation update

## Testing
- [x] Unit tests pass (`cf-gears-toolkit-security`, `cf-gears-authz-resolver-sdk`, `cf-gears-toolkit-db`, `cf-gears-tr-authz-plugin`, `cf-gears-resource-group`)
- [x] Integration tests pass (`make test-pg`: 538/538)
- [x] New PostgreSQL regression tests added
- [x] Architecture lints pass (`make dylint`)
- [x] GTS documentation validation passes (`make gts-docs`: 798 files)
- [x] Strict Clippy passes for all modified crates and targets

The full workspace `make clippy` currently stops on an unrelated pre-existing `collapsible_if` warning in `libs/rustls-corecrypto-provider/tests/handshake_smoke.rs` under local Rust 1.97.0. The modified crates pass with `-D warnings -D clippy::perf`.

## Documentation
- [x] Authorization and SecureORM documentation updated
- [ ] API documentation updated (not applicable)

## Checklist
- [x] Self-review completed
- [x] Tests pass
- [x] Commit includes DCO sign-off

## Compatibility / Versioning
This is a non-breaking security/correctness fix. New typed APIs are additive; existing public constructors and `compile_to_access_scope` remain source-compatible. Untyped native group predicates now fail closed rather than querying ambiguous membership rows.

## Related Issues
Fixes #4644

Downstream tracking: VHP-2344

> Supersedes #4618 (same head commit; the branch was renamed, which a cross-repo PR cannot follow). Review threads, their resolutions, and CI history live there.

## Summary by CodeRabbit

- **New Features**
  - Added typed resource-group membership and subtree filtering with tenant and membership-type validation.
  - Added clearer authorization errors when required capabilities are unavailable.

- **Bug Fixes**
  - Authorization now fails closed for empty, malformed, unsupported, or unadvertised group selectors.
  - Static authorization providers deny unresolved group selectors instead of ignoring them.
  - Improved handling of UUID-backed resources and textual membership identifiers.

- **Documentation**
  - Updated authorization guidance and examples for canonical resource types, group capabilities, and fail-closed behavior.


entropyshift (approved):

ffedoroff (approved):
reviewed and approved https://github.com/constructorfabric/gears-rust/pull/4618 this is clone, so also approve without review

aviator5:
The type-qualified membership SQL makes sense: identical resource IDs belonging to different types must not be conflated. However, could we discuss the separate `with_group_membership_type(...)` mapping as an API/design decision?

`ResourceType` already carries the resource type sent to AuthZ. When that value is the resource's canonical GTS type, why is it insufficient for the membership lookup as well? The documentation says a gear may use a different GTS path for policy matching, but I could not find a concrete consumer that requires those two identities to differ. Could you provide an example?

There is also the RG registry inconsistency discussed in #4618: membership types are documented/validated as external domain types without an RG prefix requirement, while the type-creation API requires `gts.cf.core.rg.type.v1~`. That limitation should be considered separately from whether an external resource needs a second GTS identity for membership.

I suggest we discuss this separately before establishing a required mapping for consumers. My preference would be to use the existing resource type for membership qualification when it is the canonical GTS ID, keeping native-group support explicit through capabilities. If a real consumer needs a different mapping, we can define and document that use case explicitly.

@fdlockgraf , @ffedoroff fyi

ffedoroff:
> The type-qualified membership SQL makes sense: identical resource IDs belonging to different types must not be conflated. However, could we discuss the separate `with_group_membership_type(...)` mapping as an API/design decision?
> 
> `ResourceType` already carries the resource type sent to AuthZ. When that value is the resource's canonical GTS type, why is it insufficient for the membership lookup as well? The documentation says a gear may use a different GTS path for policy matching, but I could not find a concrete consumer that requires those two identities to differ. Could you provide an example?
> 
> There is also the RG registry inconsistency discussed in #4618: membership types are documented/validated as external domain types without an RG prefix requirement, while the type-creation API requires `gts.cf.core.rg.type.v1~`. That limitation should be considered separately from whether an external resource needs a second GTS identity for membership.
> 
> I suggest we discuss this separately before establishing a required mapping for consumers. My preference would be to use the existing resource type for membership qualification when it is the canonical GTS ID, keeping native-group support explicit through capabilities. If a real consumer needs a different mapping, we can define and document that use case explicitly.
> 
> @fdlockgraf , @ffedoroff fyi

Agreed with @aviator5  — done in b36e0add

On your question: a consumer with diverging identities does exist. AM authorizes users
as `gts.cf.core.am.user.v1~` but registers their memberships under
`gts.cf.core.rg.type.v1~cf.core.am.user.v1~`. That is a workaround for #4052, not a
requirement, so it supports your proposal — remove the wrapper rather than model it as
an API dimension.

#4052 deserves more weight than "separate": until an external canonical type can be
registered, the only realistic consumer cannot opt in at all.

This is also consistent with keeping the RG hierarchy typed. Groups and membership rows
keep their own type, and they keep it after type storage moves from RG to Types
Registry — what moves is where types are registered, not whether RG rows carry one.

MikeFalcon77 (changes requested):
conflicts and comments

aviator5 (commented):
No issues found.

aviator5 (approved):

PR #4772, opened 2026-09-09
Types Registry implementation P0-4

### LINKED ISSUE #4624: [Types Registry P0] Phase 4 — compatibility
Parent: #4619 · Depends on: Phase 3 (#4623)

## Goal

Enforce BACKWARD compatibility and the derivation rules, fail-closed: an undecidable comparison
is rejected with its own reason rather than collapsed into "incompatible".

## Tasks

- [ ] **T17** — baseline selection (the entity's current revision for a major-only candidate,
      or the `ACTIVE`/`DELETED` definition of `vM.(n-1)~` for a minor-bearing one) compared
      through `GtsStore::compare_documents`, which resolves both sides. `Unknown` is rejected
      with its own reason.
- [ ] **T18** — identifier-derived chain validation against every managed base, the Draft-07
      dialect pin across a major, and the ADR-0015 quarantine: a stable candidate may not
      derive from or `$ref` a major-0 identifier, and a major-0 schema may not carry a
      registered Instance. `x-gts-ref` is an instance-value constraint and is outside the
      quarantine (plan.md P14).

**Observability is part of both tasks, not a follow-up** (plan.md **P16**): T16 instrumented the
admission path as it stood at the end of Phase 3, and its port carries no verdict at all.

- [ ] **T17 also** — a new `types_registry_compat_verdicts_total{verdict,forced}` over the closed
      set `compatible | incompatible | unknown`, so `Unknown` is distinguishable from
      `Incompatible` in the metrics and not only in a refusal reason, and a forced waiver is
      countable on its own. The unit span records the selected baseline, the verdict and the two
      provenance versions — identifiers stay span fields, never labels. **And the admission
      reason vocabulary gets one compile-enforced home:** today's scattered
      `ItemFailure::new("literal", …)` sites move onto `pub const`s of a `Reason` newtype, so a
      refusal a later task adds cannot compile without naming one.
- [ ] **T18 also** — each quarantine and dialect refusal carries its own reason const
      (`stable_derives_from_major_zero`, `stable_refs_major_zero`, `instance_of_major_zero`,
      `dialect_changed`), never collapsed into `invalid_schema`: an ADR-0015 refusal and a
      malformed document are different operator actions.

Evidence bar for both is T16's: rendered names, label keys and label *values* asserted against an
`InMemoryMetricExporter`; emission asserted end to end through `run_operation`; and a mutation
check that stripping the emission fails the tests.

## Exit criteria — Checkpoint 4

- [ ] The compatibility matrix passes, including the `Unknown` tier rejected with its own reason
- [ ] Provenance persisted on every revision
- [ ] Quarantine and dialect rules hold, including the two `x-gts-ref` non-cases (no preflight
      scan is needed — the release that introduces the check is the release that first persists
      an entity)
- [ ] Every verdict is counted, with `Unknown` and a forced waiver each distinguishable in the
      metrics (P16)
- [ ] Quarantine and dialect refusals each carry their own counted reason, none collapsed into
      `invalid_schema` (P16)
- [ ] Admission reasons live in one compile-enforced vocabulary: a new refusal cannot compile
      without naming one (P16)
- [ ] `make dylint` — full workspace, once for the phase
- [ ] Human review

## References

- `gears/system/types-registry/docs/p0/SPEC.md` — §7 (0.12.0 tri-state verdict), §8.6 (observability of the write path), §16.12 and §16.16, O4 (why no quarantine preflight is needed)
- `gears/system/types-registry/docs/p0/plan.md` — P16 (observability is a per-task obligation from T17 on, not a second T16)
- `gears/system/types-registry/docs/p0/todo.md` — T17–T18

Already on `main`:

- `gears/system/types-registry/docs/ADR/0003-cpt-cf-types-registry-adr-type-schema-evolution-compatibility.md`
- `gears/system/types-registry/docs/ADR/0014-cpt-cf-types-registry-adr-managed-type-schema-dialect-profile.md`
- `gears/system/types-registry/docs/ADR/0015-cpt-cf-types-registry-adr-major-zero-unstable-profile.md`

aviator5:
closes #4624

## New Features
- Added baseline-aware compatibility validation for schema admissions, including cross-minor comparisons and clear refusal reasons.
- Added optional compatibility waivers for eligible cross-minor changes; waivers are rechecked during processing and recorded with the operation.
- Added major-zero dependency quarantine and JSON Schema dialect consistency checks.
- Added compatibility verdict metrics and admission trace details.
- Added safeguards for abstract Type Schema transitions when live direct instances exist.

## Bug Fixes
- Improved handling of unreadable, missing, or invalid compatibility baselines.
- Limited compatibility diagnostics to bounded, structured messages with stable refusal reasons.


gears/system/types-registry/types-registry/src/domain/admission/reasons_tests.rs
```diff
@@ -0,0 +1,243 @@
+//! Admission reason vocabulary checks (P16 rule 3).
+//!
+//! Exhaustive production matches require codes and labels. These tests check
+//! uniqueness, round-trips, and bounded labels. [`known`] is checked against
+//! the enum source so an omitted variant cannot escape those assertions.
+//! Using `sea_orm`'s `EnumIter` would add a storage dependency to the domain.
+
+#![allow(clippy::expect_used, clippy::unwrap_used)]
```
ainetx, line 9: ### New clippy allow(expect_used/unwrap_used) lacks an inline justification comment

**Severity:** Minor

**Problem**
reasons_tests.rs adds `#![allow(clippy::expect_used, clippy::unwrap_used)]` immediately after the file's module doc comment, which describes the test file's purpose but never states why the expect/unwrap suppression itself is necessary. Other allow attributes in this codebase (and a previously confirmed finding, pr-4704/) establish the convention that a clippy allow needs its own inline justification, distinct from surrounding doc prose.

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Open gears/system/types-registry/types-registry/src/domain/admission/reasons_tests.rs.
2. Note the module doc block (lines 1-6) discusses test coverage rationale only.
3. Line 8 adds `#![allow(clippy::expect_used, clippy::unwrap_used)]` with no comment directly justifying the suppression.

**Expected behavior**
The allow attribute should be preceded by a short comment explaining why expect/unwrap are needed here (e.g., 'test fixtures deliberately panic on malformed input') per the project's established review convention.

**Actual behavior**
The allow is present with no dedicated justification, matching a previously flagged pattern in this codebase.

```text
module doc (purpose) -> #![allow(...)] (no reason) -> unwrap/expect calls throughout tests
```

**Impact**
Low risk since the suppression is scoped to a #[cfg(test)]-only module, but it perpetuates a documented review gap making it harder for reviewers/dashboards to audit why lint suppressions exist.

**Suggested correction**
Add a one-line comment directly above the `#![allow(...)]` explaining the justification (e.g., referencing that these are deliberately fail-fast test fixtures).

**How to verify**
Re-review the file and confirm the allow attribute is preceded by an explicit justification comment.

</details>

ainetx: **RESOLVED** -- Confirmed fixed as of 3717ea6635e85a0867a41066363cd0a159eb91ed -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** All new.expect()/.unwrap() calls are confined to test modules (reasons_tests.rs carries its own #![allow(...)] header; error.rs's new BaselineUnparsable fixture uses expect_err inside mod tests), consistent with the workspace convention of allowing these only in test code.
- **Verifier:** New.unwrap()/.expect() calls are confined to test code: reasons_tests.rs carries a file-level #![allow(clippy::expect_used, clippy::unwrap_used)] with a module doc explaining the purpose of the source-parsing guards, and the new expect_err in api/rest/error.rs's tests module builds a deliberately invalid-JSON fixture -- consistent with this repo's pervasive convention of blanket test-module allows (seen identically in migration_test.rs, admission_worker_test.rs, etc.).

</details>

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
```diff
@@ -381,11 +555,215 @@ fn evaluate_loaded(
...
+    if waived && verdict != CompatibilityVerdict::Compatible {
+        // Log the waiver decision; commit-time checks may still refuse the candidate.
+        // Enter the captured span because this runs inside `spawn_blocking`.
+        let _entered = reporting.span.enter();
+        tracing::warn!(
+            gts_id = %id.id(),
+            baseline_gts_id = %baseline_id,
+            baseline_revision,
+            verdict = verdict.as_str(),
+            "types_registry waived a non-compatible verdict on ADR-0004 force; the \
```
ainetx, line 724: ### Compatibility diagnostics persisted only as a truncated free-text blob

_Non-blocking review challenge -- [product]_

`diagnostics_summary` renders up to 20 diagnostics as a single joined string (`"{finding:?} at {path}"`), capped and truncated, and that string is what ends up in the refusal message / `error_payload`. Was a structured diagnostics representation (e.g. a JSON array of `{finding, path}`) considered so API consumers/tooling could parse and act on individual findings, rather than needing to regex a human-readable summary that's explicitly lossy beyond 20 entries?

**Why this is worth asking**
For a registry API where callers presumably build tooling around admission failures, baking the cap/truncation into an unstructured string forecloses programmatic consumption of the full diagnostic set and makes the 20-entry/200-byte limits a permanent API contract embedded in prose rather than a documented, adjustable interface.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
```diff
@@ -381,11 +555,215 @@ fn evaluate_loaded(
...
+    fn record(
+        &self,
+        baseline_gts_id: Option<&str>,
+        baseline_revision: Option<i32>,
+        verdict: Option<CompatibilityVerdict>,
+    ) {
+        observability::record_compat_facts(
+            self.span,
+            CompatFacts {
+                baseline: self.choice,
```
ainetx, line 639: ### Waiver re-authorization can silently downgrade `force` between acceptance and worker

_Non-blocking review challenge -- [architecture]_

Acceptance validates `force` eligibility against `select_baseline(&id, expected)` at accept time, but the worker independently recomputes `forced = accepted_force && choice.waivable()` from whatever baseline is selected when the item is actually processed. If the effective precondition/baseline classification differs by the time the worker runs (e.g. a race changes whether this is treated as a revision vs. creation), the waiver silently evaporates and the candidate is judged as if `force` were never set — with no distinct refusal reason surfaced to the caller. Is that silent downgrade intentional, or should a change in waivability between acceptance and worker processing surface as its own refusal (analogous to `BaselineUnresolvable`) rather than falling through to an ordinary compatibility refusal?

**Why this is worth asking**
A caller who explicitly requested a compatibility waiver could have their intent quietly ignored without any signal distinguishing 'waiver not honored due to race' from 'waiver honored but candidate still incompatible'. That's a meaningful product/debuggability gap for a mechanism (ADR-0004) explicitly designed to be an audited exception path.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
aviator5: Considered and deferred until a concrete consumer needs individual findings. For now, the contract is a stable `reason` plus a bounded human-readable `message`; clients should not parse the message. PRD, ADR-0003 and SPEC now explicitly reflect this decision, including that the summary limits are implementation details.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
aviator5: `select_baseline` is a pure function of the identifier and accepted precondition, both persisted and reused by the worker, including retries. A database race cannot reclassify creation as revision or change waiver eligibility. Disabling the deployment flag can intentionally clear `force`; that behavior is documented and covered by a test.

gears/system/types-registry/docs/p0/SPEC.md
```diff
@@ -334,18 +334,31 @@ out of scope):
...
-   `force`: there is no comparison to waive and no truthful `compat_forced` to record yet.
-7. ADR-0015 quarantine — refuse a stable candidate whose immediate base or `$ref` targets
-   include a major-0 identifier. `x-gts-ref` is outside the quarantine.
+6. `force` per candidate — require `allow_compatibility_force` and a waivable
+   cross-minor baseline from `compat::select_baseline`. Intra-entity revisions are
+   never waivable. Store the request on `operation_item.compat_forced`; each worker
+   pass rechecks the deployment setting and baseline eligibility. The revision's
+   `compat_forced` records the effective waiver.
+7. ADR-0015 quarantine — the worker refuses stable candidates whose immediate base,
+   `$ref` target, or conforming type is major 0. `x-gts-ref` is exempt. See below.
```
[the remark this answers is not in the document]
aviator5: Not a divergence — the two rules are deliberately separate, and both are already normative.

- The stable-reference quarantine (base / `$ref`) preserves the resolution closure of stable contracts against a floating v0 ref — PRD "Quarantine" paragraph, ADR-0015.
- The Instance-conformance refusal has a different rationale: an unstable schema cannot carry an Instance because the recorded revalidation would stop being truthful. It is stated at `PRD.md:238` (MUST NOT, with rationale), `ADR-0015` Confirmation ("registration of a registered Instance conforming to a v0 Type Schema is rejected"), `ADR-0006:109`, and the PRD checklist at line 850.

`DESIGN.md:521` calls this out explicitly: the Instance rule is *"distinct from the stable-reference quarantine"*. The implementation folds all three edge kinds into one `quarantine()` only to reuse `extract_edges`; it still distinguishes them via `StableDerivesFromMajorZero` / `StableRefsMajorZero` / `InstanceOfMajorZero`.

Folding conformance into the PRD quarantine paragraph would duplicate `PRD.md:238` and contradict `DESIGN.md:521`, so leaving as is.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
ainetx: OK, thank you

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
ainetx: OK, got it

ainetx (approved):
This is a large, carefully coordinated change spanning the compatibility engine (T17/T18), its API surface, storage, and the accompanying spec/ADR/PRD documentation, and it holds together well. The refusal contract (`reason`/`message`) is consistently threaded through the DTOs, error mapping, and docs, the worker-side quarantine and dialect-pin placement is well justified against the extracted dependency edges, and the trickier concurrency cases (abstract/derived-type transitions racing Instance creation, baseline version pinning against a moving `expected_resource_version`) are backed by explicit tests across all three database backends. Test coverage, observability (verdict/forced metrics, span fields), and the migration for `compat_forced` all line up with the acceptance criteria. Nothing blocking turned up in review — approved.

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
ainetx: If select_baseline is deterministic and the flag-disable path is the only route to silently clearing force, should that path also surface as a distinct refusal reason rather than folding into an ordinary compatibility refusal — or is the existing test asserting that the caller can't distinguish the two?

gears/system/types-registry/types-registry/src/domain/admission/unit.rs
ainetx: OK, answer accepted.

gears/system/types-registry/types-registry/src/domain/compat/baseline.rs
```diff
@@ -0,0 +1,184 @@
+//! Backward compatibility against one baseline (ADR-0003, SPEC §8.1 step 3).
+//!
+//! ```text
+//! not a Type Schema       -> no comparison; validate the Instance instead
+//! major 0                 -> no comparison (ADR-0015)
+//! vM~, creation           -> no comparison
+//! vM~, revision           -> current revision; never waivable
+//! vM.0~                   -> no comparison; opens the major
+//! vM.n~, n > 0, creation  -> current definition of vM.(n-1)~; waivable
+//! ```
```
ainetx, line 10: ### Non-ASCII punctuation (em-dash, en-dash, math symbols) added to Rust doc comments across new/edited types-registry compat and admission modules

**Severity:** Minor

**Problem**
Several newly added or retained doc comments in this diff use non-ASCII characters (em-dash '—', en-dash '–', subset symbol '⊆', ellipsis '…') purely as stylistic punctuation, with no i18n purpose. This repo has twice before confirmed the identical pattern (em-dash in Rust doc comments with no i18n intent) as a violation of the workspace's ASCII-only source policy (pr-4747/, pr-4704/).

<details>
<summary>Reproduction, impact, suggested fix, verification</summary>

**How to reproduce**
1. Open gears/system/types-registry/types-registry/src/domain/compat/baseline.rs new module doc: contains '⊆' and '—'. 2. Open src/domain/admission/unit.rs: new doc comments in CompatReporting::record and commit_revision area contain '—'. 3. Open src/domain/admission/acceptance.rs: comments contain '–' and '—' and '…'.

**Expected behavior**
New/edited source lines should use only ASCII characters in comments as in identifiers/literals, per the workspace's non_ascii_idents/non_ascii_literal policy; any legitimate non-ASCII use should be flagged as explicitly reviewed for i18n intent.

**Actual behavior**
Non-ASCII em-dash/en-dash/ellipsis/math-symbol characters were introduced in multiple new doc comments across this PR's changed files without i18n justification.

```text
PR diff --> new doc comments (baseline.rs, unit.rs, acceptance.rs) --> contain U+2014/U+2013/U+2026/U+2286 --> violates ASCII-only source policy (same pattern previously confirmed in pr-4747/, pr-4704/)
```

**Impact**
Low functional risk, but is a recurring, previously-confirmed policy violation in this exact repo; likely to fail a CI ASCII-source check if one is enforced, and is inconsistent with the team's stated convention.

**Suggested correction**
Replace the non-ASCII punctuation with ASCII equivalents (e.g. '--' or ',' for em/en-dash, '...' for ellipsis, and a spelled-out phrase like 'is a subset of' for '⊆').

**How to verify**
Grep the changed.rs files for non-ASCII byte sequences (e.g. `\u{2014}|\u{2013}|\u{2026}|\u{2286}`) and confirm none remain in comments/doc comments without an explicit i18n review note.

</details>

aviator5: Suggest discussing whether we are ok or not having non-ascii symbols in comments firstly and implement dylint rule in the 2nd case.

ainetx: This concern still stands.

<details>
<summary>Why</summary>

The finding stands: new `baseline.rs` doc comments contain `⊆` and em dashes, and `acceptance.rs` adds em dashes. The workspace has already confirmed this exact ASCII-only comment policy in `pr-4747/` and `pr-4704/`; discussing or later automating enforcement does not justify this PR's introduced violations.

</details>

PR #4776, opened 2026-09-10
feat(toolkit-db): tell the submitter when a traced outbox batch completes

striped-zebra-dev:
## Summary by CodeRabbit

* **New Features**
  * Added builder APIs for creating and enqueueing individual records and batches.
  * Added optional batch tracing with completion notifications, progress monitoring, durable status lookup, and automatic cleanup.
  * Added payload and trace validation with clearer error details.
  * Added byte-size constants to toolkit utilities.

* **Documentation**
  * Updated outbox documentation and examples for record builders, validation, batch behavior, and tracing.

* **Breaking Changes**
  * Replaced legacy enqueue signatures and batch message types with `Record` and `Records` APIs.


libs/toolkit-db/src/outbox/migrations.rs
```diff
@@ -45,6 +45,7 @@ impl MigrationTrait for CreateOutboxSchema {
         let backend = conn.get_database_backend();
         let tables = &self.tables;
 
+        create_trace(conn, backend, tables).await?;
```
MikeFalcon77, line 48: **CRITICAL**

The trace table and the new `trace_id` columns on `body` and `dead_letters` were added inside `CreateOutboxSchema`, whose name still resolves to `m001_create_toolkit_outbox_schema` (the `default_migration_name_is_preserved` test pins it). Any database that already recorded m001 skips all of it, and there is no `ALTER TABLE` anywhere in the file, so the trace table and the columns never get created. `statements.rs:490,505,510` inserts `trace_id` unconditionally, so every enqueue in mini-chat, types-registry and event-broker breaks on upgrade. This has to be a new migration (`m002_create_toolkit_outbox_trace`) with `CREATE TABLE IF NOT EXISTS` plus per-backend `ALTER TABLE ... ADD COLUMN trace_id`, returned from `outbox_migrations()` and `outbox_migrations_with_prefix()`. Fresh installs pass, which is why CI is green.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+    ///
+    /// Returns whether anyone was: `false` means the mail has been claimed and
+    /// has nowhere to go, which is the case when the guard was dropped or the
+    /// process restarted, and the mail is then correctly discarded rather than
+    /// retained.
+    pub fn deliver(&self, outcome: TraceOutcome) -> bool {
+        let Ok(mut entries) = self.entries.lock() else {
+            return false;
+        };
+        let Some(interest) = entries.get_mut(&outcome.trace) else {
```
MikeFalcon77, line 177: **HIGH**

`deliver` finds the waiter with `entries.get_mut(&outcome.trace)` and never compares the `generation` it stored on `Interest`. Duplicate traces are legal by design, `trace_status` documents that at `statements.rs:656`, so when two batches share a trace, batch A's outcome resolves batch B's subscriber while B is still in flight and A's real waiter never hears anything. Key the registry on the trace row id from `insert_trace_row` and check `generation` before handing the outcome over.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+        let generation = self
+            .generations
+            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        // A subscription taken after the outbox stopped keeps no entry, so its
+        // sender drops here and awaiting it yields `None` at once rather than
+        // waiting on a pipeline that will never run.
+        if !self.closed.load(std::sync::atomic::Ordering::Acquire)
+            && let Ok(mut entries) = self.entries.lock()
+        {
+            entries.insert(
```
MikeFalcon77, line 126: **HIGH**

`entries.insert` replaces any existing `Interest` for the same trace, so the displaced `oneshot::Sender` drops right here and the first subscriber immediately resolves to `None`, which is documented to mean this process can no longer answer. Its stall reports stop too. We need more than one interest per trace, either `HashMap<String, Vec<Interest>>` or keyed on `(trace, generation)`, with `deliver` fanning out to all live ones.

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -78,6 +100,110 @@ impl Outbox {
...
+    /// Returns `None` once the row has been swept, which for a completed trace
+    /// means it finished and was collected.
+    ///
+    /// Nothing requires a caller's traces to be unique; the most recently
+    /// enqueued one is reported.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the read fails.
+    pub async fn trace_status(
```
MikeFalcon77, line 171: **HIGH**

`trace_status` is `WHERE trace = $1` with no owner predicate (`statements.rs:667`), unlike `trace_mail`, `trace_stalled` and `trace_claim_mail`, which all scope on `owner_instance`. Anyone who knows or guesses another submitter's trace string reads that batch's queue name, entity/failure/attempt counts and the raw `last_error` text. This needs an ownership predicate bound alongside the trace, and `last_error` shouldn't go back to unscoped callers.

libs/toolkit-db/src/outbox/strategy.rs
```diff
@@ -497,10 +714,14 @@ async fn lease_guarded_ack(
...
         );
         ack_txn.rollback().await?;
+        // The claims are rolled back with everything else, and are
+        // deliberately not delivered: a caller told its batch finished must
+        // not learn it from a transaction that did not commit.
         return Ok(None);
     }
 
     ack_txn.commit().await?;
+    deliver_claimed(ctx.mailbox, claimed);
```
MikeFalcon77, line 724: **HIGH**

Nothing covers the lease-expired path where `ack_txn.rollback()` runs and the already-claimed completions must not be delivered. A regression that moved `deliver_claimed` above the commit would pass CI, and delivering only after the work is durable is the point of the whole feature. Can we add a test that subscribes to a traced batch, expires the lease before the ack with the existing `expire_lease` helper, and checks the subscription stays unresolved while `notified_at` is still NULL?

libs/toolkit-db/src/outbox/store.rs
```diff
@@ -260,22 +269,239 @@ impl<'a> OutboxStore<'a> {
...
+    /// Claim mail with no transaction of the caller's own, atomically on every
+    /// backend.
+    ///
+    /// The ack calls the primitive below inside its own transaction, so it is
+    /// already atomic there. The collector has none, and on a backend without
+    /// `RETURNING` the claim is an `UPDATE` followed by a `SELECT` - if the
+    /// second fails, the row is stamped delivered with nobody told, no record
+    /// and no retry, and the caller waits for ever. So that backend gets a
+    /// transaction; the others are one statement and need none.
+    pub(super) async fn claim_trace_mail_alone(
```
MikeFalcon77, line 430: **HIGH**

The non-`RETURNING` (MySQL) path is untested at every level: `claim_trace_mail_alone`'s transactional branch, `exec_claim_trace_mail`'s claim-outcome read, and `exec_trace_advance`'s `TraceAdvance::Unknown` branch. The store tests only assert insert SQL strings and the cross-backend statement tests only compare placeholder counts, not claim semantics. This is the same path where coderabbitai found two separate data-integrity bugs. We need MySQL/Postgres-gated tests, or recorder tests on the per-dialect statement sequence, that drive a traced batch to completion and check the claim UPDATE runs before the outcome read and the outcome is delivered once.

libs/toolkit-db/src/outbox/manager.rs
```diff
@@ -627,5 +740,11 @@ impl OutboxHandle {
     /// Cancel background tasks and join all handles. Consumes self.
     pub async fn stop(self) {
         self.tasks.shutdown().await;
+        // Resolve anyone still awaiting a completion. Dropping the senders is
+        // what makes an awaiting subscription yield `None`, which is the
+        // documented signal that this process can no longer answer. Without
+        // it the caller waits for ever: it necessarily holds the `Arc` that
+        // owns the registry, so nothing else would ever drop them.
+        self.outbox.mailbox().subscriptions().close();
```
MikeFalcon77, line 748: **MEDIUM**

`subscriptions().close()` only runs in `stop()`, but dropping the handle is a supported shutdown path since `TaskSet::drop` cancels every worker token. On that path the registry is never closed, so a caller holding its own `Arc<Outbox>` and awaiting a `TraceSubscription` waits forever, which is exactly what `Registry::close`'s doc comment says it exists to prevent. Add `impl Drop for OutboxHandle` that calls `self.outbox.mailbox().subscriptions().close()`.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+    }
+
+    /// Hand a completion to whoever is waiting for it.
+    ///
+    /// Returns whether anyone was: `false` means the mail has been claimed and
+    /// has nowhere to go, which is the case when the guard was dropped or the
+    /// process restarted, and the mail is then correctly discarded rather than
+    /// retained.
+    pub fn deliver(&self, outcome: TraceOutcome) -> bool {
+        let Ok(mut entries) = self.entries.lock() else {
```
MikeFalcon77, line 174: **MEDIUM**

Lock poisoning is swallowed at all seven sites (`deliver`, `publish_progress`, `withdraw_progress_except`, `is_idle`, `len`, `release`, `close`) through `let Ok(..) else`, `is_ok_and` and `map_or`, and none of them logs. `is_idle` then answers `true` and `deliver` answers `false`, so a single panic while the lock is held, which `PanicPolicy::CatchAndRetry` deliberately keeps the process alive through, permanently stops the notifier from querying and makes every later completion vanish. Use `.lock().unwrap_or_else(std::sync::PoisonError::into_inner)`, the guarded `HashMap` has no invariant a panic can break, and log at `error!` when recovery kicks in.

libs/toolkit-db/src/outbox/workers/notifier.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub next_look: Duration,
+}
+
+impl Notifier {
+    /// Collect one page of mail. Returns how many completions were delivered.
+    /// Returns how many completions were delivered, or the failure that
+    /// stopped it. A failure must reach the worker loop: swallowing it would
+    /// clear the backoff, so a database outage would be a failing query and a
+    /// warning every 100ms with no escalation at all.
+    async fn collect(&self) -> Result<u64, OutboxError> {
```
MikeFalcon77, line 59: **MEDIUM**

`collect` is not cancel-safe and says nothing about it. `claim_trace_mail_alone().await` commits the delivered stamp before the in-process `deliver`, so dropping this future at that await point, which any `select!`, `timeout` or `abort` added around `execute` later would do, stamps the mail delivered with nobody told and the subscriber waits until `close()`. Mark it `// NOT cancel-safe: the claim commits before the in-process handoff`, and either keep claim and deliver inside a spawned unabortable region or re-open the claim when the handoff fails.

libs/toolkit-db/src/outbox/store.rs
```diff
@@ -260,22 +269,239 @@ impl<'a> OutboxStore<'a> {
...
+        instance_id: &str,
+    ) -> Result<Option<super::trace::TraceOutcome>, DbErr> {
+        if self.dialect().supports_returning() {
+            let runner = crate::secure::SeaOrmRunner::Conn(conn);
+            return self
+                .exec_claim_trace_mail(&runner.executor(), trace_id, instance_id)
+                .await;
+        }
+
+        let txn = conn.begin().await?;
```
MikeFalcon77, line 443: **MEDIUM**

This branch opens a transaction and then leaves cleanup to `Drop` on the error and cancellation paths. A dropped sea-orm/sqlx transaction only queues a deferred rollback that runs when the connection is next driven, so the claim can sit uncommitted while holding row locks. Match on the inner result and call `txn.rollback().await` explicitly on the error path.

libs/toolkit-db/src/outbox/types.rs
```diff
@@ -97,14 +122,23 @@ pub enum OutboxError {
         found: usize,
     },
 
-    #[error("invalid queue name: '{0}'")]
-    InvalidQueueName(String),
+    #[error("a traced batch must carry at least one entity")]
+    EmptyTracedBatch,
```
MikeFalcon77, line 126: **MEDIUM**

This goes from 8 to 11 variants: three added (`EmptyTracedBatch`, `InvalidTrace`, `TraceTooLong`) and three reshaped from tuple to struct form, with `InvalidQueueName(String)`, `InvalidPayloadType(String)` and `InvalidTablePrefix(String)` all becoming `{ reason: &'static str }`. The enum isn't `#[non_exhaustive]`, so every downstream exhaustive `match` and every `InvalidQueueName(name)` pattern stops compiling, and the next variant added will break them again. Mark it `#[non_exhaustive]` while we're already taking the break, and get all three reshapes into the changelog. The crate is at `0.14.1`, so under Cargo semver this is a minor bump to `0.15.0`, and the commit needs `!` or a `BREAKING CHANGE:` footer for release-plz to cut it that way.

libs/toolkit-db/src/outbox/mod.rs
```diff
@@ -144,9 +230,11 @@ pub use handler::{
 };
 pub use manager::{OutboxBuilder, OutboxHandle};
 pub use migrations::{outbox_migrations, outbox_migrations_with_prefix};
+pub use record::{Record, RecordBuilder, RecordTarget, Records, RecordsBuilder, RecordsTarget};
+pub use subscription::{TraceProgressWatch, TraceSubscription};
+pub use trace::{TraceAdvance, TraceOutcome, TraceProgress, TraceStatus};
 pub use types::{
-    EnqueueMessage, LeaseConfig, OutboxError, OutboxMessageId, OutboxProfile, Partitions,
-    WorkerTuning,
+    LeaseConfig, OutboxError, OutboxMessageId, OutboxProfile, Partitions, WorkerTuning,
```
MikeFalcon77, line 237: **MEDIUM**

`EnqueueMessage` is dropped from the public exports with no deprecation path, and the change ships as a plain `feat(toolkit-db):`, so release-plz publishes a minor bump for what is a breaking removal. Keep a `#[deprecated(note = "use Records::to(..).push(..)")]` re-export for one release cycle and mark the commit breaking.

libs/toolkit-db/src/outbox/types.rs
```diff
@@ -550,6 +629,8 @@ pub struct OutboxProfile {
     pub processor: WorkerTuning,
     pub vacuum: WorkerTuning,
     pub reconciler: WorkerTuning,
+    pub notifier: WorkerTuning,
```
MikeFalcon77, line 632: **MEDIUM**

`OutboxProfile` is public with all fields public, so adding `notifier` and `trace_sweeper` breaks every downstream struct literal, and every future worker will break it again. Either `#[non_exhaustive]` or private fields behind per-worker setters.

libs/toolkit-db/src/outbox/trace.rs
```diff
@@ -0,0 +1,119 @@
...
+//! trace row outlives both the messages it describes and the instance that
+//! enqueued them, so a restarted process can still ask what became of work it
+//! submitted before it died.
+
+/// What a caller is told when its batch finishes.
+///
+/// Delivered once, to the instance that enqueued the batch, whichever instance
+/// processed the entities.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TraceOutcome {
```
MikeFalcon77, line 14: **MEDIUM**

`TraceOutcome`, `TraceProgress` and `TraceStatus` are public output structs with every field public and no `#[non_exhaustive]`, so any added field is breaking. The counts are `i64` as well, which makes negative entity, failure and attempt counts representable in a public type. Mark the three `#[non_exhaustive]`, expose the counts through accessors, and type them `u64` or a newtype.

libs/toolkit-db/src/outbox/mod.rs
```diff
@@ -144,9 +230,11 @@ pub use handler::{
 };
 pub use manager::{OutboxBuilder, OutboxHandle};
 pub use migrations::{outbox_migrations, outbox_migrations_with_prefix};
+pub use record::{Record, RecordBuilder, RecordTarget, Records, RecordsBuilder, RecordsTarget};
+pub use subscription::{TraceProgressWatch, TraceSubscription};
+pub use trace::{TraceAdvance, TraceOutcome, TraceProgress, TraceStatus};
```
MikeFalcon77, line 235: **MEDIUM**

`TraceAdvance` is re-exported publicly, but it's an internal ack-path signal whose `Unknown` variant only exists to encode a dialect having no `RETURNING`. It appears in no public signature, the only function returning it (`exec_trace_advance` at `store.rs:360`) is `pub(super)`, so this leaks a storage-layer detail into the crate's contract. Drop it from the `pub use trace::{...}` list and keep it `pub(crate)`.

libs/toolkit-db/src/outbox/strategy.rs
```diff
@@ -122,46 +137,191 @@ async fn read_messages(
...
+    for &(trace_id, terminal, failures) in progress {
+        let advanced = store
+            .exec_trace_advance(conn, trace_id, terminal, failures, mailbox.instance_id())
+            .await?;
+
+        // Only an advance that reached zero can be claimed. Skipping the claim
+        // otherwise spares the row every partition of the batch contends for
+        // one guarded UPDATE per ack that did not finish it. `Unknown` is the
+        // dialect that cannot say without another read, so it still tries.
+        if matches!(advanced, super::trace::TraceAdvance::StillPending)
```
MikeFalcon77, line 199: **MEDIUM**

Two negative `matches!` checks on our own `TraceAdvance` mean a variant added later silently falls through to the claim path instead of failing to compile. Make it exhaustive: `match advanced { StillPending | NotAffected => continue, Completed | Unknown => {} }`.

libs/toolkit-db/src/outbox/strategy.rs
```diff
@@ -122,46 +137,191 @@ async fn read_messages(
...
+/// reached a terminal state, and how many of those failed.
+///
+/// Only traces present in the batch appear, so an untraced batch produces
+/// nothing and issues no statement.
+fn trace_progress(
+    msgs: &[OutboxMessage],
+    trace_ids: &TraceIds,
+    upto_seq: i64,
+    failed: &HashSet<i64>,
+) -> Vec<(i64, i64, i64)> {
```
MikeFalcon77, line 153: **MEDIUM**

`trace_progress` returns `Vec<(i64, i64, i64)>` and `apply_trace_progress` unpacks it positionally, so trace id, terminal count and failure count are three interchangeable `i64`s that the compiler can't keep apart. A named `struct TraceCountdown { trace_id, terminal, failures }` passed as `&[TraceCountdown]` would stop a transposition from compiling.

libs/toolkit-db/src/outbox/store.rs
```diff
@@ -260,22 +269,239 @@ impl<'a> OutboxStore<'a> {
...
+            .query_one_raw(Statement::from_sql_and_values(
+                self.backend(),
+                self.statements.trace().claim_mail_outcome(),
+                [trace_id.into(), instance_id.into()],
+            ))
+            .await?;
+        row.as_ref().map(Self::outcome_from_row).transpose()
+    }
+
+    fn outcome_from_row(row: &sea_orm::QueryResult) -> Result<super::trace::TraceOutcome, DbErr> {
```
MikeFalcon77, line 485: **MEDIUM**

`outcome_from_row` reads by positional column index while the matching column list is written out again in `trace_claim_mail` and `trace_claim_mail_outcome`. `entities`, `failures` and `attempts` are all `i64`, so reordering any one of the three lists silently mis-assigns fields with no error anywhere. `TraceStatusRow` and `StallRow` already use `#[derive(FromQueryResult)]` with named columns; let's do the same here instead of `try_get_by_index`.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+            generations: std::sync::atomic::AtomicU64::new(0),
+            closed: std::sync::atomic::AtomicBool::new(false),
+        })
+    }
+
+    /// Register interest in a trace, before the enqueuing transaction commits.
+    ///
+    /// A completion cannot precede that commit, so there is no race: the
+    /// registration is in place before anything could deliver to it.
+    pub fn subscribe(self: &Arc<Self>, trace: &str) -> TraceSubscription {
```
MikeFalcon77, line 110: **MEDIUM**

`subscribe` takes any `&str` and never runs `validate_trace`, so a trace that `RecordBuilder::build()` would reject (empty, over 256 bytes, non-printable) still registers an interest that nothing can ever deliver to. Nothing ties the subscribed key to the enqueued one either. A validated `Trace` newtype built through `Trace::new(&str) -> Result<Trace, OutboxError>` and required by both `Outbox::subscribe` and `RecordBuilder::trace`/`RecordsBuilder::trace` covers both.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+/// and who is waiting.
+#[derive(Debug)]
+pub struct Mailbox {
+    instance_id: String,
+    subscriptions: Arc<Registry>,
+}
+
+impl Mailbox {
+    #[must_use]
+    pub fn new(instance_id: String, subscriptions: Arc<Registry>) -> Self {
```
MikeFalcon77, line 30: **MEDIUM**

`Mailbox::new` takes the instance identity as a raw `String` even though `types.rs` adds `InstanceId` for exactly this value, which forces `core.rs` to `as_str().to_owned()` it back into a primitive. Take `InstanceId` in the parameter and the field, and return `&InstanceId` from `instance_id()`.

libs/toolkit-db/src/outbox/manager.rs
```diff
@@ -414,6 +422,87 @@ impl OutboxBuilder {
...
+        };
+        let name = "notifier";
+        let worker = WorkerBuilder::new(name, ctx.cancel.clone())
+            .pacing(tuning)
+            // No timer. It schedules itself while somebody is waiting, and
+            // sleeps on this wakeup when nobody is - a subscription being
+            // taken is the only event that gives it a reason to look.
+            .notifier(outbox.mailbox().subscriptions().arrivals())
+            .notifier(Arc::clone(ctx.start_notify))
+            .listener(TracingListener)
```
MikeFalcon77, line 472: **MEDIUM**

The notifier, stall reporter and trace sweeper are built with only `TracingListener`, while the sequencer and vacuum go through `with_stats`. Completions delivered, stalls published and traces collected end up in no counter, so the stats reporter can't show that the new workers are running at all. Register a `StatsListener` for each through the existing `with_stats(..., ctx.stats_registry.as_ref(), ...)` helper.

gears/mini-chat/mini-chat/src/infra/outbox.rs
```diff
@@ -37,6 +37,22 @@ pub struct InfraOutboxEnqueuer {
...
+/// error before any statement runs.
+fn outbox_message(
+    queue: &str,
+    partition: u32,
+    payload: Vec<u8>,
+) -> Result<toolkit_db::outbox::Record<'_>, DomainError> {
+    toolkit_db::outbox::Record::to(queue, partition)
+        .payload(payload, "application/json")
+        .build()
+        .map_err(|e| DomainError::internal(format!("outbox request: {e}")))
```
MikeFalcon77, line 53: **MEDIUM**

Every `Record::build()` rejection collapses into `DomainError::internal`, so a caller-driven `OutboxError::PayloadTooLarge`, reachable through a large `AuditEnvelope` or thread summary against the 64 KiB cap, comes back as a 500 instead of a client error. `DomainError::validation` is the right mapping for `PayloadTooLarge` and `InvalidTrace`; keep `internal` for `InvalidQueueName` and `InvalidPayloadType`, which are gear config faults.

libs/toolkit-db/src/outbox/integration_tests.rs
```diff
@@ -731,151 +759,968 @@ async fn enqueue_tx_rollback_leaves_no_rows() {
...
+    assert_eq!(outcome.failures, 0);
+    assert!(outcome.is_clean());
+
+    assert!(
+        read_notified_at(&db, "cross-instance-1").await.is_some(),
+        "the claim stamps delivery so it cannot happen twice"
+    );
+
+    // A second pass finds nothing: delivery is at most once.
+    notifier.execute(&cancel).await.unwrap();
```
MikeFalcon77, line 1300: **MEDIUM**

The comment says "A second pass finds nothing: delivery is at most once", but the call's result is discarded and nothing is asserted. `waiting` was also consumed by the `await` at line 1285, so the registry is empty and the notifier's `is_idle` gate returns before issuing any query. This check can't fail. Keep a second live subscription on the same trace before the second `execute`, then assert the pass delivers nothing and leaves `notified_at` unchanged.

libs/toolkit-db/src/outbox/integration_tests.rs
```diff
@@ -979,6 +1840,197 @@ async fn enqueue_transaction_helper_no_flush_on_rollback() {
...
+        .build()
+        .unwrap();
+    outbox.enqueue_batch(&conn, batch).await.unwrap();
+
+    // The reporter is driven by hand so the test does not race its timer.
+    let mut reporter = test_stall_reporter(&outbox, &db);
+    let cancel = tokio_util::sync::CancellationToken::new();
+
+    let mut seen: Option<super::trace::TraceProgress> = None;
+    for _ in 0..200 {
```
MikeFalcon77, line 1926: **MEDIUM**

`StubbornHandler` relents after a fixed 3 retries with 10-20ms backoff while the reporter loop polls every 20ms, so the stall window can close before any poll observes it and `seen.expect(...)` panics. This passes on scheduling, not on behaviour. Have the handler keep retrying until the test signals it, for example an `AtomicBool` the test sets once it has seen the published stall.

libs/toolkit-db/src/outbox/integration_tests.rs
```diff
@@ -731,151 +759,968 @@ async fn enqueue_tx_rollback_leaves_no_rows() {
...
+    // needs is how long it has been stuck.
+    let first_seen = status.stalled_since;
+    let attempts_then = status.attempts;
+    for _ in 0..100 {
+        let later = outbox
+            .trace_status(&conn, "stuck-1")
+            .await
+            .unwrap()
+            .unwrap();
+        if later.attempts > attempts_then {
```
MikeFalcon77, line 1681: **MEDIUM**

The "stalled_since must not move with each new attempt" assertion sits inside `if later.attempts > attempts_then` inside a bounded 100-iteration loop, so if the retry count never advances the invariant is never checked and the test still goes green. Poll until `later.attempts > attempts_then`, fail if the loop runs out, then assert the equality unconditionally outside the loop.

libs/toolkit-db/src/outbox/workers/notifier.rs
```diff
@@ -0,0 +1,143 @@
...
+        Ok(delivered)
+    }
+}
+
+impl WorkerAction for Notifier {
+    type Payload = ();
+    type Error = OutboxError;
+
+    async fn execute(&mut self, _cancel: &CancellationToken) -> Result<Directive, Self::Error> {
+        if self.outbox.mailbox().subscriptions().is_idle() {
```
MikeFalcon77, line 122: **MEDIUM**

Nothing asserts the notifier's headline property: an instance with no subscriptions issues no query at all and returns `Directive::idle()`, and widens `next_look` while somebody waits. The equivalent stall-reporter gate is tested, and the query recorder harness already exists at `integration_tests.rs:1345`. Can we add a recorder test for zero statements on an empty registry, plus one showing the mail query fires once a subscription exists?

libs/toolkit-db/src/outbox/manager.rs
MikeFalcon77, line 748: **MEDIUM**

`Registry::close` is unit-tested on its own, but nothing checks that a `TraceSubscription` held across `handle.stop()` resolves to `None` instead of hanging. That's the guarantee that makes this API safe to await. Can we add an integration test for it with a timeout?

libs/toolkit-db/src/outbox/integration_tests.rs
```diff
@@ -731,151 +759,968 @@ async fn enqueue_tx_rollback_leaves_no_rows() {
...
+    }
+
+    let db = setup_db("ch2b_trace_row").await;
+    let t = make_default_test_outbox().await;
+    t.outbox.register_queue(&db, "q", 1).await.unwrap();
+
+    let conn = db.conn().unwrap();
+    let batch = Records::to("q")
+        .payload_type("text/plain")
+        .trace("import-1")
```
MikeFalcon77, line 919: **MEDIUM**

Every traced test goes through `Records` and `enqueue_batch`. The single-entity path, `Record::to(..).trace(..)` with `Outbox::enqueue`, which writes a one-entity trace row at `core.rs:380`, isn't exercised end to end at all. Can we add a test asserting the trace row's `entities` and `pending` are 1 and that the subscription gets a clean outcome?

libs/toolkit-db/src/outbox/migrations.rs
```diff
@@ -998,6 +1144,73 @@ mod tests {
...
+                "trace",
+            ] {
+                assert!(
+                    trace_columns.iter().any(|c| c == column),
+                    "the trace table is missing {column}"
+                );
+            }
+
+            let indexes = outbox_index_names(&conn).await;
+            for index in [tables.idx_trace_mail(), tables.idx_trace_key()] {
```
MikeFalcon77, line 1206: **MEDIUM**

The index list here only covers `idx_trace_mail` and `idx_trace_key`, so `idx_trace_stalled` (created at `migrations.rs:322`) isn't asserted and a migration that drops or misnames it passes. Add it, and assert the indexed columns rather than just the index names.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+    }
+}
+
+impl Future for TraceSubscription {
+    type Output = Option<TraceOutcome>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        // Already resolved: answer again rather than panicking.
+        let Some(rx) = self.rx.as_mut() else {
+            return Poll::Ready(None);
```
MikeFalcon77, line 375: **MEDIUM**

Two documented behaviours have no test: polling an already-resolved `TraceSubscription` a second time must return `None` rather than panic, which is why `rx` is an `Option`, and `subscribe` must fire the `arrivals()` notify that wakes the collector. Can we add both?

libs/toolkit-db/src/outbox/strategy.rs
```diff
@@ -122,46 +137,191 @@ async fn read_messages(
...
-    Ok(msgs)
+    Ok((msgs, trace_ids))
+}
+
+/// How much of each trace an ack is terminalizing: how many of its entities
+/// reached a terminal state, and how many of those failed.
+///
+/// Only traces present in the batch appear, so an untraced batch produces
+/// nothing and issues no statement.
+fn trace_progress(
```
MikeFalcon77, line 148: **MEDIUM**

`trace_progress` carries a documented deadlock-avoidance invariant, output sorted by trace id because `HashMap` order is the hazard, plus filtering by `upto_seq`, and it has no test. Neither do `traces_beyond` and `rejected_seqs`. The file's test module still only covers worker-id generation. Unit tests for the sort order, the `upto_seq` exclusion and per-trace failure counts would cover it.

libs/toolkit-db/src/outbox/workers/vacuum.rs
```diff
@@ -110,6 +119,13 @@ impl WorkerAction for VacuumTask {
             "vacuum: sweep complete",
         );
 
+        // Bodies were deleted, which may have been the last thing keeping a
+        // trace alive. Told conditionally: a sweep that collected nothing
+        // cannot have made any trace collectable, so it says nothing.
+        if total_deleted > 0 {
```
MikeFalcon77, line 125: **MEDIUM**

`test_collectable_traces()` at `integration_tests.rs:151` just returns a fresh `Notify`, and both call sites (`:3855`, `:5084`) pass it only to satisfy the `VacuumTask::new` signature, never awaiting `notified()`. So `if total_deleted > 0 { self.collectable_traces.notify_one(); }` is unverified in both directions, and this is the wakeup that drives the trace sweeper, so a regression here stops trace collection silently instead of failing a test. Can we assert `collectable_traces.notified()` after a sweep that deleted rows, plus a paired test that a no-op sweep sends nothing?

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -275,31 +395,41 @@ impl Outbox {
...
-            resolved.push(partition_id);
+        for item in &items {
+            resolved.push(self.resolve_partition(queue, item.partition)?);
         }
 
         let runner = db.as_seaorm();
-        let ids = Self::insert_batch(&runner, self.statements(), &resolved, items).await?;
+        let trace_id = match trace {
+            Some(trace) => {
+                let entities = i64::try_from(items.len()).unwrap_or(i64::MAX);
```
MikeFalcon77, line 423: **LOW**

A failed `i64::try_from(items.len())` substitutes `i64::MAX` as the entity count, which writes a trace row whose `pending` can never count down to zero, so the subscriber waits forever and nothing is reported. Propagate instead, `i64::try_from(items.len()).map_err(|_| OutboxError::...)?`, so an unrepresentable batch size is rejected before the trace row goes in.

libs/toolkit-db/src/outbox/types.rs
```diff
@@ -97,14 +122,23 @@ pub enum OutboxError {
         found: usize,
     },
 
-    #[error("invalid queue name: '{0}'")]
-    InvalidQueueName(String),
+    #[error("a traced batch must carry at least one entity")]
+    EmptyTracedBatch,
+
+    #[error("invalid queue name: {reason}")]
+    InvalidQueueName { reason: &'static str },
```
MikeFalcon77, line 129: **LOW**

`InvalidQueueName`, `InvalidPayloadType` and `InvalidTablePrefix` all went from `(String)` to `{ reason: &'static str }`, dropping the value that was rejected. All three carry gear-supplied config identifiers rather than end-user data, so the log-hygiene reasoning in the `validation.rs` module doc doesn't apply here, and a caller with six queues like mini-chat gets a rejection it can't attribute to a field. Keep an owned identifier next to the reason for these three and apply the value-stripping only to `InvalidTrace`, where the value really is caller-supplied.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+            } else {
+                *current = Some(progress);
+                true
+            }
+        })
+    }
+
+    /// Withdraw the stall from every watched trace outside `still_stalled`,
+    /// because those batches are moving again.
+    pub fn withdraw_progress_except(&self, still_stalled: &HashSet<String>) {
```
MikeFalcon77, line 226: **LOW**

`withdraw_progress_except` walks the whole registry and runs `send_if_modified`, a watch write lock plus waker notifications, for each entry while holding `entries`, which is the same lock `deliver` takes on every completed batch. Collect the `Arc<watch::Sender>` handles into a `Vec` under the lock, drop the guard, then publish outside the critical section.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+/// callers a borrow guard that blocks the sender if it is held across an await.
+/// [`TraceProgressWatch::current`] returns by value instead.
+pub struct TraceProgressWatch(watch::Receiver<Option<TraceProgress>>);
+
+impl TraceProgressWatch {
+    /// Wait until the stall state changes.
+    ///
+    /// Returns `false` once nothing can change it again - the batch finished,
+    /// or the subscription was dropped.
+    pub async fn changed(&mut self) -> bool {
```
MikeFalcon77, line 285: **LOW**

`changed` is a new public async fn with no cancel-safety note, and its documented use is a loop that will typically sit in a `tokio::select!` arm. It wraps `watch::Receiver::changed`, so it is cancel-safe; say so. Same for `TraceSubscription`: dropping it mid-await discards a pending completion, and that belongs on the `Future` impl.

MikeFalcon77, line 285: **LOW**

`changed` returns a bare `bool` that encodes "the channel is closed" and carries no `#[must_use]`, so `stalls.changed().await;` compiles and turns the documented `while` loop into a silent spin or an early exit. This is the same sharp edge behind coderabbitai's README finding. `#[must_use]` plus a two-variant enum (`WatchState::Changed`/`WatchState::Closed`) would prevent it.

libs/toolkit-db/src/outbox/subscription.rs
MikeFalcon77, line 375: **LOW**

Polling after the subscription resolved returns `Poll::Ready(None)`, which is also documented to mean this process can no longer answer and the caller should read the trace row. That makes a double-await bug indistinguishable from a legitimate shutdown. A dedicated enum (`Completed` / `Unanswerable` / `AlreadyTaken`) would separate them and make the re-poll loggable.

libs/toolkit-db/src/outbox/workers/notifier.rs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            Ok(rows) => rows,
+            Err(e) => {
+                warn!(error = %e, "notifier: failed to read this instance's mail");
+                return Err(e.into());
+            }
+        };
+
+        let mut delivered = 0;
+        for row in rows {
```
MikeFalcon77, line 80: **LOW**

This per-row claim loop never checks `_cancel` and does one round trip per mail row, so a cancelled shutdown has to wait out a full `batch_size` page of claims before the worker loop sees the cancellation. Break on `cancel.is_cancelled()`, take the token by name rather than `_cancel`, and leave the unclaimed rows for the next start.

libs/toolkit-db/src/outbox/strategy.rs
```diff
@@ -122,46 +137,191 @@ async fn read_messages(
...
+///
+/// Mail with nobody waiting was stamped delivered by the claim and is
+/// discarded here rather than retained: the guard was dropped, or this is not
+/// the process that asked.
+fn deliver_claimed(
+    mailbox: &super::subscription::Mailbox,
+    claimed: Vec<super::trace::TraceOutcome>,
+) {
+    for outcome in claimed {
+        let trace = outcome.trace.clone();
```
MikeFalcon77, line 228: **LOW**

`deliver` takes `TraceOutcome` by value and returns a bare `bool`, so on every early return (no entry, sender already taken) the outcome is just dropped. That forces callers to clone `outcome.trace` up front, in the ack hot path here and again at `workers/notifier.rs:91`, only so a `debug!` on the nobody-waiting branch has something to log. Hand the outcome back on failure (`Result<(), TraceOutcome>` or `Option<TraceOutcome>`) and log from that instead.

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -230,43 +356,37 @@ impl Outbox {
...
+    /// Returns an error if the queue is not registered, the partition is out of
+    /// range, or the database rejects the write.
     pub async fn enqueue(
         &self,
         db: &(impl crate::secure::DBRunner + Sync + ?Sized),
-        queue: &str,
-        partition: u32,
-        payload: Vec<u8>,
-        payload_type: &str,
+        msg: Record<'_>,
```
MikeFalcon77, line 373: **LOW**

`enqueue` and `enqueue_batch` take ownership of the `Record`/`Records` including the payload `Vec<u8>` but return a bare `OutboxError`, so a caller retrying a transient `Database` error has to rebuild or pre-clone the whole payload. Returning the consumed submission inside the error, something like `EnqueueError { source: OutboxError, record: Record<'a> }`, lets a retry reuse it.

libs/toolkit-utils/src/byte_size.rs
```diff
@@ -0,0 +1,44 @@
...
+//! the standard library offers no constant or unit type to write it any other
+//! way. These do, and they compose: `96 * MIB`, `2 * GIB`.
+//!
+//! Two families, differing only in type. A limit compared against a database
+//! counter is `i64`; a length compared against `Vec::len` is `usize`. Casting
+//! between them in a `const` buys nothing but a lint, so both exist and each
+//! is named for what it measures.
+
+/// One kibibyte, for values compared against database counters.
+pub const KIB: i64 = 1024;
```
MikeFalcon77, line 13: **LOW**

Two parallel constant families that differ only in primitive type (`MIB: i64` against `MIB_LEN: usize`) mean `64 * MIB` and `64 * KIB_LEN` are equally valid at any call site, so picking the wrong family is silent rather than a compile error. A single `ByteSize` newtype with `const fn kib/mib/gib` constructors and `as_i64()`/`as_usize()` accessors avoids duplicating the constants per primitive.

libs/toolkit-db/src/outbox/workers/vacuum.rs
```diff
@@ -44,10 +44,18 @@ pub struct VacuumTask {
...
+    collectable_traces: Arc<tokio::sync::Notify>,
 }
 
 impl VacuumTask {
-    pub fn new(db: Db, statements: Arc<OutboxStatements>, batch_size: usize) -> Self {
+    pub fn new(
+        db: Db,
+        statements: Arc<OutboxStatements>,
+        batch_size: usize,
+        collectable_traces: Arc<tokio::sync::Notify>,
```
MikeFalcon77, line 57: **LOW**

The collectable-traces wakeup is threaded as a bare `Arc<Notify>`, structurally identical to the start notify, the per-partition notifies and the registry's arrivals notify, so transposing any two of them at a `VacuumTask::new` or `.notifier(..)` call site compiles and mis-wires the pipeline silently. A newtype per channel, `struct CollectableTraces(Arc<Notify>)`, would make the compiler reject that.

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -683,27 +818,74 @@ mod tests {
...
+    #[test]
+    fn a_batch_carries_its_own_trace_and_totals() {
+        let batch = Records::to("orders")
+            .payload_type("application/json")
+            .trace("import-2026-09-08")
+            .push(0, vec![1, 2, 3])
+            .push_with_type(1, vec![4, 5], "application/vnd.legacy+json")
+            .build()
+            .unwrap();
+        assert_eq!(batch.queue(), "orders");
```
MikeFalcon77, line 884: **LOW**

`a_batch_carries_its_own_trace_and_totals` mostly reads back what it just passed to the builder (`queue()`, `len()`, `is_empty()`) and never checks the trace, which is the thing the name promises. Drop the echoed assertions and assert the trace is carried through to `into_parts` and the enqueued trace row.

libs/toolkit-utils/src/byte_size.rs
```diff
@@ -0,0 +1,44 @@
...
+pub const GIB_LEN: usize = MIB_LEN * 1024;
+
+#[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn the_two_families_agree() {
+        assert_eq!(KIB, 1024);
```
MikeFalcon77, line 37: **LOW**

`assert_eq!(KIB, 1024)`, `assert_eq!(MIB, 1_048_576)` and `assert_eq!(GIB, 1_073_741_824)` restate the constants' own literal definitions, which the compiler already guarantees. Keep the cross-family assertions that the `usize` constants equal their `i64` counterparts and drop these.

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -78,6 +100,110 @@ impl Outbox {
...
+    pub fn subscribe(&self, trace: &str) -> TraceSubscription {
+        self.mailbox.subscriptions().subscribe(trace)
+    }
+
+    /// How many traced batches this instance is currently waiting on.
+    ///
+    /// Zero means the notifier issues no query at all, so this is also the
+    /// answer to "is this instance polling for mail right now".
+    #[must_use]
+    pub fn outstanding_traces(&self) -> usize {
```
MikeFalcon77, line 129: **LOW**

`outstanding_traces` has no test, no caller, no example and no bench; a repo-wide search returns only the definition. It ships as public surface in a breaking release with nothing exercising it, while `Registry::len()` underneath it is tested. Either add a test (subscribe to a trace, assert 1, resolve or drop it, assert 0) or drop the method until something needs it.

libs/toolkit-db/src/outbox/record.rs
```diff
@@ -0,0 +1,288 @@
...
+impl<'a> Record<'a> {
+    /// Name the queue and partition this entity goes to.
+    #[must_use]
+    pub const fn to(queue: &'a str, partition: u32) -> RecordTarget<'a> {
+        RecordTarget { queue, partition }
+    }
+
+    /// The queue this entity goes to.
+    #[must_use]
+    pub const fn queue(&self) -> &'a str {
```
MikeFalcon77, line 43: **LOW**

The only `.queue()`/`.bytes()` calls in the outbox module are `batch.queue()` and `batch.bytes()` at `core.rs:884-886`, and those resolve to the `Records` accessors at `record.rs:152` and `:174`. These single-`Record` versions have no caller in the crate, its tests, its examples or its benches. Either assert them in the `Record` builder test the way `a_batch_carries_its_own_trace_and_totals` does for `Records`, or drop them.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+use std::task::{Context, Poll};
+
+use tokio::sync::{oneshot, watch};
+
+use super::trace::{TraceOutcome, TraceProgress};
+
+/// What an ack needs to deliver a completion in-process: who this instance is,
+/// and who is waiting.
+#[derive(Debug)]
+pub struct Mailbox {
```
MikeFalcon77, line 23: **LOW**

`mod subscription;` at `mod.rs:209` is private and only `TraceProgressWatch` and `TraceSubscription` are re-exported, so `Mailbox`, `Registry` and their ~15 `pub` methods are unreachable from outside the crate while advertising themselves as public API. `unreachable_pub` isn't in `[workspace.lints.rust]`, so nothing flags it automatically. `pub(crate)` would match the real visibility.

libs/toolkit-db/src/outbox/statements.rs
```diff
@@ -441,6 +575,232 @@ fn allocate_sequences(dialect: Dialect, tables: &OutboxTables) -> AllocSql {
...
+                 stalled_since = NULL, \
+                 last_error = NULL, \
+                 completed_at = CASE WHEN pending - ? <= 0 THEN {now} ELSE completed_at END, \
+                 notified_at = CASE WHEN pending - ? <= 0 \
+                                     AND notified_at IS NULL \
+                                     AND owner_instance = ? \
+                                THEN {now} ELSE notified_at END \
+             WHERE id = ? AND pending > 0",
+            tables.trace()
+        ),
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. MySQL evaluates SET left to right, so after `pending = pending - ?` the completion CASEs already see the post-ack value - they now test `pending <= 0` instead of re-subtracting the delta. Added a statement-shape test and a container test that drives a half-acked batch on real MySQL and asserts it isn't reported complete while half-done.

libs/toolkit-db/src/outbox/strategy.rs
striped-zebra-dev: Covered by a_lost_lease_before_ack_delivers_no_completion: subscribe, force the lease-lost rollback, assert the ack returns None and the subscriber stays waiting with the row untouched. The guard keys on locked_by, not locked_until, so the test steals locked_by rather than using expire_lease.

libs/toolkit-db/src/outbox/store.rs
striped-zebra-dev: Fixed - it was a real bug, not just untested. Cross-instance (owner != acker), trace_advance left notified_at NULL and MySQL's exec_claim_trace_mail only ran the notified_at IS NOT NULL SELECT, never the stamping UPDATE, so the owner's notifier never delivered. Now MySQL runs the claim_mail UPDATE with a rows_affected guard then reads the outcome, like PG's UPDATE ... RETURNING, and the redundant notified_at CASE is gone from the advance so the claim is the sole stamp. mysql_cross_instance_completion_reaches_the_submitter went red to green.

libs/toolkit-db/src/outbox/store.rs
```diff
@@ -260,22 +269,239 @@ impl<'a> OutboxStore<'a> {
...
+        // only the ack that drove `pending` to zero can see a row, because
+        // every later advance is refused by `pending > 0`.
+        let row = conn
+            .query_one_raw(Statement::from_sql_and_values(
+                self.backend(),
+                self.statements.trace().claim_mail_outcome(),
+                [trace_id.into(), instance_id.into()],
+            ))
+            .await?;
+        row.as_ref().map(Self::outcome_from_row).transpose()
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed, same MySQL claim path: exec_claim_trace_mail now runs the stamping claim_mail UPDATE (guarded, rows_affected checked) before reading the outcome, instead of the SELECT-only path. Covered by the cross-instance container test.

libs/toolkit-db/src/outbox/manager.rs
striped-zebra-dev: Added impl Drop for OutboxHandle that closes the registry, so dropping the handle resolves waiters to None like stop() does. tasks is now Option<TaskSet> so stop() can still consume it. Test asserts a subscription survives a handle drop - verified it hangs without the Drop.

striped-zebra-dev: Covered now: dropping_the_handle_resolves_a_waiting_subscription holds a live TraceSubscription across a handle drop and asserts it resolves to None - the end-to-end path, not just Registry::close in isolation.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: By design. Trace-id uniqueness is the caller's contract - the outbox never resolves collisions (no unique index, no check). Two live batches sharing a trace resolving each other's waiter is accepted; documented on the .trace() builders, subscribe, and the README.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Same as the sibling: by design. A reused trace string is the caller's to avoid; the outbox doesn't enforce uniqueness or resolve collisions.

libs/toolkit-db/src/outbox/core.rs
striped-zebra-dev: By design - the outbox is an internal library and doesn't do ownership isolation. trace_status is an unscoped lookup by trace string; the payload isn't treated as a cross-instance secret.

libs/toolkit-db/src/outbox/migrations.rs
striped-zebra-dev: By design. Greenfield - one m001 create migration, no second migration or ALTER path.

libs/toolkit-db/src/outbox/strategy.rs
striped-zebra-dev: Done. trace_progress/apply_trace_progress now use a named TraceCountdown { trace, terminal, failures } instead of the bare tuple.

libs/toolkit-db/src/outbox/core.rs
```diff
@@ -230,43 +356,37 @@ impl Outbox {
...
-        super::validation::validate_queue_name(queue)?;
-        super::validation::validate_payload_type(payload_type)?;
-        Self::validate_payload(&payload)?;
-        let partition_id = self.resolve_partition(queue, partition)?;
+        let (queue, item, trace) = msg.into_parts();
+        let partition_id = self.resolve_partition(queue, item.partition)?;
 
         let runner = db.as_seaorm();
+        let trace_id = match trace {
+            Some(trace) => Some(self.insert_trace_row(&runner, trace, queue, 1).await?),
```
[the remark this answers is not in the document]
striped-zebra-dev: By design. Atomicity is the caller's: enqueue inside your business transaction (DbTx) and the trace and body inserts commit or roll back together. A DbConn opts out of that boundary by choice; the outbox does not wrap it. A partial-failure orphan trace is self-healing - the trace sweeper collects it.

libs/toolkit-utils/src/byte_size.rs
striped-zebra-dev: By design. The two families measure different things - byte magnitudes (i64) vs buffer/length sizes (usize) - and the module doc says so. Collapsing them would force casts at the usize call sites.

libs/toolkit-db/src/outbox/statements.rs
```diff
@@ -441,6 +575,231 @@ fn allocate_sequences(dialect: Dialect, tables: &OutboxTables) -> AllocSql {
...
+            "UPDATE {} \
+             SET pending = pending - ?, \
+                 failures = failures + ?, \
+                 attempts = CASE WHEN pending <= 0 THEN attempts ELSE 0 END, \
+                 stalled_since = NULL, \
+                 last_error = NULL, \
+                 completed_at = CASE WHEN pending <= 0 THEN {now} ELSE completed_at END \
+             WHERE trace = ? AND pending > 0",
+            tables.trace()
+        ),
```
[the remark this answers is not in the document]
striped-zebra-dev: By design. Trace uniqueness is the caller's contract - the outbox never enforces it (no unique constraint, non-unique idx_trace_key by intent). Two live batches reusing a trace corrupting each other's countdown is the documented, accepted consequence; the body carries only the trace string on purpose.

libs/toolkit-db/src/outbox/types.rs
striped-zebra-dev: Added #[non_exhaustive] to OutboxError.

libs/toolkit-db/src/outbox/types.rs
striped-zebra-dev: Added #[non_exhaustive] to OutboxProfile.

libs/toolkit-db/src/outbox/trace.rs
striped-zebra-dev: Added #[non_exhaustive] to all three. Counts stay i64 - the columns are BIGINT and the values are DB-sourced non-negatives; unsigned would just fight sea-orm's mapping.

libs/toolkit-db/src/outbox/strategy.rs
striped-zebra-dev: Now an exhaustive match on TraceAdvance, so a later variant fails to compile instead of silently falling through to the claim.

libs/toolkit-db/src/outbox/mod.rs
striped-zebra-dev: Dropped TraceAdvance from the public re-export; it stays internal to the ack path.

libs/toolkit-db/src/outbox/migrations.rs
striped-zebra-dev: Added idx_trace_stalled to the asserted index list.

libs/toolkit-db/src/outbox/README.md
```diff
@@ -161,6 +173,138 @@ handle.stop().await;
...
+Waiting on a subscription tells you nothing until the batch is done, which is
+no help when it is stuck rather than slow. Ask the subscription for a progress
+receiver and you are told while it is in flight:
+
+```rust
+let waiting = outbox.subscribe("import-2026-09-08");
+let mut stalls = waiting.progress();
+tokio::spawn(async move {
+    while stalls.changed().await.is_ok() {
+        match stalls.borrow_and_update().clone() {
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed the snippet to the real API: while stalls.changed().await { ... stalls.current() }.

libs/toolkit-db/tests/outbox_trace_completion_native.rs
```diff
@@ -0,0 +1,338 @@
...
+//
+// On Postgres/SQLite the owner's notifier claim is an `UPDATE ... RETURNING`
+// that both stamps `notified_at` and returns the outcome, so a cross-instance
+// completion reaches A. On MySQL the ack path folds the delivery stamp into
+// the countdown, guarded on `owner_instance = <acking instance>`, so a
+// cross-instance ack never stamps it; and the notifier's non-RETURNING claim
+// runs only the outcome SELECT, guarded on `notified_at IS NOT NULL`, so it
+// never finds the row and never stamps it either. The completion is therefore
+// never delivered to A. This test drives that path against a real server: it
+// passes on Postgres and times out on MySQL.
```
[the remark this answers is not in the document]
striped-zebra-dev: Updated the comment. The MySQL notifier claim now runs the stamping claim_mail UPDATE before the outcome read, so it delivers on every backend and the test passes everywhere.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Documented: changed wraps watch::Receiver::changed, so it is cancel-safe - dropping the future mid-await loses no notification.

striped-zebra-dev: Added #[must_use] on changed so a loop that drops the bool warns.

libs/toolkit-db/src/outbox/workers/notifier.rs
striped-zebra-dev: The claim loop now breaks on cancel.is_cancelled(); the token is threaded through collect().

libs/toolkit-db/src/outbox/core.rs
striped-zebra-dev: Test now asserts the trace via into_parts(), not just the echoed queue/len/bytes.

libs/toolkit-utils/src/byte_size.rs
striped-zebra-dev: Dropped the tautological asserts; kept the cross-family usize checks that actually prove something.

libs/toolkit-db/src/outbox/subscription.rs
```diff
@@ -0,0 +1,629 @@
...
+        {
+            entries.insert(
+                trace.to_owned(),
+                Interest {
+                    generation,
+                    outcome: Some(tx),
+                    progress: Arc::clone(&progress),
+                },
+            );
+        }
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. subscribe now checks `closed` under the entries lock, and close() sets closed + drains under that same lock, so the mutex serializes them - a subscribe can no longer insert into a map close() already cleared.

libs/toolkit-db/src/outbox/core.rs
striped-zebra-dev: Fixed. Propagates OutboxError::TracedBatchTooLarge instead of clamping to i64::MAX, so an unrepresentable entity count is rejected before the trace row is written.

libs/toolkit-db/src/outbox/store.rs
striped-zebra-dev: Fixed. The non-RETURNING claim now rolls the txn back explicitly on error rather than leaving it to Drop; commit only on the Ok path.

libs/toolkit-db/src/outbox/workers/notifier.rs
striped-zebra-dev: Documented. collect is not cancel-safe at the claim boundary - the delivered stamp commits before the in-proc deliver - but a dropped completion is recoverable via trace_status, which the doc now points at. The per-row cancel check only avoids starting new claims on shutdown.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Fixed. Every entries-lock site now goes through one helper that recovers the poisoned guard (into_inner) and logs, so poison degrades gracefully and is observable instead of silently turning to no-op.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Fixed. Outbox::subscribe runs validate_trace and returns Result, so an empty/over-long/non-printable trace is rejected instead of registering an undeliverable interest. Unit test added.

gears/mini-chat/mini-chat/src/infra/outbox.rs
striped-zebra-dev: Fixed. PayloadTooLarge maps to DomainError::validation (4xx); other build errors stay internal. Queue and payload type are gear constants here, so that is the only caller-reachable rejection. Test added.

libs/toolkit-db/src/outbox/mod.rs
striped-zebra-dev: Greenfield - nothing depends on the old name yet, so the removal ships as a plain break with no deprecation shim.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Fixed. Mailbox::new takes InstanceId now, and core.rs passes config.instance_id directly instead of round-tripping through a String.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Moot after the redesign. TraceSubscription holds a watch::Receiver, not an Option<oneshot>, and completion() consumes self, so a second poll cannot panic. arrivals() is fired by subscribe and exercised by the new watcher tests.

striped-zebra-dev: Fixed by the redesign. completion() consumes self and the watch distinguishes a closed channel from a value, so a double-await can no longer be confused with a genuine None.

libs/toolkit-db/src/outbox/integration_tests.rs
striped-zebra-dev: Fixed. Rewritten as a_retrying_batch_is_reported_and_completes_cleanly - it drains state changes to completion rather than catching a single tick, so it no longer rides the handler's scheduling.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: By design for this crate. It lints redundant_pub_crate, which requires pub inside a private module, so unreachable_pub would fight it - pub-in-private-module is the deliberate house style. mod subscription stays private, so nothing here is reachable outside the crate.

libs/toolkit-db/src/outbox/migrations.rs
```diff
@@ -211,6 +213,137 @@ async fn create_index_if_absent(
...
+    create_index_if_absent(
+        conn,
+        backend,
+        &IndexSpec {
+            unique: false,
+            name: tables.idx_trace_stalled(),
+            table: tables.trace(),
+            columns: "owner_instance",
+            filter: stalled_filter,
+        },
```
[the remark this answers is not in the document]
striped-zebra-dev: Greenfield - no SQL/schema churn for an index tweak. The single create migration stands as-is; the retry poll is fine at this scale.

libs/toolkit-db/src/outbox/strategy.rs
striped-zebra-dev: Fixed. Added unit tests for trace_progress (cursor counting, the deadlock-avoidance sort, untraced batch), traces_beyond, and rejected_seqs.

libs/toolkit-db/src/outbox/integration_tests.rs
striped-zebra-dev: Fixed. The second pass now captures notified_at before and after and asserts it does not move, proving at-most-once delivery.

libs/toolkit-db/src/outbox/integration_tests.rs
striped-zebra-dev: Fixed. Tracks whether a further attempt was actually observed and asserts it after the loop, so the retrying_since-stability check can't pass vacuously.

libs/toolkit-db/src/outbox/workers/notifier.rs
striped-zebra-dev: Fixed. New test the_notifier_with_no_subscriptions_issues_no_query: with an empty registry, execute() issues no mail query (asserted via the statement recorder).

libs/toolkit-db/src/outbox/workers/vacuum.rs
striped-zebra-dev: Fixed. The vacuum test now holds the collectable notify and asserts it fires after a sweep that deleted rows.

libs/toolkit-db/src/outbox/integration_tests.rs
striped-zebra-dev: Fixed. New e2e test a_single_traced_record_delivers_its_completion: Record::to().trace() + Outbox::enqueue writes a one-entity trace and delivers its completion.

libs/toolkit-db/src/outbox/manager.rs
striped-zebra-dev: Fixed. Notifier/retry-reporter/trace-sweeper now return a u64 count in their Directive (completions delivered / retries reported / traces collected) and are wired through register_stats like the vacuum, so their work lands in the stats registry.

libs/toolkit-db/benches/outbox_throughput.rs
```diff
@@ -1263,11 +1477,12 @@ fn run_profile(
                         handles.iter().map(|h| Arc::clone(h.outbox())).collect();
 
                     let start = Instant::now();
-                    produce(&outboxes, &db, &queue_prefix, &profile).await;
+                    produce(&outboxes, &db, &queue_prefix, &profile, &state).await;
                     wait_for_completion(&state, timeout).await;
                     total += start.elapsed();
 
                     verify_results(&state, &profile);
+                    verify_traces(&state, &profile, timeout).await;
```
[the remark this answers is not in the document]
striped-zebra-dev: Fixed. verify_traces (which awaits the completion-delivery tasks) now runs inside the timed interval, so the traced profiles measure the completion cost. Untraced profiles have no watchers, so their timing is unchanged.

libs/toolkit-db/src/outbox/core.rs
striped-zebra-dev: Won't fix - overcomplication. Handing the record back on error would reshape the return type of the two hottest public methods for a caller that would rarely rebuild in place; callers just propagate the error.

libs/toolkit-db/src/outbox/types.rs
striped-zebra-dev: Won't fix. reason already names the field and the rule; these are gear-supplied config identifiers, and we keep the error body free of the submitted value.

libs/toolkit-db/src/outbox/strategy.rs
striped-zebra-dev: Won't fix. The only cost is one outcome.trace.clone() per completion, read solely to log the rare nobody-waiting miss; handing the value back in Err (the mpsc::SendError shape) is heavier than the debug log it saves.

libs/toolkit-db/src/outbox/core.rs
striped-zebra-dev: Kept and tested. New unit test outstanding_traces_counts_live_subscriptions: subscribe -> 1, drop -> 0. It's the read for whether this instance has in-flight traced work (zero also means the notifier issues no query).

libs/toolkit-db/src/outbox/store.rs
striped-zebra-dev: Fixed. outcome_from_row now maps a #[derive(FromQueryResult)] OutcomeRow by column name, so a reordered SELECT/RETURNING list can't silently swap two same-typed columns.

libs/toolkit-db/src/outbox/record.rs
striped-zebra-dev: Fixed. Deleted the unused single-Record queue()/bytes() accessors; only the Records versions are used.

libs/toolkit-db/src/outbox/subscription.rs
striped-zebra-dev: Fixed. clear_retries_except now collects the senders under the lock, releases it, then sends - so the per-entry watch writes no longer run while holding the mutex deliver contends for.

libs/toolkit-db/src/outbox/workers/vacuum.rs
striped-zebra-dev: Fixed. Introduced a CollectableTraces newtype (with notify()/wakeup() methods) so it can't be transposed with the other Arc<Notify> channels at construction.

MikeFalcon77 (approved):

PR #4778, opened 2026-09-10
feat(cluster): add exact-watch cache capability with watchless CAS degradation

asmith987:
## Summary by CodeRabbit

* **New Features**
  * Added support for cache backends without exact-watch capabilities.
  * Distributed locks and leader election now use timer- or polling-based fallback behavior when watches are unavailable.
  * Cache capability reporting distinguishes exact-watch support from prefix-watch support.
  * Capability requirements can now validate exact-watch support.

* **Bug Fixes**
  * Corrected capability detection for Redis caches with watch mode disabled.
  * Unsupported watch requests now return a consistent error.

* **Documentation**
  * Clarified watch support requirements, capability validation, and fallback behavior.


gears/system/cluster/cluster-conformance/src/cache.rs
```diff
@@ -306,8 +306,12 @@ pub async fn scenario_cache_009(backend: Arc<dyn ClusterCacheBackend>) {
 }
 
 /// SC-CACHE-010: TTL expiry removes the entry and emits `CacheEvent::Expired` to
-/// watchers.
+/// watchers. Capability-gated on `features().watch`: a watchless backend has no
+/// watcher to emit to (the `watch()` contract itself is asserted by SC-CACHE-012).
 pub async fn scenario_cache_010(backend: Arc<dyn ClusterCacheBackend>, time: TimeControl) {
+    if !backend.features().watch {
+        return;
+    }
```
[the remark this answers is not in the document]
asmith987: Fixed in ac15bebff. Confirmed this was the only scenario covering finite-TTL *removal* (SC-CACHE-007 is explicit `delete`; SC-CACHE-011 asserts indefinite entries *survive*), so the early return did leave watchless backends untested on expiry. The put → elapse → `get("k").is_none()` removal check now runs for every backend; only the `watch()` subscription and the `Expired` assertion stay gated on `features().watch`. Expiry is observed lazily on read (fixture `get` and Redis both return absent once past the deadline), so the removal assertion holds after the elapse without depending on a background sweep. `cache_suite_passes_against_a_watchless_memcache` exercises the new path.

gears/system/cluster/docs/DESIGN.md
```diff
@@ -312,7 +312,7 @@ All three backend traits MUST be dyn-compatible. The SDK includes compile-time a
 | `CacheCapability` | `#[non_exhaustive] enum { Linearizable, PrefixWatch }`. Per-primitive requirement enum used at resolver call sites. |
```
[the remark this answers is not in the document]
asmith987: Fixed in ac15bebff. The `CacheCapability` enum row now reads `{ Linearizable, Watch, PrefixWatch }`, and I also added the missing `CacheCapability::Watch` mapping to the §3.13 resolution table (`descriptor.features.watch == true`), matching `cache/types.rs` and `resolver.rs`.

gears/system/cluster/cluster-sdk/src/dto.rs
```diff
@@ -120,22 +120,45 @@ impl From<WireCacheConsistency> for CacheConsistency {
...
+// addition is additive for out-of-crate constructors, not a breaking change.
+#[non_exhaustive]
 pub struct WireCacheFeatures {
     /// Whether the backend natively supports prefix watches.
     pub prefix_watch: bool,
+    /// Whether the backend supports exact-key watches. `None` from a peer that
+    /// predates this field (an older server), decoded as `true` — every backend
+    /// served exact watch before the field existed, so absence means "supported"
+    /// rather than the proto3 default of `false`.
+    pub watch: Option<bool>,
```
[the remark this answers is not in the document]
asmith987: Not changing this — the PR already provides exactly what's being asked for:

- `WireCacheFeatures` is now `#[non_exhaustive]` **and** keeps `#[derive(Default)]` + `From<CacheFeatures>`. That `From`/`Default` pair *is* the migration-compatible construction path; out-of-crate callers were never expected to use a struct literal, and the one such call site in this repo (`cluster/tests/descriptor_seams.rs`) was already migrated to `WireCacheFeatures::from(CacheFeatures::new(true))`.
- Re: "treat as a semver-breaking release" — adding a field to a previously-exhaustive public struct is a breaking change, and that's handled by `cargo-semver-checks`/release-plz on publish, which will force the appropriate 0.2.x → 0.3.0 bump automatically. It's not something to hand-encode in source.

So the residual ask (breaking-release versioning) is a release-tooling concern already covered, and the migration path exists. Leaving the code as-is.

gears/system/cluster/cluster-sdk/src/dto.rs
```diff
@@ -120,22 +120,45 @@ impl From<WireCacheConsistency> for CacheConsistency {
...
+        // Enforce the `without_watch()` invariant at the wire trust boundary — a
+        // backend that cannot watch one key cannot watch a family of them — so a
+        // skewed or hand-crafted peer that pairs `watch: Some(false)` with
+        // `prefix_watch: true` is normalized to the honest state rather than
+        // admitted as the impossible one the domain constructors forbid. A
+        // conforming peer never sends it (`CacheFeatures` is constructor-only),
+        // so this only ever narrows a malformed input.
+        Self {
+            watch,
+            prefix_watch: watch && value.prefix_watch,
```
MikeFalcon77: A descriptor carrying `watch: Some(false)` with `prefix_watch: true` gets silently rewritten to `prefix_watch: false` here, and the resolver then reports `PrefixWatch` unmet against a descriptor that literally declares it supported. The contradiction is never logged, so an operator chasing a skewed peer has no trace of the narrowing.

gears/system/cluster/cluster/src/defaults/lock.rs
```diff
@@ -61,6 +61,46 @@ const MAX_CONSECUTIVE_WATCH_RESETS: u32 = 8;
 /// still prevents CPU spin, but the unusable-watch cap may eventually fire).
 const WATCH_RESUBSCRIBE_BACKOFF: Duration = Duration::from_millis(50);
 
+/// The mean poll cadence a blocking acquisition falls back to when the cache
+/// backend serves no exact watch (`features().watch == false`, e.g. redis
+/// `watch_mode: disabled`): with no event to wake on, it re-attempts the claim
+/// around this interval (jittered by [`watchless_poll_interval`]) instead. Capped
+/// by the caller's remaining budget at the call site, so a tight `timeout` still
+/// fires on time (plan D3).
+const WATCHLESS_POLL_INTERVAL: Duration = Duration::from_millis(50);
```
MikeFalcon77, line 72: With no watch, every waiter re-runs `try_acquire` (a CAS write on the lock key) every 50-100ms for the whole wait, and the interval never grows. A lock held to its full TTL with a handful of waiters turns into thousands of writes on one key, which is exactly the deployment this path targets (redis `watch_mode: disabled`). Can the interval grow, capped at something like a second, instead of staying flat?

gears/system/cluster/cluster-sdk/src/cache/types.rs
```diff
@@ -118,15 +118,36 @@ impl CacheEvent {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[non_exhaustive]
 pub struct CacheFeatures {
+    /// Whether the backend supports exact-key watches at all. Nearly every
+    /// backend does; a backend answers `false` only when it has an operator mode
+    /// in which no watch can be served (e.g. redis `watch_mode: disabled`), in
+    /// which case [`watch`](crate::cache::ClusterCacheBackend::watch) must return
+    /// [`ClusterError::Unsupported`](crate::ClusterError::Unsupported) with
+    /// `feature: "watch"`.
+    pub watch: bool,
```
MikeFalcon77: `watch` and `prefix_watch` are both `pub` on a `Copy` type, so the `!watch => !prefix_watch` invariant that `without_watch()` and `From<WireCacheFeatures>` both go out of their way to enforce can still be broken by `f.watch = false;` on a value someone already holds. `#[non_exhaustive]` blocks literal construction but not field assignment. If that pairing is genuinely impossible, private fields with `watch()`/`prefix_watch()` accessors would make it so.

gears/system/cluster/cluster-sdk/src/dto.rs
```diff
@@ -120,22 +120,45 @@ impl From<WireCacheConsistency> for CacheConsistency {
     feature = "grpc-client",
     proto_bridge(stub = "crate::grpc::stubs::profile::WireCacheFeatures")
 )]
+// Matches the domain twin `CacheFeatures`: a wire mirror expected to grow fields
+// (this one just gained `watch`) must stay `#[non_exhaustive]` so the next
+// addition is additive for out-of-crate constructors, not a breaking change.
+#[non_exhaustive]
```
MikeFalcon77, line 126: Adding `#[non_exhaustive]` to an already-public `WireCacheFeatures` breaks every out-of-crate struct literal, which is why `descriptor_seams.rs` had to change in this PR. There is also no public constructor left: `From<CacheFeatures>` is the only way in, and it can never produce `watch: None`, so downstream code cannot build the old-peer wire shape the decoder is designed around.

gears/system/cluster/cluster/src/defaults/lock.rs
```diff
@@ -154,6 +198,23 @@ impl CasBasedDistributedLockBackend {
...
+    /// `Unsupported` is a *permanent* capability answer, not a retryable close,
+    /// so it is not surfaced as an error: the caller falls back to bounded
+    /// polling. Any other error propagates.
+    ///
+    /// # Errors
+    /// Returns a non-`Unsupported` [`ClusterError`] if the subscribe fails.
+    async fn resubscribe(&self, key: &str) -> Result<Option<CacheWatch>, ClusterError> {
+        match self.cache().watch(key).await {
+            Ok(watch) => Ok(Some(watch)),
+            Err(ClusterError::Unsupported { .. }) => Ok(None),
```
MikeFalcon77: `Unsupported { .. }` ignores `feature`, so an `Unsupported` raised about anything else also reads as "this backend has no exact watch" and silently turns a blocking acquisition into a poll loop. Matching `ClusterError::Unsupported { feature: "watch" }` keeps this to the case the trait contract actually describes.

gears/system/cluster/cluster/src/defaults/leader.rs
```diff
@@ -207,8 +211,18 @@ impl CasBasedLeaderElectionBackend {
...
+            // the watch establishment cannot be missed. A backend with no
+            // exact-watch support (`features().watch == false`) answers
+            // `Unsupported`; there is then no reactive feed, so reconcile off the
+            // renewal timer alone — the same state the task reaches after an
+            // ordinary end-of-stream (`None`). This keeps election working over
+            // e.g. redis `watch_mode: disabled` instead of failing at first
+            // `enrol` (plan D3).
+            let cache_watch = match self.leases.cache().watch(&key).await {
+                Ok(watch) => Some(watch),
+                Err(ClusterError::Unsupported { .. }) => None,
```
MikeFalcon77: This degradation is silent, while `resubscribe_watch` warns for the exact same outcome later in the task's life. If the cache reported `features().watch == true` at construction but answers `Unsupported` here, the constructor warning never fired and nothing tells the operator the election is now running timer-only.

gears/system/cluster/cluster/src/defaults/lock_tests.rs
```diff
@@ -158,6 +158,57 @@ async fn blocking_lock_waits_then_acquires_on_release() {
...
+        .lock("ledger", Duration::from_secs(30), Duration::from_secs(30))
+        .await
+    else {
+        panic!("uncontended blocking lock must acquire without a cache watch");
+    };
+    assert_eq!(guard.name(), "ledger");
+}
+
+#[tokio::test]
+async fn blocking_lock_polls_to_acquire_over_a_watchless_cache() {
```
MikeFalcon77, line 182: Both new watchless tests cover acquisition only. The watchless wait can only end unsuccessfully through `poll == wait` producing `WaitStep::Elapsed`, and nothing exercises it: can we add a watchless `lock()` that times out against a held lock (expecting `LockTimeout`), and one that steals after the incumbent's TTL lapses?

gears/system/cluster/cluster/src/defaults/leader_tests.rs
```diff
@@ -91,6 +91,27 @@ async fn single_candidate_becomes_leader() {
     assert!(watch.is_leader());
 }
 
+#[tokio::test]
+async fn elects_leader_over_a_watchless_cache() {
```
MikeFalcon77, line 95: This only checks that `elect()` stops propagating `Unsupported`. The behaviour the change claims, reconciling off the renewal timer with no reactive feed, is not covered: a second candidate over a watchless cache should still take over after the incumbent's lease lapses, which is the watchless twin of `foreign_takeover_emits_lost_then_resolves`.

gears/system/cluster/cluster-sdk/src/dto_tests.rs
```diff
@@ -68,6 +74,36 @@ fn feature_flags_round_trip_through_their_mirrors() {
...
+        prefix_watch: false,
+        watch: None,
+    };
+    assert!(
+        CacheFeatures::from(absent).watch,
+        "an absent wire watch must decode as supported"
+    );
+
+    let disabled = WireCacheFeatures {
+        prefix_watch: false,
```
MikeFalcon77, line 92: `From<WireCacheFeatures>` also narrows `prefix_watch` to false when `watch` is `Some(false)`, and both cases here use `prefix_watch: false`, so that narrowing never runs. The resolver's `PrefixWatch` arm now reads the decoded value and depends on it, so a skewed peer (`watch: Some(false)`, `prefix_watch: true`) is worth an explicit case.

gears/system/cluster/plugins/redis-cluster-plugin/tests/conformance.rs
```diff
@@ -136,6 +136,46 @@ async fn cache_conformance() {
...
+/// The same `SC-CACHE-*` suite against a `watch_mode: disabled` cache, which
+/// declares `features().watch == false` and answers `Unsupported` from `watch`.
+///
+/// Proves a watchless Redis deployment passes conformance rather than panicking
+/// on `.expect("watch")`: the exact-watch scenarios (010/012/015) are
+/// capability-gated, and SC-CACHE-012 asserts the `Unsupported` contract. This is
+/// the plugin half of the exact-watch fix — the same declaration that lets an
+/// omitted `leader_election`/`lock` degrade instead of failing at first call.
+#[tokio::test]
+async fn cache_conformance_watch_mode_disabled() {
```
MikeFalcon77, line 148: This covers the cache suite in `watch_mode: disabled`, but the reported bug was a CAS lock/leader wired over such a cache failing at first use. There is no `run_lock_conformance`/`run_leader_conformance` run in that mode here, so the degradation path is only covered by the in-memory fixture, never against real redis.

gears/system/cluster/cluster/src/defaults/lock.rs
MikeFalcon77: `Unsupported` is turned into `Ok(None)` with nothing logged. The construction-time `warn_without_watch` only fires when `features().watch` is honest, so a backend that declares watch support and then answers `Unsupported` drops a blocking `lock()` into 50ms polling with no record of it anywhere.

gears/system/cluster/cluster-sdk/src/binding.rs
```diff
@@ -331,7 +331,9 @@ impl ClusterCacheBackend for UnboundCacheBackend {
     }
 
     fn features(&self) -> CacheFeatures {
-        CacheFeatures::new(false)
+        // An unbound stub can serve nothing, watch included: the weakest honest
+        // reading, matching `consistency()` above.
+        CacheFeatures::without_watch()
```
MikeFalcon77, line 336: `features()` now declares `watch: false`, but `watch()` further down returns `ProfileNotBound`, not the `Unsupported { feature: "watch" }` that the trait doc added in this PR requires from a backend declaring `watch == false`. The practical effect is that a CAS lock/leader over an unbound cache now emits the watchless warning at construction and then still fails with `ProfileNotBound`. Was keeping `CacheFeatures::new(false)` here the intent, the same way `scan_prefix` deliberately keeps `ProfileNotBound` over `Unsupported`?

gears/system/cluster/cluster/src/defaults/lock.rs
```diff
@@ -367,18 +431,41 @@ impl CasBasedDistributedLockBackend {
...
+            // jittered interval, still capped by `wait` so a tight `timeout`
+            // fires on time.
+            let step = tokio::select! {
                 // Graceful cluster shutdown: abandon the wait promptly with a
                 // terminal `Shutdown` (`cpt-cf-clst-fr-shutdown-revoke`). Held
                 // locks lapse via their deadline; this only resolves an in-flight
                 // wait.
                 () = shutdown.cancelled() => return Err(ClusterError::Shutdown),
-                waited = tokio::time::timeout(wait, watch.recv()) => waited,
+                step = async {
```
MikeFalcon77: The whole wait policy now sits in an inline async block inside `select!`: the watch branch, the poll branch, and the elapse classification in one arm. Pulling it into `async fn next_wait_step(watch: Option<&mut CacheWatch>, wait: Duration) -> WaitStep` would keep the select arms one-liners and make the two paths testable on their own.

gears/system/cluster/cluster-conformance/src/cache.rs
```diff
@@ -492,6 +522,11 @@ pub async fn scenario_cache_015(backend: Arc<dyn ClusterCacheBackend>, time: Tim
     const FIRST_EVENT_BUDGET: Duration = Duration::from_millis(500);
     const DUPLICATE_BUDGET: Duration = Duration::from_millis(250);
 
+    // Capability-gated on `features().watch` (see the doc): at-most-once delivery
+    // is a watch property, so a watchless backend does not apply.
+    if !backend.features().watch {
```
MikeFalcon77: On a watchless backend SC-CACHE-015 returns without asserting anything, so the suite reports it as passing when it ran nothing. SC-CACHE-012 already owns the `Unsupported` contract, so either surface 015 as skipped in the suite output or give the watchless branch a real assertion, the way 010 keeps its removal check.

gears/system/cluster/cluster-sdk/src/cache/resolver_tests.rs
```diff
@@ -177,6 +177,100 @@ async fn resolve_happy_path_returns_facade() {
...
+
+    let Ok(cache) = ClusterCacheV1::resolver(&hub)
+        .profile(OrdersProfile)
+        .require(CacheCapability::Watch)
+        .resolve()
+        .await
+    else {
+        panic!("a Watch requirement must resolve against a watchful descriptor");
+    };
+    assert!(cache.features().watch);
```
MikeFalcon77: `linearizable_backend()` is hardcoded to `CacheFeatures::new(true)`, so this assertion holds whatever the descriptor says, and the sibling test above makes the point that the backend is not what validation reads. The `let Ok(cache)` is what this test really checks; assert on something descriptor-derived or drop the line.

gears/system/cluster/cluster/src/defaults/lock.rs
```diff
@@ -61,6 +61,46 @@ const MAX_CONSECUTIVE_WATCH_RESETS: u32 = 8;
...
+/// A jittered watchless poll interval drawn uniformly from
+/// `[WATCHLESS_POLL_INTERVAL, 2 * WATCHLESS_POLL_INTERVAL)` — a floor of one full
+/// interval (so a poll never degenerates into a spin) plus up to one interval of
+/// jitter, so multiple waiters contending the same lock over a watchless cache
+/// desynchronize their claim attempts rather than retrying in lockstep and
+/// hammering the backend with synchronized CAS bursts. Mirrors the leader
+/// backend's `reclaim_jitter`; `ThreadRng` is entropy-seeded, so concurrent
+/// participants diverge with no explicit seeding. The caller clamps the result to
+/// its remaining `wait` budget.
+fn watchless_poll_interval() -> Duration {
```
MikeFalcon77: `watchless_poll_interval` has no test. The floor of one full interval is what keeps the watchless wait from degenerating into a spin, and a range check over a handful of draws (`>= WATCHLESS_POLL_INTERVAL` and `< 2 * WATCHLESS_POLL_INTERVAL`) is cheap to add.

gears/system/cluster/cluster/src/defaults/lock.rs
asmith987: Watchless poll now backs offs instead of a flat 50-100ms.

gears/system/cluster/cluster-sdk/src/cache/types.rs
asmith987: Made `CacheFeatures.watch` and `CacheFeatures.prefix_watch` private with by-value accessors (`watch()` / `prefix_watch()`).

gears/system/cluster/cluster-sdk/src/dto.rs
asmith987: Added a public `WireCacheFeatures::new(watch: Option<bool>, prefix_watch: bool)` so out-of-crate code can build the `watch: None` old-peer shape the decoder targets.

gears/system/cluster/cluster/src/defaults/lock.rs
asmith987: `resubscribe` now matches only `Unsupported { feature: "watch" }`, so any other `Unsupported` propagates as an error instead of silently degrading to polling.

gears/system/cluster/cluster/src/defaults/leader.rs
asmith987: `enrol` now matches only `feature: "watch"` and warns via the shared `guard::warn_runtime_watchless` when a cache that declared watch support answers `Unsupported` at runtime.

gears/system/cluster/cluster/src/defaults/lock_tests.rs
asmith987: Added two watchless tests: a `lock()` that times out against a held lock (`LockTimeout`) and one that steals the lease after the incumbent's TTL lapses.

gears/system/cluster/cluster/src/defaults/leader_tests.rs
asmith987: Added `a_follower_takes_over_after_a_lapse_over_a_watchless_cache`, exercising the timer-only takeover once the incumbent's lease lapses.

gears/system/cluster/cluster-sdk/src/dto_tests.rs
asmith987: Added an explicit skewed-peer case (`watch: Some(false)`, `prefix_watch: true`) asserting both fields decode to false.

gears/system/cluster/plugins/redis-cluster-plugin/tests/conformance.rs
asmith987: Added `cas_lock_conformance_watch_mode_disabled` (full SC-LOCK suite) and a focused `cas_leader_over_a_watchless_cache_elects_and_holds` test running the CAS lock/leader over a real `watch_mode: disabled` redis cache.

gears/system/cluster/cluster/src/defaults/lock.rs
asmith987: The runtime silent-degradation path now emits a one-time `warn` via the shared `guard::warn_runtime_watchless`, gated on the honest `features().watch()` bit.

gears/system/cluster/cluster-sdk/src/binding.rs
asmith987: Left the error as `ProfileNotBound` by design (an unbound facade reports it for every op, mirroring `scan_prefix`) and added a doc note explaining it is a "nothing is wired" signal, not a capability claim.

gears/system/cluster/cluster/src/defaults/lock.rs
asmith987: Extracted the wait policy into a free `next_wait_step(watch, poll_base, wait)` async fn, so the `select!` arm is a one-liner and both paths are unit-testable.

gears/system/cluster/cluster-conformance/src/cache.rs
asmith987: SC-CACHE-015's watchless branch now asserts the `Unsupported { feature: "watch" }` contract instead of silently returning, so a passing run always means an assertion ran.

gears/system/cluster/cluster-sdk/src/cache/resolver_tests.rs
asmith987: Dropped the tautological `assert!(cache.features().watch)` (it read the hardcoded backend, not the descriptor); the `let Ok(cache)` resolve is the real check, with the descriptor-vs-backend distinction covered by the negative sibling test.

gears/system/cluster/cluster/src/defaults/lock.rs
asmith987: Added unit tests for `watchless_poll_interval` (range/floor), `grow_watchless` (doubling + cap saturation), and `next_wait_step` tick-vs-elapse classification.

gears/system/cluster/cluster-sdk/src/dto.rs
asmith987: Added a log for this particular issue.

PR #4793, opened 2026-09-14
[WiP/Proposal] introduce GDL, the description language for gears

MikeFalcon77:
A gear today is defined by its Rust: the attribute says what it is at
runtime. What the code cannot say is what a gear is *as a component*
offered to someone assembling a system. That is what GDL adds, in two
vocabularies:

  gear.gdl     one gear as a component       (the gear author, here)
  product.gdl (or installation.gdl)  one installation from gears   (whoever composes a product)

Both are declarative. A file states facts; the resolver decides. There is
no `if`, no `for`, no `def`, no comprehension: each is refused with a
diagnostic and its own source span. Profile-specific behaviour is data, a
`profiles = ["dev", "prod"]` list on a declaration, not a branch. A
description whose output depended on how it was invoked would not be a
description.

Why a language and not more attributes: the facts split cleanly by
author. `#[toolkit::gear]` owns what the runtime needs - id,
capabilities, link-time deps, lifecycle. A descriptor owns what a
*chooser* needs - display name, category, visibility, the contracts the
gear intends to provide and consume, which Cargo features are worth
offering and where each belongs. Restating a projected fact is refused by
name (GBX0210), so neither half drifts into the other.

What it buys: from gear.gdl files plus one product.gdl, Gearbox resolves
an installation for a chosen deployment profile - which gears are in,
which applications they land in, which contract bindings stay local and
which cross an application boundary, which cluster scopes are satisfied
by which providers. It writes that as a lock and generates the
application crates from it. The descriptors are the input to all of it.

Signed-off-by: MikeFalcon77 <3686170+MikeFalcon77@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Added Gearbox metadata for numerous system, platform, example, and plugin gears, including packages, configuration, endpoints, dependencies, roles, and deployment features.
  * Composition now prevents multiple Gear Orchestrator instances in one installation.
  * Added cluster-profile support to the API contracts consumer example.

* **Documentation**
  * Added comprehensive Gearbox and GDL references, including syntax, diagnostics, profiles, feature selection, and author guidance.
  * Documented gear categories and single-directory installation requirements.

* **Chores**
  * Packaging validation now requires explicit library and binary target names.


AndrejK666:
A vendor mismatch in the corpus at `a0a42ce` (`feature/gearbox`) makes any product that uses `account-management` fail **GBX0512** in every profile.

- `account-management` selects vendor `constructorfabric` by default (`vendor_selector`).
- Its IdP plugins register under different vendors: `static-idp-plugin` = `cf` and `keycloak-idp-plugin` = `keycloak`, from their `vendor` config defaults.

It is the only host in the corpus where this happens. Every other host with a `vendor_selector` (`authn-resolver`, `authz-resolver`, `credstore`, `license-resolver`, `mini-chat`, `tenant-resolver`, `usage-collector`) selects `constructorfabric`, and each of them has at least one plugin registered under `constructorfabric`. So the consistent fix looks like giving `static-idp-plugin` the vendor `constructorfabric`, like every other `static-*` plugin. `keycloak-idp-plugin` can keep `keycloak` for products that select it on purpose.

Found while composing a product with Gearbox from Constructor Studio. As a workaround, a product can set `plugin("static-idp-plugin", config = {"vendor": "constructorfabric"})`; with that override the product resolves with 0 errors.

Issue #4625, opened 2026-08-20
[Types Registry P0] Phase 5 — batching, deletion, dry run, and dispatch


aviator5:
Parent: #4619 · Depends on: Phase 4 (#4624)

## Goal

Admit dependency-aware batches with partial success, support deletion and Dry Run, expose all
mutation routes on `/v2/`, and prove their complete submit → outbox → poll path before cutover.

P17 now splits former T27 into **T20a (mutations, this phase)** and **T22a (reads, Phase 6)**.
**T21 moves here from Phase 6.** Order: **T19 → T20 → T20a → T21 → Checkpoint 5**.
The seven-route surface becomes complete at Checkpoint 6, not Checkpoint 5.

## Tasks

- [x] **T19** — dependency-aware partial admission. Order candidates by `$ref`, derivation,
      conformance and implicit minor-predecessor edges; process one candidate per unit and
      return one outcome per identifier in submission order. Cycles, including mixed
      `$ref`/derivation and self-reference, refuse their members as `invalid_schema` without
      entity/revision/edge writes. Downstream candidates are blocked, while independent branches
      commit. Failed dependencies/predecessors yield `blocked_by_dependency` /
      `blocked_by_predecessor`; predecessor blocking takes precedence when both apply.
      Ordering is pure; implicit predecessor edges are not stored. Each blocked candidate is
      counted under its reason (P16).
- [x] **T20** — committed deletion and whole-batch Dry Run. Deletion first claims
      `entity_write_order`, then checks the positive `expected_resource_version`, locks and
      rechecks `ACTIVE` state and direct registered dependents, tombstones the entity and
      increments its version. Batch deletion orders dependents before targets; transitive-only
      dependents and `x-gts-ref` do not block. Tombstones remain exact-readable and are absent
      from lists. Dry run uses one coherent snapshot plus successful candidates' virtual
      changes; refused layers are discarded. **No entity-state writes or write-order claims
      reach storage**. Outcomes and operation completion publish atomically after the snapshot
      closes; this is no longer a per-candidate rollback-only protocol. Predicted `succeeded`
      omits `resource_version`, while `unchanged` reports the existing one. Dry-run mode enters
      the fingerprint and replay preserves the outcome fields. Metrics carry `dry_run` for
      candidate/refusal/compatibility counters and `kind` for candidate/refusal counters;
      dry runs do not observe activation-write-set histograms. Deletion reasons are named
[...]

PR #4796, opened 2026-09-14
Types Registry implementation P0-5

aviator5:
Adds T19 dependency-ordered registration batches and T20 deletion with reverse dependency ordering and dry-run support.

  Dry-run batches use a coherent in-memory simulation without entity-state writes. Includes backend, recovery, concurrency, and observability coverage, plus updated admission protocol
  documentation.

partial #4625

## Summary by CodeRabbit

* **New Features**
  * Dry-run requests now predict complete registration and deletion batches without changing entity state.
  * Batch processing respects dependencies and predecessor relationships, with per-item outcomes for blocked or cyclic candidates.
  * Active entities can be safely deleted with required version checks and dependent-entity protection.
* **Bug Fixes**
  * Dry-run outcomes align with equivalent committed executions and participate in idempotency handling.
* **Documentation**
  * Updated API and design documentation to describe current dry-run and deletion behavior.


gears/system/types-registry/types-registry/src/domain/admission/acceptance.rs
```diff
@@ -202,19 +204,9 @@ pub fn validate(
...
-    if request.dry_run {
-        // Dry Run needs a rollback-only evaluation transaction and a separate
-        // terminal-outcome write (T20). Letting it reach the ordinary creation
-        // worker makes `mark_item_succeeded` violate the dry-run result-column
-        // CHECK and strands the accepted operation in `running`.
-        return Err(AcceptanceError::DryRunNotAccepted);
-    }
+    // Registration and Deletion are the whole vocabulary; `OperationKind` is
+    // closed, so there is no third kind to refuse.
+    let deletion = request.kind == OperationKind::Deletion;
```
Artifizer, line 209: **[Medium] Deletion is now live-reachable with no owner/principal authorization (ceiling C6).**

Removing `UnsupportedOperationKind` / `DryRunNotAccepted` makes deletion reachable for any caller that reaches the submit route, including `cf.core.*` entities. The C6 ceiling is documented with rationale + owner (the `ponytail` notes in `deletion::commit_deletion` and below), so per the shared-Gears guardrail policy this is a documented deviation and merge-acceptable rather than blocking.

The one thing worth an explicit sign-off before merge: since deletion is now reachable, please confirm the **C8 "mutation routes are internal-only"** boundary still holds — that guarantee lives outside this diff and is effectively the only control standing in for authorization in P0.

aviator5: P0 retains the existing trusted, closed-environment deployment model of the in-memory registry, including for out-of-process deployment. Per-principal authorization is explicitly deferred to P1 (#4628); this PR does not weaken the existing trust boundary.

C8 is a deployment constraint: P0 mutation endpoints must remain accessible only within that trusted environment. `exposed = false` provides proxy-discovery filtering rather than complete isolation, so it should not be treated as the sole enforcement of that boundary.

Issue #4799, opened 2026-09-14
Dependabot covers only cargo: Docker base images and pinned GitHub Actions have no updater


lobster40:
## Problem

`.github/dependabot.yml` declares exactly one ecosystem:

```yaml
updates:
  - package-ecosystem: "cargo"
    directory: "/"
```

So crates are kept current and **nothing else is**. Two classes of versioned dependency have no
maintainer at all:

- **Container base images.** `FROM rust:…`, `FROM debian:…`, `FROM python:…` across four
  Dockerfiles.
- **GitHub Actions pins.** 22 SHA-pinned `uses:` references covering 23 distinct actions across the
  23 workflow files.

Pinning an action to a commit SHA is correct supply-chain practice and was deliberate (#2093,
StepSecurity hardening). But a pin only works as half of a pair: something has to move it forward.
Without the `github-actions` ecosystem, a pinned SHA is frozen permanently — including across
security releases of the action itself.

## Evidence that this is already costing us

**1. The Rust builder pin drifted two minor versions behind the toolchain.** Until #4798 both server
Dockerfiles sat on `rust:1.95.0-bookworm` while `rust-toolchain.toml` had moved to `1.97.0`, so
every image build silently rustup-installed a second toolchain. Nobody noticed because nobody owns
the pin.

**2. The same action is pinned to two different versions in the same repository:**

```
10 × uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
 3 × uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
```

Someone updated ten call sites by hand and missed three. This is precisely the failure mode a bot
does not have.

**3. Pinning was applied inconsistently.** `http-mock.Dockerfile` pinned `python` by digest while all
three `FROM debian:13.3-slim` lines had no digest at all (fixed in #4798).

**4. `.clusterfuzzlite/Dockerfile` has no tag at all:**

```dockerfile
FROM gcr.io/oss-fuzz-base/base-builder-rust
```

That is an implicit `:latest` — the fuzzing image is not reproducible from one run to the next.

**5. Stale version reference in a cache key.** `.github/workflows/fmt.yml` still uses
`shared-key: pr-fmt-stable-1.95`, left over from the 1.95 era.

## Proposal

Add the two missing ecosystems:

```yaml
  - package-ecosystem: "docker"
    directories:
      - "/testing/docker"
[...]

Issue #4255, opened 2026-07-23
[OoP-10] Gear & Platform-Host Docker Image Build & Publish Pipeline


mattgarmon:
**Parent:** [OoP](https://github.com/constructorfabric/gears-rust/issues/4102)

**Blocks:** [OoP-9](https://github.com/constructorfabric/gears-rust/issues/4111)
**Soft coordination:** [OoP-7](https://github.com/constructorfabric/gears-rust/issues/4109) (image tag/registry conventions must match chart `values.yaml` defaults)

### Description
No pipeline currently builds or publishes gear container images - [OoP-7](https://github.com/constructorfabric/gears-rust/issues/4109)'s Helm charts assume they already exist at `image.registry/image.repository:image.tag`. `mini-chat`'s Dockerfiles/Makefile targets are manual, dev-only, and single-gear.
 
Per "Platform Host Composition", we need two kinds of images: one **combined platform-host image** (trust-coupled core - `authz-resolver`, `tenant-resolver`, `resource-group`, `account-management` - plus system gears `gear-orchestrator`, `types-registry`, `credstore`, `api-gateway`, `grpc-hub`), and **minimal, separate images per OoP gear** (only the gear needed, not the full gear set).
 
Minimal per-gear images require prerequisite work: today's system gears aren't optional Cargo deps - `api_gateway`, `grpc_hub`, `gear_orchestrator`, `types_registry`, `tenant-resolver`, `authn-resolver`, `authz-resolver`, `resource_group`, `nodes_registry`, `simple_user_settings`, `credstore`, `api_egress` are all mandatory in cf-gears-example-server (only business gears like `mini-chat` are `optional = true` today). Since OoP dep resolution goes through DirectoryService + REST clients (not in-process linking), a minimal gear image should only need the target gear's crate + SDK/client crates for its deps, not full dep implementations.
 
PRD: `cpt-cf-fr-k8s-native` Design: `cpt-cf-component-k8s-packaging` ADR: `0004-cpt-cf-adr-helm-chart-organization`
Effort: Large
 
### Tasks
- [ ] 10.1 Feature-gate every remaining system gear crate as `optional = true` and `#[cfg(feature = ...)]`-gate registered_gears.rs accordingly (mirroring the existing `mini-chat`/`chat-engine` pattern).
[...]

Issue #4175, opened 2026-07-03
toolkit: env-var config overrides are silently dropped for gears with a dash in their name


cyberantonz:
## Problem

`#[toolkit::gear(name = "...")]` **requires kebab-case** gear names (snake_case is rejected). But per-deployment config overrides are read from env vars keyed by the gear name:

```
APP__gears__<gear-name>__config__<field>=...
```

For any gear name containing a dash (e.g. `analytics-api`), the env-var name contains a dash too — and such names are **silently dropped** before they ever reach the process:

- **docker compose**: a `docker-entrypoint.sh` running under `sh`/dash discards environment variables whose names are not valid shell identifiers.
- **k8s**: `envFrom` skips keys that are not C-identifiers (unless the cluster runs `RelaxedEnvironmentVariableValidation`).

The result is a nasty silent failure: every override (database URL, credentials, etc.) is stranded at its YAML default, the service dials `localhost`, and boot fails with e.g. `pool timed out` — with no hint that the env vars were eaten by the runtime.

## Real-world case

`constructorfabric/insight#1565` — the `analytics-api` gear could not receive `APP__gears__analytics-api__config__database_url` in either compose or k8s. Insight first worked around it with an underscore alias + a deep-merge fold shim in `main()`, then had to rename the gear (and eventually the whole service) to the dash-free `analytics`.

So today the only safe gear names are single words: the macro forces kebab-case, and the deploy runtimes forbid dashes — the intersection is "no multi-word gear names with env overrides".

## Suggested fixes (any of these would do)

1. **Fold underscores in the gear-key segment of the env provider**: accept `APP__gears__analytics_api__config__*` for gear `analytics-api` (map `_` → `-` when matching the gear key, since `__` is already the level separator). This keeps kebab names and makes overrides expressible as valid identifiers.
2. Alternatively, allow snake_case gear names in `#[toolkit::gear]`.
3. At minimum, document the limitation loudly (gear names with dashes cannot be configured via env vars under dash/sh entrypoints or k8s `envFrom`) and/or emit a startup warning when a `gears.<name-with-dash>` key exists.

Option 1 is the least disruptive: no existing gear renames needed, and the mapping is unambiguous as long as no two gears differ only by `-` vs `_`.

MikeFalcon77:
History of this bug, since #4115 and #4175 contradict each other and #4798/#4800
now cite both as a current limitation.

[...]

PR #4800, opened 2026-09-14
fix(docker): repair every build-from-source image and guard the pins

### LINKED ISSUE #4798: Docker: every build-from-source image is broken and make e2e-docker cannot run (no CI job builds images)
## Summary

Every Dockerfile in the repository that **builds the server from source inside the image** is broken,
and `make e2e-docker` cannot work under any circumstances. The root cause is that no CI job ever runs
`docker build`, so these files rotted undetected.

**Scope.** This is *not* "the platform cannot be containerised". Two paths are healthy and are the
pattern to follow:

| Path | Entry point | Status |
|---|---|---|
| `testing/docker/` e2e lane | `make e2e-docker` → `cf-gears.Dockerfile`, `http-mock.Dockerfile` | **broken** (A, G) |
| `testing/docker/cyberware.Dockerfile` | referenced by nothing | **broken + dead** (B–F) |
| mini-chat image, non-Linux host | `make mini-chat-docker` → `mini-chat.Dockerfile` | **broken** (B, C) |
| mini-chat image, Linux host | `make mini-chat-docker` → `mini-chat-prebuilt.Dockerfile` | works — builds on host, image only packages |
| mini-chat Helm chart | `make mini-chat-helm` | works, on top of the row above |
| standalone / OoP gears | `make oop-example`, `calculator-oop`, `cluster-oop`, file-storage `sidecar` | unaffected — has no Dockerfile at all (#4255, #4110) |

Verified on `ce2b86714` (2026-09-14).

## Defects

**A. `testing/docker/cf-gears.Dockerfile` does not exist.** Referenced by two live call sites —
`testing/docker/docker-compose.yml:19` and `tools/scripts/ci.py:335` — and absent from the working
tree and from all history (`git log --all -- testing/docker/cf-gears.Dockerfile` is empty).

**B. Both server Dockerfiles build a package that does not exist.**
`cyberware.Dockerfile:32,34` and `mini-chat.Dockerfile:43,45` run
`cargo build --release --bin cf-gears-server --package=cf-gears-server`.

```
$ cargo pkgid cf-gears-server
error: package ID specification `cf-gears-server` did not match any packages
```

The real names are `cf-gears-example-server` for both bin and package. `mini-chat.Dockerfile` then
contradicts itself at line 47 by copying `/build/target/$OUTPUT_DIR/cf-gears-example-server`.

**C. Both server Dockerfiles copy a path that does not exist.**
`COPY apps/gts-docs-validator ./apps/gts-docs-validator` (`cyberware.Dockerfile:21`,
`mini-chat.Dockerfile:22`). `apps/` contains only `cf-gears-example-server`.

**D. `cyberware.Dockerfile` copies an incomplete workspace.** It copies
`apps/ libs/ gears/ examples/ config/ proto/` but omits:
- `tools/` — yet `tools/gts-analyze` and `tools/xtask` are declared workspace members, so Cargo
  cannot load the workspace at all;
- `.cargo/config.toml` — which pins `PDFIUM_VERSION`. That pin is load-bearing: the file-parser build
  script otherwise downloads PDFIUM from `releases/latest`, and an upstream release already broke
  golden-markdown tests once (documented in that file);
- `Gears.toml`.

**E. `cyberware.Dockerfile` does not install cmake.** Its apt line installs only
`protobuf-compiler libprotobuf-dev`; `libz-ng-sys` needs cmake to build.
`mini-chat.Dockerfile:11` already does install it, so the requirement is known elsewhere in the repo.

**F. `cyberware.Dockerfile` CMD omits the `run` subcommand.**
`CMD ["/app/cf-gears-example-server", "--config", "/app/config/e2e-local.yaml"]` — the CLI requires
`run` | `check` | `migrate` to start serving.

**G. `http-mock.Dockerfile` escapes the build context.** `COPY ../e2e/testdata /app/testdata` with
compose's `context: ..` (i.e. `testing/`):

```
$ docker build -f testing/docker/http-mock.Dockerfile testing/
Step 3/9 : COPY ../e2e/testdata /app/testdata
COPY failed: forbidden path outside the build context: ../e2e/testdata
```

The mock is a `depends_on` of the API service, so this blocks compose startup independently of A.

**H. Stale env overrides in `docker-compose.yml`.** Lines 28–30 set
`APP__MODULES__api_gateway__CONFIG__*`. The config root was renamed `modules` → `gears`, so these are
no-ops; and per #4115 / #4175 an env override for a dash-named gear (`api-gateway`) cannot bind
anyway. `config/e2e-local.yaml` already sets `bind_addr: 0.0.0.0:8086`, `enable_docs: true` and
`cors_enabled: false`, so the overrides are also redundant.

**I. Process — the lane is unexercised dead code.** No workflow among the 23 files in
`.github/workflows/` runs `docker build`, `e2e-docker` or `mini-chat-docker`. `make e2e` still has
`e2e-docker` as its only prerequisite (`Makefile:987`), so the repo advertises a path that cannot run.

Prior art: this lane was last repaired in #2377 / #2090 (Jan 2026, `hyperspot.Dockerfile`) and has
regressed since the rename to gears-rust.

## Proposed fix

1. Turn the dead `cyberware.Dockerfile` into the `cf-gears.Dockerfile` that compose and `ci.py`
   already expect, built the way `mini-chat-prebuilt` proves works: `COPY . .` rather than
   enumerating workspace members — per-crate copying is exactly what rotted here. `.dockerignore`
   already excludes `target`, `.git`, `.venv`, `data`, `logs`.
2. Correct bin/package to `cf-gears-example-server`, add `cmake`, add the `run` subcommand.
3. Fix `http-mock.Dockerfile`'s COPY paths to be context-relative.
4. Fix `mini-chat.Dockerfile` (B, C).
5. Drop the stale compose env overrides (H).
6. Add a smoke CI job that runs `make e2e-docker-smoke`, otherwise this recurs (I).

## Acceptance

- `docker build -f testing/docker/cf-gears.Dockerfile .` succeeds.
- `make e2e-docker-smoke` goes green.
- A CI job exercises it on push.

lobster40:
Fixes #4798.

Every Dockerfile that builds the server from source inside the image referenced a package
(`cf-gears-server`) and a path (`apps/gts-docs-validator`) that do not exist, and the one file
`make e2e-docker` needs — `testing/docker/cf-gears.Dockerfile` — was missing from the tree and from
all history. Nothing in CI ran `docker build`, so none of it was noticed.

This does **not** touch the standalone/out-of-process path: it has no Dockerfile at all (#4255, #4110).

## Changes

**`testing/docker/cyberware.Dockerfile` → `testing/docker/cf-gears.Dockerfile`** (rename + rewrite).
`cyberware.Dockerfile` was referenced by nothing; `cf-gears.Dockerfile` is what
`docker-compose.yml:19` and `ci.py:335` already expect. The rewrite:

- builds `--bin cf-gears-example-server --package=cf-gears-example-server` (the package that exists);
- replaces the per-crate `COPY` list with `COPY . .`. Enumerating members is precisely what rotted:
  the old list omitted `tools/` (whose `gts-analyze` and `xtask` are declared workspace members, so
  Cargo could not load the workspace at all), `.cargo/config.toml` (which pins `PDFIUM_VERSION` —
  load-bearing) and `Gears.toml`. `.dockerignore` already excludes `target`, `.git`, `.github`,
  `.venv`, `data` and `logs`;
- installs `cmake` (needed by `libz-ng-sys`);
- appends the required `run` subcommand to `CMD`;
- adds `BUILD_PROFILE` (`release` default, `dev` opt-in).

**`testing/docker/http-mock.Dockerfile`** — `COPY ../e2e/...` escaped the build context, which
Docker rejects outright, so the mock could never build; and it is a `depends_on` of the API service,
so it blocked compose startup independently. Paths are now relative to the `testing/` context that
compose sets.

**`gears/mini-chat/deploy/docker/mini-chat.Dockerfile`** — same package-name and
`apps/gts-docs-validator` defects; also copies `tools/`, `.cargo/` and `Gears.toml`.

**`ENV HOME=/app` in all three server images.** Found by running the fixed image, and present in the
original files too. Every shipped config sets `server.home_dir: "~/.cf-gears"`, `useradd` runs
without `-m`, and a *numeric* `USER` does not update `HOME` — so it stayed `/root`, which uid 1000
cannot write:

```
Error: Failed to resolve server.home_dir
Caused by: 0: Failed to create home_dir
           1: Permission denied (os error 13)
```

The container exited immediately. This was invisible for mini-chat because the Helm chart never uses
the baked-in config (`helm/mini-chat/templates/configmap.yaml:11` overrides `home_dir: /app/data`),
so `make mini-chat-helm` worked while `docker run` of the same image did not. Details in
[this comment](https://github.com/constructorfabric/gears-rust/issues/4798#issuecomment-5663141059).

**`testing/docker/docker-compose.yml`** — dropped three `APP__MODULES__api_gateway__CONFIG__*`
overrides. They bound nothing: the config root was renamed `modules` → `gears`, so the `MODULES`
branch resolves to no key.

*Corrected after review by @MikeFalcon77 (see below), in c71d518.* An earlier version of this
description blamed the gear's dashed name, citing #4115 / #4175. That is wrong — **#4142** fixed
dashed gear names in June, and `APP__GEARS__API_GATEWAY__CONFIG__BIND_ADDR` binds correctly in HEAD
(`remap_gear_env_key`, `libs/toolkit/src/bootstrap/config/mod.rs:430`). It also said
`config/e2e-local.yaml` already sets all three values; it sets **two** — compose had
`CORS_ENABLED=true` against `cors_enabled: false` at `config/e2e-local.yaml:100`. Nothing changes by
deleting the dead lines, but anyone reviving them under the `GEARS` prefix would switch CORS on for
the e2e lane. That warning now lives in the file itself.

**Base-image pins.** All three `FROM debian:13.3-slim` lines were unpinned while
`http-mock.Dockerfile` pinned `python` by digest — now all are digest-pinned. Both builder stages
move to `rust:1.97.0-bookworm`, matching `rust-toolchain.toml`; they sat on `1.95.0`, so every build
rustup-installed a second toolchain.

**New `make docker-pins`** (`tools/scripts/ci.py docker-pins`) — pure stdlib text parsing, no Docker
daemon, no toolchain. Fails if a `FROM rust:<ver>` disagrees with `rust-toolchain.toml` or if any
`FROM` has lost its digest. Wired into `make check` and into an always-on CI job.
`.clusterfuzzlite/Dockerfile` is excluded by name — its base tracks the OSS-Fuzz contract (picked up
in #4799).

**`tools/scripts/ci.py`** — forwards `E2E_DOCKER_BUILD_PROFILE` as a `--build-arg`.

**`.github/workflows/e2e.yml`** — two new jobs:
- `docker-pins` — always on, ~seconds, catches pin drift on every PR;
- `e2e-docker` — runs `make e2e-docker-smoke`, scheduled/manual only, because a full in-image
  workspace build is too slow for every PR.

Without the second job this regresses again, which is the actual root cause of the whole issue.

## Verification

```
$ docker build -f testing/docker/http-mock.Dockerfile testing/
=> sha256:f6c43f06ee47…                         # was: COPY failed, forbidden path outside context

$ docker build -f testing/docker/cf-gears.Dockerfile \
    --build-arg BUILD_PROFILE=dev \
    --build-arg CARGO_FEATURES="$(cat config/e2e-features.txt)" .
=> Finished `dev` profile in 3m 32s             # was: package cf-gears-server did not match any

$ docker run -d -p 18186:8086 cf-gears-api:e2e-verify
$ curl -o /dev/null -w '%{http_code}' localhost:18186/healthz
200                                             # after 4s; was: exited, home_dir permission denied
$ curl -s localhost:18186/openapi.json | jq '.paths | length'
250
$ docker inspect <cid> --format '{{.State.Status}} restarts={{.RestartCount}}'
running restarts=0
```

`make docker-pins` was negative-tested — reverting the builder to `1.95.0` and dropping one debian
digest produces exactly:

```
testing/docker/cf-gears.Dockerfile:7:  rust 1.95.0 != rust-toolchain.toml 1.97.0
testing/docker/cf-gears.Dockerfile:45: base image is not digest-pinned: debian:13.3-slim
```

`make e2e-docker-smoke` passes end to end on this machine — image build, mock build, compose up,
health gate, `24 passed, 767 deselected` — see the comment below for the full output and for one
caveat about the size of the build layer.

## Follow-up

#4799 covers why these pins rotted in the first place — Dependabot is configured for `cargo` only,
so no Docker base image and no pinned GitHub Action has an updater.

## Summary by CodeRabbit

* **New Features**
  * Added a containerized example server for local end-to-end testing.
  * Added support for development and release build profiles in Docker builds.
  * Expanded Docker end-to-end testing to cover workspace configuration changes.
  * Added Docker image digest validation and Rust toolchain consistency checks.

* **Bug Fixes**
  * Improved container compatibility for non-root users.
  * Corrected Docker build paths and configuration handling.
  * Added Docker log capture and reliable cleanup when end-to-end tests fail.


lobster40:
## `make e2e-docker-smoke` passes

The verification left pending in the description is now done, locally on arm64 macOS:

```
== Building Docker image for E2E tests
> docker build -f testing/docker/cf-gears.Dockerfile -t cf-gears-api:e2e --build-arg CARGO_FEATURES=… --build-arg BUILD_PROFILE=dev .
   Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 33s

== Rebuilding docker-compose mock service
== Starting E2E docker-compose environment
== Waiting for API to be ready at http://localhost:8086/healthz
API is ready (after 4 attempts)

== Running pytest
================ 24 passed, 767 deselected, 2 warnings in 6.91s ================

== Stopping E2E docker-compose environment
E2E tests passed
```

That is the whole lane: image build, mock build, compose up, health gate, pytest, teardown. Before
this PR it could not reach step one — `cf-gears.Dockerfile` did not exist.

## Layer size — was a problem, now fixed in 52375d9

*(This section originally asked the reviewer to choose between two fixes. It is resolved; kept here
because the measurements are the justification for the second commit.)*

The `RUN cargo build` layer carried the whole in-container `target/` — 9.5 GB that the final image
discards but the daemon still writes during the build. `.dockerignore` does not help: it keeps the
*host's* `target/` out of the context, while this one is created inside the build. An earlier
attempt on this machine died with `no space left on device` committing that layer, and CI is no
safer — three workflows in this repo already need `jlumbroso/free-disk-space`.

The runtime stage had the same shape of waste: `chown -R` rewrote all 580 MB of the binary into a
second layer.

52375d9 deletes `target/` inside the same `RUN` (so it never enters the diff) and moves ownership
onto the `COPY` itself:

| | before | after |
|---|---|---|
| intermediate builder layer | 9.5 GB | **12 MB** |
| runtime image | 1.58 GB | **870 MB** |
| `useradd`/`chown` layer | 580 MB | 57 kB |
| build wall time | 3m33s | 2m52s (call it unchanged — run-to-run variance) |

Re-verified after the change: `healthz` 200 in 4s, `/openapi.json` 250 paths.

**On `--mount=type=cache`:** I floated it as the alternative, and it is no longer worth doing. A
cache mount would also keep `target/` out of the layer, but the layer is now 12 MB, so there is
nothing left to win there; its real benefit is incremental rebuilds, which a fresh CI runner cannot
use anyway. It would also require `ci.py` to set `DOCKER_BUILDKIT=1` and cannot be verified on a
Docker install without buildx. Not proposing it.

MikeFalcon77:
On item H - I traced the history of this in #4175:

- https://github.com/constructorfabric/gears-rust/issues/4175#issuecomment-5683405245 - how the compose lines broke, and why #4115 and #4175 look contradictory (#4175 is a duplicate of #4115, filed a week after the fix in #4142 had already merged)
- https://github.com/constructorfabric/gears-rust/issues/4175#issuecomment-5683407072 - what a final, non-per-section fix would look like

Short version, and it does not change this PR's conclusion: dropping those three
lines is right. But the reason is that the MODULES root is dead after the
modules -> gears rename, not that a dash-named gear cannot be addressed by env
var - #4142 fixed that in June, and APP__GEARS__API_GATEWAY__CONFIG__BIND_ADDR
binds correctly in HEAD today.

One value detail: compose set CORS_ENABLED=true while config/e2e-local.yaml:100
sets cors_enabled: false, so "config/e2e-local.yaml already sets all three
values" is not quite accurate. Nothing changes when the lines go away, since the
override is dead either way - just worth knowing that anyone later "fixing" them
to APP__GEARS__API_GATEWAY__* would flip CORS on.

lobster40:
Both corrections verified and applied — thank you, this was a real error on my part, not just
imprecise wording.

**On the reason.** Confirmed: #4142 *fix(toolkit-config): allow k8s env override of gear config via
underscore gear names* merged 2026-06-26, and `remap_gear_env_key`
(`libs/toolkit/src/bootstrap/config/mod.rs:430`) rewrites `_` → `-` in the segment right after
`gears`, so `APP__GEARS__API_GATEWAY__*` lands on `api-gateway`. My citation of #4115 / #4175 was
stale, and your note that #4175 duplicates #4115 a week after #4142 had merged explains how I got
there — I read the open issue and stopped.

**On the values.** Also confirmed, and worse than a nitpick: `CORS_ENABLED=true` in compose against
`cors_enabled: false` at `config/e2e-local.yaml:100`. So "already sets all three values" was wrong,
and the compose line was not a redundant restatement — it was trying to flip CORS, and had been
silently failing to.

c71d518 rewrites the comment in `docker-compose.yml` with the right reason and carries your warning
to where someone would act on it:

```yaml
      # Do not "repair" them by renaming the prefix. BIND_ADDR and ENABLE_DOCS
      # would only restate config/e2e-local.yaml:98-99, but CORS_ENABLED was
      # `true` here against `cors_enabled: false` there - so reviving the line
      # would quietly switch CORS on for the whole e2e lane.
```

The PR description is corrected too. No code change — deleting the lines was and stays right.

lobster40:
Rebased onto `main` (merge commit 427b34d). All three conflicts were base-image pins that Dependabot moved on `main` while this branch was open — worth a note, because one of them is a deliberate non-take.

**`rust:1.98.0` → kept `1.97.0`.** #4835 / #4838 bumped the builder stages to 1.98.0, but `rust-toolchain.toml` is still `channel = "1.97.0"`. That is the exact drift this branch exists to fix: the image ships one toolchain, rustup then downloads another on every build. I kept 1.97.0 so the builder matches the workspace. The `docker-pins` check added here now fails on that mismatch, so the next image bump has to move `rust-toolchain.toml` with it instead of landing alone — Dependabot re-opening this bump is expected, and it should then come with the toolchain change.

**`debian:13.3-slim` → took `13.6-slim`, re-pinned.** Took `main`'s newer tag, but the bump dropped the digest this branch had added, so I restored it at `sha256:d7e12182ce18b85b93007c1dedf31f2d29e01ccf3182cc4017c709b6259bc132` (current multi-arch index digest for the tag, resolved from the registry). Applied to `mini-chat.Dockerfile`, `mini-chat-prebuilt.Dockerfile` and `cf-gears.Dockerfile`.

**`testing/docker/cyberware.Dockerfile` — modify/delete.** Kept the delete. It is renamed to `cf-gears.Dockerfile` here, and `main`'s only changes to it were those two Dependabot bumps, both carried over to the rename target.

`make docker-pins` passes on the merged tree (4 Dockerfiles checked). No change to the substance of the PR.

.github/workflows/e2e.yml
```diff
@@ -39,6 +39,59 @@ env:
...
+  # builds on the runner, this one builds inside an image, and the two failure
+  # modes are unrelated. Until this job existed nothing in CI ran `docker build`
+  # at all, which is how every Dockerfile in the repo came to reference a
+  # package that does not exist (#4798).
+  #
+  # Scheduled/manual only — a full in-image workspace build is too slow to put
+  # on every PR. It builds with BUILD_PROFILE=dev for the same reason; the image
+  # is a functional test subject here, not a performance subject.
+  e2e-docker:
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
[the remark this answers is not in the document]
lobster40: Not fixed — the premise does not hold for this workflow.

`e2e.yml` has no `push` trigger and never has: its `on:` block is `pull_request` +
`schedule` + `workflow_dispatch`, byte-identical on `main` and on this branch (this PR's
diff to the file is +86/−0, nothing was removed). So `e2e-docker` is not "skipped for push
events" any more than every other job here is — no job in this workflow runs on push, and
adding `push` to the job condition would have no effect without also adding the trigger.

Adding the trigger is the wrong trade: `on: push` applies to the workflow, not to one job,
so the 45-minute `e2e` job would re-run on every push to `main`, repeating what already
ran on the PR. Post-merge detection is also strictly later than what `docker-changes`
already gives — `e2e-docker` runs **on the PR**, before merge, whenever the docker lane is
touched.

The real gap the finding gestures at is a change that breaks the in-image build without
touching a Dockerfile — which is exactly how #4798 happened (a package rename). The paths
filter did not cover that, so it now also matches the root workspace pins the image build
depends on: `Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `.cargo/config.toml`. Root
files only — `**/Cargo.toml` would fire the 60-minute job on nearly every Rust PR.
Anything that slips past both that filter and the PR is caught by the nightly schedule
within 24h.

The acceptance wording in #4798 ("A CI job exercises it on push") is mine and was loose;
PR-time + nightly is stronger coverage than push-time. Resolving.

Filter change in 40622006d.

.github/workflows/e2e.yml
```diff
@@ -39,6 +39,59 @@ env:
...
+
+      - name: Install Python dependencies
+        run: pip install -r testing/e2e/requirements.txt
+
+      - name: Run E2E tests in Docker
+        run: make e2e-docker-smoke
+
+      - name: Dump container logs
+        if: failure()
+        run: docker compose -f testing/docker/docker-compose.yml logs --no-color --tail=400 || true
```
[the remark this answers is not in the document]
lobster40: Already the case, and the residual hole is now closed.

`cmd_e2e` captures the compose logs **before** `docker compose down -v`, not after — see
the `Capturing docker-compose logs` step in `tools/scripts/ci.py`, which runs inside the
same `if args.docker and docker_env_started:` block, ahead of the `down`. The workflow
deliberately has no log-dump step for that reason; the `Tear down` step is only a
belt-and-braces `down -v || true` for the case where ci.py never reached its own cleanup.

The case that genuinely did lose logs was the early exit: `wait_for_health()` calls
`sys.exit(1)` on a readiness timeout, which skipped the capture and the teardown
altogether. That is fixed in the follow-up commit — the teardown moved into a `finally`
block, with `exit_code` defaulting to 1 so the capture (conditional on a non-zero status)
fires on those exits too. Resolving.

Fix in 7ca25bb26.

tools/scripts/ci.py
```diff
@@ -518,6 +594,22 @@ def cmd_e2e(args):
...
+                [
+                    "docker",
+                    "compose",
+                    "-f",
+                    "testing/docker/docker-compose.yml",
+                    "logs",
+                    "--no-color",
+                    "--tail=400",
+                ]
+            )
```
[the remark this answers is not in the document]
lobster40: Confirmed and fixed.

`wait_for_health()` exits the process on a readiness timeout, so the log capture and
`docker compose down -v` below it never ran: `make e2e-docker-smoke` left the compose
environment up locally and produced no logs for the one failure worth reading. The
local-mode server process leaked on the same path.

The teardown now lives in a `finally` block, which Python runs while the `SystemExit`
propagates, and `exit_code` starts at 1 so the log capture — conditional on a non-zero
status — fires on those early exits rather than being skipped.

Verified with the external calls stubbed out, no docker involved:

- readiness timeout → `docker compose ... logs --no-color --tail=400` captured,
  `docker compose ... down -v` ran, process exit status 1;
- clean run → no log capture, `down -v` ran, exit status 0.

Fix in 7ca25bb26.

tools/scripts/ci.py
```diff
@@ -177,10 +178,78 @@ def cmd_cfs_validate(_args):
...
+            continue
+        checked += 1
+        with open(path, encoding="utf-8") as fh:
+            for lineno, line in enumerate(fh, 1):
+                if not line.startswith("FROM "):
+                    continue
+                ref = line.split()[1]
+                if "@sha256:" not in ref:
+                    problems.append(f"{rel}:{lineno}: base image is not digest-pinned: {ref}")
+                rust = re.match(r"^rust:([0-9.]+)-", ref)
```
Artifizer, line 234: **This check will permanently block the Dependabot Docker lane.**

`.github/dependabot.yml` on `main` already has a `docker` ecosystem covering both `/testing/docker` and `/gears/mini-chat/deploy/docker` — #4799 is merged, so the "Follow-up" note in the PR description ("Dependabot is configured for `cargo` only, so no Docker base image ... has an updater") is now the inverse of reality.

And it is demonstrably working: #4838 and #4835 bumped `rust` to `1.98.0` in exactly those two directories. This branch reverts both to `1.97.0` in order to satisfy the check being added here (see the merge commit 427b34d).

The structural problem: Dependabot cannot edit `rust-toolchain.toml` — that is the `cargo` ecosystem, a different updater, and nothing correlates the two. So from the moment this merges, **every future Dependabot rust-image PR opens red and no bot can make it green.** Someone has to hand-edit `rust-toolchain.toml` into each one. That is a check one of our own automated lanes cannot structurally pass, which tends to end with the check being marked non-required and quietly ignored — the same failure mode #4798 is about.

Two ways out. I'd take the second:

1. Make the rust-version mismatch a warning (print it, and exit non-zero only for missing digests), keeping the digest half blocking.
2. Bump `rust-toolchain.toml` to `1.98.0` in this PR and keep all three images there.

Option 2 also avoids the other cost of the current approach: pinning three images *back* to an older Rust/Debian base to satisfy a lint is the opposite direction from where base-image hygiene should go, and it silently undoes two merged supply-chain updates.

The digest-pinning half of the check has no such conflict — Dependabot preserves and updates digests once they are present. It is only the `rust:<ver>` ↔ `rust-toolchain.toml` coupling that has no owner.

---

gears/mini-chat/deploy/docker/mini-chat.Dockerfile
```diff
@@ -14,19 +16,20 @@ RUN apt-get update && \
...
 
 # Copy workspace files
-COPY Cargo.toml Cargo.lock ./
+COPY Cargo.toml Cargo.lock Gears.toml ./
 COPY rust-toolchain.toml ./
+COPY .cargo ./.cargo
+COPY tools ./tools
 
 # Copy all workspace members
 COPY apps/cf-gears-example-server ./apps/cf-gears-example-server
```
Artifizer, line 25: **This file keeps the exact defect the PR identifies as the root cause, and the new CI job does not cover it.**

The reasoning in `cf-gears.Dockerfile` is right — "Enumerating members is precisely what rotted" — but it is applied only there. This file still enumerates, so it stays exposed to the same rot.

It does work today: I resolved all 118 `[workspace] members` against this COPY list (`apps/cf-gears-example-server`, `libs`, `gears`, `examples`, `config`, `proto`, `tools`) and nothing is missing, now that `tools/` is added. But that is luck of the current layout — every member outside `apps/cf-gears-example-server` happens to live under a directory that is copied wholesale. The moment someone adds a second `apps/*` member, Cargo cannot load the workspace and this image breaks exactly the way #4798 describes.

The guard added in this PR does not catch that, and worse, it looks like it does. `docker-changes` filters on `**/*.Dockerfile`, so editing this file **triggers** `e2e-docker` — but that job builds only `testing/docker/cf-gears.Dockerfile` and `http-mock.Dockerfile` (`ci.py` `cmd_e2e`, `-f testing/docker/cf-gears.Dockerfile`). A green check on a file that was never built is worse than no check at all, because the next person will trust it.

Note also that the Verification section covers `cf-gears.Dockerfile` and `http-mock.Dockerfile` only — this file's rewrite (package rename, `tools/`/`.cargo/`/`Gears.toml`, the `1.97.0` digest) has not been built anywhere, by hand or in CI.

Either:
- give this file `COPY . .` as well — it already has `cmake` and the BuildKit cache mounts, so the change is one line and the `.dockerignore` at the repo root applies equally; or
- add it to the `e2e-docker` job as a build-only step (`docker build -f gears/mini-chat/deploy/docker/mini-chat.Dockerfile .`), so the enumeration is at least exercised.

The first is cheaper and matches the conclusion this PR already reached.

---

Artifizer (commented):
Reviewed the branch locally — checked it out, ran the new tooling, and cross-checked the claims against the tree and the actual CI runs.

The core fixes hold up. I verified each of these independently:

- `rust-toolchain.toml` channel really is `1.97.0` (the `1.95.0` in the description is `workspace.package.rust-version`, the MSRV — a different field; the diff actually moves *down* from `1.98.0`)
- compose contexts resolve as claimed: `mock` → `testing/`, `file-parser-api` → repo root, so both the `COPY e2e/...` fix and `COPY . .` are correct
- `apps/gts-docs-validator` genuinely does not exist; `apps/` contains only `cf-gears-example-server`
- `config/e2e-local.yaml:98-100` is exactly `bind_addr` / `enable_docs: true` / `cors_enabled: false` — the CORS warning left in the compose file is accurate and worth having
- `make docker-pins` runs clean (4 files, 0.4s); `e2e-docker-smoke` exists; `e2e-docker` passed in **7m29s** on the last run, so the "too slow for every PR" rationale is more conservative than it needs to be

Two things I think should be resolved before merge, both left as inline comments:

1. **`docker-pins` will permanently block the Dependabot Docker lane** — the `docker` ecosystem is already live on `main` and already bumped rust to `1.98.0`; this branch reverts that to satisfy its own new check, and Dependabot cannot edit `rust-toolchain.toml` to get green again.
2. **`mini-chat.Dockerfile` keeps the enumerate-the-members pattern the PR names as the root cause**, and `e2e-docker` triggers on changes to it without ever building it.

I have a separate list of smaller items (FROM-parsing edge cases in `docker-pins`, the `Scheduled/manual only` comment contradicting its own `if:`, `needs: docker-changes` without `always()` silently skipping the nightly, a dead `pip install` step, and `.dockerignore` secret patterns now that the whole repo is the build context). Happy to add those as a second pass if useful — they are all minor and none of them block.

---

tools/scripts/ci.py
lobster40: You are right that the check as written cannot be satisfied by the lane it polices, and the description's "Follow-up" note was stale in the opposite direction — #4799 is merged, so Dependabot does own these images now. Both corrected.

**Taking your option 1, not option 2** — and the reason is a number I did not have when I wrote this.

Option 2 (move `rust-toolchain.toml` to `1.98.0` and keep all three images there) is the better shape, and I tried it first. `1.98.0` ships a new pedantic lint, `clippy::unused_async_trait_impl`, and this workspace sets `pedantic = { level = "deny", priority = -1 }`. So the bump is not a one-line change:

```
$ cargo +1.98.0 clippy --workspace --all-targets -- -D warnings
error: unused `async` for async trait impl function with no `.await` statements
   --> libs/toolkit-security/src/shared_secret.rs:96:5
error: could not compile `cf-gears-toolkit-security` (lib) due to 1 previous error
```

That run aborts on the first crate, and `toolkit-security` is foundational, so nothing downstream gets linted — it under-reports badly. Re-running with pedantic downgraded to `warn` so every crate compiles gives the real figure: **55 sites across 21 files**, concentrated in `gears/bss/ledger` (16), `libs/toolkit-db` (12), `gears/system/cluster` (9), `libs/toolkit-security` (4). Each one is a judgement call about public async surface — drop `async` and return `impl Future`, or `#[allow]` with a reason where the trait dictates the signature. That is not something to smuggle into a Dockerfile PR.

So: the mismatch is now a **warning**, missing digests still **fail**. Your point that only the version coupling has no owner is exactly the seam I cut along.

Two things to stop this becoming a warning nobody reads:

- it is emitted as a GitHub Actions annotation on the offending line, so it shows up in the PR diff rather than only in a log:
  ```
  ::warning file=testing/docker/cf-gears.Dockerfile,line=7::rust 1.98.0 != rust-toolchain.toml 1.97.0
  ```
- **#4917** owns clearing it, with the measurement and the reproduction, and `cmd_docker_pins`' docstring points at it. When that lands the check reports 0 warnings again.

The images now sit on `rust:1.98.0-bookworm` — main's side — so #4835 and #4838 are no longer reverted. `debian` moved 13.6 → 13.7 as well, but main's bump had dropped the digest, so I resolved the multi-arch index digest from the registry and restored it on all three.

I also took your smaller item about the parser while in here. `line.split()[1]` behind a `startswith("FROM ")` guard reported `FROM builder AS runtime` as an unpinned base image — a false positive on any multi-stage reference, which would have been the next way this check got switched off. It now tracks `AS` stage names, skips those and `scratch`, skips leading `--flag` tokens such as `--platform`, matches `FROM` case-insensitively and with leading whitespace, and warns rather than errors on a ref it cannot resolve statically (`$VAR`). Covered by `tools/scripts/tests/test_ci_docker_pins.py`, following the stdlib-`unittest` convention `toolkit-pr-review` already uses.

Your remaining items — the `Scheduled/manual only` comment, `needs:` without `always()`, the dead `pip install` — are addressed in the sibling thread and the commits; please do send the rest of the list.

gears/mini-chat/deploy/docker/mini-chat.Dockerfile
lobster40: Agreed on both halves, and I took both fixes rather than choosing — they close different holes.

**`COPY . .`** — your first suggestion, and it is the one that matches the conclusion this PR already reached. The file no longer enumerates members. You are right that the current list resolves cleanly today and that this is luck of the layout; one more `apps/*` member and it breaks exactly the #4798 way.

**The build-only step too**, because `COPY . .` fixes the file while the green-check problem is in the job. `docker-changes` filters on `**/*.Dockerfile`, so editing this file triggers `e2e-docker`, but `cmd_e2e` only ever built `cf-gears.Dockerfile` and `http-mock.Dockerfile`. Left alone, the next person editing any third Dockerfile gets the same misleading pass. `e2e-docker` now builds this image as well.

One detail worth recording: that step sets `DOCKER_BUILDKIT=1` explicitly. This Dockerfile uses three `RUN --mount=type=cache` directives (they predate this PR — they are on `main` today), and the legacy builder rejects that syntax outright. It is the default on current runner images, but the job should not depend silently on the runner image staying that way.

**On verification, and being straight about it:** I could not build this image locally — this machine's Docker has no `buildx` component, so any BuildKit build fails before it starts:

```
ERROR: BuildKit is enabled but the buildx component is missing or broken.
```

Which is a fair illustration of your point: a file that nothing builds does not get built by accident. The verification is therefore the new CI step rather than my laptop, and this PR touches Dockerfiles, so `docker-changes` fires and `e2e-docker` builds it on this push. I will report the result here rather than leave the claim hanging.

Also in this commit, from your smaller list: the dead `pip install -r testing/e2e/requirements.txt` step is gone — the Makefile's `py-env` prerequisite builds its own venv and installs there, so the workflow was populating an interpreter the job never used. And `needs: docker-changes` is now gated on `!cancelled()` rather than bare `always()`: same protection against a skipped or failed filter job silently taking the nightly and manual runs with it, but a cancelled run stays cancelled.

The `Scheduled/manual only` comment you flagged was already rewritten in `b06dc96` when the path filter went in — it now describes the actual `if:`. Shout if you were looking at something else.

lobster40:
## Re-verified after the merge and the review fixes

`make e2e-docker-smoke`, locally on arm64 macOS, on `63e309c`:

```
== Building Docker image for E2E tests
   Finished `release` profile [optimized + debuginfo] target(s) in 10m 24s
== Rebuilding docker-compose mock service
== Starting E2E docker-compose environment
== Waiting for API to be ready at http://localhost:8086/healthz
API is ready (after 2 attempts)
== Running pytest
================ 26 passed, 774 deselected, 2 warnings in 8.37s ================
== Stopping E2E docker-compose environment
E2E tests passed
```

`make docker-pins` on the merged tree: exit 0, 4 Dockerfiles checked, 2 warnings (the standing
`1.98.0` vs `1.97.0` mismatch, now advisory). Negative-tested by stripping a digest — exit 1 with
`base image is not digest-pinned`. Nine unit tests for the hardened `FROM` parser pass.

`mini-chat.Dockerfile` is **not** covered by that run: it uses `RUN --mount=type=cache` (three of
them, and they predate this PR — they are on `main` today), which needs BuildKit, and this machine's
Docker has no `buildx` component. The new build-only step in `e2e-docker` is what verifies it, and
this PR touches Dockerfiles so that job runs on this push.

## One honest regression to record

Keeping the images on `rust:1.98.0-bookworm` while `rust-toolchain.toml` stays `1.97.0` reintroduces
the exact cost this PR set out to remove. From the build log above:

```
Step 1/18 : FROM rust:1.98.0-bookworm@sha256:82150a52…  AS builder
...
+ cargo build --release --bin cf-gears-example-server …
info: syncing channel updates for 1.97.0-aarch64-unknown-linux-gnu
info: downloading 5 components
```

That is the "every build rustup-installs a second toolchain" defect, back again — just at
`1.98.0 → 1.97.0` instead of `1.95.0 → 1.97.0`.

I am taking that trade knowingly, and it is worth stating rather than leaving in a log:

- the alternative is pinning the images back to `1.97.0`, which reverts #4835 and #4838 and makes
  `docker-pins` a check no bot can satisfy — the objection that started this;
- this version is temporary and owned: #4917 closes it by moving the toolchain to `1.98.0`, and the
  `docker-pins` warning and its annotation are the reminder;
- it costs build time, not correctness. The lane is green either way.

It does raise #4917's priority above "nice to have" — right now every in-image build, local and CI,
pays for it. @Artifizer, if you would rather eat the Dependabot conflict than the download until
#4917 lands, say so and I will flip it; I think temporary-and-tracked beats a permanently red bot
lane, but it is a genuine trade and you framed the original one.

Artifizer (approved):

PR #4817, opened 2026-09-14
Chore/toolkit security review fixes

MikeFalcon77:
Fixes the findings from the post-review of `libs/toolkit-security`

## The two critical ones

- **An empty `ScopeConstraint` granted every row.** A conjunction over no
  filters is `TRUE`, so it was an allow-all disjunct — while the scope still
  reported itself as constrained. `toolkit-db` compiled it to `WHERE true`,
  and the PEP compiler built one from whatever predicates the PDP returned.
  Now refused at construction, at the PDP boundary, and in the SQL layer.
- **The TokenReview call had no timeout**, and it ran while holding the
  per-token single-flight lock — so a hung backend parked every caller
  presenting that token, with no deadline of their own.

## Breaking changes

Needs a release note: **`provider: kube` now requires a non-empty `audiences`
list.** An empty one disabled audience verification on both sides, so any
ServiceAccount token in the cluster authenticated as a platform peer. No
config in this repo sets it; a deployment whose YAML omits it will fail to
start.

Also: `ScopeConstraint::new` panics on an empty list (`try_new` is the
fallible form); `ScopeFilter` is `#[non_exhaustive]`; `SharedSecretInternal
Authenticator::new` → `try_new`; `build_authenticator` returns
`Result<BuiltAuthenticator, _>`; the config secret is a `SecretString`;
`outbound_marker()` carries its own identity variant; `rg_tables`/
`tenant_tables` moved to `toolkit_db::secure::schema`.

## Summary by CodeRabbit

* **Security Enhancements**
  * Anonymous authentication contexts are rejected consistently.
  * Invalid or empty shared secrets are refused during configuration.
  * Authentication backend build failures are reported instead of silently falling back.
  * Authentication requests use a configurable overall timeout.
  * Authentication cache keys no longer store plaintext credentials.

* **Authorization**
  * Empty or unsupported access constraints now fail closed.
  * Scope matching and tenant authorization behavior is more consistent across services.

* **Reliability**
  * Long credentials are cached normally, with improved expiration and cancellation handling.


libs/toolkit-security/src/access_scope.rs
```diff
@@ -758,7 +796,40 @@ impl AccessScope {
...
+    /// decision. The `contains_*` and `has_property` family reports on the
+    /// *constraint list* only, so on an allow-all scope — which has no
+    /// constraints — every one of them answers "no", which reads exactly like
+    /// deny-all at the call site. Each caller then has to remember to check
+    /// [`AccessScope::is_unconstrained`] first, and a caller that forgets denies
+    /// a request the scope actually permits.
+    ///
+    /// This folds that check in, so the unconstrained case cannot be skipped.
+    #[must_use]
+    pub fn allows_uuid(&self, property: &str, id: Uuid) -> bool {
```
aviator5: I suggest reducing the public `AccessScope` API and keeping constraint enforcement in SecureORM.

### Changes for this PR

- **Remove `allows_uuid()` and `allows_value()`.** They search for a matching value without evaluating the complete constraint. For example, given `owner_tenant_id = A AND owner_id = Alice`, `allows_uuid("owner_tenant_id", A)` returns `true` even when the object being created belongs to Bob. This confirms the tenant match, not permission to create the object. The documentation should not recommend these methods for authorization.
- **Remove `contains_value()`.** It has no production callers in this repository beyond the newly introduced `allows_value()`. Neither `contains_uuid()` nor SecureORM’s insert validation depends on it.
- **Retire `contains_uuid()` once its existing authorization callers are migrated.** SecureORM already validates inserted models against constraints using OR between constraints and AND between filters.

### Existing callers that need correction

**Resource Group: apparent constraint-enforcement bug**

The ordinary `create_group` path obtains a PDP scope, but checks tenant membership only for cross-tenant requests. The scope is then discarded, and the repository inserts with `allow_all()`. Consequently, an ordinary same-tenant creation can proceed without satisfying the returned constraints.

Pass the applicable PDP scope through to insert validation. Keep system seeding and new-tenant provisioning explicitly separate, with their own authorization contracts.

**Ledger: preserve the authorization scope through the write**

The journal repository currently constructs `AccessScope::for_tenant(entry.tenant_id)` from the row being inserted. This checks consistency with the selected tenant; it does not enforce the caller’s original PDP constraints.

The intended flow should be:

```text
PDP → AccessScope
    → service receives the scope
    → repository receives the scope
    → secure_insert / scope_with_model validates the model against it
    → INSERT
```

Do not replace the PDP scope with one derived from the input row. For derived records or deferred operations, explicitly define which authorization covers the write.

**Pricing:** audit its `contains_uuid()` write checks in the same way and remove them once equivalent constraint enforcement is confirmed.

### Delivery options

Either fix these paths in this PR, or:

1. Open separate bugs for the Resource Group, Ledger, and Pricing owners, linking the affected call sites and expected behavior.
2. Mark `contains_uuid()` as deprecated, documenting that it searches filter values and must not be used as an authorization decision.
3. Keep existing checks until their replacements are implemented.
4. Add regression tests for unauthorized tenants and matching-tenant/wrong-owner writes, while preserving legitimate internal workflows.
5. Remove the deprecated method after migration.

The replacement should be enforcement of the applicable constraints in SecureORM, rather than another tenant-membership helper.

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -370,7 +519,22 @@ impl<A: InternalAuthenticator> CachingInternalAuthenticator<A> {
             CacheLookup::Miss => {}
         }
 
-        let result = self.inner.authenticate(token).await;
+        // Bounded: `_guard` is held across this await, so an unresponsive
+        // backend would otherwise block every caller presenting this token
+        // indefinitely — they are parked on the lock, not on a call that could
+        // time out on its own.
+        let result = match tokio::time::timeout(
```
aviator5: Please apply a single deadline to the entire authentication call, including the per-token lock wait and the cache-bypass path.

Currently, three simultaneous requests with the same token and a hung backend finish after roughly **10, 20, and 30 seconds**: each gets a fresh timeout after reaching the front of the queue. All three should time out within 10 seconds of their own arrival.

Please add a concurrent-request regression test. Tokens over 16 KiB should obey the same deadline.

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -112,38 +140,158 @@ enum CacheLookup {
...
+/// cache stayed full of still-valid entries, which is exactly the sustained-load
+/// case, and it blocked every concurrent lookup while it ran.
+///
+/// The two collections are only consistent if they are updated together, so the
+/// map is private and every mutation goes through a method here.
+struct ExpiringCache {
+    entries: HashMap<String, CacheEntry>,
+    /// `(expires_at, token)` for every entry in `entries`. Ordered, so the
+    /// soonest-to-expire is the first element and expired entries are a prefix.
+    by_expiry: BTreeSet<(Instant, String)>,
```
aviator5: The new expiry index stores a second copy of each token. With 10,000 entries and a 16 KiB cacheable-token limit, the two copies can occupy approximately **312.5 MiB per cache instance**, excluding collection overhead and cached values.

Could we use a fixed-size cryptographic digest as the key in the cache, expiry index, and single-flight map? `CacheEntry` does not need the original token; it is only needed during backend validation and expiry calculation.

The existing comment explains the desire to avoid a crypto dependency in this crate. Could we accommodate the project’s approved crypto provider while keeping fixed-size keys? This would also remove the key-size reason for bypassing caching on long tokens.

libs/toolkit-security/src/access_scope.rs
MikeFalcon77: You are right, and the helper was mine. `contains_uuid` searches filter values and never evaluates a constraint, so for `[owner_tenant_id = A AND owner_id = Alice]` it returns true for the tenant alone. I wrapped that in `allows_uuid` and documented it as the way to authorize, which made it worse.

Done in this PR:

- Removed `allows_uuid`, `allows_value` and `contains_value`. No production callers; the first two were added by this PR and never adopted.
- Documented `contains_uuid` as not an authorization decision, with both ways it misleads: it reports on one filter rather than the conjunction, and it answers false for an allow-all scope and for subquery filters that do grant access.
- Pointed callers at SecureORM `validate_insert_scope`, which ANDs across a constraint and ORs across constraints.

Not done, with a reason: no `#[deprecated]` yet. The workspace builds with `-D warnings`, so the attribute breaks the build at all ~25 call sites at once, and a blanket `#[allow(deprecated)]` would just silence it. It goes on once the gates move.

On the gear call sites, I checked all three and your reading holds:

- Resource Group: `create_group` gets a PDP scope, checks tenant membership only when `target_tenant_id != tenant_id`, then drops the scope; `group_repo` inserts with `system_scope()` = `allow_all()`, and `validate_insert_scope` returns early on `is_unconstrained()`. Same-tenant creates are validated against nothing.
- Ledger: `journal_repo.rs:160` builds `AccessScope::for_tenant(entry.tenant_id)` from the row. The caller's scope is in hand at `posting/service.rs:541` and used for five neighbouring steps, but not for the insert.
- Pricing: same gate shape, but it does thread the scope into `scope_with_model` inserts, so the gate is not the only defence there.

I would rather not change three gears' authorization in this PR. Filing one issue per gear with the call sites and the expected flow, unless you want them here.

libs/toolkit-security/src/internal_auth_cache.rs
MikeFalcon77: Fixed. The timeout now wraps the whole `authenticate` call, so it covers the lock wait and not just the backend round-trip. Each caller fails within one deadline of its own arrival instead of queueing for 10s, 20s, 30s.

The inner timeout is gone. Keeping both would restart the clock for whoever just took the lock, which is the behaviour you are describing.

Cancelling at the outer deadline is safe, as you noted: the lock guard and `ReleaseTokenLockOnDrop` both clean up on drop, so an abandoned attempt leaves no lock held and no `inflight` entry.

Added `concurrent_callers_share_one_deadline_rather_than_queueing`: three callers, same token, hung backend, each must fail in under 2x the deadline. It runs on virtual time, so it costs nothing.

The oversized-token path is covered too, but differently than you suggested. The reason it bypassed the cache was that the token was the key, so a 16 KiB token cost that much per entry. Keys are a SHA-256 digest now (see the memory thread), so length changes neither the cost nor the path, and the bypass is gone. A long token goes through the normal path and obeys the same deadline.

libs/toolkit-security/src/internal_auth_cache.rs
MikeFalcon77: Done. The cache map, the expiry index and the single-flight map are now keyed on a SHA-256 digest of the token, so no map holds the credential and an entry costs the same whatever the token length.

I used `sha2`, which is already a workspace dependency, behind the same `internal-auth-cache` feature as the rest of the cache. A cryptographic digest rather than a fast hash for the reason you would expect: the input is a credential, and a collision would let one token answer for another. The module comment that argued against a crypto dependency was mine and is now wrong, so I replaced it.

`CacheEntry` never needed the token, as you said. The only places that still see the plaintext are backend validation and the `exp` parse, both of which take it as an argument.

You were right that this removes the key-size reason for the bypass: `MAX_CACHEABLE_TOKEN_LEN` and the long-token bypass are gone, and a long token is cached like any other.

`TokenKey`'s `Debug` prints two bytes of the digest, enough to correlate log lines and not much else.

MikeFalcon77:
issues are create for components:
https://github.com/constructorfabric/gears-rust/issues/4840
https://github.com/constructorfabric/gears-rust/issues/4839

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -342,20 +494,24 @@ impl<A: InternalAuthenticator> CachingInternalAuthenticator<A> {
...
             CacheLookup::Valid(identity) => return Ok(identity),
             CacheLookup::Rejected => return Err(InternalAuthNError::InvalidToken),
             CacheLookup::Miss => {}
         }
 
         // Single-flight: serialize concurrent misses for the same token so a
         // burst of calls collapses into one backend round-trip.
-        let lock = self.token_lock(token);
+        let lock = self.token_lock(key);
         let _guard = lock.lock().await;
```
aviator5: Please create `ReleaseTokenLockOnDrop` **before** `lock.lock().await`, so cancellation also cleans up callers waiting for the lock.

For example, A holds the lock and B waits for the same token. Cancel A: its cleanup keeps the entry because B still references it. Then cancel B before it resumes: B has no cleanup guard yet, so the `inflight` entry remains indefinitely. Repeating this with distinct tokens grows the map without a bound.

A regression test with 32 cancelled pairs leaves 32 entries instead of zero; moving the guard before the await makes it pass. Please add coverage for this cancellation order.

libs/toolkit-security/src/internal_auth_cache.rs
```diff
@@ -65,6 +61,19 @@ pub const MAX_TOKEN_REVIEW_CACHE_TTL: Duration = Duration::from_mins(5);
...
+/// The backend call happens while the per-token single-flight lock is held, so
+/// a backend that never answers would otherwise park every caller presenting
+/// that token for as long as it stays unresponsive — with no deadline, since
+/// the waiters are blocked on the lock rather than on the call. Bounding the
+/// call converts an indefinite hang into an `Unavailable` a caller can act on.
+///
+/// This is a request deadline, deliberately unrelated to the cache TTL: it is
+/// sized for how long a `TokenReview` round-trip may reasonably take, not for
+/// how long its answer stays usable.
+pub const TOKEN_REVIEW_BACKEND_TIMEOUT: Duration = Duration::from_secs(10);
```
aviator5: Could we rename this to `DEFAULT_AUTHENTICATION_TIMEOUT` and make `authentication_timeout` configurable, keeping 10 seconds as the default? The deadline now covers the entire authentication call, including the per-token lock wait, so `BACKEND_TIMEOUT` no longer describes its scope. Please expose an override through the authenticator's construction API and the configuration used to wire it, while preserving a single deadline for the whole call.

libs/toolkit-security/src/internal_auth_cache.rs
MikeFalcon77: Fixed. Moved `ReleaseTokenLockOnDrop` construction before `lock.lock().await`, so a waiter cancelled before it ever gets the lock still cleans up its `inflight` registration.

Added the regression test you described: 32 distinct tokens, A takes the lock and enters a slow backend call, B queues behind it on the same token, both get aborted back-to-back (no await between the two aborts, so tokio drops each future from wherever it was suspended instead of letting it run further). On the old code this leaves 32 entries in `inflight`; with the fix it's 0. Verified it actually catches the regression by running it against the reverted code first (fails 5/5) and the fixed code (passes 5/5).

libs/toolkit-security/src/internal_auth_cache.rs
MikeFalcon77: Done. Renamed it to DEFAULT_AUTHENTICATION_TIMEOUT, kept the 10s default, and added with_authentication_timeout() on CachingInternalAuthenticator to override it per instance -- still a single deadline for the whole authenticate() call, lock wait included.

Threaded an optional authentication_timeout through build_cached_k8s_authenticator (the one place provider: kube is constructed), applied via with_authentication_timeout() when Some. I did not add gear-level config fields for it -- the three call sites (oop.rs, grpc-hub/gear.rs, api-gateway/gear.rs) all pass None for now, so behavior is unchanged. Wiring a new *_timeout_secs setting into each gear's own config schema is separate scope from this PR, same reasoning as why I filed the Resource Group and Ledger scope-authorization gaps as issues instead of fixing them here. The override is available at the one construction point if a gear owner wants to expose it later.

aviator5 (approved):

Issue #4627, opened 2026-08-20
[Types Registry P0] Phase 7 — cutover and migration


aviator5:
Parent: #4619 · Depends on: Phase 6 (#4626)

## Goal

Flip the platform onto the new path and finish P0: the registry seeds only what it owns, every
other gear pushes its own declarations, the old trait and the in-memory repository are deleted,
the REST surface is promoted back onto v1, and the read path gets validators and a cache.

**T27 left this phase for Phase 5** (plan.md **P17**), so every route the promotion moves already
exists when this phase starts.

This is the largest phase and the one with the real blast radius — ~50 call sites across 20+
gears, plus the e2e suites.

## Tasks

Task identifiers below are defined in `gears/system/types-registry/docs/p0/todo.md` (and ordered in `plan.md`), which land with the first PR (Phase 0 + Phase 1).

Recommended order: **T24 → T24a → T28**. T25 / T26 float once T24 lands; T29 needs T23 only
(T27's routes have existed since Phase 5); T30 needs T24 + T26 and can run alongside T28.

- [ ] **T24 — Cutover.** types-registry stops pulling the whole process inventory: it seeds only
      what it owns (`toolkit-gts` base types and its own control-plane types) inline at `init()`,
      then starts the outbox worker, then publishes its client. Delete `switch_to_ready`, the
      `temporary`/`persistent` split, `SystemCapability::post_init` and the in-memory repository.
      From here reads are served from the database. Reads are uncached until T30.
- [ ] **T24a** — retire v1 and promote **all seven** v2 routes onto `/types-registry/v1/`, so P0
      ends on one version rather than a permanent v2. Lands directly after T24, before T28, and
      owns both changelog entries outright — the v1 `POST` break and the `GET /entities` shape
      change, one release, two entries — because this is where both breaks actually reach a v1
      caller (P17).
- [ ] **T25** *(L)* — migrate system gears and plugins: reads onto the new trait, and the ~13
      explicit `register(...)` sites onto the reconciliation helper called from `init()`. Each
      gear gates only its own readiness. Split by gear.
- [ ] **T26** *(L)* — migrate the domain gears (`bss/ledger`, `bss/rate-provider`, `mini-chat`
      and its plugins, `llm-gateway`, `model-registry`), then delete `TypesRegistryClient`, its
      models and `MockTypesRegistryClient`.
- [ ] **T28** *(L)* — move the e2e suites onto the `202` contract: submit, poll
[...]

PR #4826, opened 2026-09-15
feat(account-management): own and reconcile the platform root type

### LINKED ISSUE #4651: feat(account-management): own and reconcile the root tenant type schema
## Context

Account Management requires a concrete GTS type for the platform root tenant. Today deployments must keep two independent configuration fragments synchronized:

- `gears.account-management.config.bootstrap.root_tenant_type` selects the type;
- `gears.types-registry.config.entities` defines that type and its traits.

PR #4641 adds the missing seed to `config/quickstart.yaml` as a temporary fix. The same pattern already exists in `config/e2e-local.yaml`. This duplication is easy to miss and can leave AM running without a bootstrapped root when bootstrap is non-strict.

AM already owns the abstract `gts.cf.core.am.tenant_type.v1~` contract and the runtime invariants of the root tenant. It should therefore own the concrete root-type schema as well.

## Desired design

1. AM generates the concrete root tenant type schema.
2. The concrete GTS ID remains configurable through AM configuration (`root_tenant_type`).
3. Add an AM configuration value `root_tenant_type_idp_provisioning: bool` controlling the schema's `idp_provisioning` trait (default should be considered; `false` matches current quickstart/e2e behavior).
4. During startup, AM reconciles the desired schema with Types Registry:
   - if absent, register it;
   - if present, validate the AM-owned semantic invariants and configured `idp_provisioning` value;
   - if the existing definition conflicts, fail startup with an actionable diagnostic.
5. Once reconciliation is implemented, remove the platform-root schema from `types-registry.config.entities` in `config/quickstart.yaml` and `config/e2e-local.yaml`. Deployment-specific child types such as the E2E customer type remain deployment-owned.

## Lifecycle and correctness requirements

- Perform schema reconciliation during AM `init()`, while Types Registry is still collecting schemas before its `post_init` ready transition. Do not put it under the bootstrap saga's `strict` policy.
- Schema/config drift is always lifecycle-fatal. `bootstrap.strict: false` may continue after a root-creation saga failure, but it must not suppress an invalid root-type contract.
- Validate that the configured ID is a concrete type derived from `gts.cf.core.am.tenant_type.v1~`; AM must not register arbitrary operator-selected namespaces.
- Validate at least these effective invariants for an existing schema:
  - it belongs to the AM tenant-type chain;
  - `allowed_parent_types` is empty;
  - `idp_provisioning` equals the configured value.
- Prefer comparison of AM-owned semantic fields over raw JSON equality so harmless formatting, field ordering, or descriptive metadata does not create false drift.
- Make the `GET -> register` flow safe for concurrent AM pod startup. Identical desired definitions must converge successfully; conflicting definitions must cause the losing pod to re-read, detect drift, and fail.
- Re-read and validate the persisted schema after registration.
- Treat `root_tenant_type` and `root_tenant_type_idp_provisioning` as create-once configuration unless an explicit schema/root migration is implemented. If a root tenant already exists, verify its stored type UUID matches the configured GTS ID.
- Decide and document whether root-type reconciliation happens when built-in root bootstrap is disabled. If AM always owns the root contract, consider moving these fields out of optional `BootstrapConfig` into a dedicated root-type config section.

## Types Registry integration note

The current `TypesRegistryClient::register_type_schemas()` path uses the legacy in-memory registry, while the P0 persistent admission path is separate. The implementation should either use the authoritative persistent reconciliation API or explicitly track the migration to it; startup success must not imply persistence when only process-local state was updated.

## Acceptance criteria

- [ ] AM config exposes the root type GTS ID and `idp_provisioning` value with validation and documented defaults.
- [ ] AM builds the root-only schema from those values.
- [ ] Startup registers an absent schema and verifies the stored result.
- [ ] Startup accepts an existing semantically matching schema.
- [ ] Startup fails on mismatched `idp_provisioning`, non-empty effective parents, wrong GTS chain, or incompatible AM-owned schema fields, regardless of `bootstrap.strict`.
- [ ] Concurrent identical registration is idempotent; concurrent conflicting registration fails deterministically after re-read.
- [ ] Changing the configured type for an existing root is rejected unless explicitly migrated.
- [ ] Unit/integration tests cover absent, matching, conflicting, concurrent, registry-unavailable, and existing-root mismatch cases.
- [ ] The duplicated platform-root seed is removed from quickstart and e2e-local configs after the AM implementation lands.

Related: #4641

fdlockgraf:
## Summary

Closes #4651.

Account Management now owns the concrete platform-root tenant-type contract and reconciles it against authoritative Types Registry storage during `Gear::init`.

- introduces an independent `root_tenant_type` configuration section;
- creates the AM tenant-type base and concrete root schema when absent;
- validates existing schemas semantically instead of silently revising them;
- adopts matching unowned or legacy Types Registry-owned schemas with CAS;
- rejects schema drift, foreign ownership, deleted schemas, and existing-root ID/type drift;
- makes concurrent replica startup converge through deterministic idempotency, must-not-exist admission preconditions, and authoritative rereads;
- routes all AM reads of the configured root and its schema chain to persistent storage;
- removes the duplicated platform-root YAML seeds from quickstart and E2E configuration while retaining the deployment-owned E2E customer type.

## Configuration

```yaml
account-management:
  config:
    root_tenant_type:
      gts_id: "gts.cf.core.am.tenant_type.v1~cf.core.am.platform.v1~"
      idp_provisioning: false
```

`root_tenant_type` is independent of optional tenant bootstrap, so AM can reconcile the shared contract even when the root tenant is created out of band. `idp_provisioning` defaults to `false`.

For migration compatibility, deprecated `bootstrap.root_tenant_type` and `bootstrap.root_tenant_type_idp_provisioning` aliases remain accepted. If modern and deprecated fields are both present, they must agree.

## Reconciliation behavior

On every AM initialization with `root_tenant_type` configured:

1. Validate that the configured ID is a canonical concrete type in the `gts.cf.core.am.tenant_type.v1~` chain.
2. If a platform-root tenant already exists, verify its ID and durable `tenant_type_uuid` binding before changing registry state.
3. Reconcile the AM base schema in persistent Types Registry storage.
4. Create the concrete root schema if absent, or validate its authored and effective AM-owned invariants if present.
5. Verify active lifecycle, deterministic UUID, effective `allowed_parent_types: []`, configured `idp_provisioning`, schema shape, and ownership.
6. Reread authoritative storage after success, races, and ambiguous response loss.

Schema/configuration drift is startup-fatal and is intentionally outside `bootstrap.strict`; non-strict bootstrap cannot suppress an invalid root contract.

## Concurrency and ownership

- deterministic UUIDv5 operation idempotency keys derived from canonical content;
- must-not-exist registration preconditions;
- uniqueness in authoritative storage rather than distributed locks;
- losing replicas reread and accept only semantically matching state;
- conflicting replicas fail after reread;
- admission operations carry trusted in-process `owning_gear` attribution;
- entity ownership adoption uses resource-version CAS and fresh-content revalidation;
- existing entities owned by another gear are never taken over.

The Types Registry operation table gains a nullable `owning_gear` migration for compatibility with existing rows.

## Persistent-read transition

The legacy process-local `TypesRegistryClient` and persistent admission store remain separate today. This PR adds a transitional AM adapter that:

- reads the configured root ID/UUID and its derivation chain from authoritative storage;
- supports exact, batch, and list projections expected by existing AM consumers;
- fails closed instead of falling back to a stale legacy root copy;
- delegates unrelated deployment-owned schemas and plugin instances to the legacy client.

This makes removal of the two platform-root YAML seeds safe for current AM runtime paths without introducing a non-atomic dual write. The general platform-wide persistent-read cutover remains tracked by #4627; the adapter is marked for removal there.

## Acceptance criteria

- [x] Configurable root GTS ID and `idp_provisioning` with validation/defaults
- [x] AM-generated concrete root-only schema
- [x] Create-if-absent with authoritative verification
- [x] Validate-if-present semantic matching
- [x] Fatal IDP, effective-parent, chain, lifecycle, UUID, schema, and ownership drift
- [x] Identical concurrent starts converge; conflicting starts fail after reread
- [x] Existing root ID/type changes require explicit migration
- [x] Unit and SQLite integration coverage for absent, matching, conflicting, concurrent, unavailable, response-loss, adoption, and root mismatch scenarios
- [x] Platform-root seeds removed from `config/quickstart.yaml` and `config/e2e-local.yaml`; E2E customer seed retained

## Validation

- `cargo test -p cf-gears-account-management --lib` — **871 passed**
- Types Registry full test suite — **757 passed across 34 test binaries**
- Types Registry SDK — **53 passed**
- `cargo check --workspace --all-targets` — passed
- affected-library Clippy with `-D warnings` — passed
- PostgreSQL migration and repository suites — passed
- formatting, `git diff --check`, and quickstart/E2E YAML parsing — passed

The no-seed startup test initializes the real Types Registry, Resource Group, and Account Management gears over migrated SQLite databases. It verifies that the legacy root is absent, AM creates the persistent contract during init, and AM runtime reads resolve it through the authoritative overlay.

## Draft follow-ups / validation still outstanding

- Full quickstart/Docker E2E and dylint have not yet been run.
- Full AM reconciliation races have not yet been exercised on PostgreSQL or MySQL.
- MySQL validation was blocked locally because the container image could not be pulled within the available timeout.
- Dedicated reconciliation metrics are not included in this change.
- The shared legacy Types Registry client and v1 REST catalogue do not yet expose persistent-only entities to unrelated consumers; complete #4627 before claiming a platform-wide read cutover.

## Summary by CodeRabbit

* **New Features**
  * Added independent platform root tenant type configuration, including optional identity-provider provisioning settings.
  * Platform root tenant types are registered and validated during startup.
  * Existing platform roots are checked against configured IDs and tenant types.

* **Bug Fixes**
  * Startup now fails clearly when root bindings, tenant type definitions, or registration responses are invalid or inconsistent.
  * Tenant types must be directly derived from the supported tenant type base.
  * Bootstrap retries now report service unavailability when conflicts cannot be classified.

* **Documentation**
  * Updated bootstrap configuration names, migration guidance, and platform bootstrap documentation.


config/e2e-local.yaml
```diff
@@ -390,14 +371,16 @@ gears:
       server: "sqlite_users"
       file: "account_management.db"
     config:
+      root_type:
```
aviator5: Suggest keeping `root_tenant_type` name - it's more obvious.

gears/system/account-management/docs/DESIGN.md
```diff
@@ -1013,18 +1013,24 @@ sequenceDiagram
...
...g realm, while other metadata may trigger fresh resource creation). If the provider returns no metadata (binding established through external configuration or convention), bootstrap proceeds normally without writing a `tenant_idp_metadata` row. On subsequent starts, bootstrap detects the existing root (in `active` status) and is a no-op. A root tenant observed in `provisioning` status whose age exceeds `2 × idp_retry_timeout` triggers one synchronous in-band `deprovision_tenant` + `compensate_provisioning` pass; on confirmed cleanup the saga restarts on `no-root` and activates a fresh root within the same `run()`, otherwise the stale row is left for provisioning-reaper compensation. Successful bootstrap emits a platform audit event with `actor=system`. If IdP retries exhaust `idp_retry_timeout`, bootstrap returns `CanonicalError::ServiceUnavailable` (HTTP 503) with no row left behind (each retry's `CleanFailure` already compensated). Concurrent-replica safety is provided by the `ux_tenants_single_root` unique partial index: if two replicas race on a fresh deployment, the second insert fails the constraint and falls through to the idempotency path on its next classification attempt.
```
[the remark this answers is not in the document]
fdlockgraf: Fixed in f760f6030. Replaced the stale timeout name with `bootstrap.idp_wait_timeout` throughout the design/feature documentation and source docs.

gears/system/account-management/account-management/src/domain/root_type.rs
```diff
@@ -0,0 +1,97 @@
...
+        let type_id = self.gts_id.as_ref();
+        let parsed = GtsId::try_new(type_id)
+            .map_err(|error| format!("root_type.gts_id `{type_id}` is invalid: {error}"))?;
+        if !parsed.is_type() || parsed.id() != type_id {
+            return Err(format!(
+                "root_type.gts_id `{type_id}` must be a canonical GTS type-schema ID"
+            ));
+        }
+        let chain = parsed.chain_ids();
+        if chain.len() < 2 || chain.first().map(String::as_str) != Some(TENANT_TYPE_BASE) {
```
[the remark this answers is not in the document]
fdlockgraf: Fixed in f760f6030. `RootTypeConfig::validated_id` now requires exactly two chain segments, and the unit test covers rejection of an indirect descendant.

config/e2e-local.yaml
fdlockgraf: Agreed — renamed the independent configuration block to `root_tenant_type` in 02f345c2c and updated the examples, diagnostics, documentation, and tests consistently.

aviator5:
I suggest keeping this PR scoped to Account Management and continuing to use the existing in-memory `TypesRegistryClient`, without changes to Types Registry.

I’m currently implementing the Types Registry P0 migration. The persistent SDK and AM-specific read adapter introduced here overlap with that work and would need to be reworked during the planned cutover:

- [gears/system/types-registry/docs/p0/todo.md — T23–T26](https://github.com/constructorfabric/gears-rust/blob/main/gears/system/types-registry/docs/p0/todo.md#L1952): the shared SDK and reconciliation helper, database cutover, and consumer migration.
- [gears/system/types-registry/docs/p0/plan.md — P12](https://github.com/constructorfabric/gears-rust/blob/main/gears/system/types-registry/docs/p0/plan.md#L415): existing consumers stay on the in-memory path until T24.

For this PR, I propose keeping:

- Root-type configuration and schema generation in AM.
- Registration and semantic validation through the existing client during AM `init()`, before the registry’s ready transition.
- Fatal drift detection regardless of `bootstrap.strict`, and validation of the existing root tenant’s type binding.
- Removal of the duplicated root-type YAML seeds and tests for this behavior.

This would remove the need for the new persistent SDK, operation ownership plumbing, ownership CAS, and `RootTypeAwareRegistryClient`. AM can own the schema’s definition and validation without introducing persisted ownership attribution here.

I’ll take responsibility for migrating this AM path to the persistent registry as part of P0, preserving its create-if-absent / fail-on-drift semantics.

The interim limitations should be explicit: in-memory registration does not provide durability or coordination between replicas, and rereading it does not prove persistence. Let’s track those requirements, including restart and cross-replica tests, under [#4627](https://github.com/constructorfabric/gears-rust/issues/4627), rather than treating them as completed by this PR.

aviator5 (changes requested):

fdlockgraf:
@aviator5 changed the implementation a little - could you please take a look again?

gears/system/account-management/account-management/src/domain/bootstrap/service.rs
```diff
@@ -698,10 +703,12 @@ impl<R: TenantRepo> BootstrapService<R> {
...
-                    return BootstrapState::Terminal(Err(DomainError::internal(format!(
-                        "platform root already exists with a different id; configured root_id={} cannot be inserted (likely a config drift between platform restarts)",
-                        self.cfg.root_id
-                    ))));
+                    return BootstrapState::Terminal(Err(DomainError::RootBindingMismatch {
+                        detail: format!(
+                            "platform root already exists with a different id; configured root_id={} cannot be inserted (likely config drift between platform restarts)",
+                            self.cfg.root_id
+                        ),
+                    }));
```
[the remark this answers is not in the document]
fdlockgraf: Fixed in 5dce67b84. Retry exhaustion now returns ServiceUnavailable, while RootBindingMismatch remains reserved for drift confirmed from a repository row. Added a regression test covering all retry attempts.

gears/system/account-management/account-management/src/domain/root_type.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn omitted_idp_provisioning_defaults_to_false() {
+        let cfg: RootTypeConfig = serde_json::from_value(serde_json::json!({
+            "gts_id": "gts.cf.core.am.tenant_type.v1~cf.core.am.platform.v1~"
```
aviator5, line 67: please use `gts_id!` macros everywhere

aviator5 (approved):

fdlockgraf:
@aviator5 Fixing. Will be in the next commit

fdlockgraf:
@aviator5 Fixing. Will be in the next commit

fdlockgraf:
@aviator5 Fixing. Will be in the next commit

gears/system/account-management/account-management/src/infra/types_registry/root_type.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// Build the single AM-owned concrete root schema.
+pub fn desired_root_schema(cfg: &RootTypeConfig) -> anyhow::Result<Value> {
+    let type_id = cfg.validated_id().map_err(anyhow::Error::msg)?;
+    let parent = types_registry_sdk::GtsTypeSchema::derive_parent_type_id(type_id)
+        .ok_or_else(|| anyhow::anyhow!("root tenant type {type_id} has no concrete parent"))?;
+    Ok(json!({
+        "$id": format!("{GTS_ID_URI_PREFIX}{type_id}"),
+        "$schema": JSON_SCHEMA_DRAFT_07,
+        "description": "Platform-root tenant type owned by Account Management (no parents).",
```
diffora: **MEDIUM**

The generated `description` reads "Platform-root tenant type owned by Account Management (no parents)." while the seed it replaces — removed from `config/quickstart.yaml` and `config/e2e-local.yaml` in this same diff — reads "Platform-root tenant type (no parents)."

The registry compares whole documents (`types-registry/src/infra/storage/in_memory_repo.rs:170` and `:197`: `existing.content == *entity`, otherwise `already_exists`), so any deployment whose config still carries the old seed gets a startup-fatal conflict instead of the idempotent registration this function documents.

Emit the previously shipped `description` string verbatim so the generated document is content-equal to the seed it replaces, or record the incompatibility in an upgrade note.

gears/system/account-management/account-management/src/domain/root_type.rs
```diff
@@ -0,0 +1,98 @@
...
+use toolkit_gts::gts_id;
+use toolkit_macros::domain_model;
+
+/// Abstract AM tenant-type envelope every concrete platform root derives from.
+pub const TENANT_TYPE_BASE: &str = gts_id!("cf.core.am.tenant_type.v1~");
+
+/// AM-owned semantic contract for the concrete platform-root type.
+#[domain_model]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(default, deny_unknown_fields)]
```
diffora: **MEDIUM**

Container-level `#[serde(default)]` together with the hand-written `Default` of `gts_id: gts::GtsTypeId::new("")` lets a `root_tenant_type` block that omits `gts_id` deserialize into an empty, invalid identifier.

This re-introduces the exact empty-string sentinel this same diff deletes from `BootstrapConfig` — both the `Default` value and its `trim().is_empty()` guard in `validate()` — so a missing key is diagnosed late by `validated_id()` instead of failing at deserialization.

Drop `default` from the container attribute, keep `deny_unknown_fields`, move `#[serde(default)]` onto `idp_provisioning` alone, and delete the hand-written `Default`.

gears/system/account-management/account-management/src/gear.rs
```diff
@@ -534,6 +538,41 @@ fn check_task_join(
     }
 }
 
+fn validate_existing_root_binding(
```
diffora: **MEDIUM**

`validate_existing_root_binding` re-implements the root-binding rule that `BootstrapService::classify` already applies at `src/domain/bootstrap/service.rs:964-1004` — root-id equality, the `GtsId::try_new` parse, and `tenant_type_uuid` against the GTS-derived UUID — and both paths read `find_platform_root`.

The two copies have already drifted in their diagnostics ("existing platform root has id=X, but bootstrap.root_id=Y" here versus "platform root already exists with id X, but configured root_id is Y" there), so the same condition reports differently depending on which path detects it.

Move the rule into `domain::root_type` as a single function returning `DomainError::RootBindingMismatch` and call it from both `Gear::init` and `BootstrapService::classify`.

gears/system/account-management/account-management/src/gear.rs
```diff
@@ -664,6 +706,44 @@ impl Gear for AccountManagementGear {
...
+        // is still in configuration mode: registration stages the AM-owned
+        // document, and its system `post_init` validates the complete catalogue
+        // before stateful gears (including AM bootstrap) start.
+        if let Some(root_cfg) = root_type.as_ref() {
+            // `tenant_type_uuid` is the durable create-once binding between an
+            // existing root row and its configured GTS contract. Reject drift
+            // before mutating the process-local catalogue and before non-strict
+            // bootstrap policy can suppress it.
+            if let Some(existing_root) = repo
+                .find_platform_root(&toolkit_db::secure::AccessScope::allow_all())
```
diffora: **MEDIUM**

This is the first query against AM own database issued from `init`, and it carries neither a timeout nor a cancellation guard although `ctx.cancellation_token()` is available.

`run_init_phase` (`libs/toolkit/src/runtime/host_runtime.rs:393-407`) never polls the cancellation token between gears — unlike `run_db_phase`, which checks `self.cancel.is_cancelled()` before every gear at `host_runtime.rs:361-365` — so a stalled database pins the whole sequential startup in an await SIGTERM cannot unwind, which is the opposite of the rule this gear states at `gear.rs:913-918`.

Bound the await with `tokio::time::timeout` plus `ctx.cancellation_token()`, or move the root-binding read to the `serve()`-side path that already holds the runtime `CancellationToken`.

gears/system/account-management/account-management/src/config.rs
```diff
@@ -576,6 +649,10 @@ impl AccountManagementConfig {
         // interval / jitter / initial_delay surfaces here rather than
         // panicking inside the spawned loop on `tokio::time::sleep`.
         let integrity_err = self.integrity_check.validate().err();
+        // Root-type resolution is lifecycle-fatal independently of bootstrap's
+        // strict mode: a malformed or conflicting schema contract must never be
+        // downgraded to best-effort bootstrap behavior.
+        let root_type_err = self.resolved_root_type().err();
```
diffora, line 655: **MEDIUM**

Folding `resolved_root_type()` into `validate()` makes the comment immediately below this line false: it still states that the bootstrap sub-section is NOT validated here and that folding bootstrap validation in would make `strict = false` unreachable.

`resolved_root_type()` returns `Err` when `bootstrap.is_some()` and no root contract is supplied, so a `[bootstrap]` block missing its root type now aborts init before the strict-vs-non-strict branch runs — deliberate and pinned by `bootstrap_without_root_type_is_fatal_even_when_nonstrict`, which leaves the comment as the only incorrect part.

Rewrite the comment to carve out the root-type contract as init-fatal regardless of `bootstrap.strict`, keeping the rest of the bootstrap-validation rationale intact.

gears/system/account-management/account-management/src/config_tests.rs
```diff
@@ -7,6 +28,62 @@ fn default_validates_clean() {
...
+        ..AccountManagementConfig::default()
+    };
+    assert_eq!(
+        cfg.resolved_root_type().expect("legacy config"),
+        Some(root_type(true))
+    );
+}
+
+#[test]
+fn modern_and_legacy_root_type_conflict_is_fatal() {
```
diffora: **MEDIUM**

`modern_and_legacy_root_type_conflict_is_fatal` supplies the same `ROOT_TYPE` on both the modern and the deprecated side, so only the `idp_provisioning` conflict branch of `resolved_root_type` executes.

The `gts_id` mismatch branch never runs and `assert!(error.contains("conflicts"))` matches both messages, so deleting that branch outright would leave this test green.

Add a case whose deprecated `bootstrap.root_tenant_type` differs from `root_tenant_type.gts_id`, and assert each branch full message instead of the shared substring.

gears/system/account-management/account-management/src/domain/bootstrap/service_tests.rs
```diff
@@ -1060,15 +1113,68 @@ fn bootstrap_cfg_long_deadline() -> BootstrapConfig {
...
-async fn run_aborts_after_max_already_exists_streak_when_root_id_drifts() {
-    // Simulate a configured `root_id` that drifted away from the actual
-    // platform root: classify-by-id returns NoRoot for the configured
-    // id, but `insert_root_provisioning` collides with the existing
-    // root via the `parent_id = None` single-root invariant in the
-    // fake repo (which mirrors `ux_tenants_single_root` in production).
-    // The pair would oscillate forever; the streak cap converts that
-    // into a clean Internal after MAX_ALREADY_EXISTS_STREAK = 3 hits.
+#[tokio::test]
+async fn insert_retry_exhaustion_surfaces_service_unavailable() {
```
diffora, line 1121: **MEDIUM**

This test reaches the retry-exhaustion terminal by calling the private `step_insert` directly against a hand-built `RunCtx`, after seeding a parentless tenant with id `0xDEAD`.

With that same seeded state `run()` aborts at classification with `RootBindingMismatch` — proved by `run_rejects_existing_root_id_drift_before_insert` at line 1177 — so nothing shows this `ServiceUnavailable` terminal is reachable through the saga loop the lifecycle actually runs.

Add a `FakeTenantRepo` mode whose `find_platform_root` returns `None` while root inserts still conflict, and drive the exhaustion through `run()`.

diffora (approved):

entropyshift (approved):

PR #4863, opened 2026-09-17
fix(cluster): require lease methods on lock+leader traits

asmith987:
## Summary by CodeRabbit

* **New Features**
  * Added token-based distributed lock operations for acquiring, waiting, renewing, and releasing leases.
  * Added token-based leader-election operations for joining, renewing, and resigning.
  * Redis locks now support cross-process lease ownership with fencing to prevent stale-token updates.
  * Added per-profile limits for blocking lock waiters and a two-minute maximum wait.
* **Bug Fixes**
  * Improved handling of expired, stale, and incorrectly scoped lock tokens.
  * Ensured scoped locks and elections consistently apply configured name prefixes.
* **Tests**
  * Expanded coverage for lease lifecycle, contention, timeouts, fencing, and cross-process behavior.


gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -191,6 +208,52 @@ impl LockNames {
...
+/// `u64` drawn per acquisition. It approximates the guarantee the monotonic one
+/// gives where it actually matters (§5.8.1): a name that lapsed and was
+/// re-acquired draws a *fresh* fence, so a stale holder's token will not match the
+/// successor's lease — a probabilistic bound (a 2⁻⁶⁴ collision aside), not the
+/// monotonic column's certainty, which is more than enough for a coordination
+/// lock. It is **not** a globally monotonic third-party fencing token — ADR-002
+/// declines those (§5.4) and [`LockGuard`] exposes none — only the discriminator
+/// `renew` and `release` predicate on.
+fn fresh_fence() -> u64 {
+    rand::rng().random::<u64>()
```
MikeFalcon77, line 242: **CRITICAL**

The fence is half the authority a cross-process `renew`/`release` fences on now, and `RD-LOCK-017` shows it is the *only* half that separates two acquisitions under the same owner. `rand::rng()` is the thread-local PRNG; for lease authority material draw it from `OsRng`/`getrandom`, the same source `Uuid::new_v4` uses for the owner half in `fresh_owner`.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -455,18 +525,148 @@ impl RedisLock {
             name: logs::LOCK_ACQUIRED,
             provider = self.signals.provider(),
             lock = %name,
-            holder = %token,
+            holder = %value,
```
MikeFalcon77, line 528: **CRITICAL**

`value` is the whole `<owner>:<fence>` credential, and this PR makes it usable from a process that never saw the acquire. Before, the logged token could only be exercised by the in-process guard task; now anything that can read this DEBUG line and reach the backend seam can renew or release the lease, and the gRPC `caller.owns` check only covers the wire path. `lock = %name` already gives §5.5 its traceability.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -802,47 +988,77 @@ impl DistributedLockBackend for RedisLock {
...
+        &self,
+        name: &str,
+        owner: &str,
+        ttl: Duration,
+        timeout: Duration,
+    ) -> Result<LeaseToken, ClusterError> {
+        self.acquire_waiting_for(name, owner, ttl, timeout).await
+    }
+
+    async fn renew(&self, token: &LeaseToken, ttl: Duration) -> Result<(), ClusterError> {
```
MikeFalcon77, line 1020: **HIGH**

The new token-path `renew` never checks `self.shutdown`, while every other entry point in this backend does (`acquire_lease` at the top, `park` in its select, the guard task in its select). `RedisLockHandle::stop()` cancels the token and then calls `close_pool`, so a renew that races a graceful stop comes back as `ClusterError::Provider` instead of `ClusterError::Shutdown`, and `record_lock` then forwards it to `provider_error`, which counts an orderly shutdown as a backend fault.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -802,47 +988,77 @@ impl DistributedLockBackend for RedisLock {
...
+        )
         .instrument(span)
         .await;
-        self.signals.record_lock("lock", name, op_started, &out);
+        self.signals
+            .record_lock("renew", &token.name, started, &out);
+        out
+    }
+
+    async fn release(&self, token: &LeaseToken) -> Result<(), ClusterError> {
```
MikeFalcon77, line 1049: **HIGH**

`release` has the same gap: no `self.shutdown` check, so a release arriving after `stop()` closed the pool answers with a `Provider` error rather than `ClusterError::Shutdown`, and counts on the provider-error signal. A caller doing best-effort cleanup while the process drains gets a hard backend fault here, where the guard path returns cleanly and the trait's absence-is-Ok contract expects a definite answer.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -191,6 +208,52 @@ impl LockNames {
...
+/// A single `SET NX PX` has no counter to increment, so the fence is a random
+/// `u64` drawn per acquisition. It approximates the guarantee the monotonic one
+/// gives where it actually matters (§5.8.1): a name that lapsed and was
+/// re-acquired draws a *fresh* fence, so a stale holder's token will not match the
+/// successor's lease — a probabilistic bound (a 2⁻⁶⁴ collision aside), not the
+/// monotonic column's certainty, which is more than enough for a coordination
+/// lock. It is **not** a globally monotonic third-party fencing token — ADR-002
+/// declines those (§5.4) and [`LockGuard`] exposes none — only the discriminator
+/// `renew` and `release` predicate on.
+fn fresh_fence() -> u64 {
```
MikeFalcon77, line 241: **HIGH**

This mints the fence at random, but the trait it fills in says something stronger: `DistributedLockBackend::acquire` documents "taken over by CAS with `fence + 1`, so the previous holder's token can never match again" and `LeaseToken::fence` says "Bumped on every acquisition of `name`". The Postgres lock implements that literally (`1` or `fence + 1`), so after this PR two in-tree backends give the same field two different guarantees, and the SDK side still promises "never" where Redis gives 2^-64. The reasoning for a random draw is sound for a `SET NX PX` with no counter, but the SDK contract was not updated to admit it.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -773,27 +975,11 @@ impl DistributedLockBackend for RedisLock {
...
-        // `LockContended` lands on the bounded `result` label as `contended` and
-        // *not* on the provider-error counter: someone else holding the name is
-        // this call's answer, not a fault (`result::label`).
-        self.signals.record_lock("try_lock", name, started, &out);
-        out
+        // A fresh owner because the caller did not name one: `try_lock` holds its
+        // own lease. Built *over* `acquire` — one lease, then a guard task — so the
+        // guard path and the token path fence on the same value (DESIGN.md §5.1).
+        let token = self.acquire_once(name, &fresh_owner(), ttl).await?;
+        self.guard_from(token).await
```
MikeFalcon77, line 982: **HIGH**

`acquire_once` has already run `record_lock("try_lock", ..)` with its `Ok` by the time this line executes, so a `stop()` race inside `guard_from` is counted as a successful `try_lock` in `cluster_lock_ops_total` while the caller receives `Shutdown`. Before the refactor the shutdown re-check sat inside `try_acquire`, inside the instrumented block, so it landed on the `result` label. The `cluster.lock.abandoned` log was kept in step with the move but the metric was not.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -802,47 +988,77 @@ impl DistributedLockBackend for RedisLock {
...
     ) -> Result<LockGuard, ClusterError> {
-        // Instrumented once *around* the loop rather than inside it, following
-        // the postgres plugin's shape: a blocking acquisition is one operation
-        // however many `SET NX`s it takes, so one span covering the whole wait
-        // and one duration measuring it is what a consumer asked for. Recording
-        // per attempt would report a 30 s wait as a hundred fast operations.
+        let token = self
+            .acquire_waiting_for(name, &fresh_owner(), ttl, timeout)
+            .await?;
+        self.guard_from(token).await
```
MikeFalcon77, line 994: **HIGH**

Same split as in `try_lock`: `acquire_waiting_for` has already recorded `"lock"` as `ok`, so a `Shutdown` raised by `guard_from` here is invisible to `cluster_lock_ops_total` and the caller's error never shows up on the `result` label. The duration histogram also stops before the guard hand-off.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -455,18 +525,148 @@ impl RedisLock {
             name: logs::LOCK_ACQUIRED,
             provider = self.signals.provider(),
             lock = %name,
-            holder = %token,
+            holder = %value,
             "cluster.lock.acquired: this instance now holds the lease under this token"
         );
 
-        let (commands, guard) = LockGuard::channel(name.to_owned(), GUARD_COMMAND_BUFFER);
+        Ok(Some(LeaseToken::new(name, owner, fence)))
```
MikeFalcon77, line 532: **HIGH**

The post-`WAIT` shutdown re-check that used to sit right here is gone, and `guard_from` only restores it for the guard path. On the token path a `stop()` that races this `SET NX` now leaves the lease key in Redis and hands the token back while the pool is closing, so the name is wedged for a full TTL. At this point no caller holds the token yet, which is the same "provably ours and provably unheld" state `guard_from` abandons on, so the asymmetry with `cpt-cf-clst-fr-shutdown-ttl-cleanup` does not really apply.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
MikeFalcon77, line 1020: **HIGH**

This comment names the whole point of the token half, renewing a lease acquired through a different `RedisLock` handle, but RD-LOCK-016/017/018 all run against the single `lock` that `fixture` returns. Invariant I7 is the part a Profile-3 deployment depends on and it is the part not covered. RD-LOCK-011 already builds two independent handles for the guard path; can we do the same here, acquire on A and renew/release on B?

gears/system/cluster/cluster-sdk/src/lock/backend.rs
```diff
@@ -36,9 +36,13 @@ use crate::lock::types::LockFeatures;
 /// operate on that token. The second half exists because a remote caller must be
 /// able to renew from somewhere other than the acquiring task — and, once the lease
 /// is a record in the store, from a *replica that never saw the acquire* (§5.8.1,
-/// invariant I7). A backend should serve both from one lease rather than two
-/// mechanisms; the four lease methods are defaulted, so one that has not yet done
-/// so still compiles and reports [`ClusterError::Unsupported`] (invariant I11).
+/// invariant I7). A backend serves both from one lease rather than two mechanisms;
+/// the four lease methods are **required**, so the compiler forces every backend to
```
MikeFalcon77, line 40: **HIGH**

Making the four lease methods required turns the plugin-facing SDK trait into a breaking change for every out-of-tree `DistributedLockBackend`. `docs/DESIGN.md` invariant I11 still reads "Extensions are defaulted and dyn-safe, never required methods", ADR-011 says the `*Backend` traits "cannot acquire a required method", ADR-012 says the four lease methods are defaulted "invariant I11 kept", and `cpt-cf-clst-nfr-plugin-stability` rests on the same promise. None of those are touched in this PR, so the SDK contract and the documents that define it now disagree. If we are reversing I11 deliberately, that decision needs to land with the trait change.

gears/system/cluster/cluster-sdk/src/leader/backend.rs
```diff
@@ -103,22 +102,20 @@ pub trait LeaderElectionBackend: Send + Sync {
     /// election is an ordinary outcome, not an error: the caller becomes a
     /// follower and retries on its own cadence.
     ///
+    /// Required, not defaulted: `elect` is built over this on both sides — the
```
MikeFalcon77, line 105: **HIGH**

`join`, `renew` and `resign` are now required on the plugin-facing trait, so any out-of-tree `LeaderElectionBackend` stops compiling. ADR-012 still states these lease methods are defaulted and dyn-safe with "invariant I11 kept", and DESIGN.md's invariant table still says extensions to the `*Backend` traits are never required methods. Either those change in the same PR or the methods stay defaulted.

gears/system/cluster/cluster-sdk/src/lock/backend.rs
```diff
@@ -241,9 +216,13 @@ pub trait DistributedLockBackend: Send + Sync {
...
-/// The [`ClusterError::Unsupported`] feature name a backend without store-owned
-/// leases reports. Shared with [`LeaderElectionBackend`], since a provider
-/// implements the model for both primitives or neither.
+/// The canonical feature name for the store-owned-leases model, shared by the
+/// lock and [`LeaderElectionBackend`] primitives.
+///
+/// No default body returns it any more: the four lock lease methods and the three
+/// leader lease methods are all **required**, so a backend can no longer compile
+/// while silently lacking the model. Retained as the one spelling of the feature
+/// name for any diagnostic that still needs to name it.
```
MikeFalcon77, line 225: **HIGH**

With the default bodies gone, nothing in the workspace references `STORE_OWNED_LEASES` any more: the only two occurrences left are this definition and the `pub use` in `lock/mod.rs`. It is still public SDK surface a plugin author has to read past, for a diagnostic that does not exist.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
```diff
@@ -56,41 +59,54 @@ impl DistributedLockBackend for ScopedDistributedLockBackend {
...
         owner: &str,
         ttl: Duration,
     ) -> Result<LeaseToken, ClusterError> {
-        self.inner
+        let mut token = self
+            .inner
             .acquire(&scope::apply(&self.prefix, name), owner, ttl)
-            .await
+            .await?;
+        token.name = scope::strip(&self.prefix, &token.name).to_owned();
```
MikeFalcon77, line 76: **HIGH**

`scope::strip` is `strip_prefix(..).unwrap_or(key)`, so this silently leaves the name alone when it does not carry the prefix, while `renew`/`release` re-apply the prefix unconditionally. The pair is only correct for a token this exact wrapper produced: a token acquired through the raw backend, or through a differently-scoped view, gets a second prefix on renew and resolves to nothing, which the backend reports as `LockExpired` and is indistinguishable from a lapsed lease. Nothing in the type stops a caller from crossing the two.

gears/system/cluster/cluster-sdk/src/leader/scoped.rs
```diff
@@ -53,27 +56,40 @@ impl LeaderElectionBackend for ScopedLeaderElectionBackend {
...
         config: ElectionConfig,
     ) -> Result<Option<LeaseToken>, ClusterError> {
-        self.inner
+        let token = self
+            .inner
             .join(&scope::apply(&self.prefix, name), owner, config)
-            .await
+            .await?;
+        Ok(token.map(|mut token| {
+            token.name = scope::strip(&self.prefix, &token.name).to_owned();
```
MikeFalcon77, line 74: **HIGH**

Same asymmetry as on the lock wrapper: `scope::strip` no-ops when the prefix is absent, but `renew`/`resign` call `scope::apply` unconditionally. A token that reached this wrapper from anywhere other than its own `join` is rewritten to `prefix/prefix/name` and fails as `LockExpired`, with nothing in the type or the call to catch it.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
```diff
@@ -56,41 +59,54 @@ impl DistributedLockBackend for ScopedDistributedLockBackend {
...
         ttl: Duration,
         timeout: Duration,
     ) -> Result<LeaseToken, ClusterError> {
-        self.inner
+        let mut token = self
+            .inner
             .acquire_waiting(&scope::apply(&self.prefix, name), owner, ttl, timeout)
-            .await
+            .await?;
+        token.name = scope::strip(&self.prefix, &token.name).to_owned();
```
MikeFalcon77, line 92: **HIGH**

Nothing in the test module calls `wrapper.acquire_waiting`, so this strip is never run. If it drifts from `acquire`'s, a brokered blocking lock hands the consumer a token whose `name` still carries the prefix, and the next `renew` re-applies it, double-prefixes the key, and misses the record the acquire created. Can we add an `acquire_waiting` leg to `acquire_prepends_the_prefix_and_strips_the_returned_name`?

gears/system/cluster/cluster/src/domain/wiring_tests.rs
```diff
@@ -215,6 +215,29 @@ impl LeaderElectionBackend for MarkerLeaderElectionBackend {
         self.calls.fetch_add(1, Ordering::SeqCst);
         self.inner.elect_with_config(name, config).await
     }
+
+    // The token half counts and delegates just like the `elect` half, so
+    // `explicit_backends_override_defaults` proves the explicit binding serves the
+    // over-the-wire `Join`/`Renew`/`Resign` methods too, not only `elect`.
+    async fn join(
```
MikeFalcon77, line 222: **HIGH**

The comment says `explicit_backends_override_defaults` proves the explicit binding serves Join/Renew/Resign, but that test only calls `elect("primary")` and then asserts `leader_calls == 1`. None of these three counters is ever incremented, and adding a join call would in fact break the `== 1` assertion. Either drive the token path and raise the expected count, or drop the claim.

gears/system/cluster/cluster/src/domain/wiring_tests.rs
```diff
@@ -242,6 +265,40 @@ impl DistributedLockBackend for MarkerLockBackend {
         self.calls.fetch_add(1, Ordering::SeqCst);
         self.inner.lock(name, ttl, timeout).await
     }
+
+    // The token half counts and delegates just like the guard half, so
+    // `explicit_backends_override_defaults` proves the explicit binding serves the
+    // over-the-wire methods too, not only `try_lock`/`lock`.
+    async fn acquire(
```
MikeFalcon77, line 272: **HIGH**

Same as the leader marker above: `explicit_backends_override_defaults` calls only `try_lock` and asserts `lock_calls == 1`, so `acquire`, `acquire_waiting`, `renew` and `release` here are dead. The binding of the token path, which is the half a gear serving lock RPCs actually uses, stays unverified.

gears/system/cluster/cluster/src/config_tests.rs
```diff
@@ -468,6 +468,44 @@ impl DistributedLockBackend for FakeNativeLock {
             name: name.to_owned(),
         })
     }
+
+    // Every entry point bumps the counter (see the type doc), the token half
+    // included, so the mixed-profile test proves the *native* backend serves the
+    // over-the-wire lock methods and not the CAS default the omit path would fill.
+    async fn acquire(
```
MikeFalcon77, line 475: **HIGH**

`from_config_wires_a_mixed_backend_profile` calls only `try_lock` and then asserts `lock_calls == 1`, so none of these four token methods is reached. The comment claims the mixed-profile test proves the native backend serves the over-the-wire lock methods, and it does not: routing a natively-bound provider's token path is exactly the gap this PR is closing elsewhere.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -387,31 +451,48 @@ impl RedisLock {
...
+    /// consumer-visible token, does the re-check-and-discard in
+    /// [`guard_from`](Self::guard_from) instead.
     ///
     /// # Errors
-    /// [`ClusterError::Shutdown`] when `stop()` has run or races this, and
-    /// whatever [`map_redis_error`] makes of a failing `SET` or `WAIT`.
-    async fn try_acquire(
+    /// [`ClusterError::Shutdown`] when `stop()` has already run, and whatever
+    /// [`map_redis_error`] makes of a failing `SET` or `WAIT`.
+    async fn acquire_lease(
```
MikeFalcon77, line 475: The doc above covers the `stop()` race but not cancellation, and this future is dropped by the `tokio::time::timeout` in `attempt` and by the gRPC handler when a client disconnects or its deadline fires. A drop after the `SET NX PX` commits loses the fence along with the future, so nothing can release that lease and the name stays wedged for the full `ttl`. A `// NOT cancel-safe:` line here stating that the orphan window is the caller's `ttl` would make the story auditable.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -802,47 +988,77 @@ impl DistributedLockBackend for RedisLock {
...
+    }
+
+    async fn acquire_waiting(
+        &self,
+        name: &str,
+        owner: &str,
+        ttl: Duration,
+        timeout: Duration,
+    ) -> Result<LeaseToken, ClusterError> {
+        self.acquire_waiting_for(name, owner, ttl, timeout).await
```
MikeFalcon77, line 1017: Before this the redis backend had no `acquire_waiting`, so the gRPC `Lock` RPC never reached `wait_for_lease`. Now every wire caller parks a task re-running `SET NX` plus `PTTL` on a jittered delay capped at the 250 ms `HEARTBEAT`, which the `wake_delay` docs put at about eight `SET NX`s a second, for a timeout `api/grpc/lock.rs` clamps only at `FENCE_RETENTION_DEFAULT` (one hour). Nothing caps how many waiters one hot name can accumulate, so a contended lock becomes sustained polling load on the command pool.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
```diff
@@ -802,47 +988,77 @@ impl DistributedLockBackend for RedisLock {
...
+        &self,
+        name: &str,
+        owner: &str,
+        ttl: Duration,
+    ) -> Result<LeaseToken, ClusterError> {
+        // The same lease `try_lock` takes, minus the guard task: this caller holds
+        // the token itself and renews or releases against it from wherever it likes
+        // — the store-owned-leases half a gear serving lock RPCs over the wire needs
+        // (§6.5, §12.6). No discard on a `stop()` race here; see `acquire_lease`.
+        self.acquire_once(name, owner, ttl).await
```
MikeFalcon77, line 1007: `acquire_once` emits the `LOCK_TRY_LOCK` span and `record_lock("try_lock", ..)`, so a brokered token-path acquire is indistinguishable from an in-process `try_lock` in both the trace and `cluster_lock_ops_total`. `acquire_waiting` has the same problem under the `lock` label. That is exactly the split an operator needs when Profile 3 lock RPCs start failing and the in-process guard path is healthy.

gears/system/cluster/cluster-sdk/src/lock/backend.rs
```diff
@@ -337,46 +357,52 @@ mod tests {
...
+    /// The four lease methods are **required**, not defaulted, so a backend that
+    /// serves the guard path must serve the token path too — the store-owned-leases
+    /// half the gear serves every remote lock RPC through. The `held` flag drives
+    /// both token-path acquisitions through their real branch (acquired vs
+    /// contended/timed-out), so the assertions exercise stub behaviour rather than
+    /// echoing the stub's own input; `renew`/`release` are checked for reachability.
     #[tokio::test]
-    async fn the_lease_methods_are_defaulted_and_report_unsupported() {
-        let backend = StubBackend { held: false };
+    async fn the_lease_methods_are_implemented_not_defaulted() {
```
MikeFalcon77, line 367: Now that the four methods have no default bodies, this test runs no cluster-sdk code at all: every assertion checks `StubBackend`'s own `if self.held` branches, defined 80 lines up in this same file. The property the name claims, required rather than defaulted, is enforced by the compiler, not by these asserts, so this passes unchanged no matter what the SDK does. Either drop it or point it at something with behaviour, like `ScopedDistributedLockBackend`.

gears/system/cluster/cluster-sdk/src/leader/backend.rs
```diff
@@ -218,51 +208,78 @@ mod tests {
...
-        ));
-        assert!(matches!(
-            backend.renew(&token, Duration::from_secs(30)).await,
-            Err(ClusterError::Unsupported { .. })
-        ));
-        assert!(matches!(
-            backend.resign(&token).await,
-            Err(ClusterError::Unsupported { .. })
-        ));
+    async fn the_lease_methods_are_implemented_not_defaulted() {
```
MikeFalcon77, line 236: With the defaults gone this test executes no cluster-sdk code: the winner/loser split it asserts is `self.wins.then(...)` in `StubBackend::join` just above, written by the test itself. The doc comment says the assertions exercise stub behaviour, which is exactly the problem, and whether the methods are required is a compiler fact rather than something an assert can catch.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
```diff
@@ -173,6 +225,65 @@ mod tests {
         );
     }
 
+    /// The token path scopes the same way the guard path does on the *write* side —
+    /// the name the backend records is the `prefix/name` the wrapper composed — but,
+    /// unlike the opaque guard, the returned [`LeaseToken::name`] is stripped back to
+    /// the bare consumer name, because that field is contractually unprefixed.
+    #[tokio::test]
+    async fn acquire_prepends_the_prefix_and_strips_the_returned_name() {
```
MikeFalcon77, line 233: `RecordingBackend` has no failure mode, so neither new test covers what a caller sees when the inner backend contends. The `?` in `acquire` forwards `LockContended { name }` untouched, and that `name` is the scoped one, so the prefix leaks out on the error path while the success path is careful to strip it. Worth a case with an inner backend that returns LockContended.

gears/system/cluster/cluster-sdk/src/leader/scoped.rs
```diff
@@ -149,6 +192,67 @@ mod tests {
         );
     }
 
+    /// The token path scopes the same way `elect` does on the *write* side — the
+    /// name the backend records is the `prefix/name` the wrapper composed — but the
+    /// returned [`LeaseToken::name`] is stripped back to the bare election name,
+    /// because that field is contractually unprefixed.
+    #[tokio::test]
+    async fn join_prepends_the_prefix_and_strips_the_returned_name() {
```
MikeFalcon77, line 200: `RecordingBackend::join` always returns `Some`, so the `None` arm of the wrapper's `token.map(...)` is never run by either new test. Losing an election is the ordinary outcome per the trait docs, and nothing here pins that a follower still gets `Ok(None)` back out of the scoped wrapper. Can we add a follower case?

MikeFalcon77 (changes requested):

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: If I'm not mistaken, `rand::rng()` is cryptographically unpredictable, but I did modify the logic to use the same OS `getrandom` source as the UUID owner.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Fixed. Removed the value from the log.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Added a guard that checks to see if the shutdown token was cancelled, so `renew` is consistent with other entry points.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Added the shutdown guard to `release`.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: I've updated the docs to incorporate this information

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Fixed by moving the `try_lock` span and `record_lock` to wrap the full acquire-plus-`guard_from` hand-off (via an uninstrumented acquire core), so a `Shutdown` raised while handing out the guard now lands on the `result` label instead of being pre-counted as `ok`.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Fixed together with the `try_lock` case — `lock`'s span and duration now cover `wait_for_lease` plus `guard_from`, so a shutdown race during the guard hand-off is recorded on the `result` label and included in the op duration.

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Added tests to cover the gaps.

gears/system/cluster/cluster-sdk/src/lock/backend.rs
asmith987: I've updated all the docs to state that the four lease methods are required. This is to prevent silent runtime failures from occurring in OOP deployment profiles.

gears/system/cluster/cluster-sdk/src/leader/backend.rs
asmith987: Same as the lock trait, I've updated the documents to state the truth now.

gears/system/cluster/cluster-sdk/src/lock/backend.rs
asmith987: Removed the variable.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
asmith987: Adopted scope-identity on the token as the fix. A token presented to a different scoped view now fails.

gears/system/cluster/cluster-sdk/src/leader/scoped.rs
asmith987: Fixed. Used the same scope-identity fix as the lock.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
asmith987: Fixed. Added a test case for this.

gears/system/cluster/cluster/src/domain/wiring_tests.rs
asmith987: Updated the assertion to include the expected count so the test matches the claim.

gears/system/cluster/cluster/src/domain/wiring_tests.rs
asmith987: Extended the test to resolve the raw lock backend and calls the other methods, so the binding is now verified on the token path and not just on `try_lock`

gears/system/cluster/cluster/src/config_tests.rs
asmith987: Fixed. I updated the test to include token-path leg that calls `acquire`

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Updated the docs

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: I added a max clamp of 2 minutes to the blocking wait and a per-profile max limit of concurrent lock waiters (256).

gears/system/cluster/plugins/redis-cluster-plugin/src/lock/mod.rs
asmith987: Added new span/labels that are specifically associated to `acquire` and `acquire_waiting`, so there is no confusion.

gears/system/cluster/cluster-sdk/src/lock/backend.rs
asmith987: Dropped the test.

gears/system/cluster/cluster-sdk/src/leader/backend.rs
asmith987: Dropped the test.

gears/system/cluster/cluster-sdk/src/lock/scoped.rs
asmith987: Added a test to cover what the caller sees when the inner backend contents. Additionally, updated the token read path to strip the scope prefix from the name, so the error path returns the name similar to the success path.

gears/system/cluster/cluster-sdk/src/leader/scoped.rs
asmith987: Added the test case to cover the gap.

PR #4869, opened 2026-09-18
feat(toolkit-contract): make HTTP client connection pool and concurrency configurable

mattgarmon:
Generated REST clients previously inherited toolkit-http's defaults for the connection pool and concurrency limiter, which throttled high-concurrency gear-to-gear traffic: the idle-pool cap (32) and the client-side concurrency cap (100) were both below the ~100 concurrent requests such callers make, so connections churned and requests past 100 were shed with `HttpError::Overloaded`.
 
Expose these as tuning knobs on `ClientConfig` (and via `ClientTuning` wiring so they're settable from a gear's `client_wiring` config:
 
- `pool_max_idle_per_host` (default 256)
- `pool_idle_timeout`       (default 90s)
- `max_concurrent_requests` (default 256)
 
Defaults are raised above the inherited toolkit-http values and aligned so the idle pool can actually be exercised before the concurrency limiter sheds load. Add a matching `HttpClientBuilder::rate_limit` setter to toolkit-http (the config field previously had no builder method).
 
`build_default_http_client` now takes `&ClientConfig` instead of `require_tls: bool`, reading only the transport-construction fields; call-time fields (`timeout`, `retry`, `streaming`, `credentials`) remain applied per call by the generated client.

## Summary by CodeRabbit

* **New Features**
  * Added configurable HTTP connection-pool settings, including idle connection capacity and timeout.
  * Added maximum concurrent request limits to improve traffic control.
  * Added wiring options to customize these settings when configuring clients.
  * Added support for applying rate limits to HTTP clients.
* **Tests**
  * Added coverage for default and customized connection-pool and concurrency settings.


MikeFalcon77 (commented):
32 findings from nonstop-reviewer (run 36).

libs/toolkit-contract-macros/src/rest_contract.rs
```diff
@@ -494,7 +494,7 @@ fn generate_client_struct(model: &RestContractModel, support: &TokenStream) -> T
             ) -> ::std::result::Result<Self, ::toolkit_http::HttpError> {
                 let http = #support::runtime::client::build_default_http_client(
                     #metrics_label,
-                    config.require_tls,
+                    &config,
```
MikeFalcon77, line 497: **HIGH**

This is now the only place four `ClientConfig` fields are read (`require_tls`, both pool knobs and `max_concurrent_requests`), yet `with_http_client` below takes the same `ClientConfig` and honours none of them. A caller that sets `require_tls` and supplies its own client gets no TLS enforcement and no sign the field was dropped.

MikeFalcon77, line 497: **HIGH**

The doc on this generated `new()` still says the transport gets retry-off, `require_tls` and otel, but it now also applies the pool and concurrency knobs off `config`. That doc is what SDK users read, so it should mention that `pool_max_idle_per_host`, `pool_idle_timeout` and `max_concurrent_requests` are consumed at construction time.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -119,11 +123,21 @@ where
 #[cfg(feature = "otel")]
 pub fn build_default_http_client(
     client_type: &str,
-    require_tls: bool,
+    config: &ClientConfig,
```
MikeFalcon77, line 126: **HIGH**

`build_default_http_client` is `pub` and its second parameter went from `bool` to `&ClientConfig`, so any out-of-tree caller stops compiling. `cf-gears-toolkit-contract` is published (0.3.2), so this release needs the version bump that goes with a breaking public API change.

MikeFalcon77, line 126: **HIGH**

This takes the whole `ClientConfig` but reads only `require_tls` and the three pool/concurrency fields, and the rule that the rest must not be wired here is now kept by a hand-written comment in both cfg arms. A transport-only struct (or the four values passed explicitly) would make that split something the compiler checks instead of something a future editor has to read.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -119,11 +123,21 @@ where
...
 ) -> Result<toolkit_http::HttpClient, toolkit_http::HttpError> {
+    // Only *transport-construction* fields belong here: transport security and
+    // the connection-pool / concurrency knobs. `config.timeout`, `config.retry`,
+    // the streaming knobs, and `internal_token_provider` are applied *per call*
+    // by the generated client — do NOT wire them onto the builder or they'll be
```
MikeFalcon77, line 131: **HIGH**

This says `config.timeout` is applied per call and must not be wired onto the builder, but `HttpClientBuilder::build` unconditionally installs `TimeoutLayer::new(self.config.request_timeout)` and nothing here overrides that default of 30s. So a `ClientConfig` with `timeout` above 30s (or a wiring `timeout: "2m"`) still has its connect plus response-header phase cut at 30s, surfacing as `TransportError::Network(HttpError::Timeout(30s))`, and the same cap applies to the streaming `open_timeout`, which generated code also derives from `config.timeout`.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -119,11 +123,21 @@ where
...
+    // double-applied (or override the SDK's own retry/timeout handling).
     toolkit_http::HttpClient::builder()
         .retry(None)
-        .transport(transport_security(require_tls))
+        .transport(transport_security(config.require_tls))
+        .pool_max_idle_per_host(config.pool_max_idle_per_host)
+        .pool_idle_timeout(config.pool_idle_timeout)
+        .rate_limit(Some(toolkit_http::RateLimitConfig {
```
MikeFalcon77, line 138: **HIGH**

This installs the concurrency limiter, and a shed request comes back as `HttpError::Overloaded`, which `send_unary` folds into `TransportError::Network` via `TransportError::network`. A shed request never left the process, but the caller can't tell it apart from a mid-flight reset, so a caller that won't retry a non-idempotent `Network` failure gives up on a request that was never sent. Now that the cap is operator-tunable, `Overloaded` deserves its own `TransportError` variant.

MikeFalcon77, line 138: **HIGH**

Nothing asserts these knobs actually land on the transport. `HttpClient` exposes no accessors for them, so if `.rate_limit(...)` were dropped, or `pool_max_idle_per_host` wired to the wrong builder method, every test in the crate still passes. A behavioural test that builds a client with a tiny `max_concurrent_requests` and expects `HttpError::Overloaded` on the excess request would pin this.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -143,11 +157,19 @@ pub fn build_default_http_client(
...
-    require_tls: bool,
+    config: &ClientConfig,
 ) -> Result<toolkit_http::HttpClient, toolkit_http::HttpError> {
+    // See the `otel` sibling above: only transport-construction fields belong
+    // here; call-time fields (`timeout`, `retry`, streaming, credentials) are
+    // applied per call by the generated client, not on this builder.
     toolkit_http::HttpClient::builder()
         .retry(None)
-        .transport(transport_security(require_tls))
+        .transport(transport_security(config.require_tls))
```
MikeFalcon77, line 167: **HIGH**

This builder chain leaves `HttpClientConfig`'s default 30s `request_timeout` in place, so on a non-otel build a `ClientConfig::timeout` larger than 30s never takes effect: the attempt dies at 30s and surfaces as a `Network` error rather than `TransportError::Timeout`.

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -143,11 +157,19 @@ pub fn build_default_http_client(
...
+        .pool_max_idle_per_host(config.pool_max_idle_per_host)
```
MikeFalcon77, line 168: **HIGH**

The two cfg variants of this chain are now identical except for `.with_metrics(client_type)`. A knob added to one arm and forgotten in the other compiles fine and only misbehaves in a build with the other feature setting; can we pull the shared part into one helper that returns the builder?

libs/toolkit-contract/src/runtime/client.rs
```diff
@@ -143,11 +157,19 @@ pub fn build_default_http_client(
...
+        .pool_idle_timeout(config.pool_idle_timeout)
+        .rate_limit(Some(toolkit_http::RateLimitConfig {
```
MikeFalcon77, line 170: **HIGH**

This cfg variant repeats the same four builder calls by hand and no test covers it either, so a knob added to the `otel` sibling and forgotten here is invisible until someone reads both functions. Either add a test that builds through this configuration and checks the limiter is in place, or factor the shared builder chain into one helper both variants call.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -163,12 +163,50 @@ pub struct ClientConfig {
...
+    /// low enough that gear-to-gear calls at ~100 concurrent spent most of their
+    /// CPU in `connect(2)`). REST transport only.
+    ///
+    /// This matters for high-concurrency service-to-service calls: when the cap
+    /// is below the number of concurrent requests, hyper closes the excess
+    /// connections as they go idle and reopens them for the next request,
+    /// producing a `connect(2)` storm (ephemeral-port allocation + kernel
+    /// spinlock contention) that dominates CPU. Keep it at or above the expected
+    /// per-upstream concurrency so keep-alive actually reuses connections.
+    pub pool_max_idle_per_host: usize,
```
MikeFalcon77, line 186: **HIGH**

`ClientConfig` has all-`pub` fields and no `#[non_exhaustive]`, so each of these three additions breaks any downstream struct literal. The release is already breaking; marking the struct `#[non_exhaustive]` now keeps the next field from being another break, since `new()` plus the `with_*` chain is already the intended construction path.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -163,12 +163,50 @@ pub struct ClientConfig {
...
+    /// How long an idle HTTP/1.1 keep-alive connection is retained before it is
+    /// closed and removed from the pool. The companion of
+    /// [`pool_max_idle_per_host`](Self::pool_max_idle_per_host): that field
+    /// bounds *how many* idle connections are kept, this bounds *how long*. A
+    /// connection reaped here and reopened on the next request incurs the same
+    /// `connect(2)` cost the pool is meant to avoid, so keep this comfortably
+    /// above the gap between successive bursts to a given upstream. `None` uses
+    /// hyper-util's default. Defaults to 90s (the historical `toolkit-http`
+    /// default). REST transport only.
+    pub pool_idle_timeout: Option<Duration>,
```
MikeFalcon77, line 196: **HIGH**

`None` is not really a distinct mode here: `HttpClientBuilder::build` just skips the hyper-util setter, so hyper-util's own default applies, and that default is 90s as far as I can tell, the same value `Some(Duration::from_secs(90))` produces. `HttpClientBuilder::pool_idle_timeout` also documents `None` as "disable idle timeout (connections kept indefinitely)", which contradicts this doc, so a caller picking `None` cannot tell what they get.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -163,12 +163,50 @@ pub struct ClientConfig {
...
+    /// Maximum number of in-flight requests allowed through this client at once
+    /// (across all upstream hosts). Requests beyond the cap are shed immediately
+    /// with `toolkit_http::HttpError::Overloaded` (fail-fast, never queued).
```
MikeFalcon77, line 199: **HIGH**

"fail-fast, never queued" is not what a caller of a `#[retryable]` method sees. `send_unary` wraps the `Overloaded` coming out of `send()` in `TransportError::Network`, and `is_transient()` returns true for that, so the SDK retry loop re-issues a locally shed request up to `retry.max_attempts` times with backoff, adding load exactly when the limiter is saturated.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -163,12 +163,50 @@ pub struct ClientConfig {
...
+    /// Defaults to 256; set to `usize::MAX` to disable the limiter entirely. REST
+    /// transport only.
+    ///
+    /// Keep this at or above
+    /// [`pool_max_idle_per_host`](Self::pool_max_idle_per_host) so the idle
+    /// connection pool can be fully reused — a cap below the pool size sheds
+    /// load before keep-alive reuse takes effect. The default is aligned with it
+    /// for that reason; the inherited `toolkit-http` default (100) silently
+    /// capped exactly the high-concurrency scenario the pool tuning targets.
+    pub max_concurrent_requests: usize,
```
MikeFalcon77, line 209: **HIGH**

`usize::MAX` as the "no limiter" value sits oddly next to `pool_idle_timeout`, which spells absence as an `Option`. `toolkit_http` already models this as `rate_limit: None` / `RateLimitConfig::unlimited()`, and because `build_default_http_client` now always passes `Some(..)`, that `None` arm is unreachable from a generated client.

MikeFalcon77, line 209: **HIGH**

This file's module doc says `ClientConfig` is shared by the generated REST client and the generated gRPC client, so these hyper-specific knobs are now callable on a config headed for `GrpcClient::connect`, which reads only `base_url`, `timeout` and `require_tls`. A doc sentence saying "REST transport only" is the only thing keeping `with_max_concurrent_requests` on a gRPC client from being a silent no-op.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -191,6 +229,9 @@ impl ClientConfig {
             retry: RetryConfig::default(),
             stream_reconnect: ReconnectConfig::default(),
             require_tls: false,
+            pool_max_idle_per_host: 256,
```
MikeFalcon77, line 232: **HIGH**

These defaults raise the per-client resource ceiling a long way: 256 idle connections per upstream host (the inherited `toolkit-http` default was 32) and 256 in-flight requests (was 100). Both the pool and the limiter live on the `HttpClient`, and every generated REST client builds its own, so a process that wires a dozen clients can now hold thousands of sockets across the 90s idle window with no process-wide budget anywhere. Can we leave the default near the old value and have the gears that measured the `connect(2)` problem opt in through `ClientTuning`?

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -235,6 +276,30 @@ impl ClientConfig {
...
+    #[must_use]
+    pub fn with_pool_max_idle_per_host(mut self, max: usize) -> Self {
+        self.pool_max_idle_per_host = max;
+        self
+    }
+
+    /// Override how long idle keep-alive connections are retained (`None` uses
+    /// hyper-util's default). See [`Self::pool_idle_timeout`].
+    #[must_use]
+    pub fn with_pool_idle_timeout(mut self, timeout: Option<Duration>) -> Self {
```
MikeFalcon77, line 290: **HIGH**

The `None` case is the only behaviour this setter has beyond storing a value, and no test exercises it. `multipart_roundtrip.rs` passes `None` but asserts nothing about the resulting transport, and `ClientTuning::apply_to` only ever passes `Some`, so the documented hyper-util-default path is never checked.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -235,6 +276,30 @@ impl ClientConfig {
...
+        self.pool_idle_timeout = timeout;
+        self
+    }
+
+    /// Override the max concurrent in-flight requests (`usize::MAX` disables the
+    /// limiter). See [`Self::max_concurrent_requests`].
+    #[must_use]
+    pub fn with_max_concurrent_requests(mut self, max: usize) -> Self {
```
MikeFalcon77, line 298: **HIGH**

Nothing rejects `0` here, and `max_concurrent_requests: 0` is also reachable straight from a wiring config key. It ends up at `ConcurrencyLimitLayer::new(0)`, whose semaphore never hands out a permit, so every request through the client is shed with `HttpError::Overloaded` for the life of the process. Clamp it to at least 1, the way `toolkit-http` already clamps `buffer_capacity`.

libs/toolkit-contract/src/wiring.rs
```diff
@@ -49,6 +49,29 @@ pub struct ClientTuning {
     #[serde(default)]
     pub require_tls: Option<bool>,
 
+    /// Max idle keep-alive connections per upstream host. Raise this to at least
+    /// the expected per-upstream request concurrency so HTTP/1.1 connections are
+    /// reused instead of churned under load. Missing keeps the SDK default
+    /// ([`ClientConfig::pool_max_idle_per_host`](crate::runtime::config::ClientConfig::pool_max_idle_per_host)).
+    #[serde(default)]
+    pub pool_max_idle_per_host: Option<usize>,
```
MikeFalcon77, line 57: **HIGH**

`ClientTuning` is flattened into both the `Rest` and `Grpc` variants, so these three keys parse fine on a `transport: grpc` document and then go nowhere: the generated gRPC client reads only `base_url`, `timeout` and `require_tls`. `ClientConfig` marks the same fields "REST transport only"; that caveat is missing here, and nothing rejects or warns on a grpc wiring that sets them.

libs/toolkit-contract/src/wiring.rs
```diff
@@ -141,6 +164,15 @@ impl ClientTuning {
         if let Some(require_tls) = self.require_tls {
             cfg = cfg.with_require_tls(require_tls);
         }
+        if let Some(max) = self.pool_max_idle_per_host {
```
MikeFalcon77, line 167: **HIGH**

`pool_max_idle_per_host` is forwarded from config with no upper bound and no check against `max_concurrent_requests`. Whatever number the config file says becomes the number of idle sockets this client may hold per upstream host for the whole idle window, and a value above the concurrency cap quietly breaks the pool/concurrency relation both doc comments tell operators to keep.

libs/toolkit-contract/src/wiring.rs
```diff
@@ -141,6 +164,15 @@ impl ClientTuning {
...
+            cfg = cfg.with_pool_max_idle_per_host(max);
+        }
+        if let Some(timeout) = self.pool_idle_timeout {
+            cfg = cfg.with_pool_idle_timeout(Some(timeout));
+        }
+        if let Some(max) = self.max_concurrent_requests {
+            cfg = cfg.with_max_concurrent_requests(max);
```
MikeFalcon77, line 174: **HIGH**

Both this field's doc and `ClientConfig::max_concurrent_requests` say to keep this at or above `pool_max_idle_per_host`, but nothing enforces or tests that. A document with `max_concurrent_requests: 10` and `pool_max_idle_per_host: 512` is accepted silently and sheds load before keep-alive reuse ever kicks in, which is the exact scenario the pool tuning targets. Can we add a test pinning what happens in that case?

MikeFalcon77, line 174: **HIGH**

A `max_concurrent_requests` of 0 from a wiring config file lands here unchecked, and a 0 cap makes the built client shed every request with `HttpError::Overloaded`. `apply_to` is infallible, so there is nowhere to report the bad value and the gear simply fails all its outbound calls at runtime. Clamp 0 to 1 here, or make this conversion fallible so a bad wiring value is caught at startup.

libs/toolkit-contract/tests/multipart_roundtrip.rs
```diff
@@ -249,7 +250,10 @@ type ByteStream = Pin<
 /// a hand-written client is expected to: read the boundary out of the
 /// advertised `Content-Type`, then hand the body's byte stream to the reader.
 async fn open(url: &str) -> MultipartStream<Frame, ByteStream> {
-    let client = build_default_http_client("multipart-roundtrip-test", false).unwrap();
+    let config = ClientConfig::new(url)
+        .with_pool_idle_timeout(None)
```
MikeFalcon77, line 254: **HIGH**

`with_pool_idle_timeout(None)` changes nothing here: `None` makes the builder skip the hyper-util setter, so the pool falls back to hyper-util's own default, which is the same 90s the `ClientConfig` default already sets. What was this meant to do?

libs/toolkit-contract/tests/multipart_roundtrip.rs
```diff
@@ -249,7 +250,10 @@ type ByteStream = Pin<
...
+        .with_max_concurrent_requests(usize::MAX);
```
MikeFalcon77, line 255: **HIGH**

`open()` now builds its transport with the idle timeout off and the concurrency limiter disabled, while `frame_client()` in the same file still uses the plain defaults. Why does the hand-written half need to opt out of exactly the two knobs this PR adds? As written this half no longer runs the transport stack a generated client gets, and nothing here asserts anything about the new settings either way.

libs/toolkit-http/src/builder.rs
```diff
@@ -97,6 +97,20 @@ impl HttpClientBuilder {
...
+    /// Set the rate-limit / concurrency configuration.
+    ///
+    /// `None` disables the concurrency limiter entirely. `Some(cfg)` installs a
+    /// load-shedding [`ConcurrencyLimitLayer`](tower::limit::ConcurrencyLimitLayer)
+    /// capped at `cfg.max_concurrent_requests`; requests beyond the cap fail
+    /// fast with [`HttpError::Overloaded`](crate::error::HttpError::Overloaded)
+    /// rather than queueing. A `max_concurrent_requests` of `usize::MAX` is also
+    /// treated as unlimited (the layer is skipped).
+    #[must_use]
+    pub fn rate_limit(mut self, rate_limit: Option<RateLimitConfig>) -> Self {
```
MikeFalcon77, line 109: **HIGH**

Nothing counts a shed request. The limiter is installed outside the metrics layer in `build()`, so a request rejected here never reaches `MetricsLayer` and is absent from `http.client.request.duration`, and `map_load_shed_error` logs nothing (unlike `map_buffer_error`, which logs at error). Now that the SDK turns this on for every generated REST client, an operator has no signal that the cap is being hit.

MikeFalcon77, line 109: **HIGH**

The `usize::MAX` means unlimited behaviour documented here lives in `build()`'s `rate_limit.max_concurrent_requests < usize::MAX` guard, and no test exercises it. Callers already use it to switch the limiter off, so dropping or inverting that comparison would silently install a limiter where none is wanted with nothing failing.

MikeFalcon77, line 109: **HIGH**

This installs a concurrency cap, not a rate limit: there is no requests-per-second anywhere in `RateLimitConfig`. The doc has to say "rate-limit / concurrency" to cover for the name, and this is a brand new public method, so `concurrency_limit` would describe what it does.

MikeFalcon77, line 109: **HIGH**

`max_concurrent_requests: 0` is accepted here and gives `ConcurrencyLimitLayer::new(0)` in `build()`, which never hands out a permit, so every request through that client fails with `HttpError::Overloaded` for the client's whole lifetime. 0 is a plausible value to pass, since `max_redirects: 0` and `max_retries: 0` mean "off" elsewhere in this crate and the doc says only `usize::MAX` disables the limiter. `buffer_capacity` just below clamps 0 to 1 for exactly this class of problem.

MikeFalcon77, line 109: **HIGH**

Every other setter in this file has a test in `mod tests` below (`test_builder_retry`, `test_builder_max_body_size`, `test_builder_transport_security`), but `rate_limit` has none. `test_load_shedding_returns_overloaded_error` hand-builds the tower stack and never goes through the builder, so nothing covers this method.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -163,12 +163,50 @@ pub struct ClientConfig {
...
+    /// closed and removed from the pool. The companion of
+    /// [`pool_max_idle_per_host`](Self::pool_max_idle_per_host): that field
+    /// bounds *how many* idle connections are kept, this bounds *how long*. A
+    /// connection reaped here and reopened on the next request incurs the same
+    /// `connect(2)` cost the pool is meant to avoid, so keep this comfortably
+    /// above the gap between successive bursts to a given upstream. `None` uses
+    /// hyper-util's default. Defaults to 90s (the historical `toolkit-http`
+    /// default). REST transport only.
+    pub pool_idle_timeout: Option<Duration>,
+    /// Maximum number of in-flight requests allowed through this client at once
```
MikeFalcon77, line 197: This overstates what the cap bounds. The tower permit is released when the response future resolves, which is when the headers arrive, so the body transfer is not counted and a long-lived SSE or `multipart/mixed` stream holds no slot at all once it is open. Worth saying it caps requests waiting on response headers, otherwise operators will size it against their concurrent stream count and get no protection.

libs/toolkit-contract/src/runtime/config.rs
```diff
@@ -191,6 +229,9 @@ impl ClientConfig {
             retry: RetryConfig::default(),
             stream_reconnect: ReconnectConfig::default(),
             require_tls: false,
+            pool_max_idle_per_host: 256,
+            pool_idle_timeout: Some(Duration::from_secs(90)),
```
MikeFalcon77, line 233: These defaults now exist in two places: `toolkit-http` keeps its own `pool_max_idle_per_host: 32`, `pool_idle_timeout: 90s` and `RateLimitConfig::default()` of 100, and `build_default_http_client` always overrides all three, so the `toolkit-http` values are dead for generated clients while the 90s is copied here by hand. When either side is retuned the other drifts silently, and the doc above that quotes the `toolkit-http` numbers goes stale with it.

libs/toolkit-contract/tests/wiring_config.rs
```diff
@@ -225,5 +225,29 @@ mod runtime_conversion {
         let cfg = tuning.apply_to(endpoint);
         assert_eq!(cfg.timeout, Duration::from_secs(30));
         assert_eq!(cfg.retry.max_attempts, 3);
+        assert_eq!(cfg.pool_max_idle_per_host, 256);
+        assert_eq!(cfg.pool_idle_timeout, Some(Duration::from_secs(90)));
+        assert_eq!(cfg.max_concurrent_requests, 256);
+    }
+
+    #[test]
+    fn tuning_apply_overrides_pool_and_concurrency() {
```
MikeFalcon77, line 234: This only covers in-range values. `max_concurrent_requests: 0` is accepted by `apply_to` and ends up as `ConcurrencyLimitLayer::new(0)`, which sheds every request with `HttpError::Overloaded`, and `pool_max_idle_per_host: 0` turns off connection reuse entirely. Can we add a case for `0` on both that pins what we actually want, a parse rejection or a clamp?

libs/toolkit-contract-macros/src/rest_contract.rs
mattgarmon: Fixed - `with_http_client` doc now states transport fields (incl. `require_tls`) aren't applied to a caller-supplied client

mattgarmon: Fixed - generated `new()` doc now lists the transport fields consumed at construction.

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: No changes - release-plz auto-bumps the breaking API on release

mattgarmon: Partial - both cfg arms centralized into one `transport_builder` helper; the compiler-enforced transport-only type is deferred to #4893 with the full split

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: Fixed - wires `.timeout(config.timeout)`, so a >30s timeout is honored instead of capped at 30s

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: Fixed - added non-transient `TransportError::Overloaded`; `map_send_error` maps it instead of folding into `Network`

mattgarmon: Fixed - new integration test builds a cap-1 client and asserts the excess request returns `Overloaded`

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: Fixed - non-otel arm shares `transport_builder`, so the timeout fix applies there too

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: Fixed - extracted shared `transport_builder`; arms now differ only by `.with_metrics`

libs/toolkit-contract/src/runtime/client.rs
mattgarmon: Fixed - non-otel duplication gone (shared helper) and covered by the integration test

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Done - `ClientConfig` is now `#[non_exhaustive]`

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - doc corrected: None = hyper-util default (~90s), not "indefinite"; default is explicit `Some(90s)`

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - reworded; `Overloaded` is non-transient now, so the retry loop no longer re-issues a shed request

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Done - `max_concurrent_requests` is now `Option<usize>` (None = unlimited)

mattgarmon: Addressed - kept the shared config + added a gRPC-construction warning; compile-/parse-time enforcement (the type split) is tracked in #4893

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Changed - defaults lowered to 128/128 (from 256/256), enough to clear the measured ~100 concurrency; higher via `ClientTuning`

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - `pool_idle_timeout: None` path now tested

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - 0 clamped to 1 in the builder

libs/toolkit-contract/src/wiring.rs
mattgarmon: Fixed - added "REST transport only" caveat to the `ClientTuning` field doc

libs/toolkit-contract/src/wiring.rs
mattgarmon: Won't fix - no principled upper bound; pool < concurrency is valid (pool is per-host, limiter is process-wide)

libs/toolkit-contract/src/wiring.rs
mattgarmon: Resolved - not enforced; tests pin that both values are forwarded independently

mattgarmon: Fixed - the same 0->1 clamp covers the wiring path

libs/toolkit-contract/tests/multipart_roundtrip.rs
mattgarmon: Fixed - removed the no-op `with_pool_idle_timeout(None)` from the multipart test

libs/toolkit-contract/tests/multipart_roundtrip.rs
mattgarmon: Fixed - `open()` now uses default config (dropped the opt-outs)

libs/toolkit-http/src/builder.rs
mattgarmon: Fixed - reordered so `MetricsLayer` wraps the limiter; shed requests recorded as `error.type="overloaded"`

mattgarmon: Fixed - unlimited (`None`) path now tested

mattgarmon: Done - renamed the new builder method to `concurrency_limit`

mattgarmon: Fixed - 0 clamped to 1, mirroring `buffer_capacity`

mattgarmon: Fixed - added `test_builder_concurrency_limit` (+ none/unlimited/zero-build cases)

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - doc now says the cap bounds requests waiting on response headers, not stream duration

libs/toolkit-contract/src/runtime/config.rs
mattgarmon: Fixed - inline defaults + comment that `build_default_http_client` overrides toolkit-http's, so they can't drift (dropped the earlier constants)

libs/toolkit-contract/tests/wiring_config.rs
mattgarmon: Fixed - added a 0-case test for both knobs in `wiring_config`

libs/toolkit-contract-macros/src/rest_contract.rs
MikeFalcon77: `libs/toolkit-contract-macros/src/rest_contract.rs:514` — the generated body is still `Self { http, config }`, so `require_tls: true` plus a caller-supplied plaintext client is accepted with no runtime signal; only the rustdoc changed.
Minimal fix: mirror the generated gRPC constructor (`grpc_contract.rs:363`, which errors when `require_tls` is set and the scheme is not `https`), or `tracing::warn!` here when `config.require_tls` is set, since transport fields cannot be applied.

libs/toolkit-contract/src/runtime/client.rs
MikeFalcon77: Our claim was wrong. Versions in `libs/toolkit-contract/Cargo.toml` are release-plz-owned, not hand-edited: `release-plz.toml:40` sets `semver_check = true` workspace-wide, and every bump in that file's history comes from a `chore: release` commit (the 0.3.2 -> 0.3.3 in this range is 75eda218a).
The breaking bump is computed and applied by the release PR, so this PR should not carry it.

MikeFalcon77: `libs/toolkit-contract/src/runtime/client.rs:97` still takes the whole `&ClientConfig`, and the "call-time fields must not be wired here" rule is still only prose at client.rs:92-96 — the compiler cannot catch a future edit that wires `retry` or a streaming knob onto the builder.
Minimal change: give `transport_builder` a transport-only input (a small struct built from `ClientConfig`, or pass `require_tls`, `timeout`, the two pool fields and `max_concurrent_requests` explicitly) so the rest are not in scope.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: Only one of the two docs changed. `libs/toolkit-http/src/builder.rs:343` still reads "Set to `None` to disable idle timeout (connections kept indefinitely)", and the inline comment at builder.rs:405 repeats it, directly contradicting the new text in `libs/toolkit-contract/src/runtime/config.rs` that says `None` falls through to hyper-util's ~90s default.
A caller reading the setter they actually call still gets the wrong answer.
Minimal fix: reword builder.rs:343 and the comment at builder.rs:405 to say `None` leaves the hyper-util setter unset, so hyper-util's own default (~90s) applies.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: Reduced, not eliminated. `libs/toolkit-contract/src/runtime/config.rs:237-239` now defaults to 128/128 — still 4x toolkit-http's 32 idle-connection default, and nothing process-wide was added, so the arithmetic is unchanged in kind: a dozen generated REST clients each build their own pool and limiter, with no shared ceiling across the 90s idle window.
The ask was the opposite direction: keep the default near 32 and let the gears that measured the `connect(2)` problem raise it through `ClientTuning`, which is already wired (wiring.rs:57).
Minimal change: set `pool_max_idle_per_host` back to 32 in `ClientConfig::new`, and set 128 in the wiring of the gears that need it.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: The `None` path lost its last caller instead of gaining a test. `libs/toolkit-contract/tests/multipart_roundtrip.rs:256` now builds the default config, dropping the `with_pool_idle_timeout(None)` it used to pass, and a repo-wide grep for `with_pool_idle_timeout` returns only the setter definition (runtime/config.rs:295) and `ClientTuning::apply_to` (wiring.rs:203), which passes `Some` exclusively.
The new `tests/concurrency_limit.rs` covers the concurrency cap only; it never touches `pool_idle_timeout`.
Minimal change: add a case that calls `ClientConfig::new(..).with_pool_idle_timeout(None)` through `build_default_http_client` and asserts the client builds and serves a request.

libs/toolkit-contract/src/wiring.rs
MikeFalcon77: The limiter is not process-wide. `ConcurrencyLimitLayer` is installed inside each `HttpClient`'s own tower stack (`libs/toolkit-http/src/builder.rs:507-517`, and again at :576-583); there is no static or shared semaphore, and runtime/config.rs:196 documents the cap as "through this client at once (across all upstream hosts)".
So the stated fact does not disprove the finding: `libs/toolkit-contract/src/wiring.rs:74-76` still instructs operators to keep the cap at or above `pool_max_idle_per_host`, and wiring.rs:200 forwards the pool value with no bound and no cross-check.
Minimal change: either drop that sentence from the `ClientTuning` and `ClientConfig` docs, or have `apply_to` clamp or `warn!` when `pool_max_idle_per_host` exceeds `max_concurrent_requests`.

libs/toolkit-contract/src/wiring.rs
MikeFalcon77: The docs at HEAD still assert the invariant: `libs/toolkit-contract/src/wiring.rs:74-76` and runtime/config.rs:196-201 both tell operators to keep the cap at or above `pool_max_idle_per_host`.
The added coverage in `libs/toolkit-contract/tests/wiring_config.rs` pins 128/128 (:228), pool 512 vs cap 1000 (:249) and 0/0 (:274) — every case has cap >= pool, so the skewed `cap < pool` case is still unpinned.
Minimal change: one more `apply_to` case with `max_concurrent_requests: 10, pool_max_idle_per_host: 512` asserting the resulting `ClientConfig` values, so the documented-but-unenforced skew is recorded.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: Both declaration sites are still there; only a comment was added. `libs/toolkit-http/src/config.rs:697-698` still hardcodes `pool_idle_timeout: Some(Duration::from_secs(90))` and `pool_max_idle_per_host: 32`, `RateLimitConfig::default()` at config.rs:323 still hardcodes 100, and `libs/toolkit-contract/src/runtime/config.rs:237-239` hardcodes 128/90s/Some(128) separately — the 90s is still hand-copied across two crates.
The new comment at runtime/config.rs:235 even embeds the numbers "(32/90s/100)", adding a third thing to go stale when either side is retuned.
Minimal change: export `DEFAULT_POOL_IDLE_TIMEOUT` / `DEFAULT_POOL_MAX_IDLE_PER_HOST` from toolkit-http and have both `HttpClientConfig::default` and `ClientConfig::new` read them.

libs/toolkit-contract/src/runtime/client.rs
MikeFalcon77: `libs/toolkit-contract/src/runtime/client.rs:97` — `transport_builder` still takes `&ClientConfig`, so the transport/call-time split is still only the doc comment above it, now with `timeout` added to the set it must remember. Narrow the parameter to a transport-only struct (or pass the five values) here rather than leaving it to #4893.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: `libs/toolkit-http/src/builder.rs:343` still documents `None` as "disable idle timeout (connections kept indefinitely)", and `libs/toolkit-http/README.md:101` repeats it, so the two docs still disagree about the same value. `build` skips the hyper-util setter (`builder.rs:406`), so the ~90s default applies. Correct the builder doc and the README, or drop the `Option` and take a `Duration`.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: `libs/toolkit-contract/src/runtime/config.rs:237` — 128 idle connections per upstream host is still 4× toolkit-http's own 32 default, applied per generated client, and no process-wide budget was added. Keep `pool_max_idle_per_host` at 32 and have the gears that measured the `connect(2)` cost raise it through `ClientTuning`.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: `libs/toolkit-contract/src/runtime/config.rs:295` — no test sets `pool_idle_timeout` to `None`; `wiring.rs:203` still wraps the wiring value in `Some`, and `with_pool_idle_timeout(None)` appears nowhere in the workspace. Add a test that `build_default_http_client` succeeds with `pool_idle_timeout: None` and still reuses a pooled connection, pinning the hyper-util-default path.

libs/toolkit-contract/src/wiring.rs
MikeFalcon77: `libs/toolkit-contract/src/wiring.rs:199` still forwards `pool_max_idle_per_host` verbatim with no clamp. The relation is not moot either: the same file's doc at line 74 still tells operators to keep `max_concurrent_requests` at or above it, and the limiter lives on each `HttpClient`, so it is per-client, not process-wide. Clamp in `apply_to` or delete that instruction from the doc.

libs/toolkit-contract/src/wiring.rs
MikeFalcon77: `libs/toolkit-contract/src/wiring.rs:74` still instructs operators to keep `max_concurrent_requests` at or above `pool_max_idle_per_host`, and nothing covers the violating case — `tests/wiring_config.rs:283` only pins that each value is forwarded. Add a case with `max_concurrent_requests: 10` and `pool_max_idle_per_host: 512` pinning that it is accepted as-is, or remove the instruction from the doc.

libs/toolkit-contract/src/runtime/config.rs
MikeFalcon77: `libs/toolkit-contract/src/runtime/config.rs:238` still hardcodes `Some(Duration::from_secs(90))` next to toolkit-http's own `Some(Duration::from_secs(90))` at `libs/toolkit-http/src/config.rs:697`, and the new comment hardcodes `32/90s/100` as well — a comment cannot keep the two sides in step. Reference toolkit-http's defaults (or named constants) instead of restating the numbers.

PR #4892, opened 2026-09-21
fix(types-registry): backtick PostgreSQL/MySQL in backend test doc comments

entropyshift:
## Description

The **Clippy (full feature matrix)** workflow is failing on `main`:

https://github.com/constructorfabric/gears-rust/actions/runs/35612793094/job/106375856739

**The error.**

```
error: item in documentation is missing backticks
 --> gears/system/types-registry/types-registry/tests/compat_backends_test.rs:1:35
1 | //! T17/T18 admission behavior on PostgreSQL and MySQL.
  |                                   ^^^^^^^^^^
```

`clippy::doc_markdown` is set to `"deny"` workspace-wide via `[workspace.lints.clippy]` in the root `Cargo.toml`. Cargo turns that into a per-crate `-D clippy::doc-markdown` flag on the command line, which outranks the `#![allow(clippy::doc_markdown)]` that's already sitting at the top of both affected files — a command-line `-D` cannot be relaxed by an in-source `#[allow]`. That existing `allow` was dead code; it never actually suppressed anything.

Both files also carry `#![cfg(feature = "integration")]`, which strips the rest of the file's content when the feature is off — but it does **not** strip the module's own `//!` doc comment above it. So the lint still fires on the `--no-default-features` combo specifically, even though the file is otherwise empty in that build.

**Two files, same root cause:**

| File | Introduced | PR |
| --- | --- | --- |
| `compat_backends_test.rs` | 2026-09-11 | #4772 |
| `deletion_backends_test.rs` | 2026-09-15 | #4796 |

**Since when.** The deep matrix has been blocked earlier than this crate almost continuously since these landed — most recently by the `toolkit-utils`/`schemars` gap fixed in #4825. Once that landed, the matrix got far enough to reach this crate:

| | |
| --- | --- |
| First hit (`deletion_backends_test.rs`) | run [34982956635](https://github.com/constructorfabric/gears-rust/actions/runs/34982956635) — 2026-09-15 |
| Latest (`compat_backends_test.rs`) | run [35612793094](https://github.com/constructorfabric/gears-rust/actions/runs/35612793094) — 2026-09-21 |

**Why only these two, and not the whole repo.** A repo-wide grep turns up dozens of unbackticked `PostgreSQL`/`MySQL`/`SQLite` mentions in doc comments elsewhere (`toolkit-db`, `chat-engine`, `file-storage`, `resource-group`, `ledger`, etc.). I checked rather than assumed: two other files in this same crate, `partial_admission_backends_test.rs` and `dry_run_batch_backends_test.rs`, use the slash-joined form `PostgreSQL/MySQL` and — verified locally — do **not** trip the lint at all; clippy's word-splitter doesn't treat that joined token as a bare identifier the way it treats "PostgreSQL and MySQL" as two separate ones. So a blind find-and-replace across the repo would have touched files that were never actually broken, and would still have missed knowing whether the untouched ones are safe. I did not attempt to check the rest of the workspace: most of it has not been reached by `clippy-deep`'s fail-fast matrix in weeks (see #4825's own note on this), so there is no CI evidence either way for those, and guessing from grep alone already proved wrong twice in this crate.

**The fix.** Backtick the two standalone mentions:

```diff
-//! T17/T18 admission behavior on PostgreSQL and MySQL.
+//! T17/T18 admission behavior on `PostgreSQL` and `MySQL`.
```
```diff
-//! Deletion and dry-run checks on PostgreSQL and MySQL (T20).
+//! Deletion and dry-run checks on `PostgreSQL` and `MySQL` (T20).
```

## Testing

Verified locally with the pinned toolchain (`rust-toolchain.toml`: 1.97.0), against the exact failing command and the crate's other feature combos:

```
cargo clippy --all-targets --manifest-path gears/system/types-registry/types-registry/Cargo.toml --no-default-features -- -D warnings -D clippy::perf
cargo clippy --all-targets --manifest-path gears/system/types-registry/types-registry/Cargo.toml --no-default-features --features integration -- -D warnings -D clippy::perf
cargo clippy --all-targets --manifest-path gears/system/types-registry/types-registry/Cargo.toml --all-features -- -D warnings -D clippy::perf
```

All three reproduced the error before this change and pass clean after it. `cargo fmt --check` passes.

> **Please note:** this may not turn the workflow fully green on its own. `types-registry` is one crate out of the ~267-combo matrix, and most of the workspace hasn't been reached by a clean deep-clippy run in weeks. This change is what lets the job get past this crate to find out what's behind it.

## Type of Change

- [ ] Documentation update

## Checklist

- [x] Self-review completed

## Summary by CodeRabbit

- **Documentation**
  - Improved inline-code formatting for backend names in test documentation.
  - No functional changes.


entropyshift:
Seems to be no longer relevant after the latest changes in main

